PR #34, opened 2024-10-16
Adding community workflow

staylorTT:
Adds a workflow to label issues filed by non-tt org accounts with the "community label" and then ping a TT slack channel.

### DIFF
--- a/.github/workflows/on-community-issue.yml
+++ b/.github/workflows/on-community-issue.yml
@@ -1,0 +1,67 @@
+name: "Issue Notification"
+
+on:
+  issues:
+    types: [opened]
+  pull_request:
+    types: [opened]
+
+jobs:
+  gh-slack-bridge:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - name: Set username and number based on event type
+        id: set-variables
+        run: |
+          if [ "${{ github.event_name }}" = "issues" ]; then
+            echo "USERNAME=${{ github.event.issue.user.login }}" >> $GITHUB_ENV
+            echo "NUMBER=${{ github.event.issue.number }}" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


AleksKnezevic (approved):

.github/workflows/on-community-issue.yml
```diff
@@ -0,0 +1,67 @@
...
+          elif [ "${{ github.event_name }}" = "pull_request" ]; then
+            echo "USERNAME=${{ github.event.pull_request.user.login }}" >> $GITHUB_ENV
+            echo "NUMBER=${{ github.event.pull_request.number }}" >> $GITHUB_ENV
+            echo "EVENT_URL=${{ github.event.pull_request.html_url }}" >> $GITHUB_ENV
+            echo "EVENT_TYPE=PR" >> $GITHUB_ENV
+            echo "GH_COMMAND=pr" >> $GITHUB_ENV
+          elif [ "${{ github.event_name }}" = "push" ]; then
+            echo "USERNAME=${{ github.actor }}" >> $GITHUB_ENV
+            echo "EVENT_URL=https://github.com/${{ github.repository }}/commit/${{ github.sha }}" >> $GITHUB_ENV
+            echo "EVENT_TYPE=push" >> $GITHUB_ENV
```
mrakitaTT, line 34: Will this ever be triggered since `push` is not in the `on:` list on line 3?

mrakitaTT (approved):
LGTM except that one comment. FYI @vmilosevic if you want to take a look too.

staylorTT:
This workflow is already working as is on forge-fe and tt-mlir. Push is not needed based off the other workflows.

mrakitaTT:
@staylorTT I don't see how will those lines ever be triggered if `push` is missing from the `on:` list. See this: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#using-a-single-event

Do you actually get notifications on slack that a non-org member has pushed something? Also the message would be a bit weird in that case: `A new push has been created by a non-org member`

PR #52, opened 2024-11-05
Refactoring jax python tests

kmitrovicTT:
- Rewrote verify_module as decorator
- Decorated python tests with it
- Used black formatter to format tests
- Small refactorings and simplifications of some tests


tests/TTIR/test_scalar_add.py
kmitrovicTT, line 1: Removed completely since it seemed obsolete. Correct me if I am wrong.

tests/TTIR/test_array_add.py
```diff
@@ -3,34 +3,23 @@
...
-  def module_add(a, b):
-    return a + b
-
-  verify_module(module_add, [(2, 2), (2, 2)])
+from infrastructure import verify_test_module
 
 
-def test_3x2_array_add():
-  def module_add(a, b):
+@verify_test_module([(2, 2), (2, 2)], [(3, 2), (3, 2)])
```
AleksKnezevic, line 10: I'm not a huge fan of the input shapes being in the decorator like this and much prefer the parameterized tests as below in `test_module_add`. Since you're updating the tests, and it's not that many yet, you could change them. 

That, said, this is nitpicky and unrelated to you changes, so feel free to ignore 🙂

AleksKnezevic (approved):

tests/TTIR/test_array_add.py
kmitrovicTT: That made sense to me if tests were pytest parametrized, like 

```python
@pytest.mark.parametrize("rank", [1, 2, 3, 4, 5, 6])
def test_module_add(rank):
    input_shape = (32,) if rank == 1 else (1,) * (rank - 2) + (32, 32)

    @verify_test_module([input_shape, input_shape])
    def module_add(a, b):
        c = a + a
        d = b + b
        return c + d

    module_add() # notice the explicit call
```

but for those that were not, decorator made it possible for 

```python
def test_2x2_array_add():
  def module_add(a, b):
    return a + b

  verify_module(module_add, [(2, 2), (2, 2)])
  
def test_3x2_array_add():
  def module_add(a, b):
    return a + b

  verify_module(module_add, [(3, 2), (3, 2)])
```

to become

```python
@verify_test_module([(2, 2), (2, 2)], [(3, 2), (3, 2)])
def test_add(a, b):
    return a + b
```

which, for me, made it simple, neat, expressive and without explicit call of the nested function.

Throughout tests I left various examples, to show how decorator is flexible in its use. These two cases are a good example of its usefulness.

@mrakitaTT @nsmithtt please add some suggestions, I have no strong opinion, you and @AleksKnezevic are one of the stakeholders here.

uazizTT (approved):
You would need to rebase once the PR to unblock tt-xla is merged.

tests/TTIR/test_array_add.py
mrakitaTT: @kmitrovicTT I guess what @AleksKnezevic meant is that your decorator can be replaced by `@pytest.mark.parametrize` which is a builtin decorator and provides additional benefits, like for example marking individual parameters as `mark.xfail`. Additionally, with `parametrize` the whole test will be run multiple times and failures will be reported individually, while with your decorator test is run once with inner loop for shape variants and we can only provide one combined test failure thus require additional logging that you've added to print each failure.

Only drawback is that we have to manually call the `verify_module(input_shape)` in every test. Is there a way to combine your decorator with parametrize in order to avoid that too?

Btw I am sorry for not noticing this earlier when you've sent me the initial commit :( It looked really neat on the first glance but Aleks has a solid argument.

AleksKnezevic: Yeah, that's exactly it @mrakitaTT!

Issue #60, opened 2024-11-08
Add pre-commit CI job


mmanzoorTT:
CI is not verifying the code formatting. Add GitHub workflow to add the `pre-comm` CI pipeline.

mmanzoorTT:
https://github.com/tenstorrent/tt-torch/pull/13 adds multiple pipelines for github workflow for `tt-torch` repo. It can be used as reference to add `pre-commit` pipeline.

mmanzoorTT:
Here is another PR https://github.com/tenstorrent/tt-torch/pull/28

it first runs pre-commit and license checks; if successful then it runs the remaining jobs.

PR #66, opened 2024-11-13
Added gh CI workflows

kmitrovicTT:
Fixes https://github.com/tenstorrent/tt-xla/issues/60.

Also fixed missing tt-mlir version in automatic PR title and description.

image

### DIFF
--- a/.github/workflows/nightly-uplift.yml
+++ b/.github/workflows/nightly-uplift.yml
@@ -5,7 +5,7 @@ name: Nighty Uplift
 
 on:
   schedule:
-    - cron: '0 8 * * *'  # Runs at 08:00 UTC every day
+    - cron: '0 6 * * *'  # Runs at 06:00 UTC every day
   workflow_dispatch:  # Manual trigger
 
 jobs:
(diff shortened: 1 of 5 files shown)


.github/workflows/nightly-uplift.yml
kmitrovicTT, line 1: If no one has anything against, should I set `cron` to `06:00 UTC` which is 7AM Belgrade time? It will finish before Belgrade starts working, might be convenient.

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pre-commit
+          pip install -r requirements.txt
+
+      - name: Install pre-commit hooks
+        run: |
+          pre-commit install
```
kmitrovicTT, line 26: @mmanzoorTT tt-torch does not have these two steps, chat-gpt recommended them to me. Do you think we should have them in both repos?

mmanzoorTT: @uazizTT has implemented these workflows. He can provide more insight about it. Moreover, `pre-commit` package is part of `requirements.txt`; so adding `pip install pre-commit` seems redundant to me.

uazizTT: @kmitrovicTT I agree with @mmanzoorTT that it is redundant to add these as they are part of requirements. Also for tt-torch, I am working on a change that will pre-build all dependencies in a single build machine and then copy it over to all the test runners so that all test runners don't have to build the environment again. For that reason I think it is better to put all dependencies in a single requirements.txt file.

Issue #73, opened 2024-11-18
TT-xla not supporting scalar values


ajakovljevicTT:
Due to the fact that TTIR does not support scalar values, but internally translates them to one-dimensional arrays, the TT-xla frontend cannot test functions which return scalars, as both the lowering of the stableHLO ops and the infrastructure for testing is not able to deal with this issue.

ajakovljevicTT:
Closed as duplicate of https://github.com/tenstorrent/tt-xla/issues/81

PR #75, opened 2024-11-19
Fixes for test infrastructure to support scalars

ajakovljevicTT:
Our test infrastructure on the tt-xla side calls some jax ops (specifically `max` op), on the results of the pytests. If the test result is scalar, the op fails because it does not support that shape. This MR gives a fix for that.

However, the scalars are not fully enabled on the tt-mlir (see issue https://github.com/tenstorrent/tt-mlir/issues/1306), and test is currently marked as skipped until we merge and ingest the fix on the tt-mlir side.

### DIFF
--- a/tests/infrastructure.py
+++ b/tests/infrastructure.py
@@ -42,7 +42,11 @@ def compare_tensor_to_golden(
         if assert_on_error:
             assert ret, f"PCC is {pcc} which is less than {required_pcc}"
 
-    atol = jnp.abs(tensor - golden) if len(tensor.shape) == 0 else jnp.max(jnp.abs(tensor - golden))
+    atol = (
+        jnp.abs(tensor - golden)
+        if len(tensor.shape) == 0
+        else jnp.max(jnp.abs(tensor - golden))
+    )
     ret = ret and atol <= required_atol
     if assert_on_error:
         assert ret, f"ATOL is {atol} which is greater than {required_atol}"
--- a/tests/infrastructure.py
+++ b/tests/infrastructure.py
@@ -42,7 +42,7 @@ def compare_tensor_to_golden(
         if assert_on_error:
             assert ret, f"PCC is {pcc} which is less than {required_pcc}"
 
-    atol = jnp.max(jnp.abs(tensor - golden))
+    atol = jnp.abs(tensor - golden) if len(tensor.shape) == 0 else jnp.max(jnp.abs(tensor - golden))
     ret = ret and atol <= required_atol
     if assert_on_error:
         assert ret, f"ATOL is {atol} which is greater than {required_atol}"
--- a/tests/infrastructure.py
+++ b/tests/infrastructure.py
@@ -2,8 +2,19 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+from contextlib import contextmanager
 import jax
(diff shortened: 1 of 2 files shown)


tests/infrastructure.py
```diff
@@ -42,7 +42,11 @@ def compare_tensor_to_golden(
         if assert_on_error:
             assert ret, f"PCC is {pcc} which is less than {required_pcc}"
 
-    atol = jnp.max(jnp.abs(tensor - golden))
+    atol = (
```
kmitrovicTT, line 45: If `jnp.max(jnp.abs(tensor - golden))` will work once https://github.com/tenstorrent/tt-xla/issues/73 is fixed, leave a comment here to revert this once fixed.

kmitrovicTT (approved):

tests/infrastructure.py
ajakovljevicTT: It will not, those are separate issues (my apologies if I worded in wrongly). The test is skipped because of https://github.com/tenstorrent/tt-xla/issues/73, and I will enable it when it is fixed.

mmanzoorTT (approved):

tests/infrastructure.py
kmitrovicTT: if don't get why `jnp.max(jnp.abs(tensor - golden))` wouldn't work if `tensor` and `golden` were scalars. I just tried it

```
>>> def atol(a, b):
...     return jnp.max(jnp.abs(a-b))
... 
>>> atol(1, 2)
Array(1, dtype=int32)
>>> a = np.array([1, 2, 3])
>>> b = np.array([4, 5, 6])
>>> atol(a, b)
Array(3, dtype=int32)
```

ajakovljevicTT: The issue was that the default test behaviour was putting both the golden and silicon output on silicon, leading to the `jnp` to be ran on our silicon, which lead to issues with scalars running on tt-metal. Changed the code to force the check to be ran on cpu and reverted the change.

AleksKnezevic (approved):

tests/infrastructure.py
kmitrovicTT: Wow, that's a big deal. Let's leave this thread open for @mrakitaTT to take a look.

kmitrovicTT (approved):

tests/infrastructure.py
```diff
@@ -66,6 +73,6 @@ def verify_module(
     tt_inputs = [jax.device_put(cpu_input, tt_device) for cpu_input in cpu_inputs]
     graph = jax.jit(module)
     res = graph(*tt_inputs)
-    res_cpu = graph(*cpu_inputs)
-
+    with jax.default_device(jax.local_devices(backend="cpu")[0]):
```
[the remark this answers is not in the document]
kmitrovicTT: Or `run_on_cpu` or however you wanna name it.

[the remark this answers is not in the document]
ajakovljevicTT: Addressed the comment.

tests/infrastructure.py
mrakitaTT: Discussed offline, turns out we need to be more explicit about where we put tensors and run graphs. Andrej will open separate task to streamline this.

Issue #81, opened 2024-11-22
Fix how scalars are handled in PJRT


ajakovljevicTT:
Currently there is an issue to transfer scalars from silicon to cpu for testing due to the differing layouts, which leads to reshapes of scalars in our test infra (see PR https://github.com/tenstorrent/tt-xla/pull/75). We can address this directly in PJRT, therefore taking away that responsibility from the users of the `tt-xla` repository

PR #95, opened 2024-12-02
Added support for scalars in tt-xla

ajakovljevicTT:
Due to the fact that scalars are not supported in our TTIR dialect and instead promoted to tensors during StableHLO->TTIR conversion, the PJRT runtime did not now that return values of jax functions were scalars and not 1x1 tensors. This led to the behaviour of the python calls differing between CPU and Silicon calls, with CPU calls returning scalars as expected, and Silicon jax calls returning 1x1 tensors. 

This necessitated a workaround reshaping in the test infra, and additionally did not match with that the end user of the jax framework expected. This PR fixes that, by adding tracking of scalars on the PJRT runtime side. 

Fixes issue https://github.com/tenstorrent/tt-xla/issues/81

### DIFF
--- a/src/common/api_impl.cc
+++ b/src/common/api_impl.cc
@@ -1035,7 +1035,7 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
   assert(rt_outputs.size() == output_specs.size());
 
   for (size_t i = 0; i < output_specs.size(); ++i) {
-    bool is_scalar = client_.get_module_builder()->is_output_scalar(i);
+    bool is_scalar = client_.isOutputScalar(i);
     // PJRT expects an empty shape for scalars.
     std::vector<std::uint32_t> output_shape =
         is_scalar ? std::vector<std::uint32_t>() : output_specs[i].shape;
--- a/src/common/api_impl.cc
+++ b/src/common/api_impl.cc
@@ -1035,9 +1035,13 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
   assert(rt_outputs.size() == output_specs.size());
 
   for (size_t i = 0; i < output_specs.size(); ++i) {
+    bool is_scalar = client_.get_module_builder()->is_output_scalar(i);
+    // PJRT expects an empty shape for scalars.
+    std::vector<std::uint32_t> output_shape =
+        is_scalar ? std::vector<std::uint32_t>() : output_specs[i].shape;
     auto result_buffer = std::make_unique<BufferInstance>(
-        *this->addressable_devices_[dev_index], rt_outputs[i],
-        output_specs[i].shape, output_specs[i].stride);
+        *this->addressable_devices_[dev_index], rt_outputs[i], output_shape,
+        output_specs[i].stride);
     result_buffer->setType(
         convertElementTypeToBufferType(output_specs[i].dataType));
     DLOG_F(INFO, "Runtime output id: %d", result_buffer->unique_id());
--- a/src/common/api_impl.h
+++ b/src/common/api_impl.h
@@ -370,9 +370,9 @@ public:
   // Advances the timeline, returning (current, next) time point values.
   std::tuple<uint64_t, uint64_t> AdvanceTimeline();
 
-  // Returns the module builder used for this ClientInstance.
-  const ModuleBuilder *get_module_builder() const {
-    return module_builder_.get();
+  // Checks if the output on the i-th index is a scalar.
+  bool isOutputScalar(const int index) const {
+    return module_builder_->isOutputScalar(index);
   }
 
 protected:
--- a/src/common/api_impl.h
+++ b/src/common/api_impl.h
@@ -370,6 +370,11 @@ public:
   // Advances the timeline, returning (current, next) time point values.
   std::tuple<uint64_t, uint64_t> AdvanceTimeline();
 
+  // Returns the module builder used for this ClientInstance.
+  const ModuleBuilder *get_module_builder() const {
+    return module_builder_.get();
+  }
+
 protected:
   std::string cached_platform_name_;
   std::string cached_platform_version_;
--- a/src/common/api_impl.h
+++ b/src/common/api_impl.h
@@ -371,7 +371,7 @@ public:
   std::tuple<uint64_t, uint64_t> AdvanceTimeline();
 
   // Checks if the output on the i-th index is a scalar.
-  bool isOutputScalar(const int index) const {
+  bool isOutputScalar(const size_t index) const {
     return module_builder_->isOutputScalar(index);
   }
 
--- a/src/common/module_builder.cc
+++ b/src/common/module_builder.cc
@@ -66,38 +66,16 @@ ModuleBuilder::ModuleBuilder()
   m_context->appendDialectRegistry(registry);
 }
 
-static bool isScalarType(mlir::Type type) {
-  if (mlir::isa<mlir::FloatType>(type) || mlir::isa<mlir::IntegerType>(type)) {
-    return true;
-  }
-  if (auto tensorType = mlir::dyn_cast<mlir::RankedTensorType>(type)) {
-    return tensorType.getRank() == 0;
-  }
-  return false;
-}
-
-void ModuleBuilder::collectOutputTypes(mlir::ModuleOp &&module) {
-  m_is_output_scalar.clear();
-  for (auto &op : module.getOps()) {
-    if (auto funcOp = mlir::dyn_cast<mlir::func::FuncOp>(op)) {
-      // We care only for return ops of public functions, as that are the ones
-      // that will produce results in the flatbuffer.
-      if (funcOp.isPublic()) {
-        funcOp.walk([&](mlir::Operation *op) {
-          if (auto returnOp = mlir::dyn_cast<mlir::func::ReturnOp>(op)) {
-            for (auto operand : returnOp->getOperands()) {
-              m_is_output_scalar.push_back(isScalarType(operand.getType()));
-            }
-          }
-        });
-      }
-    }
-  }
+bool ModuleBuilder::isOutputScalar(int index) const {
+  assert(index < m_is_output_scalar.size() && "Index of output out of range.");
+  return m_is_output_scalar[index];
 }
 
 tt_pjrt_status ModuleBuilder::buildModule(const std::string_view &code,
                                           const std::string_view &format) {
   DLOG_F(LOG_DEBUG, "ModuleBuilder::buildModule");
+
+  m_status = tt_pjrt_status::kSuccess;
 
   mlir::OwningOpRef<mlir::ModuleOp> mlir_module = createVHLOModule(code);
   if (!tt_pjrt_status_is_ok(m_status)) {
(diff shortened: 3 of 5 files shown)


src/common/module_builder.h
```diff
@@ -33,6 +33,8 @@ class ModuleBuilder {
 
   size_t getNumOutputs() const { return m_num_outputs; };
 
+  bool is_output_scalar(int index) const { return m_is_output_scalar[index]; }
```
kmitrovicTT, line 36: Use camelCaseStyle.

src/common/module_builder.h
```diff
@@ -51,6 +53,10 @@ class ModuleBuilder {
   void
   createFlatbufferBinary(const mlir::OwningOpRef<mlir::ModuleOp> &mlir_module);
 
+  // Fills up the m_is_output_scalar array with information is the output type
+  // scalar or not.
+  void collectOutputTypes(mlir::ModuleOp &&module);
+
   // Prints module to console for debug purposes.
   static void print_module(mlir::OwningOpRef<mlir::ModuleOp> &mlir_module);
```
kmitrovicTT, line 61: Change this to camelCase as well if it is not used in too many places and will bloat PR.

src/common/module_builder.cc
```diff
@@ -65,6 +66,36 @@ ModuleBuilder::ModuleBuilder()
...
+      // that will produce results in the flatbuffer.
+      if (funcOp.isPublic()) {
+        funcOp.walk([&](mlir::Operation *op) {
+          if (mlir::func::ReturnOp return_op =
+                  mlir::dyn_cast<mlir::func::ReturnOp>(op)) {
+            for (auto operand : op->getOperands()) {
+              m_is_output_scalar.push_back(isScalarType(operand.getType()));
+            }
+          }
+        });
```
kmitrovicTT, line 93: I can't figure out where `});` is coming from.

src/common/module_builder.cc
```diff
@@ -65,6 +66,36 @@ ModuleBuilder::ModuleBuilder()
...
+  if (auto tensorType = mlir::dyn_cast<mlir::RankedTensorType>(type)) {
+    return tensorType.getRank() == 0;
+  }
+  return false;
+}
+
+void ModuleBuilder::collectOutputTypes(mlir::ModuleOp &&module) {
+  m_is_output_scalar.clear();
+  for (auto &op : module.getOps()) {
+    if (auto funcOp = mlir::cast<mlir::func::FuncOp>(op)) {
```
kmitrovicTT, line 82: `mlir::cast` will assert if `op` cannot be casted to `FuncOp`. Use `dyn_cast` like below instead.

You can use `auto` down below to shorten the statement. It is obvious what the type will be from the cast.

src/common/module_builder.cc
```diff
@@ -65,6 +66,36 @@ ModuleBuilder::ModuleBuilder()
...
+      // We care only for return ops of public functions, as that are the ones
+      // that will produce results in the flatbuffer.
+      if (funcOp.isPublic()) {
```
kmitrovicTT, line 85: What does it mean for op to be public?

src/common/api_impl.h
```diff
@@ -364,6 +364,11 @@ class ClientInstance {
   // Advances the timeline, returning (current, next) time point values.
   std::tuple<uint64_t, uint64_t> AdvanceTimeline();
 
+  // Returns the module builder used for this ClientInstance.
+  const ModuleBuilder *get_module_builder() const {
+    return module_builder_.get();
+  }
```
kmitrovicTT, line 370: You don't have to return module builder, you can to

```cpp
bool is_output_scalar(int index) const { return module_builder_.is_output_scalar(); }
```

src/common/module_builder.cc
ajakovljevicTT: Function will be public if it was compiled to be directly used by the frontend, and hence will have output buffers, which means having output buffers. If the function is related to some jax optimisation, compilation specific (as is the `jax.lax` remainder case), it will be private.

kmitrovicTT (approved):

ajakovljevicTT:
@AleksKnezevic do you have the time to take a look, interested in your opinion. Thanks!

AleksKnezevic (approved):

src/common/module_builder.cc
```diff
@@ -65,6 +66,35 @@ ModuleBuilder::ModuleBuilder()
...
+    return true;
+  }
+  if (auto tensorType = mlir::dyn_cast<mlir::RankedTensorType>(type)) {
+    return tensorType.getRank() == 0;
+  }
+  return false;
+}
+
+void ModuleBuilder::collectOutputTypes(mlir::ModuleOp &&module) {
+  m_is_output_scalar.clear();
```
mrakitaTT, line 80: Good catch, you just made me realize I've introduced a bug with not reseting `m_status` at the beginning of `buildModule` function. Can you please fix that too?

src/common/module_builder.cc
```diff
@@ -65,6 +66,35 @@ ModuleBuilder::ModuleBuilder()
   m_context->appendDialectRegistry(registry);
 }
 
+static bool isScalarType(mlir::Type type) {
```
mrakitaTT, line 69: Let's try to keep function declarations and definitions organized in top->down order in which they are first called, to facilitate easier code reading. If we have functions A and B, and A calls A1 and A2, I am fine with both `A, B, A1, A2` and `A, A1, A2, B` order. I usually group them closer together when they are locally called so I would choose `A, A1, A2, B`, but for example if A2 was called also from B (like `printModule` is called from many functions) then I would move it at the bottom.

mrakitaTT (changes requested):

src/common/module_builder.cc
```diff
@@ -65,6 +66,35 @@ ModuleBuilder::ModuleBuilder()
...
+  }
+  if (auto tensorType = mlir::dyn_cast<mlir::RankedTensorType>(type)) {
+    return tensorType.getRank() == 0;
+  }
+  return false;
+}
+
+void ModuleBuilder::collectOutputTypes(mlir::ModuleOp &&module) {
+  m_is_output_scalar.clear();
+  for (auto &op : module.getOps()) {
```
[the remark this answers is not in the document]
ajakovljevicTT: Thanks for pointing this out! I've changed the function to a walk across ops. To do that, I did some of the stuff differently then your other comments. Please let me know if that's ok and if I need to change anything.

src/common/api_impl.h
```diff
@@ -364,6 +364,11 @@ class ClientInstance {
   // Advances the timeline, returning (current, next) time point values.
   std::tuple<uint64_t, uint64_t> AdvanceTimeline();
 
+  // Checks if the output on the i-th index is a scalar.
+  bool isOutputScalar(const int index) const {
```
mrakitaTT, line 368: Use `size_t` instead of int

src/common/module_builder.cc
```diff
@@ -65,10 +66,17 @@ ModuleBuilder::ModuleBuilder()
   m_context->appendDialectRegistry(registry);
 }
 
+bool ModuleBuilder::isOutputScalar(int index) const {
```
mrakitaTT, line 69: Use `size_t` instead of int, in declaration too

src/common/module_builder.cc
```diff
@@ -65,10 +66,17 @@ ModuleBuilder::ModuleBuilder()
...
+  assert(index < m_is_output_scalar.size() && "Index of output out of range.");
```
mrakitaTT, line 70: Nit: Change to `Output index is out of range`. Also assert/error messages shouldn't be terminated with a dot.

src/common/module_builder.cc
```diff
@@ -192,9 +241,15 @@ void ModuleBuilder::createFlatbufferBinary(
   tt::runtime::Binary runtime_binary_handle(m_flatbuffer_binary);
   m_num_inputs = runtime_binary_handle.getProgramInputs(0).size();
   m_num_outputs = runtime_binary_handle.getProgramOutputs(0).size();
+
+  if (m_num_outputs != m_is_output_scalar.size()) {
+    DLOG_F(ERROR, "The number of return types does not match ");
+    m_status = tt_pjrt_status::kInternal;
+    return;
```
mrakitaTT, line 248: No need for `return`

src/common/module_builder.cc
```diff
@@ -192,9 +241,15 @@ void ModuleBuilder::createFlatbufferBinary(
   tt::runtime::Binary runtime_binary_handle(m_flatbuffer_binary);
   m_num_inputs = runtime_binary_handle.getProgramInputs(0).size();
   m_num_outputs = runtime_binary_handle.getProgramOutputs(0).size();
+
+  if (m_num_outputs != m_is_output_scalar.size()) {
+    DLOG_F(ERROR, "The number of return types does not match ");
```
mrakitaTT, line 246: I would change the message to: `"Created flatbuffer binary contains different number of outputs ({%0}) than expected ({%1})", m_num_outputs , m_is_output_scalar.size()`. I am not sure how exactly formatting is done in loguru to insert these numbers, you can look it up.

src/common/module_builder.cc
```diff
@@ -135,7 +145,46 @@ void ModuleBuilder::convertFromVHLOToSHLO(
...
+        m_is_output_scalar.push_back(isScalarType(returnType));
+      }
+    }
+
+    if (moduleOp) {
+      if (moduleOp == module.get()) {
+        return;
+      }
+      collectOutputTypes(moduleOp);
+    }
```
mrakitaTT, line 175: You don't need this since walk works in a recursive fashion itself. Also this would re-trigger clearing of the `m_is_output_scalar` and log from the beginning.

src/common/module_builder.cc
```diff
@@ -135,7 +145,46 @@ void ModuleBuilder::convertFromVHLOToSHLO(
...
+
+  m_is_output_scalar.clear();
+
+  module.get().walk([&](mlir::Operation *op) {
+    mlir::func::FuncOp funcOp = mlir::dyn_cast<mlir::func::FuncOp>(op);
+    mlir::ModuleOp moduleOp = mlir::dyn_cast<mlir::ModuleOp>(op);
+
+    // We care only about return ops of public functions, as that are the ones
+    // that will produce results in the flatbuffer.
+    if (funcOp && funcOp.isPublic()) {
```
mrakitaTT, line 163: Change to `if (!condition) return`, see more: https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

src/common/module_builder.cc
ajakovljevicTT: I did not know that walk is recursive by itself, thanks for noticing this! Fixed.

src/common/api_impl.cc
kmitrovicTT, line 1: FYI You are gonna hit a conflict after rebase since `submit` API has been updated.

ajakovljevicTT: Thanks for the heads up! Rebased successfully in order to check tests now.

src/common/module_builder.cc
```diff
@@ -65,10 +66,17 @@ ModuleBuilder::ModuleBuilder()
   m_context->appendDialectRegistry(registry);
 }
 
+bool ModuleBuilder::isOutputScalar(size_t index) const {
```
mrakitaTT, line 69: Missing `const` for `index`

mrakitaTT (approved):

Issue #93, opened 2024-11-29
Handle comparison of tensors with different types in test infra


mrakitaTT:
Our compiler currently doesn't support all data types that Jax(StableHLO) supports and it does internal conversion of types. That can lead to situations where output tensors from silicon and from golden are having different data types. We need to enable comparison of tensors with different types by doing explicit data type conversions before comparison. That should be done in one place, in test infra, to avoid having to do these conversions in each test that depends on it (for example `test_compare.py` tests that convert booleans to floats).

mrakitaTT:
Fixed by #100

PR #100, opened 2024-12-03
Improved test infrastructure.

### LINKED ISSUE #80: Add more explicit generation of test tensors regarding devices in the test infrastructure
Currently, we rely on flags such as `on_device` in `random_input_tensor` in `tests/infrastructure.py` and on default behaviour of jax regarding whether certain ops are ran on TT silicon or the cpu. This should be more explicit, as it is not immediately obvious where a certain op executes or the device on which the array lives. Some of the abstractions used in this effort might even be useful to third-party users of the tt-xla repository.

kmitrovicTT:
- Provided layer which uses jax to connect to various devices
- Provided layer which uses jax to run payload on various devices
- Provided infra which makes it easy to unit test either simple ops, or graphs or even full models

Fixes #80, #93.

### DIFF
--- a/src/common/api_impl.h
+++ b/src/common/api_impl.h
@@ -144,10 +144,11 @@ public:
   }
 
   std::string_view kind_string() { return kind_string_; }
-  std::string_view debug_string() { return debug_string_; }
-  std::string_view user_string() {
+  std::string_view debug_string() { return to_string(); }
+  std::string_view to_string() {
     std::stringstream ss;
-    ss << "TTDevice(id=" << device_id() << ")";
+    ss << kind_string_ << "(id=" << device_id() << ", arch=" << arch_string_
+       << ")";
     user_string_ = ss.str();
     return user_string_;
   }
@@ -160,8 +161,9 @@ public:
 
 private:
   int client_id_;
-  std::string kind_string_ = "wormhole";
-  std::string debug_string_ = "debug_string";
+  std::string kind_string_ = "TTDevice";
+  // TODO should not be hardcoded
+  std::string arch_string_ = "Wormhole";
   std::string user_string_ = "";
 };
 
--- a/tests/infra/device_connector.py
+++ b/tests/infra/device_connector.py
@@ -1,0 +1,110 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+from enum import Enum
+from typing import Sequence
+
+import jax
+import jax._src.xla_bridge as xb
+
+
+class DeviceType(Enum):
+    CPU = "cpu"
+    TT = "tt"
+    GPU = "gpu"
+
+
+class DeviceConnector:
+    """
+    Singleton class providing connections to devices on which jax commands will be
+    executed.
+
+    As a singleton it is instantiated only once, thus making sure that PJRT plugin is
+    registered exactly once. Registration needs to happen before any other jax commands
+    are executed. Registering it multiple times would cause error.
+
+    Do not instantiate this class. Use provided global instance.
+    """
+
+    _instance = None
+
+    def __new__(cls, *args, **kwargs):
+        # Ensure that only one instance of the class is created.
+        if cls._instance is None:
+            cls._instance = super().__new__(cls, *args, **kwargs)
+
+        return cls._instance
+
+    def __init__(
+        self, tt_pjrt_plugin_relpath: str = "build/src/tt/pjrt_plugin_tt.so"
+    ) -> None:
+        # We need to ensure __init__ body is executed once. It will be called each time
+        # `DeviceConnector()` is called.
+        if self._is_initialized():
+            return
+
+        self._initialized = False
+
+        plugin_path = os.path.join(os.getcwd(), tt_pjrt_plugin_relpath)
+        if not os.path.exists(plugin_path):
+            raise FileNotFoundError(
+                f"Could not find tt_pjrt C API plugin at {plugin_path}"
+            )
+
+        self._plugin_path = plugin_path
+        self._initialize_backend()
+
+    def connect_tt_device(self) -> jax.Device:
+        """Returns TTDevice handle."""
+        return self.connect_device(DeviceType.TT)
+
+    def connect_cpu(self) -> jax.Device:
+        """Returns CPUDevice handle."""
+        return self.connect_device(DeviceType.CPU)
+
+    def connect_gpu(self) -> jax.Device:
+        """Returns GPUDevice handle."""
+        return self.connect_device(DeviceType.GPU)
+
+    def connect_device(self, device_type: DeviceType) -> jax.Device:
+        """Returns handle for device identified by `device_type`."""
+        return jax.devices(device_type.value)[0]
+
+    def _supported_devices(self) -> Sequence[DeviceType]:
+        """Returns list of supported device types."""
+        # TODO support GPU
+        return [DeviceType.CPU, DeviceType.TT]
+
+    def _supported_devices_str(self) -> str:
+        """Returns comma separated list of supported devices as a string."""
+        # Note no space, only comma.
+        return ",".join([device.value for device in self._supported_devices()])
+
+    def _initialize_backend(self) -> None:
+        """
+        Registers TT plugin which will make TTDevice available in jax.
+
+        Needs to be called before any other jax command.
+        """
+        xb.register_plugin(
+            DeviceType.TT.value,
+            priority=500,
+            library_path=self._plugin_path,
+            options=None,
+        )
+        jax.config.update("jax_platforms", self._supported_devices_str())
+
+        self._initialized = True
+
+    def _is_initialized(self) -> bool:
+        """Checks if connector is already initialized."""
+        if hasattr(self, "_initialized") and self._initialized == True:
+            return True
+
+        return False
+
+
+# Global instance of DeviceConnector to be used from other modules.
+connector = DeviceConnector()
--- a/tests/infra/device_runner.py
+++ b/tests/infra/device_runner.py
@@ -1,0 +1,112 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Callable, Sequence
+
+import jax
+
+from .device_connector import DeviceType, connector
+from .test_module import TestModule
+
+
+class DeviceRunner:
+    @staticmethod
+    def run_on_tt_device(module: TestModule) -> jax.Array:
+        """Runs test module on TT device."""
+        return DeviceRunner._run_on_device(module, DeviceType.TT)
+
+    @staticmethod
+    def run_on_cpu(module: TestModule) -> jax.Array:
+        """Runs test module on CPU."""
+        return DeviceRunner._run_on_device(module, DeviceType.CPU)
+
+    @staticmethod
+    def run_on_gpu(module: TestModule) -> jax.Array:
+        """Runs test module on GPU."""
+        raise NotImplementedError("Support for GPUs not implemented")
+
+    @staticmethod
+    def put_on_tt_device(*tensors: jax.Array) -> Sequence[jax.Array]:
+        """Puts `tensors` on TT device."""
+        return DeviceRunner._put_on_device(tensors, DeviceType.TT)
+
+    @staticmethod
+    def put_on_cpu(*tensors: jax.Array) -> Sequence[jax.Array]:
+        """Puts `tensors` on CPU."""
+        return DeviceRunner._put_on_device(tensors, DeviceType.CPU)
+
+    @staticmethod
+    def put_on_gpu(*tensors: jax.Array) -> Sequence[jax.Array]:
+        """Puts `tensors` on GPU."""
+        raise NotImplementedError("Support for GPUs not implemented")
+
+    @staticmethod
+    def _run_on_device(
+        module: TestModule,
+        device_type: DeviceType,
+        raise_ex_if_jit_failed: bool = False,
+    ) -> jax.Array:
+        """
+        Runs test module on device.
+
+        If jitted graph fails to run, will try to run non-jitted graph if
+        `raise_ex_if_jit_failed` is False.
+        """
+        device = connector.connect_device(device_type)
+
+        # TODO is there a better way to check if function can be jitted than runtime fail?
+        try:
+            graph = module.get_jit_graph()
+            inputs = DeviceRunner._put_on_device(module.get_inputs(), device_type)
+
+            with jax.default_device(device):
+                return graph(*inputs)
+
+        except Exception as e:
+            if raise_ex_if_jit_failed:
+                raise e
+
+            with jax.default_device(device):
+                return module()
+
+    @staticmethod
+    def _put_on_device(
+        tensors: Sequence[jax.Array], device_type: DeviceType
+    ) -> Sequence[jax.Array]:
+        """Puts `tensors` on device identified by `device_type`."""
+        device = connector.connect_device(device_type)
+        return [jax.device_put(t, device) for t in tensors]
+
+
+# ----- Convenience decorators -----
+
+
+def run_on_tt_device(f: Callable):
+    """Runs any decorated function on TT device."""
+
+    def wrapper(*args, **kwargs):
+        module = TestModule(f, args=args, kwargs=kwargs)
+        return DeviceRunner.run_on_tt_device(module)
+
+    return wrapper
+
+
+def run_on_cpu(f: Callable):
+    """Runs any decorated function on CPU."""
+
+    def wrapper(*args, **kwargs):
+        module = TestModule(f, args=args, kwargs=kwargs)
+        return DeviceRunner.run_on_cpu(module)
+
+    return wrapper
+
+
+def run_on_gpu(f: Callable):
+    """Runs any decorated function on GPU."""
+
+    def wrapper(*args, **kwargs):
+        module = TestModule(f, args=args, kwargs=kwargs)
+        return DeviceRunner.run_on_gpu(module)
+
+    return wrapper
--- a/tests/infra/module_tester.py
+++ b/tests/infra/module_tester.py
@@ -1,0 +1,141 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from enum import Enum
+from typing import Callable, Sequence, Union
+
+import jax
+
+from .device_runner import DeviceRunner
+from .test_model import TestModel
+from .test_module import TestModule
+from .utils import (
+    compare_allclose,
+    compare_atol,
+    compare_equal,
+    compare_pcc,
+    random_tensor,
+)
+
+
+class TestType(Enum):
+    INFERENCE = "inference"
+    TRAINING = "training"
+
+
+class ComparisonMetric(Enum):
+    EQUAL = "equal"
+    PCC = "pcc"
+    ATOL = "atol"
+    ALLCLOSE = "allclose"
+
+
+class ModuleTester:
+    """
+    Class providing infrastructure for testing test modules.
+
+    Testing consists of comparing output results of running provided test module on
+    different devices. Frequently, run on CPU or GPU is taken as a ground truth
+    ("golden"), while custom device run is compared to it.
+
+    It supports testing inference and training modes.
+    """
+
+    def __init__(
+        self,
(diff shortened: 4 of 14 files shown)


tests/infra/module_tester.py
```diff
@@ -0,0 +1,141 @@
...
+    tester = ModuleTester(comparison_metric=comparison_metric)
+    module = (
+        f.as_test_module(inputs)
+        if isinstance(f, TestModel)
+        else TestModule(f, args=inputs)
+    )
+    return tester(module)
+
+
+def test(
```
kmitrovicTT, line 124: Prefixing public functions with `test` will cause trouble with `pytest -v tests`. If you have a better name for this and function below, please suggest @mrakitaTT @ajakovljevicTT .

kmitrovicTT: I also need you opinion on https://github.com/tenstorrent/tt-xla/pull/101/files#r1869054172.

tests/infra/device_connector.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def __new__(cls, *args, **kwargs):
+        # Ensure that only one instance of the class is created.
+        if cls._instance is None:
+            cls._instance = super().__new__(cls, *args, **kwargs)
+
+        return cls._instance
+
+    def __init__(
+        self, tt_pjrt_plugin_relpath: str = "build/src/tt/pjrt_plugin_tt.so"
```
ajakovljevicTT, line 41: Should the string be put in some global variable instead of just here plainly? Something like:
TT_PJRT_PLUGIN_PATH="build/src/tt/pjrt_plugin_tt.so"

tests/models/test_simple_nn.py
```diff
@@ -0,0 +1,71 @@
...
+from typing import Sequence
+
+import jax
+import jax.numpy as jnp
+import pytest
+from infra.module_tester import ComparisonMetric, test, test_with_random_inputs
+from infra.test_model import TestModel
+
+
+class SimpleNN(TestModel):  # TODO what's the benefit of inheriting nnx.Module?
```
kmitrovicTT, line 14: @mrakitaTT per [flax docs](https://flax.readthedocs.io/en/latest/api_reference/flax.nnx/module.html) I saw no reason to inherit from `nnx.Module`. It runs just fine as is now. Inheriting from `nnx.Module`, if I understood correctly, can mainly help with nested submodules and training mode. We have none of those for now.

This inheritance, if needed, could be provided through inheritance chain by letting `TestModel` inherit from `nnx.Module`, or concrete test model classes can use multiple inheritance since `TestModel` is only an interface class.

tests/infra/device_connector.py
```diff
@@ -0,0 +1,110 @@
...
+class DeviceConnector:
+    """
+    Singleton class providing connections to devices on which jax commands will be
+    executed.
+
+    As a singleton it is instantiated only once, thus making sure that PJRT plugin is
+    registered exactly once. Registration needs to happen before any other jax commands
+    are executed. Registering it multiple times would cause error.
+
+    Do not instantiate this class. Use provided global instance.
```
mrakitaTT, line 28: I think there are some established ways to create a singleton class in python which protect against this, and which avoid having a global variable for instance. Can you please explore those options?

tests/infra/device_runner.py
```diff
@@ -0,0 +1,112 @@
...
+    @staticmethod
+    def put_on_gpu(*tensors: jax.Array) -> Sequence[jax.Array]:
+        """Puts `tensors` on GPU."""
+        raise NotImplementedError("Support for GPUs not implemented")
+
+    @staticmethod
+    def _run_on_device(
+        module: TestModule,
+        device_type: DeviceType,
+        raise_ex_if_jit_failed: bool = False,
```
mrakitaTT, line 48: I think we always want to run jitted graph and to error out if it fails. I guess you wanted to make this method universal so we can use it outside of test module runs, for example in function to create random test inputs. I think the better option than is to create two methods, one which runs jitted graph, and one which doesn't, and then to use the other one explicitly for those specific cases outside of test module runs.

That will also remove the need for this try/except below.

tests/infra/module_tester.py
```diff
@@ -0,0 +1,141 @@
...
+
+        Call tester with passed test module and it will run tests on it.
+        """
+        return self._test(module)
+
+    def _test(self, module: TestModule) -> bool:
+        if self._test_type == TestType.INFERENCE:
+            return self._test_inference(module)
+        else:
+            return self._test_training(module)
```
mrakitaTT, line 65: This is a bit different for ops/graphs and for models. For ops/graphs we want to just run the module and verify results, i.e. inference only. For models we want to test both eval and training modes. We should discuss if we want to separate the op tester from model tester completely.

tests/infra/module_tester.py
mrakitaTT: Couple of thoughts here:
- We should not have a `.sh` file to run tests which we have to maintain, it is better if we can use plain `pytest` call to run tests. Therefore we should avoid using `test` prefix, and use something like `run_test`.
- It would be better to not even have global functions like this, and instead for this method to be inside the ModuleTester or ModelTester. I really like how tt-torch [ModelTester](https://github.com/tenstorrent/tt-torch/blob/main/tests/utils.py#L15) has separate functions to load model, compile model, run the model in eval/train modes, verify results, etc. Tester for ops does not need to have any of this, which is why I think we are shooting ourselves in the foot here by trying to make something universal to work for both ops and models testing. Let's talk more about this today.
- In the future tt-xla will be used for other frameworks too, so we should create a bit different structure for tests (not as part of this PR though):
```
tests/
  infra/
    (infra files)
  jax/
    ops/
    graphs/
    models/
    test_data_types.py
    ...
```

tests/infra/test_module.py
```diff
@@ -0,0 +1,54 @@
...
+        self._args = args
+        self._kwargs = kwargs if kwargs is not None else {}
+
+        self._inputs = tuple(self._args) + tuple(self._kwargs.values())
+
+    def get_inputs(self) -> Sequence[Any]:
+        return self._inputs
+
+    def get_jit_graph(self):
+        return jax.jit(self._f)
```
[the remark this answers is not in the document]
mrakitaTT: Yeah, I've just made a same remark in my previous comment, we should have separate functions to load model, compile model, run forward pass, run backward pass, test model, etc.

tests/models/test_simple_nn.py
```diff
@@ -0,0 +1,71 @@
...
+        b1 = jax.numpy.zeros(b1_shape)
+
+        return [act, w0, b0, w1, b1]
+
+
+def test_simple_nn():
+    model = SimpleNN.get_model()
+    inputs = SimpleNN.get_model_inputs()
+
+    assert test(model, inputs, ComparisonMetric.PCC)
```
mrakitaTT, line 53: This should all happen inside the ModelTester. Our task when writing tests should only be to implement the `load_model` and `load_inputs` and then when we run the tester it should automatically load the model, load the inputs, compile it, run it in eval/train modes, verify results.

tests/models/test_simple_nn.py
```diff
@@ -0,0 +1,71 @@
...
+        act_shape, w0_shape, b0_shape, w1_shape, b1_shape = (
+            (32, 784),
+            (784, 128),
+            (1, 128),
+            (128, 128),
+            (1, 128),
+        )
+
+        act = jax.numpy.ones(act_shape)
+        w0 = jax.numpy.ones(w0_shape)
```
mrakitaTT, line 41: Weights are not really model inputs. We should load weights into the model as part of the `load_model` func. Models from huggingface usually come with weights packed.

tests/models/test_simple_nn.py
mrakitaTT: We definitely don't want our ModelTester to inherit that, nor use ModuleTester as an ML module. ML module should be created and returned from the `load_model` function. That will usually happen by loading model from some repo like hugging face, but for the models that we manually make (for example by finding code on github or in ML paper) we should have a separate file implementing the model, and that class which implements the model might inherit from `nnx.Module` if we use flax to implement the model.

mrakitaTT (commented):
I've added some thoughts in comments, let's talk more today. Lot of these things will become much clearer when we start adding first models, so I wouldn't rush with pushing in these changes yet.

src/common/api_impl.h
```diff
@@ -160,8 +161,9 @@ class DeviceDescription {
 
 private:
   int client_id_;
-  std::string kind_string_ = "wormhole";
-  std::string debug_string_ = "debug_string";
+  std::string kind_string_ = "TTDevice";
+  // TODO should not be hardcoded
+  std::string arch_string_ = "Wormhole";
```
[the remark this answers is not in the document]
kmitrovicTT: It's not a change that impacts anything in pjrt plugin, it was set to dummy value anyway. When you print `jax.Array.device` we were getting `debug`. Now we get `TTDevice(Wormhole)`

tests/jax/ops/test_add.py
```diff
@@ -0,0 +1,22 @@
...
+
+def add(x: jax.Array, y: jax.Array) -> jax.Array:
+    return jax.numpy.add(x, y)
+
+
+@pytest.mark.parametrize(
+    ["x_shape", "y_shape"],
+    [
+        [(32, 32), (32, 32)],
+        # [(64, 64), (64, 64)],
```
mrakitaTT, line 18: Remove dead code

tests/jax/models/test_simple_nn.py
```diff
@@ -0,0 +1,71 @@
...
+    @staticmethod
+    def _get_model_forward_pass_method_name() -> str:
+        return "__call__"
+
+
+@pytest.fixture
+def comparison_config() -> ComparisonConfig:
+    config = ComparisonConfig()
+    config.disable_all()
+    config.pcc.enable()
```
mrakitaTT, line 61: Are model tests failing other metrics?

tests/jax/models/test_simple_nn.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # @override
+    @staticmethod
+    def _get_model_as_workload() -> Workload:
+        return Workload(SimpleNNTester._get_model(), SimpleNNTester._get_model_inputs())
+
+    # @override
+    @staticmethod
+    def _get_model_forward_pass_method_name() -> str:
+        return "__call__"
```
mrakitaTT, line 54: Do we need to override these? Seems like this will be a default behavior. `_get_model_forward_pass_method_name` already is, and `_get_model_as_workload` is unimplemented in the base but should probably be implemented in this way.

tests/jax/models/test_simple_nn.py
```diff
@@ -0,0 +1,71 @@
...
+            (1, 128),
+        )
+
+        act = jax.numpy.ones(act_shape)
+        w0 = jax.numpy.ones(w0_shape)
+        b0 = jax.numpy.ones(b0_shape)
+        w1 = jax.numpy.ones(w1_shape)
+        b1 = jax.numpy.zeros(b1_shape)
+
+        return [act, w0, b0, w1, b1]
```
mrakitaTT, line 44: This should return only `act = jax.numpy.ones(act_shape)` and it should not be called `act` but something like `input_features`. Inside `SimpleNN` we should "load weights" by constructing the `w0, b0, w1, b1` arrays.

tests/infra/utils.py
```diff
@@ -0,0 +1,74 @@
...
+from typing import Any, Callable, Mapping, Optional, Sequence, Union
+
+import jax
+import jax.numpy as jnp
+from flax import linen, nnx
+
+
+@dataclass
+class Workload:
+    f: Callable
```
mrakitaTT, line 16: Can we have a bit more descriptive name than `f`? :)

tests/infra/base_model_tester.py
```diff
@@ -0,0 +1,117 @@
...
+
+from .base_tester import BaseTester
+from .comparison import ComparisonConfig
+from .device_runner import DeviceRunner
+from .utils import Model, Workload
+
+
+class TestType(Enum):
+    INFERENCE = "inference"
+    TRAINING = "training"
```
mrakitaTT, line 22: We should be testing both inference and training, so we shouldn't have this enum. Training tests should just `pass` for now with a TODO in this base class to implement them later.

tests/infra/base_model_tester.py
```diff
@@ -0,0 +1,117 @@
...
+        self,
+        comparison_config: ComparisonConfig,
+        test_type: TestType = TestType.INFERENCE,
+    ) -> None:
+        super().__init__(comparison_config)
+        self._test_type = test_type
+
+    @staticmethod
+    @abstractmethod
+    def _get_model_as_workload() -> Workload:
```
mrakitaTT, line 47: We should still have `get_model` and `load_model_inputs`, they should just be wrapped into workload inside this function, and this function shouldn't have to be overrided in every test, instead tests will just override and implement those former two functions.

tests/jax/models/test_simple_nn.py
kmitrovicTT: Atol is 64.  PCC and allclose pass.

kmitrovicTT: Remember that I already ran into a similar situation https://github.com/tenstorrent/tt-metal/issues/15130

mrakitaTT: That's wild that atol is failing with such a high difference, while allclose passes. Allclose is calculated as:
```
absolute(a - b) <= (atol + rtol * absolute(b))
```

which tells me that the only way for this to happen is either:
1. Output activations get insanely high (>=6400) so that `64 <= 0.01 + 0.01 * activation`
2. We are running comparison functions on device instead of cpu, which might cause incorrect calculations

Looking at the way you initialized `ExampleModel` weights, it seems like the 1st case is the root cause. Inputs are randomly initialized in `[0, 1]` range, first matmul gives `784*[0, 1] + 1 = [1, 785]`. Second matmul gives `128 * [1, 785] + 0 = [128, 100480]`. You should initialize weights with random values in some smaller range like `[-0.01, 0.01]`, biases too. Then you will be able to enable atol comparison too.

Additionally, I think we should make sure that the comparison functions are running on CPU, I'll add a separate comment for that in `BaseTester`.

tests/jax/models/flax_distil_bert_for_masked_lm/test_flax_distil_bert_for_masked_lm.py
```diff
@@ -0,0 +1,87 @@
...
+    # @override
+    @staticmethod
+    def _get_input_activations() -> Sequence[jax.Array]:
+        tokenizer = AutoTokenizer.from_pretrained(MODEL)
+        inputs = tokenizer("Hello [MASK].", return_tensors="np")
+        return [inputs["input_ids"]]
+
+    # @override
+    def _get_forward_method_kwargs(self) -> Dict[str, jax.Array]:
+        activations = self._get_input_activations()
```
mrakitaTT, line 40: Rename variable to `inputs_activations` or `inputs`

tests/infra/model_tester.py
```diff
@@ -0,0 +1,177 @@
...
+    _get_forward_method_name() -> str # Optional, has default behaviour.
+    # One of or both:
+    _get_forward_method_args(self) -> Sequence[Any] # Optional, has default behaviour.
+    _get_forward_method_kwargs(self) -> Mapping[str, Any] # Optional, has default behaviour.
+    ```
+    """
+
+    def __init__(
+        self,
+        comparison_config: ComparisonConfig,
```
mrakitaTT, line 42: allow default initialization of this ` = ComparisonConfig(),`

tests/infra/base_tester.py
```diff
@@ -0,0 +1,72 @@
...
+        if self._comparison_config.equal.enabled:
+            compare_equal(device_output, golden_output)
+        if self._comparison_config.atol.enabled:
+            compare_atol(device_output, golden_output, self._comparison_config.atol)
+        if self._comparison_config.pcc.enabled:
+            compare_pcc(device_output, golden_output, self._comparison_config.pcc)
+        if self._comparison_config.allclose.enabled:
+            compare_allclose(
+                device_output, golden_output, self._comparison_config.allclose
+            )
```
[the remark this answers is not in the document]
kmitrovicTT: It will run on CPU due to both tensors being placed on CPU two lines above

```python
        device_output, golden_output = DeviceRunner.put_tensors_on_cpu(
            device_out, golden_out
        )
```

but, to make this more explicit, I can define

```python
def run_on_cpu(f: Callable):
    def wrapper(*args, **kwargs):
        workload = Workload(f, args, kwargs)
        return DeviceRunner.run_on_cpu(workload)

    return wrapper
```

and decorate every `compare` function with it. 

Wdy think?

src/common/api_impl.h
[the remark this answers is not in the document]
kmitrovicTT: Oh, I see.

This is not that deep, this is just vaguely defined test. If I change it to

```python
def test_to_device():
    cpu_array = random_input_tensor((32, 32))
    device = jax.devices("tt")[0]  # explicitly use "tt" backend
    tt_array = jax.device_put(cpu_array, device)
    assert tt_array.device.device_kind == "TTDevice"  ##  "TTDevice" instead of "wormhole"
```

it passes. 

Due to change in `tests/conftest.py` I had to make, _order_ of devices registered with jax is now `[cpu, tt]` and previously it was `[tt, cpu]`. I am saying that because doing `jax.devices()` with no arg will return the firstly registered one, which shouldn't be relied upon. We should explicitly say which backend.

String change is not that big of a deal. I will still need to change the `jax.devices("tt")[0] ` part, regardless of if I revert PJRT changes or not. 

Do you still feel like I should revert it?

tests/infra/utils.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def as_mlir_module(self) -> str:
+        """
+        Returns workload as mlir module string.
+
+        Note that workload.executable must be the result of jit, otherwise empty string
+        will be returned.
+        """
+        try:
+            s = export.export(self.executable)(*self.args, **self.kwargs).mlir_module()
```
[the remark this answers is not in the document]
kmitrovicTT: Yeah, you are right, it actually bothered me that I put jax specific stuff in Workload.

I redefined it as a util function down below

```python
def workload_as_mlir_module(workload: Workload, framework: Framework = Framework.JAX) -> str
```

similar to what I did with `random_tensor`.

tests/jax/models/test_simple_nn.py
kmitrovicTT: You are right.

Expanded `random_tensor` util function with `minval` and `maxval` and redefined biases and weights in `ExampleModel` as

```python
self.w0 = random_tensor(w0_shape, minval=-0.01, maxval=0.01)
self.w1 = random_tensor(w1_shape, minval=-0.01, maxval=0.01)
self.b0 = random_tensor(b0_shape, minval=-0.01, maxval=0.01)
self.b1 = random_tensor(b1_shape, minval=-0.01, maxval=0.01)
```

Now it passes everything. Removed `ComparisonConfig` fixture since it is unneeded.

src/common/api_impl.h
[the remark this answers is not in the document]
mrakitaTT: I see, you are right it makes sense to update the test and leave the changes. I would still open an issue for this:
> We should understand better how these are used for, [here](https://github.com/openxla/xla/blob/2da0f3ed4ba9cb4159f9f2bd6f20b32c71ed6b97/xla/pjrt/pjrt_device_description.h#L67) is how XLA defines these.

and add a TODO comment above your changes in `api_impl.h` to rethink those, with a link to that issue.

tests/infra/base_tester.py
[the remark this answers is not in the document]
mrakitaTT: Better to be explicit, I know Andrej had some issues where things were running on TT device even though inputs are in host memory. I like the decorator idea 👍

tests/jax/models/example_model/model.py
```diff
@@ -16,16 +17,17 @@ def __init__(self) -> None:
...
+        self.w1 = random_tensor(w1_shape, minval=-0.01, maxval=0.01)
+        self.b0 = random_tensor(b0_shape, minval=-0.01, maxval=0.01)
+        self.b1 = random_tensor(b1_shape, minval=-0.01, maxval=0.01)
 
     def __call__(
         self, act: jax.Array, w0: jax.Array, b0: jax.Array, w1: jax.Array, b1: jax.Array
     ) -> jax.Array:
         # Note how activations, weights and biases are directly passed to the forward
-        # method. `self` is not accessed.
+        # method as inputs, `self`` is not accessed. Otherwise they would be embedded
```
mrakitaTT, line 29: Nitpick typo, two ` after self (wouldn't leave this comment if there weren't other unresolved comments)

PR #106, opened 2024-12-07
Add PJRT plugin attributes for StableHLO version

### LINKED ISSUE #97: Add PJRT attribute for StableHLO version to tt-xla plugin
Hello- I was going to try building and contributing but hit some setup issues and ran out of time, so figure I'll at least give some code pointers to what I intended to add, should be fairly trivial.

If a PJRT plugin has an attribute for `stablehlo_current_version`, then JAX will precisely downgrade the IR to the plugin's version. Without the attribute, JAX uses 12w IR downgrade, meaning newer features can't be used or integrated for several months, similarly an older than 12w plugin will have more stability if it lets JAX know it's precise version so it downgrades more than 12w - note support >12w isn't guaranteed by JAX but historically has been fairly stable.

To add attr support we'll need an impl for `PJRT_Plugin_Attributes`:
https://github.com/tenstorrent/tt-xla/blob/d770c943987cf251f5f239249d55fad76dc5ba34/src/common/api_impl.cc#L1098

A template for specifying these attrs can be found in:
https://github.com/openxla/xla/blob/4220451d1582817f80dac19a4760374b666cbce5/xla/pjrt/c/pjrt_c_api_wrapper_impl.cc#L353

Implementation where the attrs are held statically so they can be safely accessed by frameworks here, the `<0>/<1>` is used to ensure that each string has a unique storage:
https://github.com/openxla/xla/blob/4220451d1582817f80dac19a4760374b666cbce5/xla/pjrt/c/pjrt_c_api_helpers.cc#L637

The `stablehlo_current_version` impl is commented out for a bug currently, but once JAX makes a new release (this week or next) that is the impl that should be used, so perhaps hold off on adding this until there is a >=0.4.36 release on https://github.com/jax-ml/jax/releases

mrakitaTT:
Fixes #97

### DIFF
--- a/src/common/api_impl.cc
+++ b/src/common/api_impl.cc
@@ -10,6 +10,7 @@
 
 #include "common/api_impl.h"
 
+// c++ standard library includes
 #include <cassert>
 #include <cstring>
 #include <iostream>
@@ -17,7 +18,9 @@
 #include <sstream>
 #include <utility>
 
+// tt-xla includes
 #include "common/module_builder.h"
+#include "common/plugin_attributes.h"
 #include "common/status.h"
 
 namespace tt::pjrt {
@@ -1095,12 +1098,7 @@ void BindMonomorphicApi(PJRT_Api *api) {
     return nullptr;
   };
 
-  api->PJRT_Plugin_Attributes =
-      +[](PJRT_Plugin_Attributes_Args *args) -> PJRT_Error * {
-    DLOG_F(LOG_DEBUG, "PJRT_Plugin_Attributes");
-    args->num_attributes = 0;
-    return nullptr;
-  };
+  api->PJRT_Plugin_Attributes = InitializePluginAttributes;
 
   // Bind by object types.
   BufferInstance::BindApi(api);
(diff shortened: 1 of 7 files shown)


GleasonK (commented):
LGTM! Reviewed from phone but names and data types look correct, and use of deserialize pipeline looks right.

src/common/api_impl.cc
```diff
@@ -1095,12 +1098,7 @@ void BindMonomorphicApi(PJRT_Api *api) {
     return nullptr;
   };
 
-  api->PJRT_Plugin_Attributes =
-      +[](PJRT_Plugin_Attributes_Args *args) -> PJRT_Error * {
-    DLOG_F(LOG_DEBUG, "PJRT_Plugin_Attributes");
-    args->num_attributes = 0;
-    return nullptr;
-  };
+  api->PJRT_Plugin_Attributes = InitializePluginAttributes;
```
ajakovljevicTT, line 1101: Should we move this to the `PluginAttributes::BindApi(api)` function, to be in line with the other bindings? We could do the same for `PJRT_Plugin_Initialize`

mrakitaTT: That's a good question, I wasn't sure if we want to couple API bindings with these classes as it was currently done so I've deferred that decision for later. Could you please add this to the list you are compiling, of things to consider for API implementation changes (to rethink where to do API bindings)?

ajakovljevicTT: Okay, I will do that!

PR #116, opened 2024-12-11
Refactoring api_impl.cc and api_impl.h code

ajakovljevicTT:
The `tt-xla` repository currently contains particularly large files `api_impl.cc` and `api_impl.h` that handle multiple functionalities through different classes. This issue is for refactoring `api_impl.cc` and `api_impl.h`. Splitting this file into smaller, logically organized files, with each file containing a single class or closely related classes. Other than that, no significant changed to the logic or features of the specific functions.


src/tt/client.h
```diff
@@ -10,7 +10,7 @@
 #ifndef IREE_PJRT_PLUGIN_PJRT_CPU_CLIENT_H_
 #define IREE_PJRT_PLUGIN_PJRT_CPU_CLIENT_H_
 
-#include "common/api_impl.h"
+#include "common/pjrt_implementation/client_instance.h"
```
AleksKnezevic, line 13: Thanks for doing this! Since we only have one client, does it make sense to eliminate the tt and common directories and just have everything live in pjrt_implementation?

mrakitaTT: Thank you for suggestion Aleks, it makes sense for sure, we plan to change the folder structure and reorganize the code further, this PR is just one step in this process - to split the `api_impl.h/cc` into multiple files so we can read and understand the code easier. We will make further changes in some of the next PRs.

kmitrovicTT:
Is this a good moment to also do some stylistic changes, like renaming functions and attributes to either use `camelCaseStyle` or `underscore_style`, spacing them out properly and leaving comments in headers, for example instead of

```cpp
  static void BindApi(PJRT_Api *api);

  // iree_hal_buffer_view_t* buffer_view() { return buffer_view_.get(); }
  DeviceInstance &device() { return device_; }
  tt_pjrt_status AsyncDeallocate();
  tt_pjrt_status Delete();
```

doing 

```cpp
  // Comment
  static void bindApi(PJRT_Api *api);

  // Comment
  DeviceInstance &device() { return device_; }

  // Comment
  tt_pjrt_status asyncDeallocate();

  // Comment
  tt_pjrt_status delete();
```

and so on? Nothing that changes functionality, just coding style.

inc/common/pjrt_implementation/buffer_instance.h
```diff
@@ -0,0 +1,81 @@
+// SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+#include "common/pjrt_implementation/event_instance.h"
+#include "common/status.h"
+#include "tt/runtime/runtime.h"
+#include "xla/pjrt/c/pjrt_c_api.h"
```
mrakitaTT, line 7: I'd expect that lot more headers need to include this, `bindings.h` for example. Let's be explicit about includes (which we are not forwarding).

Also, this is a third party include so it needs to be in a separate group.

inc/common/pjrt_implementation/executable_image.h
```diff
@@ -0,0 +1,68 @@
+// SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// This file incorporates work covered by the following copyright and permission
+// notice:
+// SPDX-FileCopyrightText: Copyright 2023 The IREE Authors
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// https://llvm.org/LICENSE.txt
```
mrakitaTT, line 9: Why is only this header having IREE copyright and others don't?

src/common/CMakeLists.txt
```diff
@@ -29,6 +38,7 @@ target_include_directories(TTPJRTCommon PUBLIC
     ${PROJECT_SOURCE_DIR}/third_party/tt-mlir/src/tt-mlir/include
     ${PROJECT_SOURCE_DIR}/third_party/tt-mlir/src/tt-mlir/runtime/include
     ${PROJECT_SOURCE_DIR}/third_party/tt-mlir/src/tt-mlir-build/stablehlo/
+    ${PROJECT_SOURCE_DIR}/inc
```
mrakitaTT, line 41: Move to top so include folders are in alphabetical order

src/common/pjrt_implementation/bindings.cc
mrakitaTT, line 1: .cc files also need headers

src/common/pjrt_implementation/bindings.cc
```diff
@@ -0,0 +1,45 @@
+#include "common/pjrt_implementation/bindings.h"
```
mrakitaTT, line 1: Separate main header of a .cc file with a blank line from other headers (in other files too)

src/common/pjrt_implementation/buffer_instance.cc
```diff
@@ -0,0 +1,215 @@
+
+
+#include "common/pjrt_implementation/buffer_instance.h"
+#include "common/pjrt_implementation/device_instance.h"
+#include "common/pjrt_implementation/utils.h"
+#include "xla/pjrt/c/pjrt_c_api.h"
```
mrakitaTT, line 6: This include can probably be removed from all .cc files

mrakitaTT:
> Is this a good moment to also do some stylistic changes, like renaming functions and attributes to either use `camelCaseStyle` or `underscore_style`, spacing them out properly and leaving comments in headers

@kmitrovicTT We will do it in some of the following PRs, class by class, as we get better understanding of them and how we want to change them.

inc/common/pjrt_implementation/executable_image.h
ajakovljevicTT: My mistake, added IREE copyright to the other files.

inc/common/pjrt_implementation/api_bindings.h
```diff
@@ -0,0 +1,58 @@
...
+
+namespace tt::pjrt {
+
+// Top-level API bindings.
+void BindMonomorphicApi(PJRT_Api *api);
+
+static void BindUndefineds(PJRT_Api *api);
+
+template <typename PlatformTy, typename ClientInstanceTy>
+static void BindApi(PJRT_Api *api) {
```
mrakitaTT, line 27: I think you can remove `static` for these functions

inc/common/pjrt_implementation/buffer_instance.h
mrakitaTT: You left this one unresolved, maybe my comment was a bit unclear. I was referring to `xla/pjrt/c/pjrt_c_api.h` include. It is a third party include, so it should be in a separate group, similar to `tt/runtime/runtime.h`. Structures from that file are used in most of the pjrt_imlementation .h files so you should explicitly include it in those files instead of relying on transitive includes. For example `api_bindings.h` is using `PJRT_Api` which is defined in `xla/pjrt/c/pjrt_c_api.h`, which you are not including there, but your build is passing because of transitive include through `common/pjrt_implementation/error_instance.h`.

src/common/pjrt_implementation/utils.cc
```diff
@@ -0,0 +1,59 @@
+#include "common/pjrt_implementation/utils.h"
```
mrakitaTT, line 1: Copyright header missing here, not sure how pre-commit spdx checks are not catching this

src/common/pjrt_implementation/executable_image.cc
```diff
@@ -0,0 +1,128 @@
...
+  };
+  api->PJRT_Executable_OutputMemoryKinds =
+      +[](PJRT_Executable_OutputMemoryKinds_Args *args) -> PJRT_Error * {
+    DLOG_F(LOG_DEBUG, "ExecutableImage::PJRT_Executable_OutputMemoryKinds");
+    return ErrorInstance::MakeError(tt_pjrt_status::kUnimplemented);
+  };
+}
+
+//===----------------------------------------------------------------------===//
+// Top-level API binding.
```
mrakitaTT, line 125: Seems like a stray comment left from the original `api_impl.cc`, probably meant to be in `api_bindings.h`

src/common/pjrt_implementation/buffer_instance.cc
```diff
@@ -0,0 +1,225 @@
+
+
```
mrakitaTT, line 2: Remove blank lines

mrakitaTT (approved):

src/common/pjrt_implementation/utils.cc
mrakitaTT: Can you please run pre-commit checks locally before fixing this one, to see if it works correctly? Maybe it doesn't work for .cc extension. If it doesn't work, please open a new issue for this.

mrakitaTT: On more suspect thing here is that you've changed `(c)` to `©` in these headers, but `.github/check-spdx.yaml` expects `(c)`, and you didn't get errors.

ajakovljevicTT: In talks with @vmilosevic, he pointed out that the (c)/© distinction and the specific header is only important when `spdx` is making its own licence header. Otherwise, it just checks if one exists. Nevertheless, I will change the character and add the licence header where needed.

PR #124, opened 2024-12-18
Add workflow concurrency check

vmilosevic:
Add concurrency check, new workflows triggered on the same branch should cancel old ones in progress.
https://github.com/tenstorrent/tt-mlir/issues/1619

### DIFF
--- a/.github/workflows/on-pr.yml
+++ b/.github/workflows/on-pr.yml
@@ -5,6 +5,10 @@ on:
   pull_request:
     branches: [ "main" ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   pre-commit:
     uses: ./.github/workflows/pre-commit.yml


kmitrovicTT (approved):

.github/workflows/on-pr.yml
```diff
@@ -5,6 +5,10 @@ on:
   pull_request:
     branches: [ "main" ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
mrakitaTT, line 10: There might be cases when someone wants the old workflow to finish in order to compare results with the new commit, but I guess we can manually rerun the workflow in GH Actions tab?

mrakitaTT (approved):
Nice, I love this change! I used to manually cancel the old workflows when I push new commits, glad we could automate it :)

.github/workflows/on-pr.yml
vmilosevic: Yes, I guess in that case he would need to wait for the workflow to finish before pushing the next change.
Also, you can rerun canceled workflows, and start only the canceled tasks to resume.

PR #129, opened 2024-12-23
Add the initial batch of JAX model tests

sgligorijevicTT:


tests/jax/models/Albertv2/base/test_albert_base.py
```diff
@@ -0,0 +1,77 @@
...
+import jax
+import pytest
+from flax import linen as nn
+from infra import ComparisonConfig, ModelTester, RunMode
+from transformers import AutoTokenizer, FlaxAlbertForMaskedLM
+
+MODEL = "albert/albert-base-v2"
+
+
+class FlaxAlbertForMaskedLMTester(ModelTester):
```
kmitrovicTT, line 16: I would rename all of these classes to match the `MODEL`. For example, I'd rename this one to `AlbertBaseTester`. Similar for others.

tests/jax/models/Albertv2/base/test_albert_base.py
```diff
@@ -0,0 +1,77 @@
...
+    """Tester for Albert model with a `language modeling` head on top."""
+
+    # @override
+    @staticmethod
+    def _get_model() -> nn.Module:
+        return FlaxAlbertForMaskedLM.from_pretrained(MODEL)
+
+   # @override
```
kmitrovicTT, line 24: This comment is misaligned.

tests/infra/device_runner.py
```diff
@@ -116,9 +116,12 @@ def _safely_put_workload_on_device(
...
         for key, value in workload.kwargs.items():
-            try:
-                value_on_device = jax.device_put(value, device)
-            except:
+            if (isinstance(value, dict) or isinstance(value, list) or isinstance(value, tuple) or isinstance(value, jax.Array)):
+                try:
+                    value_on_device = jax.device_put(value, device)
+                except:
+                    value_on_device = value
+            else:
```
kmitrovicTT, line 124: Why do this when `jax.device_put` will raise an exception if it cannot handle `value`? That was the whole point of putting it in a `try-except` block.

tests/infra/base_tester.py
```diff
@@ -32,10 +32,12 @@ def __init__(
     ) -> None:
         self._comparison_config = comparison_config
 
-    @staticmethod
-    def _compile(executable: Callable) -> Callable:
+    def _compile(self, executable: Callable) -> Callable:
         """Sets up `executable` for just-in-time compile."""
-        return jax.jit(executable)
+        return jax.jit(executable, static_argnames=self._get_static_argnames())
```
kmitrovicTT, line 37: These `static args` are rather interesting, leaving docs [here](https://jax.readthedocs.io/en/latest/_autosummary/jax.jit.html) for reference.

We should not be providing `_get_static_argnames` in this base class for others to overload, mainly because overload mechanism is only used in model tests for now, ops and graphs don't need it.

What I suggest is to expand `_compile` method with additonal `static_argnames: Sequence[str]` parameter. Move the definition of `_get_static_argname` to `ModelTester`. Store them in some private variable like we did with others in `_init_model_hooks` and then pass them to `super()._compile`. 

I currently don't see a need for op and graph tests to use this feature, but if needed, we can easily expand `Workload` by one additional argument which will help us carry it wherever needed. Let's not do that for now, what I described above should suffice.

tests/infra/base_tester.py
```diff
@@ -32,10 +32,12 @@ def __init__(
...
+
+    def _get_static_argnames(self) -> Sequence[str]:
+        return []
```
kmitrovicTT, line 40: As said above, move this to `ModelTester` and provide a docstring by pulling out a sentence or two from `jax.jit` [docs](https://jax.readthedocs.io/en/latest/_autosummary/jax.jit.html).

tests/jax/models/Albertv2/base/test_albert_base.py
```diff
@@ -0,0 +1,77 @@
...
+    
+    # @override
+    def _get_forward_method_args(self):
+        return self._get_input_activations()
+
+    # @override
+    def _get_forward_method_kwargs(self) -> Dict[str, jax.Array]:
+        assert hasattr(self._model, "params")
+        return {"params": self._model.params}
+
```
kmitrovicTT, line 44: There really is no need to always override both of these methods, as I explained in `tests/jax/models/example_model/mixed_args_and_kwargs/test_example_model_mixed_args_and_kwargs.py`.

If there is no `*` or any other mechanism in model's forward method signature that forces us to call this model as `model(input_activations, ..., params=params)`, that is to pass `input_activations` as args and `params` as kwargs, overriding only  `_get_forward_method_kwargs` (since you don't have to skip any positional args that might occur between args you are trying to pass) will be sufficient.

That being said, I'd suggest implementing only `_get_forward_method_kwargs` in all models in this PR.

tests/jax/models/Distilbert/test_distilbert.py
```diff
@@ -34,15 +34,15 @@ def _get_input_activations() -> Sequence[jax.Array]:
...
 
     # @override
     def _get_forward_method_kwargs(self) -> Dict[str, jax.Array]:
-        input_activations = self._get_input_activations()
-
-        assert len(input_activations) == 1
         assert hasattr(self._model, "params")
-
-        return {"input_ids": input_activations[0], "params": self._model.params}
+        return {"params": self._model.params}
```
kmitrovicTT, line 45: As described in a comment above, this change won't be necessary. 

However, I am fine removing unused `ComparisonConfig` and renaming this class.

tests/jax/models/MNIST_CNN/test_mnistcnn.py
```diff
@@ -0,0 +1,89 @@
...
+    
+    # @override
+    def _get_forward_method_args(self):
+        inp = self._get_input_activations()
+        init_model = self._model.init(jax.random.PRNGKey(42), inp, train=False)
+        return [init_model, inp]
+
+    # @override
+    def _get_forward_method_kwargs(self) -> Dict[str, jax.Array]:
+        return {"train": False}
```
kmitrovicTT, line 44: This is a perfect example where we _need_ to override both `_get_forward_method_args` and `_get_forward_method_kwargs` due to the way you defined `CNN.__call__`.

tests/jax/models/MNIST_CNN/CNN.py
```diff
@@ -0,0 +1,29 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import jax
+from flax import linen as nn
+
+class CNN(nn.Module):
+    @nn.compact
+    def __call__(self, x, *, train: bool):
```
kmitrovicTT, line 10: This `*` might be completely unneeded here (not sure if inheriting `nn.Module` or `@nn.compact` decorator force it though), but leave it as is since it provides a perfect example of a tester which uses every part of override-interface.

tests/jax/models/MNIST_CNN/test_mnistcnn.py
```diff
@@ -0,0 +1,89 @@
...
+    
+    # @override
+    def _get_static_argnames(self):
+        return ['train']
```
kmitrovicTT, line 48: Also, this is a good example of having `train` as a kwarg and also a static arg used in `jax.jit`. Good job!

tests/jax/models/Roberta/test_roberta.py
```diff
@@ -0,0 +1,91 @@
...
+
+# ----- Fixtures -----
+
+
+@pytest.fixture
+def comparison_config() -> ComparisonConfig:
+    config = ComparisonConfig()
+    config.disable_all()
+    config.pcc.enable()
+    return config
```
kmitrovicTT, line 60: It is ok if some model passes only one metric, it can happen. But since you skip both tests down below (for now) you can remove this completely since we can't know if it passes all other metrics as well.

tests/jax/models/Albertv2/base/test_albert_base.py
kmitrovicTT: Or, to be 1:1 match with huggingface model name, I'd rename it to `AlbertBaseV2Tester`.

tests/jax/models/Albertv2/base/test_albert_base.py
mrakitaTT, line 1: Albert has v1 and v2, so we could have a folder hierarchy like:
```
Albert/
    v1/
        base/
        large/
        ...
    v2/
        base/
        large/
        ...
```

tests/infra/base_tester.py
mrakitaTT: Let's have a brief talk about this tomorrow

tests/jax/models/Albertv2/base/test_albert_base.py
```diff
@@ -0,0 +1,77 @@
...
+   # @override
+    @staticmethod
+    def _get_forward_method_name() -> str:
+        return "__call__"
+
+    # @override
+    @staticmethod
+    def _get_input_activations() -> Sequence[jax.Array]:
+        tokenizer = AutoTokenizer.from_pretrained(MODEL)
+        inputs = tokenizer("Hello [MASK].", return_tensors="np")
```
mrakitaTT, line 33: Reminder to talk about masks

tests/jax/models/Albertv2/xxlarge/test_albert_xxlarge.py
```diff
@@ -0,0 +1,77 @@
...
+import jax
+import pytest
+from flax import linen as nn
+from infra import ComparisonConfig, ModelTester, RunMode
+from transformers import AutoTokenizer, FlaxAlbertForMaskedLM
+
+MODEL = "albert/albert-xxlarge-v2"
+
+
+class FlaxAlbertForMaskedLMTester(ModelTester):
```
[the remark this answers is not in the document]
kmitrovicTT: Take note that in order to handle relative imports, which makes sense within one model's test, you'd have to add `__init__.py` files, depending on the structure of your test.

Take a look how `tests/jax/models/example_model` does it, where all testers have the same model.

tests/jax/models/Roberta/test_roberta.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # @override
+    @staticmethod
+    def _get_model() -> nn.Module:
+        return FlaxRobertaForMaskedLM.from_pretrained(MODEL)
+
+    # @override
+    @staticmethod
+    def _get_forward_method_name() -> str:
+        return "__call__"
```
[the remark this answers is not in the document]
kmitrovicTT: Yeah, I left these in examples I provided intentionally just to show everything that should be overridden if default implementation does not suffice.

tests/jax/models/MNIST_CNN/test_mnistcnn.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # @override
+    @staticmethod
+    def _get_input_activations() -> Sequence[jax.Array]:
+        img = jnp.ones((28, 28, 1))
+        return img
+    
+    # @override
+    def _get_forward_method_args(self):
+        inp = self._get_input_activations()
```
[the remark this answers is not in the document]
kmitrovicTT: I left it like that to make it super clear that that method should only be concerned with forming input activations, maybe by fetching them from somewhere, or making raw tensors, images, whatever.

We cannot know (although it most certainly will be the first one) which argument to model's forward method is input activations, nor its name. For example in `FlaxDistilBertPreTrainedModel`  input activations are called `input_ids` which is not super intuitive for me, although it is the first arg in `__call__`.  

Thus I left this to be as flexible as possible. 

We could maybe make default behaviour of `_get_forward_method_args` be

```python
def _get_forward_method_args(self) -> Sequence[Any]:
    return [self._get_input_activations()]    # instead of []
```

if that makes any sense. 

But you can't do something like this for `_get_forward_method_kwargs` since you never know the name of the arg refering to input activations, as I showed above. Thus I'd leave it as is since it enforces you to be explicit and carefully respect model's call interface.

tests/jax/models/MNIST_CNN/test_mnistcnn.py
```diff
@@ -0,0 +1,89 @@
...
+        init_model = self._model.init(jax.random.PRNGKey(42), inp, train=False)
```
kmitrovicTT, line 39: I am confused by this line. Can you explain a bit?

tests/jax/models/GPT2/test_gpt2.py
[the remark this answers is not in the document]
sgligorijevicTT: That is unlikely, I don't think OpenAI ever release GPT 1 weights, and they certainly won't release GPT3 or GPT4 weights.
Besides, GPT2 specifically is important enough to warrant being treated as a distinct model, and not a specific version within a broader series. However, I will refactor the structure to account for different sizes within the GPT2 series.

tests/jax/models/Albertv2/base/test_albert_base.py
sgligorijevicTT: I have refactored Albert tester classes to make them useable for multiple tests, now adding the exact model name as part of the class name no longer makes sense.
Additionally, we might have to have multiple tester classes for the same family of models depending on how they're used(masked language modeling vs sentiment analysis), so it makes sense to mirror the huggingface model class name in the tester name. An alternative is to come up with new names for the same broader family of tests.
Additionally, in this case, v1 and v2 should be able to reuse the tester so it might make more sense to keep it like this and later move it once v1 models are added.

tests/infra/device_runner.py
sgligorijevicTT: After some discussion offline and a few iterations, we have agreed to guard using both static args and the try-catch.

tests/infra/base_tester.py
sgligorijevicTT: Implemented the suggested changes

tests/jax/models/Roberta/test_roberta.py
sgligorijevicTT: good catch, it was indeed a remnant of older versions of the infra

tests/jax/models/MNIST_CNN/CNN.py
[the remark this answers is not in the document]
sgligorijevicTT: I have reorganized the folders, MNIST MLP will be moved and possibly upgraded in a later PR.

tests/jax/models/MNIST_CNN/CNN.py
```diff
@@ -0,0 +1,29 @@
...
+        x = nn.max_pool(x, window_shape=(2, 2), strides=(2, 2))
+
+        x = x.reshape((x.shape[0], -1))
+        x = nn.Dense(features=256)(x)
+        x = nn.BatchNorm(use_running_average=not train, epsilon=1e-5)(x)
+        x = nn.relu(x)
+        x = nn.Dense(features=128)(x)
+        x = nn.BatchNorm(use_running_average=not train, epsilon=1e-5)(x)
+        x = nn.relu(x)
+        x = nn.Dense(features=10)(x)
```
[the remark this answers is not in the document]
sgligorijevicTT: Adding softmax introduced a new bug, I left it in the modified code but commented out.

tests/jax/models/MNIST_CNN/test_mnistcnn.py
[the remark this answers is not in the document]
sgligorijevicTT: A convention(abeit an often broken one) in JAX broadly is to have model take parameters first. Having inputs as the first argument is not a good default. It also has the benefit of not getting in the way if the user wants to be explicit and only go trough kwargs.

tests/jax/models/MNIST_CNN/test_mnistcnn.py
```diff
@@ -0,0 +1,89 @@
...
+    comparison_config: ComparisonConfig,
+) -> MNISTCNNTester:
+    return MNISTCNNTester(comparison_config, RunMode.TRAINING)
+
+
+# ----- Tests -----
+
+
+
+def test_mnist_inference(
```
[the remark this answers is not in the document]
sgligorijevicTT: :pensive: I wish. 
Good catch.

tests/jax/models/MNIST_CNN/CNN.py
```diff
@@ -0,0 +1,29 @@
...
+        return x
```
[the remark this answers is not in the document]
sgligorijevicTT: That was the symbol for missing newline at the end of file yep. fixed.

tests/jax/models/MNIST_CNN/test_mnistcnn.py
sgligorijevicTT: Flax linen API expects the user to create the weights manually with a call to `init`. I added a comment to clarify.

tests/jax/models/Albert/v2/tester.py
```diff
@@ -0,0 +1,53 @@
...
+        self._model_name = model_name
+
+    # @override
+    def _get_model(self) -> nn.Module:
+        return FlaxAlbertForMaskedLM.from_pretrained(self._model_name)
+
+    # @override
+    @staticmethod
+    def _get_forward_method_name() -> str:
+        return "__call__"
```
mrakitaTT, line 32: No need to override this method since this is the default behavior

tests/jax/models/GPT2/test_gpt2.py
```diff
@@ -0,0 +1,66 @@
...
+from infra import ModelTester, RunMode
+from transformers import AutoTokenizer, FlaxGPT2LMHeadModel
+
+MODEL_PATH = "openai-community/gpt2"
+
+
+class GPT2Tester(ModelTester):
+    """Tester for GPT2 for autoregressive text generation."""
+
+    @staticmethod
```
mrakitaTT, line 19: Forgot to remove here. Please make sure that you've removed `@staticmethod` in all places where overriding methods which shouldn't be static, and add `# @override`

tests/jax/models/MNIST/CNN/MNIST_CNN.py
```diff
@@ -0,0 +1,33 @@
...
+        x = x.reshape((x.shape[0], -1))
+        x = nn.Dense(features=256)(x)
+        x = nn.BatchNorm(use_running_average=not train, epsilon=1e-5)(x)
+        x = nn.relu(x)
+
+        x = nn.Dense(features=128)(x)
+        x = nn.BatchNorm(use_running_average=not train, epsilon=1e-5)(x)
+        x = nn.relu(x)
+        x = nn.Dense(features=10)(x)
+        # TODO(stefan): This causes a failure, uncomment once fixed.
```
mrakitaTT, line 31: Is the failure something related to tt-xla or is it failure later in compilation/runtime? If it is related to tt-xla, please open a new GH issue and paste the link in this TODO comment. If it is issue in compilation/runtime then you don't have to comment this out, you can just update the skip reason in the tester.

tests/jax/models/Albert/v2/base/test_albert_base.py
mrakitaTT, line 1: Missing `v2` from the tester files names in `Albert/v2`

tests/jax/models/Albertv2/base/test_albert_base.py
mrakitaTT: > I have refactored Albert tester classes to make them useable for multiple tests, now adding the exact model name as part of the class name no longer makes sense.

In this case you can just remove the variant name from the tester name, so you get the `AlbertV2Tester` as a name for the common tester.

> Additionally, we might have to have multiple tester classes for the same family of models depending on how they're used(masked language modeling vs sentiment analysis)

That's a good argument and in that case we can just append additional context to the tester name, for example AlbertV2MaskedLMTester`.

> Additionally, we might have to have multiple tester classes for the same family of models depending on how they're used(masked language modeling vs sentiment analysis), so it makes sense to mirror the huggingface model class name(and not the model's path on HF Hub) in the tester name. An alternative is to come up with new names for the same broader family of tests.

I wouldn't mirror the huggingface model class name because it is not standardized and depends on the author who added it to HF.

tests/jax/models/MNIST_CNN/CNN.py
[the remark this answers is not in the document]
mrakitaTT: Couple of things to note:
- PEP [discourages](https://peps.python.org/pep-0008/#package-and-module-names) use of capital letters in package/module names, and discourages use of underscores in package names. Transformers library [uses](https://github.com/huggingface/transformers/tree/main/src/transformers/models) lower letter names, though they use underscores. I think we should change all model folder names to lowercase too.
- You can just call implementation files `model_implementation.py` instead of full model name since they are already in the appropriate subfolder. Tester files are different because pytest depends on them.
- Missed the underscore in `tests/jax/models/MNIST/CNN/test_mnistcnn.py` :)

tests/jax/models/MNIST_CNN/CNN.py
```diff
@@ -0,0 +1,29 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import jax
+from flax import linen as nn
+
+class CNN(nn.Module):
```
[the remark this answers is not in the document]
mrakitaTT: [Shouldn't use](https://peps.python.org/pep-0008/#class-names) underscore in class names, I would rename `MNIST_CNN` to `MNISTCNNModel`. I know it is a bit unreadable but wcyd :)

tests/jax/models/MNIST_CNN/test_mnistcnn.py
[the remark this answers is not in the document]
mrakitaTT: Got it, makes sense.

tests/jax/models/GPT2/test_gpt2.py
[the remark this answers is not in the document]
mrakitaTT: You are right. I wanted to avoid discussions what makes models distinct enough and to simply sort them in folders semantically, but then someone can make an argument that albert could be under bert folder etc so we would have discussions anyways :)

What do you think about these guidelines: If models are architecturally different enough they should be in separate folders under `test/jax/models`. Good measure for that is if they can reuse the same tester, because then we want them to be in the same folder so they can reuse the common tester.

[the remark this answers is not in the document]
sgligorijevicTT: I think this will require some bikeshedding offline, but for now that's a good guideline.

tests/jax/models/MNIST/CNN/MNIST_CNN.py
sgligorijevicTT: It's a segfault deep inside LLVM, possibly some fusion or decomposition gone awry. Initially I didn't want to keep the change as in that case we no longer have any tests that fail on broadcasting specifically, but ideally that too would be covered by another test. I'm changing this and bumping up tests with BNs and broadcasting on my todo list.

tests/jax/models/Roberta/test_roberta.py
mrakitaTT, line 1: Forgot to lowercase `Roberta` folder

tests/jax/models/Roberta/test_roberta.py
```diff
@@ -0,0 +1,71 @@
...
+
+from typing import Dict, Sequence
+
+import jax
+import pytest
+from flax import linen as nn
+from infra import ModelTester, RunMode
+from transformers import AutoTokenizer, FlaxRobertaForMaskedLM
+
+MODEL_PATH = "FacebookAI/roberta-base"
```
mrakitaTT, line 13: Seems like Roberta also has base, large, etc versions. Don't do it in this PR, but maybe in some of the future PRs you can cover other versions too, similar as you did for Albert.

tests/jax/models/mnist/cnn/model_implementation.py
```diff
@@ -0,0 +1,32 @@
...
+        x = x.reshape((x.shape[0], -1))
+        x = nn.Dense(features=256)(x)
+        x = nn.BatchNorm(use_running_average=not train, epsilon=1e-5)(x)
+        x = nn.relu(x)
+
+        x = nn.Dense(features=128)(x)
+        x = nn.BatchNorm(use_running_average=not train, epsilon=1e-5)(x)
+        x = nn.relu(x)
+        x = nn.Dense(features=10)(x)
+        x = nn.softmax(x)
```
mrakitaTT, line 31: I would add blank lines above and below these two lines to separate the final fc layer too, similar to other layers

tests/jax/models/mnist/cnn/test_mnist_cnn.py
```diff
@@ -0,0 +1,84 @@
...
+    def _get_model(self) -> nn.Module:
+        return MNISTCNNModel()
+
+    # @override
+    def _get_forward_method_name(self) -> str:
+        return "apply"
+
+    # @override
+    def _get_input_activations(self) -> Sequence[jax.Array]:
+        img = jnp.ones((28, 28, 1))
```
mrakitaTT, line 29: Shouldn't this be `(1, 28, 28)`?

mrakitaTT (approved):

tests/jax/models/mnist/cnn/test_mnist_cnn.py
sgligorijevicTT: `(28, 28, 1)` is an example of unbatched input, input is in `(H,W,C)` and channels is 1 because MNIST is in grayscale. Batched input would look like `(1, 28, 28, 1)` for example. I will add a batch dimension and a comment to clarify.

mrakitaTT (approved):

mrakitaTT:
🎉  Great work @sgligorijevicTT!

PR #131, opened 2024-12-24
Provided one global DeviceConnector instance. Moved typedefs from utils to a new file.

kmitrovicTT:
Also moved `Workload` to a new file.

### DIFF
--- a/tests/infra/types.py
+++ b/tests/infra/types.py
@@ -1,0 +1,23 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from enum import Enum
+from typing import Union
+
+import jax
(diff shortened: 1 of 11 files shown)


tests/infra/types.py
```diff
@@ -0,0 +1,39 @@
...
+Tensor = Union[jax.Array]
+
+# Convenience alias. Could be used to represent nnx.Module, torch.nn.Module, etc.
+# NOTE nnx.Module is the newest API, linen.Module is legacy but it is used in some
+# huggingface models.
+Model = Union[nnx.Module, linen.Module]
+
+
+@dataclass
+class Workload:
```
mrakitaTT, line 22: Having this class defined in `types.py` somehow doesn't seem right, maybe move to a separate module?

mrakitaTT (approved):

PR #136, opened 2024-12-27
Deleted old test infra and moved old tests to new infra.

kmitrovicTT:
- Changed shapes in most of the tests to (32, 32). A follow up PR will be sent to
parametrize them to the shapes used in models.
- Added empty files with TODOs in them for all ops that we should have tests for
- Minor changes in test infra to accommodate needs for writing these tests


tests/jax/test_data_types.py
```diff
@@ -5,17 +5,20 @@
 # RUN: %PYTHON %s | FileCheck %s
 
 
-import jax
 import jax.numpy as jnp
 
 # Currently, tt::runtime only support float32, bfloat16, uint16, and uint32
 
 
+# TODO this test needs to be rewritten, this is hard to figure out.
```
mrakitaTT, line 13: Change to "TODO Investigate the purpose of this test and if we can test it in a different way."

tests/jax/test_device_initialization.py
```diff
@@ -0,0 +1,95 @@
...
+    PJRT plugin is registered during DeviceConnector instantiation. By providing global
+    singleton instance of the connector in device_connector.py we solved the problem of
+    automatic plugin registration.
+
+    We can simulate it here by calling the constructor directly.
+
+    NOTE we cannot manually do `xla_bridge.register_plugin` here because it will
+    report `PJRT_Api already exists for device type tt`.
+    """
+    DeviceConnector()
```
mrakitaTT, line 27: Wouldn't `from infra.device_connector import device_connector` force this initialization to happen? I don't think you need this fixture

tests/jax/test_scalar_types.py
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import jax
+import pytest
+from infra import run_op_test_with_random_inputs
+
+# TODO not sure what this is supposed to test. Rethink.
```
mrakitaTT, line 9: Change to `TODO We should put here tests which operate on scalars, with different types of ops (for example one unary, one binary etc). Test below is remnant from the old test set, reconsider keeping it.`

tests/jax/ops/test_broadcast.py
```diff
@@ -0,0 +1,23 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import jax.numpy as jnp
+import pytest
+from infra import run_op_test_with_random_inputs
+
+# TODO this skip reason is from old infra. Currently it fails with
+# `error: type of return operand 0 ('tensor<2x1xf32>') doesn't match function result type ('tensor<2x4xf32>') in function @main`.
+# Check why this happens and what is the state of this test in old infra (skip reason is old).
```
mrakitaTT, line 11: Hm, how did pre-commit script allow this comment to be over 80 characters O.o If you are not sure too, please open new issue for us to investigate this later.

tests/jax/ops/test_comparison_operands.py
```diff
@@ -0,0 +1,77 @@
...
+
+# NOTE TTNN does not support boolean data type, so bfloat16 is used instead.
+# Hence the output of comparison operation is bfloat16. JAX can not perform any
+# computation due to mismatch in output data type (in testing infrastructure).
+# The following tests explicitly convert data type of comparison operation
+# output for the verification purposes.
+# TODO Remove this workaround once the data type issue is resolved.
+# https://github.com/tenstorrent/tt-xla/issues/93
+
+# This should work in new infra due to match_dtypes CHECK TODO
```
mrakitaTT, line 21: I'm pretty sure you can just remove it right away (all comments above and `convert_output_to_bfloat16` fixture), this should be solved with your infra changes.

tests/jax/ops/test_comparison_operands.py
```diff
@@ -0,0 +1,77 @@
...
+        [(64, 64), (64, 64)],
+    ],
+)
+def test_add(x_shape: tuple, y_shape: tuple):
+    run_op_test_with_random_inputs(equal, [x_shape, y_shape])
+    run_op_test_with_random_inputs(not_equal, [x_shape, y_shape])
+    run_op_test_with_random_inputs(greater, [x_shape, y_shape])
+    run_op_test_with_random_inputs(greater_or_equal, [x_shape, y_shape])
+    run_op_test_with_random_inputs(less, [x_shape, y_shape])
+    run_op_test_with_random_inputs(less_or_equal, [x_shape, y_shape])
```
mrakitaTT, line 77: We should split this file into multiple files, one per each comparison op, to be consistent with other ops

tests/jax/ops/test_constant.py
```diff
@@ -0,0 +1,37 @@
...
+    def module_constant_zeros(a):
+        zeros = jnp.zeros(a.shape)
+        return zeros
+
+    def module_constant_ones(a):
+        ones = jnp.ones(a.shape)
+        return ones
+
+    run_op_test_with_random_inputs(module_constant_zeros, input_shapes)
+    run_op_test_with_random_inputs(module_constant_ones, input_shapes)
```
mrakitaTT, line 25: We should split this into two tests

tests/jax/ops/test_constant.py
```diff
@@ -0,0 +1,37 @@
...
+
+import jax.numpy as jnp
+import pytest
+from infra import run_op_test_with_random_inputs
+
+
+# TODO this skip reason is from old infra. Currently it fails with
+# `error: type of return operand 0 ('tensor<1xf32>') doesn't match function result type ('tensor<3x3xf32>') in function @main`.
+# Check why this happens and what is the state of this test in old infra (skip reason is old).
+@pytest.mark.parametrize("input_shapes", [[(3, 3)]])
```
mrakitaTT, line 13: Should add `(1, 1)` variant into `input_shapes`

tests/jax/ops/test_constant.py
```diff
@@ -0,0 +1,37 @@
...
+    def module_constant_ones(a):
+        ones = jnp.ones(a.shape)
+        return ones
+
+    run_op_test_with_random_inputs(module_constant_zeros, input_shapes)
+    run_op_test_with_random_inputs(module_constant_ones, input_shapes)
+
+
+# TODO this was copied from old infra. It does nothing with `a` argument. Figure out what
+# was meant to be tested this way.
```
mrakitaTT, line 29: This meant to test creation of constant tensor which has different values inside, which is not currently supported in TTNN, that's why op fails to legalize in TTIR->TTNN conversion. It doesn't need `a` argument, I think the problem was that in the old infra you had to pass some input shape to the function which runs test, so it creates input tensor with that shape, so we had to put some dummy input tensor. I think that with your infra you can just do this, but you would have to allow `Workload` to accept `None` args:
```
@pytest.mark.skip("Fails due to: error: failed to legalize operation 'ttir.constant'")
def test_constant_op_multi_value():
    def module_constant_multi():
        multi = jnp.array([[1, 2], [3, 4]], dtype=jnp.float32)
        return multi

    run_op_test(module_constant_multi)
```

Not that I've changed the test name to `test_constant_op_multi_value`.

tests/jax/ops/test_conv1d.py
```diff
@@ -0,0 +1,47 @@
...
+from infra import ComparisonConfig, random_tensor, run_op_test
+
+
+@pytest.fixture
+def comparison_config() -> ComparisonConfig:
+    config = ComparisonConfig()
+    config.disable_all()
+    config.pcc.enable()
+    config.pcc.required_pcc = 0.95
+    return config
```
mrakitaTT, line 16: Please add some TODO to investigate why `conv1d` has such poor precision. Same in `conv2d` test.

tests/jax/ops/test_conv2d.py
mrakitaTT, line 1: StableHLO has only one `convolution` op, so splitting these tests into two files is not desirable from the perspective that we want to have one test file per each StableHLO op. We should also name these files according to stablehlo ops' names ([stablehlo spec](https://openxla.org/stablehlo/spec)). Currently their website is broken so it is not easy to see list of all ops, so I suggest you open new issue to rename these files and organize them to be one file per stablehlo op.

tests/jax/ops/test_negation.py
```diff
@@ -0,0 +1,17 @@
...
+import pytest
+from infra import run_op_test_with_random_inputs
+
+
+@pytest.mark.parametrize("x_shape", [(32, 32), (64, 64)])
+def test_negation(x_shape: tuple):
+    def negation(x: jax.Array) -> jax.Array:
+        return jnp.negative(x)
+
+    # Try both negative and positive values.
```
mrakitaTT, line 16: Change to "Trying" instead of "Try", this way it seems like a TODO comment :) Same for `tests/jax/ops/test_sign.py`

tests/jax/ops/test_reduce.py
```diff
@@ -0,0 +1,34 @@
...
+from infra import ComparisonConfig, run_op_test_with_random_inputs
+
+
+@pytest.fixture
+def comparison_config() -> ComparisonConfig:
+    config = ComparisonConfig()
+    config.disable_all()
+    config.pcc.enable()
+    config.pcc.required_pcc = 0.95
+    return config
```
mrakitaTT, line 17: Add TODO to investigate why we need such comparison config

tests/jax/ops/test_reduce.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def reduce_sum(x: jax.Array) -> jax.Array:
+        return jnp.sum(x)
+
+    run_op_test_with_random_inputs(
+        reduce_max, [x_shape], comparison_config=comparison_config
+    )
+    run_op_test_with_random_inputs(
+        reduce_sum, [x_shape], comparison_config=comparison_config
+    )
```
mrakitaTT, line 34: Should be split into separate tests, because these reduces will be used with different shapes in models so the test parametrization should grow separately

mrakitaTT (approved):
Approving since all comments are trivial to resolve and I am going on vacation. Feel free to merge when you resolve them.

tests/jax/ops/test_comparison_operands.py
mrakitaTT: Actually you can fix this later as part of the new issue which I mentioned here: https://github.com/tenstorrent/tt-xla/pull/136#discussion_r1898779830

tests/jax/graphs/test_relu.py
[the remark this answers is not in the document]
kmitrovicTT: > Remove the test_maximum_as_relu test since it is testing the same thing as the ops/test_maximum.py

I explicitly added these two versions of ReLU since 

```python
jnp.maximum(random_tensor(x_shape), jnp.zeros(y_shape)) # (1)
```

works but

```python
jnp.maximum(random_tensor(x_shape), 0) # (2)
```

fails due to unsupported broadcast. But they are basically equivalent since the underlying `stablehlo` [function](https://openxla.org/stablehlo/spec#maximum) works with tensors indeed, not tensor and a scalar. First case `(1)` from above maps directly to `stablehlo.maximum`, but second case maps to a graph with `stablehlo.broadcast_in_dim` preceding `stablehlo.maximum`. 

I think we should have example (1) somewhere since it is an op test, not a graph test. Maybe we could expand `test_maximum.py` 

```python
def test_maximum(x_shape: tuple, y_shape: tuple):
    def maximum(x: jax.Array, y: jax.Array) -> jax.Array:
        return jnp.maximum(x, y)

    run_op_test_with_random_inputs(maximum, [x_shape, y_shape])
    # Special case where maximum op is used as ReLU activation function where second
    # argument is a tensor of zeros.
    run_op_test(maximum, [random_tensor(x_shape), jnp.zeros(y_shape)])
```

which tests this specific edge case of maximum used as ReLU.

tests/jax/graphs/test_simple_gradient.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import jax
+import pytest
+from infra import run_graph_test_with_random_inputs
+
+
+@pytest.mark.parametrize("x_shape", [(32, 32), (64, 64)])
+# @pytest.mark.skip("Inputs to eltwise binary must be tilized")
```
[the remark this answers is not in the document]
kmitrovicTT: Actually forgot to remove the comment, it works, we shouldn't skip it.

tests/jax/graphs/test_simple_regression.py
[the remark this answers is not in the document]
kmitrovicTT: Will handle this issue separately.

tests/jax/graphs/test_simple_regression.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+@pytest.mark.parametrize(
+    ["weights", "bias", "X", "y"], [[(1, 2), (1, 1), (2, 1), (1, 1)]]
+)
+@pytest.mark.skip("failed to legalize operation 'stablehlo.dot_general'")
+def test_simple_regression(weights, bias, X, y):
+    def simple_regression(weights, bias, X, y):
+        def loss(weights, bias, X, y):
+            predict = X.dot(weights) + bias if bias is not None else X.dot(weights)
```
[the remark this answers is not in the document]
kmitrovicTT: Wow, good question, made me wonder.

Usually, when we have a parameter which influences control flow, it must be a [static arg](https://jax.readthedocs.io/en/latest/_autosummary/jax.jit.html). 

> Calling the jitted function with different values for these constants [static args] will trigger recompilation

But in this example we always have some bias, never `None`.  So I tweaked the infra a bit to allow me to set bias to `None`

```python
@pytest.mark.parametrize(
    ["weights", "bias", "X", "y"], [[(1, 2), (1, 1), (2, 1), (1, 1)]]
)
# @pytest.mark.skip("failed to legalize operation 'stablehlo.dot_general'")
def test_simple_regression(weights, bias, X, y):
    def simple_regression(weights, bias, X, y):
        def loss(weights, bias, X, y):
            predict = X.dot(weights) + bias if bias is not None else X.dot(weights)
            return ((predict - y) ** 2).sum()

        # Compute gradient and update weights.
        weights -= jax.grad(loss)(weights, bias, X, y)
        return weights

    run_graph_test_with_random_inputs(simple_regression, [weights, bias, X, y])
    run_graph_test_with_random_inputs(simple_regression, [weights, None, X, y])
```

When bias exists it produces 

```mlir
module @jit_simple_regression attributes {jax.uses_shape_polymorphism = false, mhlo.num_partitions = 1 : i32, mhlo.num_replicas = 1 : i32} {
  func.func public @main(%arg0: tensor<1x2xf32> {mhlo.layout_mode = "{1,0}", mhlo.sharding = "{replicated}"} loc("weights"), %arg1: tensor<1x1xf32> {mhlo.layout_mode = "{1,0}", mhlo.sharding = "{replicated}"} loc("bias"), %arg2: tensor<2x1xf32> {mhlo.layout_mode = "{1,0}", mhlo.sharding = "{replicated}"} loc("X"), %arg3: tensor<1x1xf32> {mhlo.layout_mode = "{1,0}", mhlo.sharding = "{replicated}"} loc("y")) -> (tensor<1x2xf32> {jax.result_info = "", mhlo.layout_mode = "default"}) {
    %0 = stablehlo.dot_general %arg2, %arg0, contracting_dims = [1] x [0] : (tensor<2x1xf32>, tensor<1x2xf32>) -> tensor<2x2xf32> loc(#loc48)
    %1 = stablehlo.broadcast_in_dim %arg1, dims = [0, 1] : (tensor<1x1xf32>) -> tensor<2x2xf32> loc(#loc49)
    %2 = stablehlo.add %0, %1 : tensor<2x2xf32> loc(#loc49)
    %3 = stablehlo.broadcast_in_dim %arg3, dims = [0, 1] : (tensor<1x1xf32>) -> tensor<2x2xf32> loc(#loc50)
    %4 = stablehlo.subtract %2, %3 : tensor<2x2xf32> loc(#loc50)
    %5 = call @integer_pow(%4) : (tensor<2x2xf32>) -> tensor<2x2xf32> loc(#loc51)
    %cst = stablehlo.constant dense<2.000000e+00> : tensor<f32> loc(#loc)
    %6 = stablehlo.broadcast_in_dim %cst, dims = [] : (tensor<f32>) -> tensor<2x2xf32> loc(#loc52)
    %7 = stablehlo.multiply %6, %5 : tensor<2x2xf32> loc(#loc52)
    %cst_0 = stablehlo.constant dense<1.000000e+00> : tensor<f32> loc(#loc)
    %8 = stablehlo.broadcast_in_dim %cst_0, dims = [] : (tensor<f32>) -> tensor<2x2xf32> loc(#loc53)
    %9 = stablehlo.multiply %8, %7 : tensor<2x2xf32> loc(#loc52)
    %10 = stablehlo.dot_general %9, %arg2, contracting_dims = [0] x [0] : (tensor<2x2xf32>, tensor<2x1xf32>) -> tensor<2x1xf32> loc(#loc54)
    %11 = stablehlo.transpose %10, dims = [1, 0] : (tensor<2x1xf32>) -> tensor<1x2xf32> loc(#loc55)
    %12 = stablehlo.subtract %arg0, %11 : tensor<1x2xf32> loc(#loc56)
    return %12 : tensor<1x2xf32> loc(#loc)
  } loc(#loc)
  func.func private @integer_pow(%arg0: tensor<2x2xf32> loc("jit(simple_regression)/jit(main)/integer_pow[y=1]"(#loc46))) -> tensor<2x2xf32> {
    return %arg0 : tensor<2x2xf32> loc(#loc51)
  } loc(#loc51)
} loc(#loc)
```

when bias is `None` it produces

```mlir
module @jit_simple_regression attributes {jax.uses_shape_polymorphism = false, mhlo.num_partitions = 1 : i32, mhlo.num_replicas = 1 : i32} {
  func.func public @main(%arg0: tensor<1x2xf32> {mhlo.layout_mode = "{1,0}", mhlo.sharding = "{replicated}"} loc("weights"), %arg1: tensor<2x1xf32> {mhlo.layout_mode = "{1,0}", mhlo.sharding = "{replicated}"} loc("X"), %arg2: tensor<1x1xf32> {mhlo.layout_mode = "{1,0}", mhlo.sharding = "{replicated}"} loc("y")) -> (tensor<1x2xf32> {jax.result_info = "", mhlo.layout_mode = "default"}) {
    %0 = stablehlo.dot_general %arg1, %arg0, contracting_dims = [1] x [0] : (tensor<2x1xf32>, tensor<1x2xf32>) -> tensor<2x2xf32> loc(#loc47)
    %1 = stablehlo.broadcast_in_dim %arg2, dims = [0, 1] : (tensor<1x1xf32>) -> tensor<2x2xf32> loc(#loc48)
    %2 = stablehlo.subtract %0, %1 : tensor<2x2xf32> loc(#loc48)
    %3 = call @integer_pow(%2) : (tensor<2x2xf32>) -> tensor<2x2xf32> loc(#loc49)
    %cst = stablehlo.constant dense<2.000000e+00> : tensor<f32> loc(#loc)
    %4 = stablehlo.broadcast_in_dim %cst, dims = [] : (tensor<f32>) -> tensor<2x2xf32> loc(#loc50)
    %5 = stablehlo.multiply %4, %3 : tensor<2x2xf32> loc(#loc50)
    %cst_0 = stablehlo.constant dense<1.000000e+00> : tensor<f32> loc(#loc)
    %6 = stablehlo.broadcast_in_dim %cst_0, dims = [] : (tensor<f32>) -> tensor<2x2xf32> loc(#loc51)
    %7 = stablehlo.multiply %6, %5 : tensor<2x2xf32> loc(#loc50)
    %8 = stablehlo.dot_general %7, %arg1, contracting_dims = [0] x [0] : (tensor<2x2xf32>, tensor<2x1xf32>) -> tensor<2x1xf32> loc(#loc52)
    %9 = stablehlo.transpose %8, dims = [1, 0] : (tensor<2x1xf32>) -> tensor<1x2xf32> loc(#loc53)
    %10 = stablehlo.subtract %arg0, %9 : tensor<1x2xf32> loc(#loc54)
    return %10 : tensor<1x2xf32> loc(#loc)
  } loc(#loc)
  func.func private @integer_pow(%arg0: tensor<2x2xf32> loc("jit(simple_regression)/jit(main)/integer_pow[y=1]"(#loc45))) -> tensor<2x2xf32> {
    return %arg0 : tensor<2x2xf32> loc(#loc49)
  } loc(#loc49)
} loc(#loc)
```

It seems like `if bias is None` evaluation happens before mlir graph is created, thus there is no trace of this conditional in resulting graph, but we rather get two distinct graphs as if we wrote two separate examples, one which does just `X.dot(weights)` and one which does `X.dot(weights) + bias`.

ChatGPTs comment on this:

> Does the if end up in the StableHLO Graph?

>Yes, the if condition does influence the StableHLO graph in the sense that JAX's tracing will include both potential operations (the one with bias and the one without bias) as part of the computation, unless you specifically control for it via jax.lax.cond or similar control flow mechanisms that JAX can handle more explicitly in JIT-compiled code.

> However, JAX's tracing is smart enough to only compile the path that gets executed at runtime. So, if the bias is not None, JAX will compile the branch with X.dot(weights) + bias. If bias is None, it will compile the branch with X.dot(weights). This behavior is dynamic but happens at trace time, and it can be seen in the compiled StableHLO graph as the set of operations that JAX decides to include based on the conditional check during execution.

tests/jax/test_device_initialization.py
kmitrovicTT: Nope, since `device_connector` instance is not actually used anywhere in this test. It would be an unused import.

I left it like this in order to explain where the device initialization (that is plugin registration) actually happens, to lead someone reading the test to the code that does it.

Previously, plugin registration was done here explicitly. Now, due to the way we decided to resolve it, it happens during global `device_connector` object instantiation. That would be completely hidden away from this test (which is called `test_device_initialization.py`) if I had not done this.

tests/jax/graphs/test_simple_regression.py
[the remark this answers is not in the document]
sgligorijevicTT: By the looks of things, JAX has a special case implemented to not replace `None` with tracers, but that makes a lot of sense.
One thing to note is that static is needed to make array **values** available at compile time, shapes and data types are stored in tracers and always available, in other words they are always static.
Checks `x is None` and `x is not None` are semantically type checks, despite being spelled as value checks (well, actually identity checks, but because `None` is the only object of type `NoneType` and with only one possible value, all 3 checks are equivalent). In other words, it is as if you had tracers with `dtype=NoneType` and unspecified shape, and did tests of the form `x.dtype != NoneType`. But by not replacing anything, jax makes existing code have the intended meaning. Incredibly clever!

tests/jax/ops/test_broadcast.py
kmitrovicTT: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#comments

It is a bit annoying that you have to take care of it yourself. I think clang formats comments as well. Not sure why black doesn't.

I reformatted it a bit.

tests/jax/ops/test_comparison_operands.py
kmitrovicTT: I am getting some weird errors. Tried to debug it briefly, but no luck. Opened up separate [issue](https://github.com/tenstorrent/tt-xla/issues/156) to investigate this.

tests/jax/ops/test_comparison_operands.py
kmitrovicTT: Makes sense. I left TODO and opened an [issue](https://github.com/tenstorrent/tt-xla/issues/156) just for this file. I kept it together since they all need `convert...` decorator to work for now since I wasn't able to [resolve](https://github.com/tenstorrent/tt-xla/pull/136#discussion_r1898776761) this quickly.

tests/jax/ops/test_constant.py
kmitrovicTT: Is (3, 3) legit here or should we go with (32, 32)?

tests/jax/ops/test_constant.py
kmitrovicTT: Works out of the box if you do `run_op_test(module_constant_multi, [])` since infra expects args to be iterable down the pipeline, so this additional empty list argument allows us to run it without changes to infra.

tests/jax/ops/test_conv2d.py
kmitrovicTT: Got it.

Did it in this PR since it is trivial, and it would be nice if we had a complete list of ops we want to support asap.

So now we have 47 tests:

```bash
>> find tests/jax/ops -type f -exec basename {} \; | sort
test_abs.py
test_add.py
test_and.py
test_broadcast_in_dim.py
test_cbrt.py
test_ceil.py
test_clamp.py
test_compare.py
test_concatenate.py
test_constant.py
test_convert.py
test_convolution.py
test_cosine.py
test_divide.py
test_dot_general.py
test_dynamic_iota.py
test_exponential.py
test_exponential_minus_one.py
test_floor.py
test_gather.py
test_iota.py
test_log.py
test_log_plus_one.py
test_logistic.py
test_maximum.py
test_minimum.py
test_multiply.py
test_negate.py
test_not.py
test_or.py
test_power.py
test_reduce.py
test_reduce_window.py
test_remainder.py
test_reshape.py
test_reverse.py
test_rsqrt.py
test_scatter.py
test_select.py
test_sign.py
test_sine.py
test_slice.py
test_sqrt.py
test_subtract.py
test_tanh.py
test_transpose.py
test_xor.py
```

which matches [this list](https://docs.google.com/spreadsheets/d/15JbTmYDccsJgeD_mJa-V6vvyD3e6IGCKZvJhLpkZr5I/edit?gid=0#gid=0) of ops with priority 0 (and some extra which were there already.

tests/jax/graphs/test_simple_regression.py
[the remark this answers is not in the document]
kmitrovicTT: Wow. Thanks @sgligorijevicTT for making this clear.

Issue #9, opened 2024-09-20
Enable N300 


AleksKnezevic:
Currently tests fail on N300, please fix and enable CI

nsmithtt:
@mrakitaTT, we're getting closer to running more multi-device tests and we'd like to leverage Jax to do so.  Do you know if this is still an issue?

mrakitaTT:
@nsmithtt no one has worked on this yet and we still have a hardcoded `int device_info_count_ = 1;` hack in order to be able to use n300 devices to run tests. The last tech sync when Wooseok was presenting the multi device progress I was wondering how is he getting the hlo graphs and if he is using tt-xla for his work, but haven't yet gotten to asking him.

@ajakovljevicTT this might be a good PJRT task for you or Sanja, can you please investigate what fails when you disable the device count hack and how device is selected in PJRT etc?

nsmithtt:
I think we got multi device stableHLO IR from a multi-device google collab and have been working directly from the stablehlo.  Or @wooseokTT did you hack PJRT to make it work for a one off test?

wooseokTT:
@nsmithtt @mrakitaTT I have local fix that uses device ids instead of the hardcoded "1". Then, I used matmul python code to generate stableHLO mlir using tt-xla.

ajakovljevicTT:
@wooseokTT Would you mind sharing that fix in a branch? Thanks!

wooseokTT:
@ajakovljevicTT  It's not a big change. I just fixed code to the point where I can run multidevice, and didn't test any other things. 

diff --git a/src/common/api_impl.cc b/src/common/api_impl.cc
index 8684839..e799f3b 100644
--- a/src/common/api_impl.cc
+++ b/src/common/api_impl.cc
@@ -20,6 +20,8 @@
 
 namespace tt::pjrt {
 
+#define TEST_MESH 1
+
 std::pair<tt::target::DataType, size_t>
 MapBufferTypeToElementType(PJRT_Buffer_Type buffer_type) {
   switch (buffer_type) {
@@ -653,9 +655,13 @@ void ClientInstance::BindApi(PJRT_Api *api) {
 tt_pjrt_status ClientInstance::PopulateDevices() {
   DLOG_F(LOG_DEBUG, "ClientInstance::PopulateDevices");
   auto [system_desc, chip_ids] = tt::runtime::getCurrentSystemDesc();
+#if TEST_MESH
+  int device_info_count_ = chip_ids.size();
[...]

PR #145, opened 2025-01-03
Enabling non-hardcoded run of our test infrastructure

ajakovljevicTT:
As per issue https://github.com/tenstorrent/tt-xla/issues/9, our test infra had a problem when detecting two TT devices, as is the case on the two-chip N300, which was workarounded by hardcoding which devices to use. This PR fixes that, instead just setting that the number of devices that the backend uses is capped to 1. Additionally, there was a bug in our old infra which ran the tests on cpu instead of a TT device, so this PR also fixes that.

Fixes https://github.com/tenstorrent/tt-xla/issues/9.

### DIFF
--- a/inc/common/pjrt_implementation/device_description.h
+++ b/inc/common/pjrt_implementation/device_description.h
@@ -20,7 +20,7 @@ namespace tt::pjrt {
 class DeviceDescription {
 
 public:
-  DeviceDescription(int32_t client_id) : client_id_(client_id) {};
+  DeviceDescription(int32_t device_id) : m_device_id(device_id) {};
   ~DeviceDescription();
   operator PJRT_DeviceDescription *() {
     return reinterpret_cast<PJRT_DeviceDescription *>(this);
@@ -35,21 +35,18 @@ public:
   std::string_view debug_string() { return to_string(); }
   std::string_view to_string() {
     std::stringstream ss;
-    ss << kind_string_ << "(id=" << device_id() << ", arch=" << arch_string_
+    ss << kind_string_ << "(id=" << getDeviceId() << ", arch=" << arch_string_
        << ")";
     user_string_ = ss.str();
     return user_string_;
   }
 
-  // TODO
-  int64_t device_id() { return 0; }
-
-  int client_id() { return client_id_; }
+  int getDeviceId() { return m_device_id; }
 
   int process_index() { return 0; }
 
 private:
-  int client_id_;
+  int m_device_id;
 
   // TODO We should understand better how these are used.
   // See https://github.com/tenstorrent/tt-xla/issues/125
--- a/inc/common/pjrt_implementation/executable_image.h
+++ b/inc/common/pjrt_implementation/executable_image.h
@@ -26,7 +26,9 @@ class ExecutableImage {
 
 public:
   ExecutableImage(const tt::runtime::Binary &binary, std::string code,
-                  const std::vector<bool> &is_output_scalar);
+                  const std::vector<bool> &is_output_scalar,
+                  size_t num_addressable_devices);
+
   operator PJRT_Executable *() {
     return reinterpret_cast<PJRT_Executable *>(this);
   }
@@ -53,6 +55,10 @@ public:
   // Checks if the output on the i-th index is a scalar.
   bool isOutputScalar(size_t index) const;
 
+  const size_t get_num_addressable_devices() const {
+    return num_addressable_devices;
+  }
+
 private:
   // The reference count. Must be disposed when reaching zero.
   std::atomic<int> m_ref_count;
@@ -65,6 +71,7 @@ private:
 
   size_t m_arg_count;
   size_t m_result_count;
+  size_t num_addressable_devices;
 
   // For every output, holds if the type is a scalar or not.
   std::vector<bool> m_is_output_scalar;
--- a/src/common/module_builder.h
+++ b/src/common/module_builder.h
@@ -36,6 +36,10 @@ public:
     return m_is_output_scalar;
   };
 
+  // This needs to return the number of addressable devices from the StableHLO
+  // code. Currently hardcoded to one, as we only support one-chip execution.
+  size_t getNumAddressableDevices() const { return 1; }
+
 private:
   // Creates VHLO module from the input program code.
   mlir::OwningOpRef<mlir::ModuleOp>
--- a/src/common/pjrt_implementation/client_instance.cc
+++ b/src/common/pjrt_implementation/client_instance.cc
@@ -164,12 +164,10 @@ void ClientInstance::BindApi(PJRT_Api *api) {
 tt_pjrt_status ClientInstance::PopulateDevices() {
   DLOG_F(LOG_DEBUG, "ClientInstance::PopulateDevices");
   auto [system_desc, chip_ids] = tt::runtime::getCurrentSystemDesc();
-  int device_info_count_ =
-      1; // TODO: revert to chip_ids.size(); once
-         // https://github.com/tenstorrent/tt-xla/issues/9 is fixed
+  int devices_count = chip_ids.size();
 
-  devices_.resize(device_info_count_);
-  for (size_t i = 0; i < device_info_count_; ++i) {
+  devices_.resize(devices_count);
+  for (size_t i = 0; i < devices_count; ++i) {
     devices_[i] = new DeviceInstance(i, *this);
   }
 
@@ -197,7 +195,8 @@ PJRT_Error *ClientInstance::Compile(const PJRT_Program *program,
       *this,
       new ExecutableImage(module_builder_->getBinary(),
                           std::string(program->code, program->code_size),
-                          module_builder_->getIsOutputScalar()),
+                          module_builder_->getIsOutputScalar(),
+                          module_builder_->getNumAddressableDevices()),
       addressable_devices_);
   *out_executable = executable.release();
 
--- a/src/common/pjrt_implementation/loaded_executable_instance.cc
+++ b/src/common/pjrt_implementation/loaded_executable_instance.cc
@@ -10,6 +10,8 @@
 
 #include "common/pjrt_implementation/loaded_executable_instance.h"
 
+#include <unordered_set>
+
 #include "common/pjrt_implementation/buffer_instance.h"
 #include "common/pjrt_implementation/client_instance.h"
 #include "common/pjrt_implementation/error_instance.h"
@@ -32,12 +34,15 @@ void LoadedExecutableInstance::BindApi(PJRT_Api *api) {
     DLOG_F(
         LOG_DEBUG,
         "LoadedExecutableInstance::PJRT_LoadedExecutable_AddressableDevices");
-    const std::vector<DeviceInstance *> &devices =
-        LoadedExecutableInstance::Unwrap(args->executable)
-            ->addressable_devices();
+    LoadedExecutableInstance *loaded_executable =
+        LoadedExecutableInstance::Unwrap(args->executable);
+    const std::vector<DeviceInstance *> &addressable_devices =
+        loaded_executable->addressable_devices();
+    int num_addressable_devices =
+        loaded_executable->image_->get_num_addressable_devices();
     args->addressable_devices = const_cast<PJRT_Device **>(
-        reinterpret_cast<PJRT_Device *const *>(devices.data()));
-    args->num_addressable_devices = devices.size();
+        reinterpret_cast<PJRT_Device *const *>(addressable_devices.data()));
+    args->num_addressable_devices = num_addressable_devices;
     return nullptr;
   };
   api->PJRT_LoadedExecutable_Delete =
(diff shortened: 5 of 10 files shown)


kmitrovicTT:
@ajakovljevicTT please rebase and ping me again to take a look. That bug should be gone now that old infra is gone.

inc/common/pjrt_implementation/device_description.h
```diff
@@ -42,14 +43,16 @@ class DeviceDescription {
   }
 
   // TODO
```
kmitrovicTT, line 45: What does this refer to? Remove, we need to redo everything in any case, especially naming and spacing.

src/common/module_builder.h
```diff
@@ -35,6 +35,10 @@ class ModuleBuilder {
 
   bool isOutputScalar(size_t index) const;
 
+  // This needs to return the number of addressable devices from the StableHLO
+  // code.
+  size_t getNumberOfAddressibleDevices() const;
```
kmitrovicTT, line 40: You can just implement it here right away, it is too short to have any benefit of moving it to .cc. Similar as what was done with the methods above. Make sure to copy that comment though.

I like that it is fully verbose, but I suggest renaming it to `getNumAddressableDevices`, to match `getNum...` getters from above.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -32,12 +34,15 @@ void LoadedExecutableInstance::BindApi(PJRT_Api *api) {
     DLOG_F(
         LOG_DEBUG,
         "LoadedExecutableInstance::PJRT_LoadedExecutable_AddressableDevices");
-    const std::vector<DeviceInstance *> &devices =
+    auto &addressable_devices =
```
kmitrovicTT, line 37: Any particular reason why this was changed to `auto`? I don't mind it, but it is not that cumbersome and it makes the type more explicit.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -32,12 +34,15 @@ void LoadedExecutableInstance::BindApi(PJRT_Api *api) {
...
         LoadedExecutableInstance::Unwrap(args->executable)
             ->addressable_devices();
+    int num_addressable_devices =
+        LoadedExecutableInstance::Unwrap(args->executable)
+            ->image_->get_num_addresible_devices();
```
kmitrovicTT, line 42: Typo: `addressable`.

kmitrovicTT: Maybe add `assert num_addressable_devices == addressable_devices.size()`.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -87,13 +90,23 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
   rt_inputs.reserve(args->num_args);
 
+  std::unordered_set<int> device_ids;
+
   for (size_t i = 0; i < args->num_args; ++i) {
     BufferInstance *buffer =
         BufferInstance::Unwrap(args->argument_lists[dev_index][i]);
     rt_inputs.emplace_back(buffer->tensor());
+    device_ids.insert(
+        chip_ids[buffer->device().device_description()->device_id()]);
```
kmitrovicTT, line 100: Maybe explain this a bit, since it is a long statement. Maybe even extract this index into a var.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -77,8 +82,6 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
   DLOG_F(LOG_DEBUG, "LoadedExecutableInstance::Execute");
 
   auto [system_desc, chip_ids] = tt::runtime::getCurrentSystemDesc();
-  int dev_0 = chip_ids[0];
-  tt::runtime::Device device = tt::runtime::openDevice({dev_0});
 
   assert(args->num_devices == 1);
   int dev_index = 0;
```
kmitrovicTT, line 87: Why is this hardcoded?

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -77,8 +82,6 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
   DLOG_F(LOG_DEBUG, "LoadedExecutableInstance::Execute");
 
   auto [system_desc, chip_ids] = tt::runtime::getCurrentSystemDesc();
-  int dev_0 = chip_ids[0];
-  tt::runtime::Device device = tt::runtime::openDevice({dev_0});
 
   assert(args->num_devices == 1);
```
kmitrovicTT, line 86: What's the difference between `num_devices` and `num_addressable_devices`?

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -87,13 +90,23 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
   for (size_t i = 0; i < args->num_args; ++i) {
     BufferInstance *buffer =
         BufferInstance::Unwrap(args->argument_lists[dev_index][i]);
     rt_inputs.emplace_back(buffer->tensor());
+    device_ids.insert(
+        chip_ids[buffer->device().device_description()->device_id()]);
     DLOG_F(INFO, "Runtime input id: %d", buffer->unique_id());
   }
 
+  assert(device_ids.size() == 1);
```
kmitrovicTT, line 104: Why?

tests/infra/device_connector.py
```diff
@@ -81,9 +81,16 @@ def connect_gpu(self) -> jax.Device:
         """Returns GPUDevice handle."""
         return self.connect_device(DeviceType.GPU)
 
-    def connect_device(self, device_type: DeviceType) -> jax.Device:
+    def _number_of_devices(self, device_type: DeviceType) -> int:
+        """Returns the number of devices of specifed type."""
+        return len(jax.devices(device_type.value))
```
kmitrovicTT, line 86: Move below with private methods. Rephrase a bit `Returns the number of available devices of specified type.`

tests/infra/device_connector.py
```diff
@@ -69,9 +69,9 @@ def is_initialized(self) -> bool:
 
         return False
 
-    def connect_tt_device(self) -> jax.Device:
+    def connect_tt_device(self, num_device: int = 0) -> jax.Device:
```
kmitrovicTT, line 72: Does it make more sense to call this `device_num`?

tests/infra/device_connector.py
```diff
@@ -81,9 +81,16 @@ def connect_gpu(self) -> jax.Device:
...
 
-    def connect_device(self, device_type: DeviceType) -> jax.Device:
+    def _number_of_devices(self, device_type: DeviceType) -> int:
+        """Returns the number of devices of specifed type."""
+        return len(jax.devices(device_type.value))
+
+    def connect_device(
+        self, device_type: DeviceType, num_device: int = 0
+    ) -> jax.Device:
         """Returns handle for device identified by `device_type`."""
```
kmitrovicTT, line 91:
```
"""
Returns handle for device identified by `device_type`. 

If there are multiple available devices of `device_type`, `device_num` makes it
possible to choose between them. By default, returns first available device.
"""
```

tests/infra/device_runner.py
```diff
@@ -64,18 +64,22 @@ def put_tensors_on_gpu(*tensors: Tensor) -> Sequence[Tensor]:
...
 
     @staticmethod
-    def _run_on_device(device_type: DeviceType, workload: Workload) -> Tensor:
+    def _run_on_device(
+        device_type: DeviceType, workload: Workload, num_device: int = 0
+    ) -> Tensor:
         """Runs `workload` on device identified by `device_type`."""
-        device_workload = DeviceRunner._put_on_device(device_type, workload)
+        device_workload = DeviceRunner._put_on_device(device_type, workload, num_device)
         device = device_connector.connect_device(device_type)
```
kmitrovicTT, line 72: Don't you wanna pass `device_num` to `connect_device` as well?

src/common/pjrt_implementation/loaded_executable_instance.cc
ajakovljevicTT: `num_devices`, passed as the argument of this function is the total number of devices for execution. It was inferred from `num_addressable_devices` in the xla bridge code.

src/common/pjrt_implementation/loaded_executable_instance.cc
ajakovljevicTT: Just a sanity check, since we only support execution on one chip currently. Note that this corresponds to the number of chips that the specific instance will run on, not the whole number of chips.

kmitrovicTT: Add comment `Sanity check, we only support execution on one chip currently.`

src/common/pjrt_implementation/loaded_executable_instance.cc
kmitrovicTT: Inferred how? Equal to it?

inc/common/pjrt_implementation/device_description.h
```diff
@@ -41,15 +42,16 @@ class DeviceDescription {
...
+  int64_t device_id() { return device_id_; }
 
   int client_id() { return client_id_; }
 
   int process_index() { return 0; }
 
 private:
+  static int static_device_id;
   int client_id_;
+  int device_id_;
```
mrakitaTT, line 54: I am not sure that we actually need both of these, let's discuss tomorrow

src/common/module_builder.h
```diff
@@ -35,6 +35,10 @@ class ModuleBuilder {
 
   bool isOutputScalar(size_t index) const;
 
+  // This needs to return the number of addressable devices from the StableHLO
+  // code. Currently hardcoded to one, as we only support one-chip execution.
+  size_t getNumAddressableDevices() const { return 1; }
```
mrakitaTT, line 40: Let's discuss this tomorrow too, I am interested to see how is it enough to know just the number of addressable devices and not the actual IDs list

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -164,9 +164,7 @@ void ClientInstance::BindApi(PJRT_Api *api) {
 tt_pjrt_status ClientInstance::PopulateDevices() {
   DLOG_F(LOG_DEBUG, "ClientInstance::PopulateDevices");
   auto [system_desc, chip_ids] = tt::runtime::getCurrentSystemDesc();
-  int device_info_count_ =
-      1; // TODO: revert to chip_ids.size(); once
-         // https://github.com/tenstorrent/tt-xla/issues/9 is fixed
+  int device_info_count_ = chip_ids.size();
```
mrakitaTT, line 167: I would rename this to `devices_count`

inc/common/pjrt_implementation/executable_image.h
```diff
@@ -61,6 +67,7 @@ class ExecutableImage {
 
   size_t arg_count;
   size_t result_count;
+  size_t num_addressable_devices;
```
mrakitaTT, line 70: I would rename this to `addressable_devices_count`, and also in names of getter functions

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -32,12 +34,15 @@ void LoadedExecutableInstance::BindApi(PJRT_Api *api) {
     DLOG_F(
         LOG_DEBUG,
         "LoadedExecutableInstance::PJRT_LoadedExecutable_AddressableDevices");
-    const std::vector<DeviceInstance *> &devices =
+    const std::vector<DeviceInstance *> &addressable_devices =
         LoadedExecutableInstance::Unwrap(args->executable)
             ->addressable_devices();
+    int num_addressable_devices =
+        LoadedExecutableInstance::Unwrap(args->executable)
```
mrakitaTT, line 41: Unwrap once into `LoadedExecutableInstance *` variable, and then use it

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -77,23 +82,34 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+        buffer->device().device_description()->device_id();
+    device_ids.insert(chip_ids[buffer_device_id]);
     DLOG_F(INFO, "Runtime input id: %d", buffer->unique_id());
   }
 
+  assert(device_ids.size() == 1);
+
+  std::vector<int> device_ids_vector(device_ids.begin(), device_ids.end());
+
+  tt::runtime::Device device = tt::runtime::openDevice(device_ids_vector);
```
mrakitaTT, line 111: reminder to talk about this too

inc/common/pjrt_implementation/executable_image.h
ajakovljevicTT: This parameter is called `num_addressable_devices` in the PJRT API, so it might be confusing to name it differently in our implementation.

src/common/module_builder.h
ajakovljevicTT: Discussed offline, closing

src/common/pjrt_implementation/loaded_executable_instance.cc
ajakovljevicTT: Discussed offline, closing

inc/common/pjrt_implementation/device_description.h
ajakovljevicTT: Fixed after offline discussion

PR #165, opened 2025-01-15
Removed hardcoded device description debug strings

### LINKED ISSUE #125: Revisit DeviceDescripton class implementation
In our class `DeviceDescription` we have some hardcoded values for private attributes and their matching getters which should be better understood and revised. 

[Here](https://github.com/openxla/xla/blob/2da0f3ed4ba9cb4159f9f2bd6f20b32c71ed6b97/xla/pjrt/pjrt_device_description.h#L67) is how XLA defines them for `PjRtDeviceDescription`.

sdjukicTT:
Instead of a hardcoded string that describes the device architecture ("Wormhole"), we get it from the system description.

`kind_string()` is changed to return that string instead of "TTDevice" because it better fits the description of what `device_kind` represents:  a vendor-dependent string that uniquely identifies the kind of device, e.g., "Tesla V100-SXM2-16GB".

closes #125


inc/common/pjrt_implementation/device_instance.h
```diff
@@ -8,6 +8,7 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 // https://llvm.org/LICENSE.txt
 
+#include "types_generated.h"
```
kmitrovicTT, line 11: Needed?

inc/common/pjrt_implementation/device_description.h
```diff
@@ -32,14 +49,8 @@ class DeviceDescription {
...
-  std::string_view debug_string() { return to_string(); }
-  std::string_view to_string() {
-    std::stringstream ss;
-    ss << kind_string_ << "(id=" << device_id() << ", arch=" << arch_string_
-       << ")";
-    user_string_ = ss.str();
-    return user_string_;
-  }
+  std::string_view debug_string() { return user_string_; }
+  std::string_view to_string() { return user_string_; }
```
kmitrovicTT, line 53: Do we need both of these if they return the same?

inc/common/pjrt_implementation/device_instance.h
sdjukicTT: You're right, it isn't needed, it is included through `device_description.h`. I removed it.

inc/common/pjrt_implementation/device_description.h
sdjukicTT: We might want to make them different in the future, they map to different PJRT functions.

inc/common/pjrt_implementation/device_description.h
```diff
@@ -10,6 +10,7 @@
 
 #include <sstream>
 
+#include "types_generated.h"
```
kmitrovicTT, line 13: Needed?

kmitrovicTT (approved):

inc/common/pjrt_implementation/device_description.h
```diff
@@ -20,7 +21,13 @@ namespace tt::pjrt {
 class DeviceDescription {
 
 public:
-  DeviceDescription(int32_t client_id) : client_id_(client_id) {};
+  DeviceDescription(int32_t client_id, tt::target::Arch arch)
```
ajakovljevicTT, line 24: Move to the .cc file

sdjukicTT: Moved.

inc/common/pjrt_implementation/device_description.h
sdjukicTT: Yes, for `tt::target::Arch` and `tt::target::EnumNameArch`.

src/common/pjrt_implementation/device_description.cc
```diff
@@ -14,6 +14,13 @@
 
 namespace tt::pjrt {
 
+DeviceDescription::DeviceDescription(int32_t client_id, tt::target::Arch arch)
+    : client_id_(client_id) {
+  kind_string_ = target::EnumNameArch(arch);
+  std::stringstream ss;
+  ss << "TTDevice(id=" << device_id() << ", arch=" << kind_string_ << ")";
+  user_string_ = ss.str();
+}
```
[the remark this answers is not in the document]
sdjukicTT: Added.

ajakovljevicTT (approved):

inc/common/pjrt_implementation/device_description.h
mrakitaTT, line 13: This include is from tt-mlir, let's move it into separate group with `// tt-mlir includes` comment

inc/common/pjrt_implementation/device_description.h
mrakitaTT, line 1: I know this isn't in the scope of this PR, but since this file is very small and we are now changing most of it, can you please do these changes too:
- prefix all member variables with `m_`
- change function names to be in camel case starting with first letter small (like `toString`)
- add blank lines between variables declarations, and functions declarations

inc/common/pjrt_implementation/device_description.h
```diff
@@ -32,14 +33,8 @@ class DeviceDescription {
   }
 
   std::string_view kind_string() { return kind_string_; }
```
mrakitaTT, line 35: I would rename this function to `getDeviceKind`. Please also add a header comment:
```
Returns a vendor-dependent string that uniquely identifies the kind of device, e.g. `Wormhole_b0`.
```

inc/common/pjrt_implementation/device_description.h
```diff
@@ -32,14 +33,8 @@ class DeviceDescription {
...
-  std::string_view debug_string() { return to_string(); }
-  std::string_view to_string() {
-    std::stringstream ss;
-    ss << kind_string_ << "(id=" << device_id() << ", arch=" << arch_string_
-       << ")";
-    user_string_ = ss.str();
-    return user_string_;
-  }
+  std::string_view debug_string() { return user_string_; }
```
mrakitaTT, line 36: I would rename this function to `toDebugString`. Please also add a header comment:
```
Returns a debug string suitable for logging when errors occur. Should be verbose enough to describe the current device unambiguously.
```

inc/common/pjrt_implementation/device_description.h
```diff
@@ -32,14 +33,8 @@ class DeviceDescription {
...
+  std::string_view to_string() { return user_string_; }
```
mrakitaTT, line 37: I would rename this function to `toString`. Please also add a header comment:
```
Returns a device description string suitable for reading by end users, should be reasonably terse.
```

inc/common/pjrt_implementation/device_description.h
```diff
@@ -51,11 +46,8 @@ class DeviceDescription {
 private:
   int client_id_;
 
-  // TODO We should understand better how these are used.
-  // See https://github.com/tenstorrent/tt-xla/issues/125
-  std::string kind_string_ = "TTDevice";
-  std::string arch_string_ = "Wormhole";
-  std::string user_string_ = "";
+  std::string kind_string_;
```
mrakitaTT, line 49: I would rename this to `m_device_kind`

inc/common/pjrt_implementation/device_description.h
mrakitaTT, line 37: Please also change all three functions to return `std::string` instead of `std::string_view`, no need to return string views, they will be created from those strings in the PJRT API calls.

src/common/pjrt_implementation/device_description.cc
```diff
@@ -14,20 +14,28 @@
 
 namespace tt::pjrt {
 
+DeviceDescription::DeviceDescription(int32_t client_id, tt::target::Arch arch)
+    : m_client_id(client_id) {
+  m_device_kind = tt::target::EnumNameArch(arch);
```
mrakitaTT, line 19: Move into initializer list above (with `m_client_id`). It is a good practice because it avoids reinitialization, if the non-primitive-type field is not initialized in the initializer list then it will be initialized with default constructor first, before constructor body is executed where you then reinitialize it.

src/common/pjrt_implementation/device_description.cc
```diff
@@ -14,20 +14,28 @@
...
+  std::stringstream ss;
+  ss << "TTDevice(id=" << deviceId() << ", arch=" << m_device_kind << ")";
```
mrakitaTT, line 21: Seems like you used some non-unicode characters here, maybe copy/paste somehow caused it, try to retype these characters

mrakitaTT (approved):

inc/common/pjrt_implementation/device_description.h
```diff
@@ -31,31 +34,31 @@ class DeviceDescription {
...
-  std::string_view to_string() {
-    std::stringstream ss;
-    ss << kind_string_ << "(id=" << device_id() << ", arch=" << arch_string_
-       << ")";
-    user_string_ = ss.str();
-    return user_string_;
-  }
+  // Returns a vendor-dependent string that uniquely identifies the kind of
+  // device, e.g. `Wormhole_b0`.
+  std::string &getDeviceKind() { return m_device_kind; }
```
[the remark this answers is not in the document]
mrakitaTT: In other two functions below too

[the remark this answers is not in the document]
sdjukicTT: Added

inc/common/pjrt_implementation/device_description.h
```diff
@@ -31,31 +34,31 @@ class DeviceDescription {
...
 
   // TODO
-  int64_t device_id() { return 0; }
+  int64_t deviceId() { return 0; }
 
-  int client_id() { return client_id_; }
+  int clientId() { return m_client_id; }
 
-  int process_index() { return 0; }
+  int processIndex() { return 0; }
```
mrakitaTT, line 54: For these you can just add `const` after function name since they return primitive types, like:
```
int64_t deviceId() const { return 0; }
```

And please prefix them with `get` while you are changing them already. Sorry for dragging your PR with this!

src/common/pjrt_implementation/device_description.cc
sdjukicTT: I think the issue was that `string_view` was initialized by a temporary `string` value returned by methods `getDeviceKind()`, `toDebugString()` and `toString()`, which it outlived.

mrakitaTT: Good catch!

mrakitaTT (approved):

PR #174, opened 2025-01-17
Support for superset integration

kmitrovicTT:
- Recorded relevant properties of op and model tests
- Added compile/runtime fail to xfail/skip messages of model inference tests
- Split test_compare into multiple functions
- Added lambda function in each parametrization to generate more informative test descriptions (instead of `some_test[xshape0-yshape0]` it will report `some_test[(32, 32)-(32, 32)]`).

### DIFF
--- a/tests/jax/models/albert/v2/base/test_albert_base.py
+++ b/tests/jax/models/albert/v2/base/test_albert_base.py
@@ -11,7 +11,7 @@ from utils import record_model_test_properties, runtime_fail
 from ..tester import AlbertV2Tester
 
 MODEL_PATH = "albert/albert-base-v2"
-MODEL_NAME = MODEL_PATH.split("/")[1]
+MODEL_NAME = "albert-v2-base"
 
 
 # ----- Fixtures -----
--- a/tests/jax/models/albert/v2/base/test_albert_base.py
+++ b/tests/jax/models/albert/v2/base/test_albert_base.py
@@ -2,12 +2,16 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+from typing import Callable
+
 import pytest
 from infra import RunMode
+from utils import record_model_test_properties, runtime_fail
 
 from ..tester import AlbertV2Tester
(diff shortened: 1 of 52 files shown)


tests/jax/models/albert/v2/base/test_albert_base.py
```diff
@@ -2,12 +2,16 @@
...
+from typing import Callable
+
 import pytest
 from infra import RunMode
+from utils import record_model_test_properties, runtime_fail
 
 from ..tester import AlbertV2Tester
 
 MODEL_PATH = "albert/albert-base-v2"
+MODEL_NAME = MODEL_PATH.split("/")[1]
```
mrakitaTT, line 14: I don't like coupling model name with model path, nor using `class.__qualname__`. We should just manually define `MODEL_NAME` string in each test, by combining names of the parent folders, for example in this case: `MODEL_NAME = "albert-v2-base"`.

mrakitaTT (approved):

tests/jax/models/mnist/mlp/test_mnist_mlp.py
```diff
@@ -78,12 +79,18 @@ def training_tester(request) -> MNISTMLPTester:
 )
 def test_mnist_inference(
     inference_tester: MNISTMLPTester,
+    record_tt_xla_property: Callable,
```
mrakitaTT, line 82: I've just realized that this test is parametrized, so it should also have `ids=lambda val: f"{val}"` so we can see proper version that failed in the tabular views. Please also rename the test to `test_mnist_mlp_inference`, and below to `test_mnist_mlp_training`.

kmitrovicTT: Done.

tests/jax/models/albert/v2/base/test_albert_base.py
kmitrovicTT: Done.

PR #175, opened 2025-01-17
Tests for dot_general op

umalesTT:


tests/jax/ops/test_dot_general.py
```diff
@@ -11,14 +11,14 @@
...
+        [(1, 32), (1, 32)],
+        [(1, 32, 64), (1, 32, 32)],
+        [(2, 32, 64), (2, 32, 64)],
+        [(2, 16, 32, 64), (2, 16, 64, 32)],
     ],
 )
 def test_dot_general(x_shape: tuple, y_shape: tuple):
     def dot_general(x: jax.Array, y: jax.Array) -> jax.Array:
-        return jnp.dot(x, y)
+        return jax.lax.dot_general(x, y, dimension_numbers=((1, 1), (0, 0)))
```
kmitrovicTT, line 22: We need to thoroughly cover every edge case for this op. For example, now it is missing the cases where it works like a regular matmul. Please add those cases as well, parametrize `dimension_numbers` if needed, and add short comments above parameter cases which better describe what that test case is trying to hit.

umalesTT: In a most recent commit, I coded additional tests and organized them in 3 groups, with each group commented.
I did this because I found that parametrizing dimension_numbers is not trivial.

kmitrovicTT (approved):

PR #180, opened 2025-01-20
Mark failing tests as xfail and unmark passing tests

### LINKED ISSUE #169: Mark tests as xfail instead of skip
For tests that we skipped due to an issue, we should mark as xfail, which will unexpectedly pass when the issue gets fixed, either by uplift or direct fix commit.

sgligorijevicTT:
I have marked all failing tests I could as xfail. There is a single hard segfault in model tests that has to remain skipped. I have also updated failure reason where appropriate.

Fixes #169


tests/jax/models/llama/tester.py
```diff
@@ -21,8 +21,8 @@ def __init__(
         comparison_config: ComparisonConfig = ComparisonConfig(),
         run_mode: RunMode = RunMode.INFERENCE,
     ) -> None:
-        super().__init__(comparison_config, run_mode)
         self._model_name = model_name
+        super().__init__(comparison_config, run_mode)
```
mrakitaTT, line 25: Is this change necessary?

mrakitaTT (approved):

tests/jax/models/llama/tester.py
sgligorijevicTT: Yes, it is. The constructor of the base class reads the `_model_name` field, so it has to be initialized before calling `super()`. I'm not sure how I didn't catch this in testing when I sent the original PR, maybe it got messed up during a refactor and I forgot to unmark as skip?

sgligorijevicTT:
Good news, the most recent uplift fixed a few tests.
Bad news, that same uplift broke a test that was just unmarked as skip in this PR 😔 .

PR #183, opened 2025-01-21
Fix MLPMixer test

sgligorijevicTT:
In my original test implementation, the model was instantiated slightly differently from how that is done in the original repo. This PR changes that. It also has the side effect of bypassing the error in MLIR and allowing the model to compile.

### DIFF
--- a/tests/jax/models/mlpmixer/test_mlpmixer.py
+++ b/tests/jax/models/mlpmixer/test_mlpmixer.py
@@ -88,7 +88,12 @@ def training_tester() -> MlpMixerTester:
 
 
 @pytest.mark.skip(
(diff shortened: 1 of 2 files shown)


tests/jax/models/mlpmixer/test_mlpmixer.py
```diff
@@ -93,8 +88,8 @@ def training_tester() -> MlpMixerTester:
 
 
 @pytest.mark.skip(
-    reason="error: failed to legalize operation 'ttir.convolution' that was explicitly marked illegal"
-)
+    reason="Statically allocated circular buffers in program 16 clash with L1 buffers on core range [(x=0,y=0) - (x=6,y=0)]. L1 buffer allocated at 475136 and static circular buffer region ends at 951136"
+)  # segfault
```
kmitrovicTT, line 92: Raise an issue for this and link it here.

kmitrovicTT (approved):

PR #189, opened 2025-01-24
Make pcc allclose tunable for each test. Add test for MLP training.

umalesTT:
When tensors are too close, pcc will output NaN values. Therefore, for each test it should be possible to separately tune the threshold of allclose.rtol and allclose.atol below which pcc won't be calculated and therefore test will be able to pass. 

This is applied in test for MLP neural network training with MSE loss and SGD optimizer.


tests/infra/comparison.py
```diff
@@ -34,17 +34,18 @@ class AtolConfig(ConfigBase):
...
 @dataclass
 class AllcloseConfig(ConfigBase):
     rtol: float = 1e-2
     atol: float = 1e-2
 
 
+@dataclass
+class PccConfig(ConfigBase):
+    required_pcc: float = 0.99
+    allclose: AllcloseConfig = AllcloseConfig()
```
kmitrovicTT, line 46: Just leave a comment above this explaining why we need it.

kmitrovicTT (approved):

tests/infra/comparison.py
umalesTT: Done

PR #190, opened 2025-01-27
Correct the tt-mlir path in get-docker-tag.sh

vmilosevic:
Script get-docker-tag.sh was incorrectly generating docker tag, it didn't take into account tt-mlir docker tag change.
Correct the path to match path where cmake ExternalProject_Add clones tt-mlir to gix the docker tag.

### DIFF
--- a/.github/get-docker-tag.sh
+++ b/.github/get-docker-tag.sh
@@ -11,13 +11,15 @@ set -e
 
 # Execute this in a separate bash process
 (
-    # Read tt-mlir version from third_party/CMakeLists.txt and clone third_party/tt-mlir
+    # Read tt-mlir version from third_party/CMakeLists.txt
+    # clone tt-mlir version to tmp/third_party/tt-mlir
     # Get the MLIR docker tag
+    TT_MLIR_PATH=third_party/tt-mlir/src/tt-mlir
     TT_MLIR_VERSION=$(grep -oP 'set\(TT_MLIR_VERSION "\K[^"]+' third_party/CMakeLists.txt)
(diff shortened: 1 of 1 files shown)


.github/get-docker-tag.sh
```diff
@@ -8,26 +8,24 @@
...
+MLIR_DOCKER_TAG=$(
+    # Read tt-mlir version from third_party/CMakeLists.txt
+    # clone tt-mlir version to tmp/third_party/tt-mlir
     # Get the MLIR docker tag
+    TT_MLIR_PATH=tmp/third_party/tt-mlir
     TT_MLIR_VERSION=$(grep -oP 'set\(TT_MLIR_VERSION "\K[^"]+' third_party/CMakeLists.txt)
-    if [ ! -d "third_party/tt-mlir" ]; then
-        git clone https://github.com/tenstorrent/tt-mlir.git third_party/tt-mlir --quiet
+    if [ ! -d $TT_MLIR_PATH ]; then
+        git clone https://github.com/tenstorrent/tt-mlir.git $TT_MLIR_PATH --quiet
```
mrakitaTT, line 18: Maybe we can wget only files that `get-docker-tag.sh` needs to avoid cloning the whole repo?

vmilosevic: I thought about this, but it is a bit tricky because we want to call script github/get-docker-tag.sh from `mlir` and it can take into account multiple files. It looks at mlir docker files and env/CmakeList.txt but it could be changed to look at others.
Mlir checkout is not that big ~20MB, but if it grows much larger we can look for other options.

PR #197, opened 2025-01-28
Refactoring of ModuleBuilder class

sdjukicTT:
### Problem description
Vector `m_is_output_scalar`, that for every output of a binary program, holds if the type is scalar, is related to a binary program and therefore to a specific `ExecutableImage`. We don't store it there with the binary program, but only in `ModuleBuilder` class. 
During `LoadedExecutableInstance::Execute` it is accessed through `ClientInstance` which has a `ModuleBuilder`, and this assumes the `m_is_output_scalar` is the same one we made during `ClientInstance::Compile` when we made the binary. But this can we false, if another program was compiled with the same Client in the meantime, the values are invalid because they have been overwritten.

### What's changed
In `ModuleBuilder`, `m_flatbuffer_binary` is changed to be `tt::runtime::Binary` instead of `std::shared_ptr<void>`. This is because we always cast it to `tt::runtime::Binary`, which has the shared pointer handle inside.

`m_is_output_scalar` has a getter method, so it could be passed to `ExecutableImage` during `ClientInstance::Compile` and stored there together with the binary program. Now in `LoadedExecutableInstance::Execute` we check if a specific output is scalar through the executable image, and not through the client, which solves the overwriting problem.

`m_num_inputs` and `m_num_outputs` are removed from `ModuleBuilder` because they are used only to check if the number of outputs matches the expected number (from `m_is_output_scalar`) and to be passed to `ExecutableImage`. Now `ExecutableImage` gets the number of inputs and outputs from the binary, and does the same check.

### Checklist

### DIFF
--- a/inc/common/pjrt_implementation/executable_image.h
+++ b/inc/common/pjrt_implementation/executable_image.h
@@ -33,38 +33,38 @@ public:
   }
   static void BindApi(PJRT_Api *api);
 
-  void AddRef() { ref_count.fetch_add(1); }
+  void AddRef() { m_ref_count.fetch_add(1); }
   void DecRef() {
-    if (ref_count.fetch_sub(1) == 0) {
+    if (m_ref_count.fetch_sub(1) == 0) {
       delete this;
     }
   }
 
-  const size_t get_arg_count() const { return arg_count; }
+  const size_t get_arg_count() const { return m_arg_count; }
 
-  const size_t get_result_count() const { return result_count; }
+  const size_t get_result_count() const { return m_result_count; }
 
-  const tt::runtime::Binary &get_binary() const { return binary; }
+  const tt::runtime::Binary &get_binary() const { return m_binary; }
 
-  const std::string &get_code() const { return code; }
+  const std::string &get_code() const { return m_code; }
 
   bool isOutputScalar(size_t index) const;
 
 private:
   // The reference count. Must be disposed when reaching zero.
-  std::atomic<int> ref_count;
+  std::atomic<int> m_ref_count;
 
   // Raw compiler output.
-  tt::runtime::Binary binary;
+  tt::runtime::Binary m_binary;
 
   // Original code fed to the compiler. Stored for debugging.
-  const std::string code;
+  const std::string m_code;
 
-  size_t arg_count;
-  size_t result_count;
+  size_t m_arg_count;
+  size_t m_result_count;
 
   // For every output, holds if the type is a scalar or not.
-  std::vector<bool> is_output_scalar;
+  std::vector<bool> m_is_output_scalar;
 };
 
 } // namespace tt::pjrt
--- a/inc/common/pjrt_implementation/executable_image.h
+++ b/inc/common/pjrt_implementation/executable_image.h
@@ -12,6 +12,7 @@
 #include <memory>
 #include <string>
 
+#include "tt/runtime/types.h"
 #include "xla/pjrt/c/pjrt_c_api.h"
 
 #ifndef TT_XLA_INC_COMMON_PJRT_IMPLEMENTATION_EXECUTABLE_IMAGE_H_
@@ -22,10 +23,10 @@ namespace tt::pjrt {
 class ExecutableImage {
 
 public:
-  ExecutableImage(std::shared_ptr<void> binary, std::string code,
+  ExecutableImage(const tt::runtime::Binary &binary, std::string code,
                   size_t arg_count, size_t result_count)
-      : ref_count(1), binary(std::move(binary)), code(code),
-        arg_count(arg_count), result_count(result_count) {}
+      : ref_count(1), binary(binary), code(code), arg_count(arg_count),
+        result_count(result_count) {}
   operator PJRT_Executable *() {
     return reinterpret_cast<PJRT_Executable *>(this);
   }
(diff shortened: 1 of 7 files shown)


inc/common/pjrt_implementation/executable_image.h
```diff
@@ -45,22 +44,27 @@ class ExecutableImage {
...
+  tt::runtime::Binary binary;
 
   // Original code fed to the compiler. Stored for debugging.
   const std::string code;
 
   size_t arg_count;
   size_t result_count;
+
+  // For every output, holds if the type is a scalar or not.
+  std::vector<bool> is_output_scalar;
```
ajakovljevicTT, line 67: I know it was not the scope of this PR, but can you put the `m_` markers for class fields in the `ExecutableImage` class? We eventually should do it for all classes, but lets start with this one, since you already touched it.

sdjukicTT: Added.

mrakitaTT:
I plan to take a look tonight/tomorrow, please wait on me :)

ajakovljevicTT:
It seems that the rebase to main added multiple commits and changed additional files. Try to squash and force push.

ajakovljevicTT (approved):

src/common/pjrt_implementation/executable_image.cc
```diff
@@ -121,4 +137,9 @@ void ExecutableImage::BindApi(PJRT_Api *api) {
   };
 }
 
+bool ExecutableImage::isOutputScalar(const size_t index) const {
+  assert(index < is_output_scalar.size() && "Output index out of range");
```
umalesTT, line 141: Shouldn't it be m_is_output_scalar in this assert?

umalesTT, line 141: Shouldn't it be m_is_output_scalar in this assert?

umalesTT: I overlooked commit 7845a7b, sorry.

src/common/module_builder.cc
```diff
@@ -219,23 +214,11 @@ void ModuleBuilder::createFlatbufferBinary(
     const mlir::OwningOpRef<mlir::ModuleOp> &mlir_module) {
   m_flatbuffer_binary = mlir::tt::ttnn::ttnnToFlatbuffer(mlir_module.get());
 
-  if (m_flatbuffer_binary == nullptr) {
+  if (m_flatbuffer_binary.handle == nullptr) {
     DLOG_F(ERROR, "Failed to generate flatbuffer binary");
     m_status = tt_pjrt_status::kInternal;
     return;
```
mrakitaTT, line 220: `return` is not needed here anymore

src/common/module_builder.h
```diff
@@ -17,6 +17,7 @@
 
 // tt-xla includes
 #include "status.h"
+#include "tt/runtime/types.h"
```
mrakitaTT, line 20: This one is from tt-mlir so it should be under `// tt-mlir includes` group, above tt-xla includes. [Here](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) is the explanation regarding the order of includes.

src/common/pjrt_implementation/executable_image.cc
```diff
@@ -19,6 +19,22 @@ namespace tt::pjrt {
...
+ExecutableImage::ExecutableImage(const tt::runtime::Binary &binary,
+                                 std::string code,
+                                 const std::vector<bool> &is_output_scalar)
+    : m_ref_count(1), m_binary(binary), m_code(code),
+      m_is_output_scalar(is_output_scalar) {
+  m_arg_count = m_binary.getProgramInputs(0).size();
+  m_result_count = m_binary.getProgramOutputs(0).size();
+
+  if (m_result_count != m_is_output_scalar.size()) {
+    DLOG_F(ERROR,
```
mrakitaTT, line 31: Please add a TODO comment above this log that we should throw error instead, otherwise execution will continue and then crash later. We should rethink error handling in tt-xla, maybe we can have a specialized error class inheriting `std::runtime_error` which contains field to hold `tt_pjrt_status`, so we throw it from internal logic and then catch it in top level API functions and return status from the exception. FYI @ajakovljevicTT

src/common/pjrt_implementation/executable_image.cc
```diff
@@ -48,7 +64,7 @@ void ExecutableImage::BindApi(PJRT_Api *api) {
       +[](PJRT_Executable_NumOutputs_Args *args) -> PJRT_Error * {
     DLOG_F(LOG_DEBUG, "ExecutableImage::PJRT_Executable_NumOutputs");
     ExecutableImage *exec = ExecutableImage::Unwrap(args->executable);
-    args->num_outputs = exec->result_count;
+    args->num_outputs = exec->m_result_count;
```
mrakitaTT, line 67: Use getters to access fields, in other places below too

src/common/pjrt_implementation/executable_image.cc
```diff
@@ -19,6 +19,22 @@ namespace tt::pjrt {
 
 const std::string_view kMlirFormat = "mlir";
 
+ExecutableImage::ExecutableImage(const tt::runtime::Binary &binary,
+                                 std::string code,
+                                 const std::vector<bool> &is_output_scalar)
+    : m_ref_count(1), m_binary(binary), m_code(code),
+      m_is_output_scalar(is_output_scalar) {
+  m_arg_count = m_binary.getProgramInputs(0).size();
+  m_result_count = m_binary.getProgramOutputs(0).size();
```
mrakitaTT, line 28: Initialize these too with other fields above in initializer list (just use `binary` instead of `m_binary`)

mrakitaTT (approved):

Issue #203, opened 2025-01-29
Add more MNIST tests


mrakitaTT:
We should add MNIST mlp test, [this](https://github.com/tenstorrent/tt-xla/blob/c633eb23b9f2e9eba079812e2b0643556bb33c10/tests/TTIR/test_mnist.py#L63) is how it looked like in the old infra.

We should also add another MNIST cnn test following [this](https://github.com/tenstorrent/tt-torch/blob/main/tests/models/mnist/test_mnist.py#L15) architecture.

sgligorijevicTT:
2/2 PRs addressing this were merged, closing

PR #205, opened 2025-01-29
Add MNIST MLP test

sgligorijevicTT:
### Ticket
This addresses one part of #203 

### Problem description
Test for a simple MLP for MNIST was missing

### What's changed
Added a new test

### Checklist


tests/jax/models/mnist/mlp/test_mnist_mlp.py
```diff
@@ -0,0 +1,90 @@
...
+    "inference_tester",
+    [
+        (128,),
+        (128, 128),
+        (192, 128),
+        (128, 128, 128),
+        (256, 128, 64),
+    ],
+    indirect=True,
+)
```
mrakitaTT, line 76: I think this will bias our model metrics dashboard into thinking we are running 5 models (@kmitrovicTT to confirm). We will have multiple tests per model in the future (for different NLP tasks) so we should think of a way to count them all together as 1 model passing in our high level model metrics.

If this is not easily achievable in the meantime we can just choose one configuration for the MNIST MLP model, for example same as [forge model](https://github.com/tenstorrent/tt-forge-fe/blob/d4079889ad0f019914d1351d769ee9568b0da6c6/forge/test/mlir/mnist/utils.py#L17), and comment out the rest. Btw please add anyways one parametrization that will match the Forge model.

sgligorijevicTT: Fair point, I will comment out the current configurations and add the same one as forge as the only one.

mrakitaTT: Talked with Kristijan, parametrization won't be counted as 5 models since the test name will still be the same, sorry about false alarm, you can uncomment the parametrization :)

If we had multiple tests per model that would be counted as multiple models, but we have a (hacky) way to support it in the future by parsing model name from test file path. We won't yet implement that because it is not yet clear if we are going to keep using the same data pipeline logic.

tests/jax/models/mnist/mlp/test_mnist_mlp.py
```diff
@@ -0,0 +1,89 @@
...
+    return MNISTMLPTester(request.param, run_mode=RunMode.TRAINING)
+
+
+# ----- Tests -----
+
+
+@pytest.mark.parametrize(
+    "inference_tester",
+    [
+        (512,),
```
mrakitaTT, line 70: Shouldn't it be (512, 512) per [Forge implementation](https://github.com/tenstorrent/tt-forge-fe/blob/d4079889ad0f019914d1351d769ee9568b0da6c6/forge/test/mlir/mnist/utils.py#L17)?

sgligorijevicTT: Oopsie, yeah.

tests/jax/models/mnist/mlp/test_mnist_mlp.py
```diff
@@ -0,0 +1,89 @@
...
+
+@pytest.mark.parametrize(
+    "inference_tester",
+    [
+        (512, 512),
+        # (128,),
+        # (128, 128),
+        # (192, 128),
+        # (128, 128, 128),
+        # (256, 128, 64),
```
mrakitaTT, line 75: You can uncomment this now (more info [here](https://github.com/tenstorrent/tt-xla/pull/205#discussion_r1937090814)), and then please move the `(512, 512),` below `(192, 128),`

mrakitaTT (approved):

mrakitaTT (approved):

Issue #202, opened 2025-01-29
Switch runtime tensor creation to owned buffers


ajakovljevicTT:
Currently, in the PJRT backend, we create tensors which are just borrowed, giving us issues when used potentially deallocated memory. As runtime now has an api for create tensors owned by the runtime, we should switch to that one.

Issue #200, opened 2025-01-29
Unflattened for loop gives undefined behavior after its first iteration, when run on device


umalesTT:
This issue appeared during MLP training for MNIST dataset, however for clarity sake below is the simplified version of the code with just minimizing the function x^2 using SGD - issue is the same for both. 
In this first code, with unflattened for loop, after one or two iterations behavior becomes undefined and final f(x) is calculated to be inf (while cpu gives correct results). 

```
import jax
import jax.numpy as jnp
from jax import random

import os
import sys
import jax._src.xla_bridge as xb

def init_device():
    backend = "tt"
    path = os.path.join(os.path.dirname(__file__), "../build/src/tt/pjrt_plugin_tt.so")
    if not os.path.exists(path):
        raise FileNotFoundError(f"Could not find tt_pjrt C API plugin at {path}")

    print("Loading tt_pjrt C API plugin", file=sys.stderr)
    xb.discover_pjrt_plugins()

    plugin = xb.register_plugin("tt", priority=500, library_path=path, options=None)
    print("Loaded", file=sys.stderr)
    jax.config.update("jax_platforms", "tt,cpu")

init_device()
current_device = jax.devices()[0]

def f(x):
    return x**2

@jax.jit
def sgd_optimization(x, learning_rate):
    grad = jax.grad(f)(x)
    x = x - learning_rate * grad
    return x

x_init = 5.0  # Starting point
learning_rate = 0.1
num_iterations = 10

x = x_init

for i in range(num_iterations):
    x_old = x
    x = sgd_optimization(x, learning_rate)
    print(f"Iteration {i}, x = {x}, f(x) = {f(x)}")

# Final result
print(f"Optimized x = {x}, f(x) = {f(x)}")
```

However, if the for loop is flattened and implemented in the below fashion, results are correct and coincide with those of cpu. 

```
import jax
import jax.numpy as jnp
from jax import random

import os
import sys
import jax._src.xla_bridge as xb

def init_device():
    backend = "tt"
    path = os.path.join(os.path.dirname(__file__), "../build/src/tt/pjrt_plugin_tt.so")
    if not os.path.exists(path):
        raise FileNotFoundError(f"Could not find tt_pjrt C API plugin at {path}")

    print("Loading tt_pjrt C API plugin", file=sys.stderr)
    xb.discover_pjrt_plugins()

    plugin = xb.register_plugin("tt", priority=500, library_path=path, options=None)
[...]

PR #220, opened 2025-01-31
Changing tensor creation

ajakovljevicTT:
### Ticket
Fixes https://github.com/tenstorrent/tt-xla/issues/202
Fixes https://github.com/tenstorrent/tt-xla/issues/200

### Problem description
Since we were given raw pointers to make tensors for in PJRT, we had no control over their ownership. This PR allocates new memory and copies the contents of the pointer, so now PJRT has full control of its memory.

### DIFF
--- a/inc/common/pjrt_implementation/buffer_instance.h
+++ b/inc/common/pjrt_implementation/buffer_instance.h
@@ -23,7 +23,7 @@ class DeviceInstance;
 
 class BufferInstance {
 public:
-  BufferInstance(DeviceInstance &device, tt::runtime::Tensor tensor,
+  BufferInstance(DeviceInstance &device, std::unique_ptr<tt::runtime::Tensor>& tensor,
                  std::vector<std::uint32_t> shape,
                  std::vector<std::uint32_t> stride);
   BufferInstance(DeviceInstance &device);
@@ -44,7 +44,7 @@ public:
     // the hook to get an unsafe pointer (avoids a copy).
     return false;
   }
-  tt::runtime::Tensor tensor() { return tensor_.value(); }
+  tt::runtime::Tensor tensor() { return *tensor_; }
 
   PJRT_Error *GetMemoryLayout(PJRT_Buffer_GetMemoryLayout_Args *args);
   // Gets the required host size in bytes to copy to host.
@@ -74,7 +74,7 @@ private:
   // API elements that must have the same lifetime as BufferInstance.
   std::vector<int64_t> dims_;
   std::vector<std::uint32_t> stride_;
-  std::optional<tt::runtime::Tensor> tensor_;
+  std::unique_ptr<tt::runtime::Tensor> tensor_;
 
   std::vector<int64_t> minor_to_major_;
   std::vector<int64_t> tile_dims_;
--- a/inc/common/pjrt_implementation/buffer_instance.h
+++ b/inc/common/pjrt_implementation/buffer_instance.h
@@ -26,7 +26,8 @@ public:
   BufferInstance(DeviceInstance &device,
                  std::unique_ptr<tt::runtime::Tensor> &tensor,
                  std::vector<std::uint32_t> shape,
-                 std::vector<std::uint32_t> stride);
+                 std::vector<std::uint32_t> stride,
+                 std::shared_ptr<void> host_buffer_ptr);
   BufferInstance(DeviceInstance &device);
   ~BufferInstance();
   operator PJRT_Buffer *() { return reinterpret_cast<PJRT_Buffer *>(this); }
@@ -85,7 +86,11 @@ private:
   std::optional<PJRT_Buffer_Type> DataType;
 
   // OnReady event - currently not used.
-  EventInstance *on_ready_event_;
+  std::shared_ptr<EventInstance> on_ready_event_;
+
+  // Pointer to the host memory used to create this buffer, if buffer is created
+  // on device, the value of this pointer is nullptr.
+  std::shared_ptr<void> host_buffer_ptr_;
 };
 
 } // namespace tt::pjrt
--- a/inc/common/pjrt_implementation/buffer_instance.h
+++ b/inc/common/pjrt_implementation/buffer_instance.h
@@ -91,11 +91,12 @@ private:
   // OnReady event - currently not used.
   EventInstance *on_ready_event_;
 
-  // Pointer to the host memory used to create this buffer, if buffer is created
+  // Pointer to the host memory used to create this buffer. 
+  // If buffer is created
   // on device, the value of this pointer is nullptr. It is necessary to keep
   // track of this memory since the runtime will not clean it, and we need to
   // pass the shared pointer to the runtime.
-  std::shared_ptr<void> host_buffer_ptr_ = nullptr;
+  std::shared_ptr<void> host_buffer_ptr_;
 };
 
 } // namespace tt::pjrt
--- a/inc/common/pjrt_implementation/buffer_instance.h
+++ b/inc/common/pjrt_implementation/buffer_instance.h
@@ -91,7 +91,7 @@ private:
   // OnReady event - currently not used.
   EventInstance *on_ready_event_;
 
-  // Pointer to the host memory used to create this buffer. 
+  // Pointer to the host memory used to create this buffer.
   // If buffer is created
   // on device, the value of this pointer is nullptr. It is necessary to keep
   // track of this memory since the runtime will not clean it, and we need to
--- a/inc/common/pjrt_implementation/buffer_instance.h
+++ b/inc/common/pjrt_implementation/buffer_instance.h
@@ -23,7 +23,8 @@ class DeviceInstance;
 
 class BufferInstance {
 public:
-  BufferInstance(DeviceInstance &device, std::unique_ptr<tt::runtime::Tensor>& tensor,
+  BufferInstance(DeviceInstance &device,
+                 std::unique_ptr<tt::runtime::Tensor> &tensor,
                  std::vector<std::uint32_t> shape,
                  std::vector<std::uint32_t> stride);
   BufferInstance(DeviceInstance &device);
--- a/inc/common/pjrt_implementation/buffer_instance.h
+++ b/inc/common/pjrt_implementation/buffer_instance.h
@@ -23,10 +23,13 @@ class DeviceInstance;
 
 class BufferInstance {
 public:
-  BufferInstance(DeviceInstance &device,
-                 std::unique_ptr<tt::runtime::Tensor> &tensor,
-                 std::vector<std::uint32_t> shape,
-                 std::vector<std::uint32_t> stride,
+  BufferInstance(DeviceInstance &device, tt::runtime::Tensor &tensor,
+                 const std::vector<std::uint32_t> &shape,
+                 const std::vector<std::uint32_t> &stride);
+
+  BufferInstance(DeviceInstance &device, tt::runtime::Tensor &tensor,
+                 const std::vector<std::uint32_t> &shape,
+                 const std::vector<std::uint32_t> &stride,
                  std::shared_ptr<void> host_buffer_ptr);
   BufferInstance(DeviceInstance &device);
   ~BufferInstance();
@@ -46,7 +49,7 @@ public:
     // the hook to get an unsafe pointer (avoids a copy).
     return false;
   }
-  tt::runtime::Tensor tensor() { return *tensor_; }
+  const tt::runtime::Tensor &getTensor() const { return tensor_; }
 
   PJRT_Error *GetMemoryLayout(PJRT_Buffer_GetMemoryLayout_Args *args);
   // Gets the required host size in bytes to copy to host.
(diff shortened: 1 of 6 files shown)


src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -106,8 +106,10 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
     // PJRT expects an empty shape for scalars.
```
umalesTT, line 106: In https://github.com/tenstorrent/tt-xla/pull/197 isOutputScalar function is deleted from client_instance.h and added to executable_image.h
Could this result in conflict between these 2 PRs?

src/common/pjrt_implementation/device_instance.cc
```diff
@@ -93,4 +92,24 @@ tt_pjrt_status DeviceInstance::HostBufferToDevice(
...
+  }
+  return size * element_size;
+}
+
+std::unique_ptr<tt::runtime::Tensor> DeviceInstance::MakeDeviceTensor(
+    const void *data, std::vector<std::uint32_t> &shape,
+    std::vector<std::uint32_t> &strides, size_t element_size,
+    tt::target::DataType element_type) {
+  size_t tensor_size = getSize(shape, element_size);
+  std::shared_ptr<void> new_memory(new char[tensor_size], [](void *) {});
```
sgligorijevicTT, line 109: This leaks memory.
We create a shared pointer with a deleter that does nothing and then pass it's memory into a runtime tensor factory. The runtime will assume memory is borrowed and won't clean up, nor will our shared_ptr, and the memory is leaked.

I should also note this code somewhat duplicates logic of `createOwnedTensor` function from the rt which isn't exposed to us.

I still maintain that the best fixes would be to either use `createOwnedTensor` or the new runtime stitching APIs, but it is what it is :pensive:

ajakovljevicTT: Nice catch @sgligorijevicTT! I've changed the code so that the BufferInstance now holds the pointer to host memory, so this should be fixed now.

inc/common/pjrt_implementation/buffer_instance.h
```diff
@@ -74,14 +76,18 @@ class BufferInstance {
...
 
   std::vector<int64_t> minor_to_major_;
   std::vector<int64_t> tile_dims_;
   std::vector<size_t> tile_dim_sizes_;
 
   // Underlying datatype of tensor.
   std::optional<PJRT_Buffer_Type> DataType;
+
+  // Pointer to the host memory used to create this buffer, if buffer is created
+  // on device, the value of this pointer is nullptr.
```
kmitrovicTT, line 89: Might be a good idea to set it to nullptr by default.

src/common/pjrt_implementation/buffer_instance.cc
```diff
@@ -19,12 +19,13 @@ int BufferInstance::id_counter_ = 0;
 BufferInstance::~BufferInstance() = default;
 
 BufferInstance::BufferInstance(DeviceInstance &device,
-                               tt::runtime::Tensor tensor,
+                               std::unique_ptr<tt::runtime::Tensor> &tensor,
```
kmitrovicTT, line 22: Mark as `&&` to indicate ownership will be moved.

src/common/pjrt_implementation/buffer_instance.cc
```diff
@@ -19,12 +19,13 @@ int BufferInstance::id_counter_ = 0;
...
                                std::vector<std::uint32_t> shape,
-                               std::vector<std::uint32_t> stride)
-    : device_(device) {
+                               std::vector<std::uint32_t> stride,
+                               std::shared_ptr<void> host_buffer_ptr)
+    : device_(device), host_buffer_ptr_(host_buffer_ptr) {
```
kmitrovicTT, line 26: you can add `tensor_(std::move(tensor))` to the initializer list.

src/common/pjrt_implementation/device_instance.cc
```diff
@@ -93,4 +90,28 @@ tt_pjrt_status DeviceInstance::HostBufferToDevice(
   return tt_pjrt_status::kSuccess;
 }
 
+size_t DeviceInstance::getSize(const std::vector<std::uint32_t> &shape,
+                               size_t element_size) {
+  size_t size = 1;
+  for (auto dim : shape) {
+    size *= dim;
+  }
+  return size * element_size;
```
kmitrovicTT, line 99: You could do `size = std::accumulate(shape.begin(), shape.end(), 1, std::multiplies<std::uint32>())` if you prefer

inc/common/pjrt_implementation/device_instance.h
```diff
@@ -67,6 +69,14 @@ class DeviceInstance {
 private:
   tt_pjrt_status OpenDevice();
 
+  size_t getSize(const std::vector<std::uint32_t> &shape, size_t element_size);
```
kmitrovicTT, line 72: Since class is called `DeviceInstance` `getSize` might indicate you are calculating size of the device itself.

Rename to `calculateTensorSize` or `calculateBufferSize` (whichever you find makes more sense) and make it static (it has no use of `this`).

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -106,9 +106,11 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
     std::vector<std::uint32_t> output_shape =
         is_scalar ? std::vector<std::uint32_t>() : output_specs[i].shape;
+    std::unique_ptr<tt::runtime::Tensor> tensor_ptr =
+        std::make_unique<tt::runtime::Tensor>(rt_outputs[i]);
     auto result_buffer = std::make_unique<BufferInstance>(
-        *this->addressable_devices_[dev_index], rt_outputs[i], output_shape,
-        output_specs[i].stride);
+        *this->addressable_devices_[dev_index], tensor_ptr, output_shape,
+        output_specs[i].stride, nullptr);
```
kmitrovicTT, line 113: You can make it optional in declaration or have it be `nullptr` by default, as I already mentioned.

inc/common/pjrt_implementation/buffer_instance.h
```diff
@@ -23,9 +23,11 @@ class DeviceInstance;
 
 class BufferInstance {
 public:
-  BufferInstance(DeviceInstance &device, tt::runtime::Tensor tensor,
+  BufferInstance(DeviceInstance &device,
+                 std::unique_ptr<tt::runtime::Tensor> &tensor,
```
kmitrovicTT, line 27: Don't forget to add `&&` here as well.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -106,9 +106,11 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
     // PJRT expects an empty shape for scalars.
     std::vector<std::uint32_t> output_shape =
         is_scalar ? std::vector<std::uint32_t>() : output_specs[i].shape;
+    std::unique_ptr<tt::runtime::Tensor> tensor_ptr =
+        std::make_unique<tt::runtime::Tensor>(rt_outputs[i]);
```
kmitrovicTT, line 110: So this is invoking copy constructor, right? And down below, created unique ptr is moved into the buffer. What happens with `rt_outputs`? To they still point to something relevant?

Please space out this cluttered loop by adding some empty lines where it makes sense.

inc/common/pjrt_implementation/device_instance.h
```diff
@@ -67,6 +69,14 @@ class DeviceInstance {
...
+
+  BufferInstance *MakeDeviceBuffer(const void *data_ptr,
```
kmitrovicTT, line 74: Please add some comment above.

src/common/pjrt_implementation/device_instance.cc
```diff
@@ -93,4 +90,28 @@ tt_pjrt_status DeviceInstance::HostBufferToDevice(
...
+  size_t tensor_size = getSize(shape, element_size);
+  std::shared_ptr<void> new_memory(new char[tensor_size], [](void *ptr) {
+    delete[] static_cast<char *>(ptr);
+  });
+  std::memcpy(new_memory.get(), data, tensor_size);
+  std::unique_ptr<tt::runtime::Tensor> device_tensor =
+      std::make_unique<tt::runtime::Tensor>(tt::runtime::createTensor(
+          new_memory, shape, strides, element_size, element_type));
+  return new BufferInstance(*this, device_tensor, shape, strides, new_memory);
+}
```
kmitrovicTT, line 115: Here are some chatgpt suggestions on how to make this function more clear:

- Added `const` to vector parameters
- Spacing between lines, comments
- changed `char` to `std::uint8_t` since it directly associates with a size of a byte

```cpp
BufferInstance* DeviceInstance::MakeDeviceBuffer(
    const void* data, 
    const std::vector<std::uint32_t>& shape,
    const std::vector<std::uint32_t>& strides,
    size_t element_size,
    tt::target::DataType element_type) {
  // Calculate total size of the tensor in bytes
  size_t tensor_size = getSize(shape, element_size);

  // Allocate memory for the tensor and manage its lifetime with a custom deleter
  auto new_memory = std::shared_ptr<void>(new std::uint8_t[tensor_size], [](void* ptr) {
    delete[] static_cast<std::uint8_t*>(ptr);
  });

  // Copy the provided data into the newly allocated memory
  std::memcpy(new_memory.get(), data, tensor_size);

  // Create the Tensor using the allocated memory and provided details
  auto device_tensor = std::make_unique<tt::runtime::Tensor>(
      tt::runtime::createTensor(new_memory, shape, strides, element_size, element_type));

  // Return a new BufferInstance encapsulating the Tensor
  return new BufferInstance(*this, std::move(device_tensor), shape, strides, new_memory);
}
```

I would also implement a small `allocateMemoryOnHost` function to wrap this ugly `new_memory` and `memcpy` stuff.

sgligorijevicTT: If we're being really pedantic about types, there is also `std::byte` since C++17.

src/common/pjrt_implementation/buffer_instance.cc
kmitrovicTT: Actually, no, this should be declared as `std::unique_ptr<tt::runtime::Tensor> tensor` to indicate you are transfering ownership. And down below where you call this constructor, you should either do `std::move(tensor)` or construct it in-place (automatic object).

Please read https://stackoverflow.com/questions/8114276/how-do-i-pass-a-unique-ptr-argument-to-a-constructor-or-a-function/8114913#8114913 before changing anything about `&&` I mentioned.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -77,6 +77,33 @@ void LoadedExecutableInstance::BindApi(PJRT_Api *api) {
...
+LoadedExecutableInstance::getTensorSize(const std::vector<std::uint32_t> &shape,
+                                        size_t element_size) {
+  size_t size = 1;
+  for (auto dim : shape) {
+    size *= dim;
+  }
+  return size * element_size;
+}
+
+std::unique_ptr<BufferInstance> LoadedExecutableInstance::createOutputBuffer(
```
mrakitaTT, line 90: Talked offline, should revert this, open issue to figure out more elegant way to handle this in runtime, and disable the failing test with the comment linking that issue

src/common/pjrt_implementation/buffer_instance.cc
```diff
@@ -132,7 +133,7 @@ void BufferInstance::BindApi(PJRT_Api *api) {
       +[](PJRT_Buffer_ReadyEvent_Args *args) -> PJRT_Error * {
     DLOG_F(LOG_DEBUG, "BufferInstance::PJRT_Buffer_ReadyEvent");
     BufferInstance *buffer = BufferInstance::Unwrap(args->buffer);
-    buffer->on_ready_event_ = new EventInstance();
+    buffer->on_ready_event_ = std::make_shared<EventInstance>();
     args->event = *buffer->on_ready_event_;
```
mrakitaTT, line 137: Since the caller is responsible for destroying the event, and since the caller can use this event after the buffer is destroyed, we don't have any other option than to release the ownership to caller and to just store the pointer to event on our side. The event should definitely outlive the buffer so we should be safe to use it. I would do this:
```
std::unique_ptr<EventInstance> onReadyEvent = std::make_unique<EventInstance>();
buffer->on_ready_event_ = onReadyEvent.get();
// Releasing the ownership to the PJRT API caller since the caller is
// responsible for calling PJRT_Event_Destroy on event.
args->event = onReadyEvent.release();
```

tests/jax/ops/test_convert.py
```diff
@@ -25,6 +25,18 @@ def conditionally_skip(from_dtype: DTypeLike, to_dtype: DTypeLike):
     Extracted here in order not to pollute the test function.
     """
     # ---------- Atol comparison failed ----------
+    # When no conversion is required, a no-op MLIR graph is created.
+    # However, due to input tensor ownership issues, the output tensor
+    # returned by the MLIR runtime will reference the same data as the input.
+    # If the input tensor is deallocated, the output tensor will lose access
+    # to valid data and may contain garbage.
+    # See issue #244 for more details.
```
mrakitaTT, line 33: The issue #244 is just to use the buffer semantics to avoid having to workaround these issues every time. We should have another issue that tracks making changes in runtime to facilitate more elegant way of working around the ownership issues. Can you please open one and reference that one here? You should also reference it above in the comment for `MakeDeviceBuffer` function in `inc/common/pjrt_implementation/device_instance.h`.

mrakitaTT (approved):

PR #221, opened 2025-01-31
Switch to using purely dynamic libs for tt-mlir

vwellsTT:
### Problem description
We want to change our linking strategy for tt-mlir to use dynamic linking. For some reason, this breaks CI for this repo. We fix by making stablehlo deps embedded in dylib, and also remove various static tt-mlir lib linking here (since components have been folded into the shared lib).

Related to: https://github.com/tenstorrent/tt-mlir/pull/1947

### What's changed
Simplifies how we import tt-mlir by folding everything into shared-lib target--remove links to all static tt-mlir libs, including stablehlo + shardy.

Checklist
[X] successful on-pr run

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -46,76 +46,13 @@ target_include_directories(TTPJRTCommon PUBLIC
     ${TTMLIR_TOOLCHAIN_DIR}/src/stablehlo
 )
 
-set(STABLEHLO_LIBS
-StablehloBase
-StablehloReferenceIndex
-StablehloReferenceErrors
-StablehloReferenceElement
-StablehloReferenceAxes
-StablehloReferenceValue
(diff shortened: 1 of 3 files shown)


kmitrovicTT:
Looks much simpler!

@ajakovljevicTT @mrakitaTT might this cause problems for us with PJRT for some reason? I don't see any potential problems, but you got deeper into PJRT than me, especially @ajakovljevicTT

ajakovljevicTT:
> Looks much simpler!
> 
> @ajakovljevicTT @mrakitaTT might this cause problems for us with PJRT for some reason? I don't see any potential problems, but you got deeper into PJRT than me, especially @ajakovljevicTT

Seems ok to me, if it compiles and the tests pass. After this merges, I will go through the PJRT implementation code and potentially erase any not-needed includes.

Issue #211, opened 2025-01-29
Simple tests that demonstrate "tensor" parallel and "batch" parallel


nsmithtt:
That is:
- some tests that demonstrate how to cut tensors on XY dims, i.e. tensor parallel
- some tests that demonstrate how to cut tensors on batch dim, i.e. batch parallel
- Potentially some combination of both

PR #231, opened 2025-02-04
Adding infrastructure for multichip tests

ajakovljevicTT:
### Ticket
Part of issue https://github.com/tenstorrent/tt-xla/issues/211

### Problem description
Integrating new tester class, called `MultichipTester`, into our testing infrastructure in`tt-xla`. It will receive the input and output shardings from the user, the function to be tested, and the golden function which will run on CPU, in a non-sharded mode. It runs the received data and compared the results.


tests/infra/multichip_tester.py
```diff
@@ -0,0 +1,119 @@
...
+            module_sharded,
+            out_shardings=output_sharding,
+            static_argnames=static_argnames,
+        )
+
+    def test(self, workload: Workload, cpu_workload: Workload) -> None:
+        """
+        Runs test by running `workload` on TT device and CPU and comparing the results.
+        """
+        compiled_executable = self._compile(workload.executable)
```
kmitrovicTT, line 58: Better yet

```python
compiled_device_workload = MultichipWorkload(
    self._compile_for_device(device_workload.executable),
    workload.args,
    workload.kwargs,
    mesh=self.mesh,
    in_specs=self.in_specs,
)

compiled_cpu_workload = Workload(
    self._compile_for_cpu(cpu_workload.executable), 
    cpu_workload.args, 
    cpu_workload.kwargs,
)

tt_res = DeviceRunner.run_on_multichip_device(compiled_device_workload)
cpu_res = DeviceRunner.run_on_cpu(compiled_cpu_workload)

self._compare(tt_res, cpu_res)
```

tests/infra/multichip_tester.py
```diff
@@ -0,0 +1,119 @@
...
+        Tests `f` by running it with random inputs in range [`minval`, `maxval`) on
+        TT device and CPU and comparing the results.
+        """
+        inputs = [
+            jax.random.uniform(
+                key=jax.random.key(0), shape=shape, minval=minval, maxval=maxval
+            )
+            for shape in input_shapes
+        ]
+        workload = Workload(f, inputs)
```
kmitrovicTT, line 98: Make a distinction everywhere between `cpu_workload` and `device_workload`.

Also, I don't mind how it is now, it just bothers me a bit that we create a regular Workload here and prepack it into MultichipWorkload later on. Maybe could be done right away, if it does not pollute the `test_with_random_inputs` function.

tests/infra/multichip_tester.py
```diff
@@ -0,0 +1,119 @@
...
+    def __init__(
+        self,
+        mesh: jax.Mesh,
+        in_specs: tuple,
+        out_specs: jax.sharding.PartitionSpec,
+        comparison_config: ComparisonConfig = ComparisonConfig(),
+    ) -> None:
+        self.mesh = mesh
+        self.in_specs = in_specs
+        self.out_specs = out_specs
```
kmitrovicTT, line 31: Explain all of these, as they are public, in class docstring.

tests/infra/device_runner.py
```diff
@@ -2,22 +2,28 @@
...
-import inspect
-
 
 class DeviceRunner:
     """
     Class providing methods to put and run workload on any supported device.
     """
 
+    @staticmethod
+    def run_manual(workload: Workload) -> Tensor:
```
kmitrovicTT, line 23: Would `run_on_multichip_tt_device` be a better name? Bring it down with `run_on_tt_device`.

Argument is `MultichipWorkload`.

tests/infra/device_runner.py
```diff
@@ -2,22 +2,28 @@
...
+        """Runs `workload` on TT device."""
+        return DeviceRunner._run_manual(workload)
```
kmitrovicTT, line 25: I created `_run_on_device` for all public `run` functions to use, but since we don't have anything else but tt device multichip run, just leave the implementation inside, no need for `_run_manual`.

tests/infra/device_runner.py
```diff
@@ -63,6 +69,51 @@ def put_tensors_on_gpu(*tensors: Tensor) -> Sequence[Tensor]:
...
+            args_on_device,
+            kwargs_on_device,
+            mesh=multichip_workload.mesh,
+            in_specs=multichip_workload.in_specs,
+        )
+
+    @staticmethod
+    def _run_manual(workload: Workload) -> Tensor:
+        """Runs `workload` on a device."""
+        return workload.execute().block_until_ready()
```
kmitrovicTT, line 115: You can do `put_with_none_sharding` here.

I am curious why does this not need the `with` statement and connection to the device.

tests/infra/device_runner.py
```diff
@@ -63,6 +69,51 @@ def put_tensors_on_gpu(*tensors: Tensor) -> Sequence[Tensor]:
         """Puts `tensors` on GPU."""
         raise NotImplementedError("Support for GPUs not implemented")
 
+    @staticmethod
+    def put_with_none_sharding(
+        multichip_workload: MultichipWorkload,
```
kmitrovicTT, line 74: Tip: if you remove the comma at the end, this will be a one line.

tests/infra/device_runner.py
```diff
@@ -63,6 +69,51 @@ def put_tensors_on_gpu(*tensors: Tensor) -> Sequence[Tensor]:
         """Puts `tensors` on GPU."""
         raise NotImplementedError("Support for GPUs not implemented")
 
+    @staticmethod
+    def put_with_none_sharding(
```
kmitrovicTT, line 73: Please come up with a better name and better description. This seems to resemble `_safely_put_workload_on_device` a lot.

tests/infra/multichip_tester.py
```diff
@@ -0,0 +1,137 @@
...
+        super().__init__(comparison_config)
```
kmitrovicTT, line 43: Bring to the top of the function.

tests/infra/multichip_tester.py
```diff
@@ -0,0 +1,137 @@
...
+            module_sharded,
+            out_shardings=output_sharding,
+            static_argnames=static_argnames,
+        )
+
+    def test(
+        self, multichip_workload: MultichipWorkload, cpu_workload: Workload
+    ) -> None:
+        """
+        Runs test by running `workload` on TT device and CPU and comparing the results.
```
kmitrovicTT, line 69: There are two different workloads now. Update.

tests/infra/multichip_tester.py
```diff
@@ -0,0 +1,137 @@
...
+        Tests an input executable with random inputs in range [`minval`, `maxval`) by running it on a mesh of
+        TT devices and comparing it to output of the cpu executable ran with the same input.
+        """
+        inputs = [
+            jax.random.uniform(
+                key=jax.random.key(0), shape=shape, minval=minval, maxval=maxval
+            )
+            for shape in input_shapes
+        ]
+        multichip_workload = MultichipWorkload(
```
kmitrovicTT, line 110: Call it just `device_workload`. The entire class is `Multichip` so it is obvious device is multichip.

tests/jax/multichip/manual/all_gather.py
```diff
@@ -0,0 +1,28 @@
...
+def test_all_gather(x_shape: tuple):
+    def fwd(batch):
+        act = jax.lax.all_gather(batch, "batch", axis=0, tiled=True)
+        return act
+
+    def golden_fwd(batch):
+        return jnp.tile(batch, (2, 1))
+
+    in_specs = (make_partition_spec(("batch")),)
+    out_specs = make_partition_spec(("batch"))
```
kmitrovicTT, line 24: Much nicer, no low-level jax exposed.

tests/infra/workload.py
```diff
@@ -27,3 +28,13 @@ def __post_init__(self):
...
+
+
+@dataclass
+class MultichipWorkload(Workload):
+    """
+    An extension of the Workload dataclass that includes a mesh and partition specs, necessary for multichip sharding.
+    """
+
+    mesh: jax.sharding.Mesh = None
+    in_specs: Sequence[jax.sharding.PartitionSpec] = None
```
kmitrovicTT, line 40: If these must exist in every multichip workload, don't make them optional (by setting None by default). Unless you have a check somewhere that relies on it.

tests/jax/multichip/manual/unary_eltwise.py
```diff
@@ -0,0 +1,34 @@
...
+
+        stitched_result = b1 + b2
+        return stitched_result
+
+    in_specs = (make_partition_spec((("x", "y"))),)
+    out_specs = make_partition_spec((None, None))
+
+    run_multichip_test_with_random_inputs(
+        fwd, fwd_single_device, [x_shape], (1, 2), ("x", "y"), in_specs, out_specs
+    )
```
kmitrovicTT, line 34: Consider parametrizing axis names, since you use them in three places.

tests/jax/multichip/manual/unary_eltwise.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def fwd_single_device(a_block):
+        a1, a2 = jnp.split(a_block, 2, axis=1)
+
+        b1, b2 = jnp.negative(a1), jnp.negative(a2)
+
+        stitched_result = b1 + b2
+        return stitched_result
+
+    in_specs = (make_partition_spec((("x", "y"))),)
```
kmitrovicTT, line 29: Pretty sure you have one extra `( )` around `x, y` that does nothing.

tests/infra/workload.py
```diff
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from dataclasses import dataclass
+import jax
```
kmitrovicTT, line 6: `from jax.sharding import Mesh, PartitionSpec` will save you from double `jax.sharding.` down below. :)

tests/infra/multichip_tester.py
```diff
@@ -0,0 +1,137 @@
...
+    minval: float = 0.0,
+    maxval: float = 1.0,
+    comparison_config: ComparisonConfig = ComparisonConfig(),
+) -> None:
+    """
+    Tests an input executable with random inputs in range [`minval`, `maxval`) by running it on a mesh of
+    TT devices and comparing it to output of the cpu executable ran with the same input.
+    """
+    mesh = device_connector.get_tt_device_mesh(mesh_shape, axis_names)
+    tester = MultichipTester(mesh, in_specs, out_specs, comparison_config)
```
kmitrovicTT, line 134: I would just go one level down and hide `device_connector` usage from user (since they directly call this function in tests). Pass `mesh_shape` and `axis_name` to `MultichipTester` and use them to create mesh where you need it (maybe in `__init__`).

tests/infra/multichip_tester.py
```diff
@@ -0,0 +1,137 @@
...
+            static_argnames=static_argnames,
+        )
+
+    def test(
+        self, multichip_workload: MultichipWorkload, cpu_workload: Workload
+    ) -> None:
+        """
+        Runs test by running `workload` on TT device and CPU and comparing the results.
+        """
+        multichip_compiled_workload = MultichipWorkload(
```
kmitrovicTT, line 71: `compiled_device_workload`

tests/infra/multichip_tester.py
```diff
@@ -0,0 +1,137 @@
...
+            self._compile_for_device(multichip_workload.executable),
+            multichip_workload.args,
+            multichip_workload.kwargs,
+            mesh=self.mesh,
+            in_specs=self.in_specs,
+        )
+
+        cpu_compiled_workload = Workload(
```
kmitrovicTT, line 79: `compiled_cpu_workload`

tests/infra/device_connector.py
```diff
@@ -69,6 +69,10 @@ def is_initialized(self) -> bool:
 
         return False
 
+    def get_tt_device_mesh(self, shape: tuple, axis_names: tuple) -> jax.sharding.Mesh:
+        tt_devices = jax.devices(DeviceType.TT.value)
+        return jax.make_mesh(shape, axis_names, devices=tt_devices)
```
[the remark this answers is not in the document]
ajakovljevicTT: the `jax.make_mesh` will throw with an exception if the number of devices is different from what the `shape` expects.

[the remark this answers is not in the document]
mrakitaTT: Just curious, what if the number of devices is larger than what mesh needs, will that throw exception too or will that take a subset of devices?

tests/infra/multichip_tester.py
```diff
@@ -0,0 +1,140 @@
...
+        module_sharded = shard_map(
+            executable,
+            mesh=self.device_mesh,
+            in_specs=self.in_specs,
+            out_specs=self.out_specs,
+        )
+        output_sharding = NamedSharding(self.device_mesh, self.out_specs)
+        return jax.jit(
+            module_sharded,
+            out_shardings=output_sharding,
```
mrakitaTT, line 65: Is it necessary to define `out_shardings` for `jax.jit` if `mesh` and `out_specs` are already set in module's `shard_map`?

tests/infra/device_runner.py
```diff
@@ -82,6 +102,49 @@ def _put_on_device(
...
         return DeviceRunner._safely_put_workload_on_device(workload, device)
 
+    @staticmethod
+    def _put_multichip_workload_on_device(
+        multichip_workload: MultichipWorkload,
+    ) -> MultichipWorkload:
+        """Gives the workload inputs shardings, necessary for multichip workloads"""
+        args_on_device = []
+        spec_index = 0
+        # TODO: It might necessary to put a try-except block here, but holding that off until we come across a case where it's needed.
```
mrakitaTT, line 112: Wrap lines at 80 characters length

tests/infra/device_runner.py
```diff
@@ -74,6 +82,18 @@ def _run_on_device(
         with jax.default_device(device):
             return device_workload.execute()
 
+    @staticmethod
+    def _put_sharded_tensor_on_multichip_device(
+        tensor: Tensor, mesh: jax.sharding.Mesh, in_spec: jax.sharding.PartitionSpec
+    ) -> Tensor:
+        """
+        Needed for multichip: Uses put_device to give inputs shardings.
+        We just put dummy sharding equal to none, since jax needs to have some notion of sharding when running graph with these buffers as input.
```
mrakitaTT, line 91: Wrap lines at 80 characters length

tests/infra/device_runner.py
```diff
@@ -74,6 +82,18 @@ def _run_on_device(
...
+        """
+        none_tuple = (None,) * len(in_spec)
+        none_spec = PartitionSpec(*none_tuple)
+        return jax.device_put(tensor, NamedSharding(mesh, none_spec), may_alias=True)
```
mrakitaTT, line 95: If we fix the issue with protobuff schemas in PJRT, will we be able to use the `in_spec` here instead of `none_spec`? You should put some TODO about this in the comment above

mrakitaTT (approved):

tests/infra/device_runner.py
ajakovljevicTT: If we fix the issue there, we could just not give any spec, and let the compiler infer from stableHLO code. Will ad a TODO comment here.

tests/infra/multichip_tester.py
ajakovljevicTT: I am not sure, and wanted to avoid anything I might have overlooked.

tests/infra/device_runner.py
ajakovljevicTT: Just leaving a note here: This wasn't caught by the pre-commit hook, we should add a check for that.

ajakovljevicTT: Will add that check in the next PR and change all of the files, does that make sense?

tests/infra/device_runner.py
```diff
@@ -23,6 +23,14 @@ def run_on_tt_device(workload: Workload, device_num: int = 0) -> Tensor:
         """Runs `workload` on TT device."""
         return DeviceRunner._run_on_device(workload, DeviceType.TT, device_num)
 
+    @staticmethod
+    def run_on_multichip_device(multichip_workload: MultichipWorkload) -> Tensor:
+        """Runs `workload` on a multichip device."""
+        sharded_workload = DeviceRunner._put_multichip_workload_on_device(
+            multichip_workload
+        )
+        return sharded_workload.execute().block_until_ready()
```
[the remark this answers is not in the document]
ajakovljevicTT: You are correct, this was copied from code that did not do the compare as you described, omitting `block_until_ready`.

tests/infra/multichip_tester.py
ajakovljevicTT: I would leave it here for now, and then check if it actually needed when we add (and run) more tests. What do you think?

tests/infra/device_runner.py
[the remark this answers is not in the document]
kmitrovicTT: Yeah, I forgot to comment on this. We should either block until ready in both multichip and regular case or omit it and rely on `device_put` to handle it.

tests/infra/device_runner.py
ajakovljevicTT: Added change for all files in this PR: https://github.com/tenstorrent/tt-xla/pull/278

kmitrovicTT: https://github.com/tenstorrent/tt-xla/pull/278#pullrequestreview-2640570097

PR #235, opened 2025-02-05
Fixing attribute name in the ExecutableImage class and adding an event to BufferInstance

ajakovljevicTT:
My previous merge had an error in the naming of `num_addressable_devices` attribute, so this PR changes that. In addition, the multichip xla execution requires that the BufferInstance has a `ready_event`, even if it now used used, so I also added this in this PR.

### DIFF
--- a/src/common/pjrt_implementation/buffer_instance.cc
+++ b/src/common/pjrt_implementation/buffer_instance.cc
@@ -131,7 +131,9 @@ void BufferInstance::BindApi(PJRT_Api *api) {
   api->PJRT_Buffer_ReadyEvent =
       +[](PJRT_Buffer_ReadyEvent_Args *args) -> PJRT_Error * {
     DLOG_F(LOG_DEBUG, "BufferInstance::PJRT_Buffer_ReadyEvent");
-    args->event = *(new EventInstance());
+    BufferInstance *buffer = BufferInstance::Unwrap(args->buffer);
+    buffer->on_ready_event_ = new EventInstance();
+    args->event = *buffer->on_ready_event_;
     return nullptr;
   };
   // TODO: Rework the API to be Aliases(b1, b2) to let the plugin explicitly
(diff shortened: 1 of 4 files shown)


src/common/pjrt_implementation/buffer_instance.cc
```diff
@@ -131,6 +131,7 @@ void BufferInstance::BindApi(PJRT_Api *api) {
   api->PJRT_Buffer_ReadyEvent =
       +[](PJRT_Buffer_ReadyEvent_Args *args) -> PJRT_Error * {
     DLOG_F(LOG_DEBUG, "BufferInstance::PJRT_Buffer_ReadyEvent");
+    args->event = *(new EventInstance());
```
sdjukicTT, line 134: Should we store a pointer in our BufferInstance class?

ajakovljevicTT: You are correct, although we don't use it, it might be benefitial to have it. Addressed.

kmitrovicTT (approved):

src/common/pjrt_implementation/buffer_instance.cc
```diff
@@ -131,6 +131,9 @@ void BufferInstance::BindApi(PJRT_Api *api) {
   api->PJRT_Buffer_ReadyEvent =
       +[](PJRT_Buffer_ReadyEvent_Args *args) -> PJRT_Error * {
     DLOG_F(LOG_DEBUG, "BufferInstance::PJRT_Buffer_ReadyEvent");
+    BufferInstance *buffer = BufferInstance::Unwrap(args->buffer);
+    buffer->on_ready_event_ = new EventInstance();
```
mrakitaTT, line 135: @ajakovljevicTT Everytime you write a naked `new` you should have a **really** good reason. This will cause a memory leak since you are not deleting it in `BufferInstance` destructor, forgetting to do so is the main reason to avoid using naked new :) Can you please fix this with changes in #220?

ajakovljevicTT: Noted, will fix it in the aforementioned PR

mrakitaTT: My bad, I've just found out that PJRT API callers are responsible with destroying events, though we should still avoid using naked new, I will add further comments in #220

Issue #178, opened 2025-01-20
Metal error inside a test failing to compile


sgligorijevicTT:
Running the model test for llama will print a metal error message

```
                 Always |    FATAL | Unsupported data type
```

despite failing to compile with `failed to legalize operation 'stablehlo.reduce'`.

sgligorijevicTT:
Update, this happens when our plugin has support for creating a tensor of a specific type but the support is missing inside the mlir runtime. In that case the right type gets passed down when putting tensors on the device and the runtime catches the mismatch early(as opposed to what happens when the support is missing inside our plugin, which is that a tensor of wrong type is created and then execution has a high probability of faulting due to size mismatches)

PR #236, opened 2025-02-05
Add BLOOM tests

sgligorijevicTT:
### Problem description
We need more tests

### What's changed
Added test for BLOOM family of LLMs

### Checklist

### DIFF
--- a/tests/jax/models/bloom/bloom_1b1/test_1b1.py
+++ b/tests/jax/models/bloom/bloom_1b1/test_1b1.py
@@ -1,0 +1,50 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Callable
+
+import pytest
+from infra import ModelTester, RunMode
+from utils import record_model_test_properties
+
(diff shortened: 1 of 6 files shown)


tests/jax/models/bloom/bloom_1b1/test_1b1.py
```diff
@@ -0,0 +1,50 @@
...
+
+@pytest.fixture
+def training_tester() -> BloomTester:
+    return BloomTester(ModelTester, run_mode=RunMode.TRAINING)
+
+
+# ----- Tests -----
+
+
+@pytest.mark.skip(reason="Unsupported data type")  # segfault
```
kmitrovicTT, line 33: Does this fail in our runtime or in metal? Would you classify it as a compile or runtime fail?

sgligorijevicTT: Looks like this one is trickier than I thought. After some investigation, it appears like that error message is not directly related to the crash(seems to be another instance of #178 ), and the crash happens with no useful logs or trace somewhere during the compile. I would still keep this error message as the reason, as I'm almost certain the root cause is the same between this error and the segfault. Will mark as compile failure.

Issue #164, opened 2025-01-15
Implementing PJRT functions needed for multichip run


ajakovljevicTT:
When giving jax a sharded array to put on device, it uses `PJRT_Executable_OutputElementTypes` and `PJRT_Executable_OutputDimensions` functions to get information about the number, dimensions and type of program outputs. This information needs to be gathered from the stableHLO code that jax has generated, which can be accessed from the `module_buffer` class.

ajakovljevicTT:
Closed by https://github.com/tenstorrent/tt-xla/pull/238

PR #238, opened 2025-02-05
Adding implementation for PJRT_Executable_OutputElementTypes and PJRT_Executable_OutputDimensions

sdjukicTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/164

### Problem description
`PJRT_Executable_OutputElementTypes` and `PJRT_Executable_OutputDimensions` were unimplemented.

### What's changed
In `LoadedExecutableInstance::Execute` we got the shape for every `PJRT_Buffer` output from `binary.getProgramOutputs` and `image_->isOutputScalar`. Now we store the shape in `ExecutableImage` class and have a getter for the final shape, and not `isOutputScalar`.

Adding a vector that stores `PJRT_Buffer_Type` for every output in `ExecutableImage`, as well as a getter, for the implementation of `PJRT_Executable_OutputElementTypes`.

`PJRT_Executable_OutputDimensions` requires the dimensions to be in a different format. It returns an array of output dimension sizes and an array of concatenated dimensions, so we also store that type of pointer in `ExecutableImage`. We store it because we are responsible for memory deallocation, and not the caller of `PJRT_Executable_OutputDimensions`. But because concatenated dimensions are rarely needed, these fields are `nullptr` until `get_output_dims_concatenated` is called.

### Notes
It is possible that `PJRT_Executable_OutputElementTypes` should return only different unique PJRT_Buffer_Types, and not a type for every output.

We might want to store information about stride of every output in `ExecutableImage` as well, so `binary.getProgramOutput` wouldn't have to be called in `LoadedExecutableInstance::Execute`, and all the information about outputs would be at one place.

### Checklist

### DIFF
--- a/inc/common/pjrt_implementation/executable_image.h
+++ b/inc/common/pjrt_implementation/executable_image.h
@@ -56,22 +56,26 @@ public:
     return m_num_addressable_devices;
   }
 
-  const std::vector<std::uint32_t> &get_output_shape(size_t index) const;
+  const std::vector<std::uint32_t> &get_output_shape(const size_t index) const;
 
+  PJRT_Buffer_Type *get_output_types() { return m_output_types.data(); }
+
+  size_t get_num_output_types() const { return m_output_types.size(); }
+
+private:
+  // Retrieves pointers to the concatenated list of output dimensions and the
+  // corresponding list of ranks (number of dimensions per output tensor).
   void get_output_dims_concatenated(const size_t **dim_sizes,
                                     const int64_t **dims);
 
-  PJRT_Buffer_Type *get_output_types() { return m_output_types.data(); }
-  size_t num_output_types() { return m_output_types.size(); }
-
-private:
-  // Checks whether m_output_dim_sizes and m_output_dims_concatenated have been
-  // filled.
-  bool areOutputDimsConcatinated() const {
+  // Checks if the concatenated output dimensions and the corresponding rank
+  // list have been initialized.
+  bool areOutputDimsConcatenated() const {
     return m_output_dim_sizes && m_output_dims_concatenated;
   }
 
-  // Fills the m_output_dim_sizes and m_output_dims_concatenated array.
+  // Populates the concatenated list of output dimensions and the corresponding
+  // list of ranks.
   void populateOutputDimsConcatenated();
 
   // The reference count. Must be disposed when reaching zero.
@@ -96,12 +100,12 @@ private:
   // For every output, holds a list of its dimensions.
   std::vector<std::vector<uint32_t>> m_output_dims;
 
-  // For every output, holds how many dimensions it has. Nullptr until getter
-  // get_output_dims_concatenated is called.
+  // For every output, stores rank (number of dimensions). Nullptr until its
+  // getter is called.
   std::unique_ptr<size_t[]> m_output_dim_sizes;
 
-  // Holds all output dimensions concatenated. Nullptr until getter
-  // get_output_dims_concatenated is called.
+  // Stores all output dimensions concatenated in a flat array. Nullptr until
+  // its getter is called.
   std::unique_ptr<int64_t[]> m_output_dims_concatenated;
 };
 
--- a/inc/common/pjrt_implementation/executable_image.h
+++ b/inc/common/pjrt_implementation/executable_image.h
@@ -10,6 +10,7 @@
 
 #include <atomic>
 #include <string>
+#include <vector>
 
 #include "xla/pjrt/c/pjrt_c_api.h"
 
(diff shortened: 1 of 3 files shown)


src/common/pjrt_implementation/executable_image.cc
```diff
@@ -139,9 +164,35 @@ void ExecutableImage::BindApi(PJRT_Api *api) {
...
-  return m_is_output_scalar[index];
+const std::vector<std::uint32_t> &
+ExecutableImage::get_output_shape(const size_t index) const {
+  assert(index < m_output_dims.size() && "Output index out of range");
+  return m_output_dims[index];
+}
+
+void ExecutableImage::get_output_dims_concatenated(const size_t **dim_sizes,
+                                                   const int64_t **dims) {
+  if (!m_output_dim_sizes || !m_output_dims_concatenated) {
```
ajakovljevicTT, line 175: It's a bit easier to follow this if you do the reverse:
```
if (m_output_dim_sizes && m_output_dims_concatenated)
{
    /* trivial changes */
    return;
}
/* do your algorithm */
```

sdjukicTT: The idea was to populate `m_output_dim_sizes` and `m_output_dims_concatenated` the first time the getter `get_output_dims_concatenated ` is called, and to just return them every time after that. To make the code more readable, I made a separate method `populateOutputDimsConcatenated`. Both `m_output_dim_sizes` and `m_output_dims_concatenated` are filled in the same method because they are only used together, and because we can easily calculate length needed for `m_output_dims_concatenated` in the same loop we fill `m_output_dim_sizes` in.

src/common/pjrt_implementation/executable_image.cc
```diff
@@ -139,9 +164,39 @@ void ExecutableImage::BindApi(PJRT_Api *api) {
...
+    for (size_t j = 0; j < m_output_dim_sizes[i]; j++) {
+      m_output_dims_concatenated[dims_index + j] = m_output_dims[i][j];
+    }
+    dims_index += m_output_dim_sizes[i];
+  }
+}
+
+void ExecutableImage::get_output_dims_concatenated(const size_t **dim_sizes,
+                                                   const int64_t **dims) {
+  if (!m_output_dim_sizes || !m_output_dims_concatenated) {
```
ajakovljevicTT, line 194: Nit: extract this check into a separate function, something like: `areOutputDimsConcatinated()`

inc/common/pjrt_implementation/executable_image.h
```diff
@@ -51,14 +52,28 @@ class ExecutableImage {
...
+  const std::vector<std::uint32_t> &get_output_shape(size_t index) const;
+
+  void get_output_dims_concatenated(const size_t **dim_sizes,
+                                    const int64_t **dims);
+
+  PJRT_Buffer_Type *get_output_types() { return m_output_types.data(); }
+  size_t num_output_types() { return m_output_types.size(); }
+
 private:
+  // Checks whether m_output_dim_sizes and m_output_dims_concatenated have been
```
ajakovljevicTT, line 68: Describe this in a bit more detail, something like: "Check whether the information on output dimension has been populated".

ajakovljevicTT (approved):

src/common/pjrt_implementation/executable_image.cc
```diff
@@ -100,11 +116,61 @@ void ExecutableImage::BindApi(PJRT_Api *api) {
     }
     return nullptr;
   };
+  api->PJRT_Executable_OutputElementTypes =
+      +[](PJRT_Executable_OutputElementTypes_Args *args) -> PJRT_Error * {
+    DLOG_F(LOG_DEBUG, "ExecutableImage::PJRT_Executable_OutputElementTypes");
+    ExecutableImage *exec = ExecutableImage::Unwrap(args->executable);
+    // There is a possibility that this method should return unique types, and
+    // not a type for every output.
```
mrakitaTT, line 124: Looking at the xla implementations [here](https://github.com/openxla/xla/blob/e80e15bce6ff55495cc78c68b04d01ac12a9d399/xla/pjrt/c/pjrt_c_api_wrapper_impl.cc#L146) and [here](https://github.com/openxla/xla/blob/61bc5db987176c96274b5130f31edbb9675b287e/xla/pjrt/pjrt_executable.cc#L374), it doesn't seem to be the case. Did you find clues somewhere else that it might need to return unique types?

mrakitaTT:
>We might want to store information about stride of every output in ExecutableImage as well, so binary.getProgramOutput wouldn't have to be called in LoadedExecutableInstance::Execute, and all the information about outputs would be at one place.

Makes sense to me, do you think you can do this as part of this PR?

inc/common/pjrt_implementation/executable_image.h
sdjukicTT: Changed.

src/common/pjrt_implementation/executable_image.cc
```diff
@@ -100,11 +116,61 @@ void ExecutableImage::BindApi(PJRT_Api *api) {
...
+
+    args->num_outputs = exec->get_result_count();
+    exec->get_output_dims_concatenated(&args->dim_sizes, &args->dims);
+
+    return nullptr;
+  };
+}
+
+const std::vector<std::uint32_t> &
+ExecutableImage::get_output_shape(const size_t index) const {
```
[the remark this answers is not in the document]
sdjukicTT: Understood, changed.

src/common/pjrt_implementation/executable_image.cc
sdjukicTT: You are right, I only took a shallow look. Removed the comment.

mrakitaTT (approved):

PR #242, opened 2025-02-06
Set HF_HOME folder

vmilosevic:
### Ticket
N/A

### Problem description
Loading models from Hugging face can be slow for large models, we want to use shared volume to optimize loding HF models.

### What's changed
Added volume mount for docker containers /mnt/dockercache:/mnt/dockercache
Set Huggingface cache folder HF_HOME=/mnt/dockercache/huggingface
With this HF models cache will be on a shared volume, used by tt-forge-fe, tt-torch and tt-xla runners.

### Checklist


kmitrovicTT (approved):

mrakitaTT (approved):
Great, thanks for making this change, we just discussed couple days ago that we should do this :)

Issue #241, opened 2025-02-06
Fix types inside tests


sgligorijevicTT:
Some type annotations inside our test infrastructure and tests are not correct.

PR #243, opened 2025-02-06
Update type annotations

sgligorijevicTT:
### Ticket
#241 

### Problem description
Some type annotations were incorrect

### What's changed
I updated `_get_input_activations` to return `Any` in the base class, because activations might also be a dictionary. In particular, some huggingface classes follow a convention where the output of some preprocessing class is a dict that can be unpacked directly into the call of a model class.
I corrected the wrong `flax.linen.Module` type annotations from `_get_model` inside testers for models on huggingface. Huggingface classes are not themselves modules, but rather have a module as one of their members. I also removed the type annotations inside base, this rendered the `Model` type alias redundant.

### Checklist

### DIFF
--- a/tests/infra/model_tester.py
+++ b/tests/infra/model_tester.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from enum import Enum
-from typing import Any, Callable, Mapping, Sequence
+from typing import Any, Callable, Dict, Mapping, Sequence, Union
 
 from flax import linen, nnx
 from transformers.modeling_flax_utils import FlaxPreTrainedModel
@@ -14,6 +14,7 @@ from transformers.modeling_flax_utils import FlaxPreTrainedModel
 from .base_tester import BaseTester
 from .comparison import ComparisonConfig
 from .device_runner import DeviceRunner
+from .types import Model
 from .workload import Workload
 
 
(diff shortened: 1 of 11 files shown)


tests/infra/types.py
```diff
@@ -6,16 +6,10 @@
...
 import jax
-from flax import linen, nnx
 
 # Convenience alias. Could be used to represent jax.Array, torch.Tensor, np.ndarray, etc.
 Tensor = Union[jax.Array]
 
-# Convenience alias. Could be used to represent nnx.Module, torch.nn.Module, etc.
-# NOTE nnx.Module is the newest API, linen.Module is legacy but it is used in some
-# huggingface models.
-Model = Union[nnx.Module, linen.Module]
```
kmitrovicTT, line 17: > Huggingface classes are not themselves modules, but rather have a module as one of their members

Do they have a common class which we can add to this Union instead of removing it completely?

I like to be explicit about types, it's my cpp OCD. `Any` is just too broad.

tests/infra/model_tester.py
```diff
@@ -80,12 +79,12 @@ def _init_model_hooks(self) -> None:
...
 
     @abstractmethod
-    def _get_model(self) -> Model:
+    def _get_model(self) -> Any:
         """Returns model instance."""
         raise NotImplementedError("Subclasses must implement this method.")
 
     @abstractmethod
-    def _get_input_activations(self) -> Sequence[Any]:
+    def _get_input_activations(self) -> Any:
```
kmitrovicTT, line 87: Same here. Should we do `Union[Dict, Sequence[Any]]` instead?

tests/infra/types.py
sgligorijevicTT: Fair point, there is FlaxPreTrainedModel in HF that serves a close enough role. I will add it back.

kmitrovicTT (approved):

PR #251, opened 2025-02-12
Support PyTrees as model outputs in test infra

sgligorijevicTT:
### Ticket
N/A

### Problem description
Some model implementations, primarily from huggingface, return classes with multiple fields as outputs instead of a single jax Array. This causes type errors during comparisons. It is also possible other models might want to have multiple outputs.

### What's changed
This PR changes the test infrastructure to work with arbitrary PyTrees instead of a single tensor, using tree utilities from jax. That includes any custom class a model wants to return as long as it's properly registered as a PyTree node(which includes HF classes).

### Checklist
- [X] Update type annotations where appropriate

### DIFF
--- a/tests/infra/base_tester.py
+++ b/tests/infra/base_tester.py
@@ -8,6 +8,7 @@ from abc import ABC
 from typing import Callable, Sequence
 
 import jax
+from jaxtyping import PyTree
 
 from .comparison import (
     ComparisonConfig,
@@ -61,7 +62,7 @@ class BaseTester(ABC):
                 device_output, golden_output, self._comparison_config.allclose
             )
 
-    def _match_data_types(self, *tensors: Tensor) -> Sequence[Tensor]:
+    def _match_data_types(self, *tensors: PyTree) -> PyTree:
         """
         Casts all tensors to float32 if not already in that format.
 
--- a/tests/infra/base_tester.py
+++ b/tests/infra/base_tester.py
@@ -67,7 +67,9 @@ class BaseTester(ABC):
 
         Tensors need to be in same data format in order to compare them.
         """
-        return [
-            tensor.astype("float32") if tensor.dtype.str != "float32" else tensor
-            for tensor in tensors
-        ]
+        return jax.tree.map(
+            lambda tensor: tensor.astype("float32")
+            if tensor.dtype.str != "float32"
+            else tensor,
+            tensors,
+        )
--- a/tests/infra/comparison.py
+++ b/tests/infra/comparison.py
@@ -75,25 +75,20 @@ class ComparisonConfig:
 
 @run_on_cpu
 def compare_equal(device_output: Tensor, golden_output: Tensor) -> None:
-    assert isinstance(device_output, jax.Array) and isinstance(
-        golden_output, jax.Array
-    ), f"Currently only jax.Array is supported"
-
-    eq = (device_output == golden_output).all()
-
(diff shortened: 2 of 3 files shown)


tests/infra/base_tester.py
```diff
@@ -67,7 +67,9 @@ def _match_data_types(self, *tensors: Tensor) -> Sequence[Tensor]:
...
-        return [
-            tensor.astype("float32") if tensor.dtype.str != "float32" else tensor
-            for tensor in tensors
-        ]
+        return jax.tree.map(
+            lambda tensor: tensor.astype("float32")
+            if tensor.dtype.str != "float32"
+            else tensor,
+            tensors,
+        )
```
kmitrovicTT, line 75: What is the return type of this?

tests/infra/comparison.py
```diff
@@ -74,26 +74,21 @@ def disable_all(self) -> None:
...
-    assert isinstance(device_output, jax.Array) and isinstance(
-        golden_output, jax.Array
-    ), f"Currently only jax.Array is supported"
-
-    eq = (device_output == golden_output).all()
-
-    assert eq, f"Equal comparison failed."
+def compare_equal(device_output: PyTree, golden_output: PyTree) -> None:
+    passed = jax.tree.map(lambda x, y: (x == y).all(), device_output, golden_output)
+    assert jax.tree.all(passed), f"Equal comparison failed."
```
kmitrovicTT, line 79: Take note that is now heavily coupled to jax. `Tensor` alias served as a way to abstract the concrete framework underneath. Now that you use `Pytree`, it is solely a jax thing. Even asserts have been removed. Same for functions below.

This `Pytree` structure is so weird to me. How can `device_output` be a `Pytree`? It is a tensor by design.

sgligorijevicTT: > How can `device_output` be a `Pytree`? It is a tensor by design.

This caught me off guard too.
Models can return anything, and specifically huggingface classes take full advantage of that by returning a custom class( example: [FlaxMaskedLMOutput](https://huggingface.co/docs/transformers/v4.49.0/en/main_classes/output#transformers.modeling_flax_outputs.FlaxMaskedLMOutput) is returned by [FlaxAlbertForMaskedLM](https://huggingface.co/docs/transformers/en/model_doc/albert#transformers.FlaxAlbertForMaskedLM.__call__) ). This is so they can return multiple tensors(for example, specifying `output_hidden_states=True` to a model causes it to also return outputs of every transformer block alongside the logits).

Regarding generality across frameworks, `jax.tree` functions *should* be agnostic to what exactly is inside the PyTree, and as long as the lambda passed to some `jax.tree` function can handle the leaf nodes everything should work. This code is also reusable in cases other frameworks return standard containers of custom tensor types. Main point of concern would be if models from other frameworks start returning custom dataclasses too.

tests/infra/base_tester.py
sgligorijevicTT: `tree.map` return a pytree with identical structure as the input pytree except with all leaf nodes replaced by the output type of the lambda.

kmitrovicTT: So is the return type hint of this function ok?

tests/infra/comparison.py
kmitrovicTT: We'll deal with it if that issue arises.

kmitrovicTT (approved):

Issue #229, opened 2025-02-04
Return custom num addressable devices in PJRT_LoadedExecutable_AddressableDevices


ajakovljevicTT:
Currently, we are hardcoding the number of addressable devices to be returned when `PJRT_LoadedExecutable_AddressableDevices`, to be able to support multichip, this has to be custom, with this information inferred from the StableHLO code, possibly via the `mhlo.num_partitions = 2` attribute. This will require changes to the `ExecutableImage::get_num_addressable_devices() ` function. It also might necessitate changing this part of the code that returns concrete addressable devices, but is currently not tested: 
```
args->addressable_devices = const_cast<PJRT_Device **>(
        reinterpret_cast<PJRT_Device *const *>(addressable_devices.data()));
```

ajakovljevicTT:
Closed by https://github.com/tenstorrent/tt-xla/pull/262

PR #262, opened 2025-02-18
Getting number of addressable devices from the mlir module instead of hardcoding

sdjukicTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/229

### Problem description
Number of addressable devices is hardcoded to 1, which is a problem for multichip.

### What's changed
- Now we get the number of addressable devices from the `mhlo.num_partitions` attribute in VHLO module.
- Changed device_id to be the same as chip_id from `getCurrentSystemDesc`, for less convoluted code.
- Moved `num_addressable_devices` from `ExecutableImage` to `LoadedExecutableInstance` because it is related to `addressable_devices_` of the `LoadedExecutableInstance`.

### Checklist

### DIFF
--- a/inc/common/pjrt_implementation/loaded_executable_instance.h
+++ b/inc/common/pjrt_implementation/loaded_executable_instance.h
@@ -29,9 +29,11 @@ class LoadedExecutableInstance {
 public:
   LoadedExecutableInstance(
       ClientInstance &client, ExecutableImage *image,
-      const std::vector<DeviceInstance *> &addressable_devices)
+      const std::vector<DeviceInstance *> &addressable_devices,
+      size_t num_devices_to_utilize)
       : client_(client), image_(image),
-        addressable_devices_(addressable_devices) {}
+        addressable_devices_(addressable_devices),
+        num_devices_to_utilize_(num_devices_to_utilize) {}
   ~LoadedExecutableInstance();
 
   operator PJRT_LoadedExecutable *() {
@@ -46,6 +48,8 @@ public:
     return addressable_devices_;
   }
 
+  size_t get_num_devices_to_utilize() const { return num_devices_to_utilize_; }
+
   // Loads all executables to addressable devices.
   tt_pjrt_status LoadAll();
 
@@ -59,6 +63,7 @@ private:
   ClientInstance &client_;
   ExecutableImage *image_; // Ref-counted semantics.
   std::vector<DeviceInstance *> addressable_devices_;
+  size_t num_devices_to_utilize_;
   std::vector<ResidentExecutable> resident_executables_;
 };
 
--- a/src/common/module_builder.cc
+++ b/src/common/module_builder.cc
@@ -7,7 +7,6 @@
 
 // c++ standard library includes
 #include <cstdlib>
-#include <iostream>
 
 // loguru includes
 #include "loguru/loguru.hpp"
@@ -85,6 +84,8 @@ tt_pjrt_status ModuleBuilder::buildModule(const std::string_view &code,
     return m_status;
   }
 
+  collectNumDevicesToUtilize(mlir_module);
+
   convertFromVHLOToSHLO(mlir_module);
   if (!tt_pjrt_status_is_ok(m_status)) {
     return m_status;
@@ -124,6 +125,24 @@ ModuleBuilder::createVHLOModule(const std::string_view &code) {
   printModule(vhlo_module);
 
   return vhlo_module;
+}
+
+void ModuleBuilder::collectNumDevicesToUtilize(
+    mlir::OwningOpRef<mlir::ModuleOp> &mlir_module) {
+  auto num_partitions_attr =
+      mlir_module->getOperation()->getAttrOfType<mlir::IntegerAttr>(
+          "mhlo.num_partitions");
+  auto num_replicas_attr =
+      mlir_module->getOperation()->getAttrOfType<mlir::IntegerAttr>(
+          "mhlo.num_replicas");
+  if (num_partitions_attr && num_replicas_attr) {
+    m_num_devices_to_utilize =
+        num_partitions_attr.getInt() * num_replicas_attr.getInt();
+  } else {
+    m_num_devices_to_utilize = 1;
+    DLOG_F(WARNING, "mhlo.num_partitions, mhlo.num_replicas not found, using "
+                    "default number of devices: 1");
+  }
 }
 
 void ModuleBuilder::convertFromVHLOToSHLO(
--- a/src/common/pjrt_implementation/client_instance.cc
+++ b/src/common/pjrt_implementation/client_instance.cc
@@ -163,8 +163,8 @@ tt_pjrt_status ClientInstance::PopulateDevices() {
 
   devices_.resize(devices_count);
   for (size_t i = 0; i < devices_count; ++i) {
-    devices_[i] =
-        new DeviceInstance(i, *this, system_desc->chip_descs()->Get(i)->arch());
+    devices_[i] = new DeviceInstance(chip_ids[i], *this,
+                                     system_desc->chip_descs()->Get(i)->arch());
   }
 
   // For now, just make all devices addressable.
@@ -191,11 +191,9 @@ PJRT_Error *ClientInstance::Compile(const PJRT_Program *program,
       *this,
       new ExecutableImage(module_builder_->getBinary(),
                           std::string(program->code, program->code_size),
-                          module_builder_->getIsOutputScalar(),
-                          module_builder_->getNumAddressableDevices()),
-      addressable_devices_);
+                          module_builder_->getIsOutputScalar()),
+      addressable_devices_, module_builder_->getNumDevicesToUtilize());
   *out_executable = executable.release();
-
   return nullptr;
 }
 
--- a/src/common/pjrt_implementation/loaded_executable_instance.cc
+++ b/src/common/pjrt_implementation/loaded_executable_instance.cc
@@ -36,10 +36,14 @@ void LoadedExecutableInstance::BindApi(PJRT_Api *api) {
         "LoadedExecutableInstance::PJRT_LoadedExecutable_AddressableDevices");
     LoadedExecutableInstance *loaded_executable =
         LoadedExecutableInstance::Unwrap(args->executable);
+    // TODO: Set addressable devices in the loaded executable class to only the
+    // devices being utilized, rather than all addressable devices. This way,
+    // the number of devices will be determined by the list length instead of a
+    // separate field in the class.
     const std::vector<DeviceInstance *> &addressable_devices =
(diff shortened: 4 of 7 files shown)


src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -81,8 +81,6 @@ tt_pjrt_status
 LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
   DLOG_F(LOG_DEBUG, "LoadedExecutableInstance::Execute");
 
-  auto [system_desc, chip_ids] = tt::runtime::getCurrentSystemDesc();
-
   // Sanity check, as we only support execution on one chip currently.
   assert(args->num_devices == 1);
```
ajakovljevicTT, line 85: This should also be changed now, no?

sdjukicTT: I removed the unnecessary assert.

inc/common/pjrt_implementation/loaded_executable_instance.h
```diff
@@ -59,6 +65,7 @@ class LoadedExecutableInstance {
   ClientInstance &client_;
   ExecutableImage *image_; // Ref-counted semantics.
   std::vector<DeviceInstance *> addressable_devices_;
+  size_t num_addressable_devices_;
```
kmitrovicTT, line 68: If `num_addressable_devices_` must be equal to `addressable_devices_.size()` at all times, do we really need it?

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -163,8 +163,8 @@ tt_pjrt_status ClientInstance::PopulateDevices() {
 
   devices_.resize(devices_count);
   for (size_t i = 0; i < devices_count; ++i) {
-    devices_[i] =
-        new DeviceInstance(i, *this, system_desc->chip_descs()->Get(i)->arch());
+    devices_[i] = new DeviceInstance(chip_ids[i], *this,
+                                     system_desc->chip_descs()->Get(i)->arch());
```
kmitrovicTT, line 167: This seems like a weird line break.

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -191,9 +191,8 @@ PJRT_Error *ClientInstance::Compile(const PJRT_Program *program,
       *this,
       new ExecutableImage(module_builder_->getBinary(),
                           std::string(program->code, program->code_size),
-                          module_builder_->getIsOutputScalar(),
-                          module_builder_->getNumAddressableDevices()),
-      addressable_devices_);
+                          module_builder_->getIsOutputScalar()),
+      addressable_devices_, module_builder_->getNumAddressableDevices());
```
kmitrovicTT, line 195: This also.

inc/common/pjrt_implementation/loaded_executable_instance.h
sdjukicTT: They aren't equal. `addressable_devices_` represents total chips, and `num_addressable_devices_` tells us how many of them we are using in the particular loaded executable.

kmitrovicTT: Can we call them `num_devices_` or `num_used_devices_`? I get that `addressable` means the same as `used` in this context, but just to avoid this confusion with `addressable_devices_`.

kmitrovicTT (approved):
Fix comments and you're good to go as far as I am concerned.

inc/common/pjrt_implementation/loaded_executable_instance.h
mrakitaTT: Yeah, I had the similar pending comment, to rename from `num_addressable_devices_` into `m_num_devices_to_utilize`, but then I started wondering how JAX and PJRT actually use the addressable devices. This lead me to a realisation that we have two different groups of "addressable devices":
1. `PJRT_Client_AddressableDevices` - "Returns a list of devices that are addressable from the client. Addressable devices are those that the client can issue commands to. All devices are addressable in a single-process environment."
2. `PJRT_LoadedExecutable_AddressableDevices` - "Returns a list of devices this executable will run on."

This means that we shouldn't just pass the full `addressable_devices` list from client to the `LoadedExecutableInstance`, instead we should get the exact subset of devices that the executable needs to run on from the StableHLO. Then we won't need `num_addressable_devices` anymore, the size of the vector will be the num. FYI @ajakovljevicTT

src/common/module_builder.cc
```diff
@@ -118,6 +120,19 @@ ModuleBuilder::createVHLOModule(const std::string_view &code) {
   return vhlo_module;
 }
 
+void ModuleBuilder::collectNumAddressableDevices(
+    mlir::OwningOpRef<mlir::ModuleOp> &mlir_module) {
+  if (auto attr = mlir_module->getOperation()->getAttrOfType<mlir::IntegerAttr>(
+          "mhlo.num_partitions")) {
```
mrakitaTT, line 126: @ajakovljevicTT do we need `mhlo.num_replicas * mhlo.num_partitions` here? Take a look at [this](https://github.com/openxla/xla/blob/main/xla/pjrt/pjrt_client.h#L1300) and [this](https://android.googlesource.com/platform/external/tensorflow/+/6b511124eb0/tensorflow/compiler/xla/pjrt/tpu_client.cc#110).

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -70,10 +70,8 @@ tt_pjrt_status
 LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
   DLOG_F(LOG_DEBUG, "LoadedExecutableInstance::Execute");
 
-  auto [system_desc, chip_ids] = tt::runtime::getCurrentSystemDesc();
-
   // Sanity check, as we only support execution on one chip currently.
-  assert(args->num_devices == 1);
+  assert(args->num_devices == num_addressable_devices_);
```
mrakitaTT, line 74: Seems to me that we need more changes than this in this function to make it work for multichip, for example we still have a hardcoded `dev_index = 0`. Let's sync offline (@ajakovljevicTT)

mrakitaTT: Looking at the comments in the `PJRT_LoadedExecutable_Execute_Args` it seems like we should either:
1. Run on the addressable devices passed to the `LoadedExecutable` from the compiler, where `num_devices` should match the size of the addressable devices list
2. Or run on the `execute_device` specified in `PJRT_LoadedExecutable_Execute_Args`

src/common/module_builder.cc
ajakovljevicTT: You are correct, thanks for catching this! In our examples, `mhlo.num_replicas` was always 1, so it was not a problem, but yes. @sdjukicTT please change.

src/common/pjrt_implementation/loaded_executable_instance.cc
ajakovljevicTT: @mrakitaTT This change is scoped so that it only adressess the `addressable_devices` number. The PR that I will put out soon (that relies on this one and some of the other PR by @sdjukicTT), will execute on devices that are read from the list of `PJRT_Buffer`s that the xla gives us in the `Execute()` function.

inc/common/pjrt_implementation/loaded_executable_instance.h
ajakovljevicTT: @kmitrovicTT I agree that the naming is a bit weird since we had just adopted what xla and PJRT use.

@mrakitaTT yes, I had this exact dilemma at some point, but the devices for compilation are passed through extra compilation options (protobuf), and without it, we have no idea what exact devices the  `LoadedExecutableInstance` will run on. Since we can only rely on the things we can see from the StableHLO code, we can get the number of devices and later get the exact devices for execution by checking what the are the devices of individual `BufferInstance`s.

src/common/module_builder.cc
```diff
@@ -118,6 +120,19 @@ ModuleBuilder::createVHLOModule(const std::string_view &code) {
...
+    m_num_addressable_devices = attr.getInt();
+  } else {
+    m_num_addressable_devices = 1;
+    DLOG_F(
+        WARNING,
```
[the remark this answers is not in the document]
sdjukicTT: It does get set.

inc/common/pjrt_implementation/loaded_executable_instance.h
mrakitaTT: Discussed offline, we can't get the exact list of devices yet from the compile options but since we are currently returning all devices as addressable devices this shouldn't be a problem. Let's then just rename occurrences of `num_addressable_devices` to `num_devices_to_utilize` in `ModuleBuilder` and `LoadedExecutableInstance` to make it more obvious what that number actually counts.

src/common/pjrt_implementation/loaded_executable_instance.cc
mrakitaTT: Ok, let's just not forget to add some TODO to support in the future the scenario when `execute_device` is specified.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -70,10 +70,8 @@ tt_pjrt_status
 LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
   DLOG_F(LOG_DEBUG, "LoadedExecutableInstance::Execute");
 
-  auto [system_desc, chip_ids] = tt::runtime::getCurrentSystemDesc();
-
   // Sanity check, as we only support execution on one chip currently.
```
mrakitaTT, line 73: This comment is now outdated, should edit to explain that we are making sure that the number of devices matches the number of devices we counted from the SHLO module.

sdjukicTT: Updated.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -39,7 +39,7 @@ void LoadedExecutableInstance::BindApi(PJRT_Api *api) {
     const std::vector<DeviceInstance *> &addressable_devices =
         loaded_executable->addressable_devices();
     int num_addressable_devices =
-        loaded_executable->image_->get_num_addressable_devices();
+        loaded_executable->get_num_devices_to_utilize();
```
mrakitaTT, line 42: Currently `num_devices_to_utilize` can be different from the `addressable_devices.size()`, since we pass the full list of addressable devices to the `LoadedExecutableInstance` instead of list of devices to utilize. Let's add a TODO here to make notion of that issue.

mrakitaTT (approved):

Issue #201, opened 2025-01-29
Explicitly give mesh shape and systemDescPath to the pipeline option in ttir->ttnn MLIR conversion.


ajakovljevicTT:
Currently, we rely on having the default options set when calling the `createTTIRToTTNNBackendPipeline` in our `ModuleBuilder`, which translates `TTIR` to `TTNN` mlir dialect. This will not be possible for multichip configurations, as we need to explicitly say which for which system descriptor and what shape we will compile our model. Currently, and intermediate solution is given like this:
```
  mlir::tt::ttnn::TTIRToTTNNBackendPipelineOptions options;
  options.systemDescPath="system_desc.ttsys"; // This is manually copied to our tt-xla repo
  options.meshShape = {1,2};
  mlir::tt::ttnn::createTTIRToTTNNBackendPipeline(ttir_to_ttnn_pm, options);
```
We need to find a way to get these system descriptors specifically, along with the correct mesh shape.

ajakovljevicTT:
Closed by https://github.com/tenstorrent/tt-xla/pull/262

PR #265, opened 2025-02-19
Storing system descriptor

sdjukicTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/201

### Problem description
Default system descriptor used in `TTIRToTTNNBackendPipeline` can't be used for multichip, actual current system descriptor is needed.

### What's changed
We store the system descriptor and then set `options.systemDescPath`.

### Checklist


src/common/module_builder.cc
```diff
@@ -197,6 +197,7 @@ void ModuleBuilder::convertFromTTIRToTTNN(
   mlir::PassManager ttir_to_ttnn_pm(mlir_module.get()->getName());
 
   mlir::tt::ttnn::TTIRToTTNNBackendPipelineOptions options;
+  options.systemDescPath = "system_desc.ttsys";
```
ajakovljevicTT, line 200: Can we extract this path into client_instance or module_builder static attribute? Also, it might be nice to have a folder instead of just storing it in root. That way, we could even delete the system_desc when ClientInstance.destroy() is called, leaving less things in memory.

sdjukicTT: Why a folder?

src/common/module_builder.h
```diff
@@ -40,6 +40,8 @@ class ModuleBuilder {
   // code. Currently hardcoded to one, as we only support one-chip execution.
   size_t getNumAddressableDevices() const { return 1; }
 
+  static constexpr char const *system_desc_path = "system_desc.ttsys";
```
ajakovljevicTT, line 43: can this be a `std::string`?

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -27,6 +27,7 @@ ClientInstance::ClientInstance(std::unique_ptr<Platform> platform)
 }
 
 ClientInstance::~ClientInstance() {
+  std::remove(ModuleBuilder::system_desc_path);
```
ajakovljevicTT, line 30: nit: put this line line below the log

src/common/module_builder.cc
ajakovljevicTT: I think we should be very of keeping thing in root, and this adds another opportunity to be more verbose (for example, by calling the folder "system_descriptors"). However, I might be wrong. @mrakitaTT do you have an opinion on this?

src/common/module_builder.h
```diff
@@ -40,6 +40,8 @@ class ModuleBuilder {
   // code. Currently hardcoded to one, as we only support one-chip execution.
   size_t getNumAddressableDevices() const { return 1; }
 
+  static constexpr std::string_view system_desc_path = "system_desc.ttsys";
```
kmitrovicTT, line 43: Add comment above.

kmitrovicTT (approved):

src/common/module_builder.h
mrakitaTT: Yes this can be a plain string, there is no benefit to keep it as string_view. See more info [here](https://abseil.io/docs/cpp/guides/strings) regarding string vs string_view (std string_view is implementing the same thing as abseil's).

mrakitaTT: Ideally, we would store `tt::runtime::SystemDesc` object inside the `ClientInstance` and pass it through the `Compile` function to the `ModuleBuilder` which would then pass it to the `createTTIRToTTNNBackendPipeline`. Since tt-mlir currently only supports passing a path to the descriptor and is not easy to change to support passing already parsed object, as a temporary solution we need to save it on the disk as pass the path to the compiler.

I propose these changes:
- Create a member variable `tt::runtime::SystemDesc m_system_descriptor` in `ClientInstance` which will be set in the `ClientInstance::PopulateDevices()` function from the `system_desc` variable.
- Create a member variable `std::string m_cached_system_descriptor_path` which should be set inside the `ClientInstance` constructor to the combination of `std::filesystem::temp_directory_path()` directory and some file name that should be unique from other programs, for example `tt_pjrt_system_descriptor` plus maybe name of the device architecture, and maybe even some client id if there is some in pjrt structures.
- In the `ClientInstance::PopulateDevices()` initialize `m_system_descriptor` with `system_desc`, and then store it into `m_cached_system_descriptor_path` with a TODO comment to remove that once the support in tt-mlir is done to pass the system descriptor object. Check if `store` method checks for errors, if not we should check.
- In the `ClientInstance::Compile` function pass the `m_cached_system_descriptor_path` to the `module_builder_->buildModule` call.
- In the `ClientInstance::~ClientInstance()` remove the cached system descriptor, as you already do.

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -28,6 +28,7 @@ ClientInstance::ClientInstance(std::unique_ptr<Platform> platform)
 
 ClientInstance::~ClientInstance() {
   DLOG_F(LOG_DEBUG, "ClientInstance::~ClientInstance");
+  std::remove(ModuleBuilder::system_desc_path.data());
```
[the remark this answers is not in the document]
mrakitaTT: `std::remove` is cleaning up a file stored in that path

src/common/module_builder.cc
mrakitaTT: Touched on this topic in my [other comment](https://github.com/tenstorrent/tt-xla/pull/265#discussion_r1978379993)

system_desc.ttsys
mrakitaTT, line 1: Forgot to comment, this file shouldn't be checked in. Btw how did you generate it in the repo root, does the `store` function generate descriptor in path relative to the current working dir?

src/common/module_builder.h
sdjukicTT: Implemented these changes with few exceptions:
- `store` method doesn't check for errors, but I don't know how we should check. `store` method uses `std::ofstream::write`, I think the default is that exceptions aren't thrown in case of failure, and `store` method doesn't set exceptions.
- Didn't add name of the device architecture to system descriptor path name because we initialize this string in `ClientInstance` constructor, and get the name of the architecture later, from the system descriptor in `ClientInstance::PopulateDevices()`.
- Didn't add client id to system descriptor path name because `ClientInstance` doesn't have that field currently.

system_desc.ttsys
sdjukicTT: `store` method uses `std::ofstream` which works like that.

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -164,8 +171,11 @@ void ClientInstance::BindApi(PJRT_Api *api) {
 tt_pjrt_status ClientInstance::PopulateDevices() {
   DLOG_F(LOG_DEBUG, "ClientInstance::PopulateDevices");
   auto [system_desc, chip_ids] = tt::runtime::getCurrentSystemDesc();
-  int devices_count = chip_ids.size();
 
+  system_descriptor_ = system_desc;
+  system_descriptor_.store(cached_system_descriptor_path_.data());
```
mrakitaTT, line 176: Check after this if `std::filesystem::exists(cached_system_descriptor_path_)` and if it doesn't, do:
```
DLOG_F(ERROR, "Failed to store the system descriptor to the disk using path: %s",
    cached_system_descriptor_path_.c_str());
```

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -21,13 +24,17 @@ namespace tt::pjrt {
 //===----------------------------------------------------------------------===//
 
 ClientInstance::ClientInstance(std::unique_ptr<Platform> platform)
-    : platform_(std::move(platform)) {
+    : platform_(std::move(platform)), system_descriptor_(nullptr) {
   DLOG_F(LOG_DEBUG, "ClientInstance::ClientInstance");
   module_builder_ = std::make_unique<ModuleBuilder>();
+  cached_system_descriptor_path_ =
+      std::filesystem::temp_directory_path().concat(
+          "/tt_pjrt_system_descriptor");
```
mrakitaTT, line 32: Please add a TODO comment here explaining that this name would need to be unique to avoid clashes between multiple clients, but since we plan soon to remove the need for storing the descriptor to the disk we are leaving it simple like this until then (or until it causes problems).

mrakitaTT (approved):

src/common/module_builder.cc
```diff
@@ -193,10 +195,12 @@ void ModuleBuilder::convertFromSHLOToTTIR(
 }
 
 void ModuleBuilder::convertFromTTIRToTTNN(
-    mlir::OwningOpRef<mlir::ModuleOp> &mlir_module) {
+    mlir::OwningOpRef<mlir::ModuleOp> &mlir_module,
+    const std::string &system_descriptor_path) {
```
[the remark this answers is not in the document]
sdjukicTT: Fixed, thank you!

Issue #269, opened 2025-02-20
Consume tt-mlir build requirements when setting up venv


mrakitaTT:
Currently `pybind11` and `nanobind` are required to build tt-mlir, but they are not explicitly stated in tt-mlir `env/build-requirements.txt`, they get transiently installed through other dependencies. This makes Frontends oblivious to those unspecified tt-mlir dependencies, which causes problems because FEs don't use the same venv as tt-mlir.

PR (TODO: add link) adds `pybind11` and `nanobind` explicitly into tt-mlir `env/build-requirements.txt`. After that change lands, we should make a change on FEs to make their venv setup scripts install these dependencies from `third-party/tt-mlir/env/build-requirements.txt` instead of manually stating the same list of dependencies. This might cause issues if FEs and tt-mlir require different versions of some requirements, but I guess we will have to handle these case by case by finding the version they both can work with.

mrakitaTT:
Fixed by #270

PR #270, opened 2025-02-20
Consume tt-mlir build requirements when setting up venv

mrakitaTT:
### Ticket
#269

### Problem description
Currently `pybind11` and `nanobind` are required to build tt-mlir, but they are not explicitly stated in tt-mlir `env/build-requirements.txt`, they get transiently installed through other dependencies. This makes Frontends oblivious to those unspecified tt-mlir dependencies, which causes problems because FEs don't use the same venv as tt-mlir.

PR https://github.com/tenstorrent/tt-mlir/pull/2225 adds `pybind11` and `nanobind` explicitly into tt-mlir `env/build-requirements.txt`. After that change lands, we should make a change on FEs to make their venv setup scripts install these dependencies from `third-party/tt-mlir/env/build-requirements.txt` instead of manually stating the same list of dependencies. This might cause issues if FEs and tt-mlir require different versions of some requirements, but I guess we will have to handle these case by case by finding the version they both can work with.

### What's changed
Changed venv activate script to consume dependencies from tt-mlir `env/build-requirements.txt`, and removed redundant dependencies from `requirements.txt`

### Checklist


venv/install_ttmlir_requirements.sh
```diff
@@ -0,0 +1,26 @@
...
+# Exit immediately if a command exits with a non-zero status
+set -e
+
+source ${TTPJRT_SOURCE_DIR}/venv/bin/activate
+
+TT_MLIR_ENV_DIR=${TTPJRT_SOURCE_DIR}/third_party/tt-mlir/src/tt-mlir/env
+
+# Install llvm requirements
+LLVM_VERSION=$(grep -oP 'set\(LLVM_PROJECT_VERSION "\K[^"]+' ${TT_MLIR_ENV_DIR}/CMakeLists.txt)
+REQUIREMENTS_CASHE_DIR=${TTPJRT_SOURCE_DIR}/venv/bin/requirements_cashe
```
kmitrovicTT, line 17: Typo: `CACHE`

third_party/CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ else()
     ExternalProject_Add(
         tt-mlir
         PREFIX ${TTPJRT_SOURCE_DIR}/third_party/tt-mlir
+        # Installing the python dependencies before the build
+        PATCH_COMMAND TTPJRT_SOURCE_DIR=${TTPJRT_SOURCE_DIR} bash ${TTPJRT_SOURCE_DIR}/venv/install_ttmlir_requirements.sh
```
kmitrovicTT, line 34: Do we wanna call it `TTPJRT_SOURCE_DIR` or `TTXLA_SOURCE_DIR` or something else?

venv/install_ttmlir_requirements.sh
kmitrovicTT, line 1: Could this file be generalized so that all frontends can use it?

kmitrovicTT (approved):

third_party/CMakeLists.txt
mrakitaTT: Yeah, I thought about renaming to `TTXLA` but opted out of it because we use `TT_PJRT` in other cmake files, didn't want to deal with that in this PR.

venv/install_ttmlir_requirements.sh
mrakitaTT: I think it will have to be frontend custom, because each frontend has their own build flow, but I think this code can be easily reused (and I plan to forward this PR to other frontends)

venv/install_ttmlir_requirements.sh
```diff
@@ -0,0 +1,26 @@
...
+# Exit immediately if a command exits with a non-zero status
+set -e
+
+source ${TTPJRT_SOURCE_DIR}/venv/bin/activate
+
+TT_MLIR_ENV_DIR=${TTPJRT_SOURCE_DIR}/third_party/tt-mlir/src/tt-mlir/env
+
+# Install llvm requirements
+LLVM_VERSION=$(grep -oP 'set\(LLVM_PROJECT_VERSION "\K[^"]+' ${TT_MLIR_ENV_DIR}/CMakeLists.txt)
+REQUIREMENTS_CACHE_DIR=${TTPJRT_SOURCE_DIR}/venv/bin/requirements_cashe
```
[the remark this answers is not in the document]
mrakitaTT: dang, good catch, tnx, will fix in next PR

PR #278, opened 2025-02-25
Setting the max line length of python files to 80

ajakovljevicTT:
Setting the max line length to 80 of our python files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,6 +4,7 @@ repos:
     hooks:
       - id: black
         language_version: python3
+        args: [--line-length=80]
   - repo: https://github.com/pre-commit/mirrors-clang-format
     rev: v18.1.7
     hooks:
--- a/examples/simple_regression.py
+++ b/examples/simple_regression.py
@@ -16,14 +16,20 @@ import jax._src.xla_bridge as xb
 # program will execute on tt device if not specified otherwise.
 def initialize():
     backend = "tt"
-    path = os.path.join(os.path.dirname(__file__), "../build/src/tt/pjrt_plugin_tt.so")
+    path = os.path.join(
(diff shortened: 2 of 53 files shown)


kmitrovicTT (commented):
From https://pre-commit.com/

> The default line length setting in Python, particularly when using the pre-commit framework with the black formatter, is 88 characters. This is specified by black as its default line length, which is designed to promote readability while accommodating common display widths.

Let's just be consistent across projects if possible. Check what tt-mlir is using (and other frontends preferably). I think they all inherit this default behaviour specified by pre commit configuration.

ajakovljevicTT:
> From https://pre-commit.com/
> 
> > The default line length setting in Python, particularly when using the pre-commit framework with the black formatter, is 88 characters. This is specified by black as its default line length, which is designed to promote readability while accommodating common display widths.
> 
> Let's just be consistent across projects if possible. Check what tt-mlir is using (and other frontends preferably). I think they all inherit this default behaviour specified by pre commit configuration.

@kmitrovicTT I have put the 80 here as that was advised by @mrakitaTT in another PR for docstrings. I have checked other frontends and tt-mlir. `TT-forge` has line length 120, and `tt-torch` and `tt-mlir` leave the default options.

mrakitaTT:
@ajakovljevicTT My bad, when I posted that comment I thought 80 characters is a default limit. Let's do what `tt-torch` and `tt-mlir` are doing and leave the default options. Though I am not sure why is that not already applied by our pre-commit, any ideas?

mrakitaTT:
I am also confused by 80 characters limit defined here: https://github.com/tenstorrent/tt-xla/blob/main/.style.yapf Maybe that file is introducing these issues in formatting

vmilosevic (approved):

ajakovljevicTT:
@mrakitaTT should we close this PR for now then?

PR #286, opened 2025-02-27
Add a test for prelayernorm roberta

sgligorijevicTT:
### Problem description
We need more tests

### What's changed
I added a test for RobertaPreLayerNorm architecture.

### Checklist

### DIFF
--- a/tests/jax/models/roberta_prelayernorm/efficient_mlm_m0_40/test_efficient_mlm_m0_40.py
+++ b/tests/jax/models/roberta_prelayernorm/efficient_mlm_m0_40/test_efficient_mlm_m0_40.py
@@ -1,0 +1,98 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Callable, Dict, Sequence
+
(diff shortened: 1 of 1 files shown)


tests/jax/models/roberta_prelayernorm/efficient_mlm_m0_40/test_efficient_mlm_m0_40.py
kmitrovicTT, line 1: This doesn't fall into `models/roberta` folder we have already?

kmitrovicTT (approved):

tests/jax/models/roberta_prelayernorm/efficient_mlm_m0_40/test_efficient_mlm_m0_40.py
sgligorijevicTT: There is a small architectural difference between Roberta and RobertaPreLayerNorm (specifically placement of layernorms), it might make sense to merge them into the same folder, but then again we separated out models with smaller architectural differences in other cases. Besides, I was mildly confused that the weights for this model were called "efficient_mlm", so I think it's useful to leave a mention of "RobertaPreLayerNorm" somewhere.

PR #290, opened 2025-02-28
Do not trigger CI when .md files are changed

nsumrakTT:
files to ignore: *.md, *.gitignore, *LICENSE*
closes https://github.com/tenstorrent/tt-xla/issues/295


.github/workflows/on-pr.yml
```diff
@@ -9,6 +9,9 @@ on:
         type: string
   pull_request:
     branches: [ "main" ]
+    paths-ignore:
```
mrakitaTT, line 12: Can we please also add `LICENSE` file to this list? It doesn't have .md extension

.github/workflows/on-pr.yml
```diff
@@ -30,7 +33,7 @@ jobs:
       test_mark: 'push'
 
   check-all-green:
-    if: always()
+    if: always() && needs.build-and-test.result != 'skipped'
```
mrakitaTT, line 36: I'm trying to understand how this works. If I understand correctly, the `paths-ignore` filter will prevent the whole `on-pr` workflow to be run when only .md files are changed, so how will then `check-all-green` job even be triggered? I.e. how can we have a situation where `build-and-test` job is skipped but `check-all-green` is not?

nsumrakTT: check-all-green has always() because it is mandatory to pass so PR could be merged. I've added line which will (hopefully) make it green for .md or .gitignore files

nsumrakTT:
When merged, I had to test it further

.github/workflows/on-pr.yml
mrakitaTT: Oh, I thought that when `on:` conditions are not met `jobs:` section wont be run at all, are you saying that when you put `if: always()` on a job it will be run even when `on:` conditions are not met?

.github/workflows/on-pr.yml
```diff
@@ -21,8 +21,35 @@ jobs:
...
+            if [[ ! $file =~ ^.*\.(md|gitignore)$ && $file != *"LICENSE"* ]]; then
+              all_match=false
+              break
+            fi
+          done
+          if [ "$all_match" = true ]; then
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "skip=false" >> "$GITHUB_OUTPUT"
+          fi
```
mrakitaTT, line 48: Can this be replaced with `echo "skip="$all_match >> "$GITHUB_OUTPUT"`?

mrakitaTT (approved):

.github/workflows/on-pr.yml
nsumrakTT: I did it in this way because we might need to emit some data for collection in case build is skipped. I'll check with @vmilosevic and change it accordingly.

PR #291, opened 2025-02-28
Add pytest push markers to models that were missing it

sgligorijevicTT:
### Problem description
I forgot to add push markers to some model tests I added

### What's changed
I added the markers


kmitrovicTT (approved):
Please check if after the next nightly these models appear in https://superset.tenstorrent.com/superset/dashboard/64/

Issue #295, opened 2025-03-03
Do not trigger CI if *only .md, .gitignore or LICENSE files are changed


nsumrakTT:
@mrakitaTT : "I've noticed that when we change only .md files in tt-xla repo (probably other repos too) the full CI gets kicked off, unnecessarily running device tests. Should find a way to disable running build and test when only .md files are changed."

PR #298, opened 2025-03-04
Check if workflow is triggered by PR to prevent errors

nsumrakTT:
### Ticket
#295

### Problem description
tt-mlir uplift breaks as it is not PR but manually triggered 

### What's changed
Check if PR


mrakitaTT:
> tt-mlir uplift breaks as it is not PR but manually triggered

Can you please explain more on this, looking at the code I am not sure I understand how your previous change broke the uplift nor what is this change supposed to fix, it's been a long day :)

Now I am thinking that maybe the better idea in the first place was to split the `on-pr` workflow into two workflows:
- `on-pr-check` containing `check-all-green` job which will always be run on `pull_request` and `workflow_dispatch`
- `on-pr` containing the other jobs which will have the `paths-ignore` parameter.

@nsumrakTT @vmilosevic what do you think?

nsumrakTT:
> > tt-mlir uplift breaks as it is not PR but manually triggered
> 
> Can you please explain more on this, looking at the code I am not sure I understand how your previous change broke the uplift nor what is this change supposed to fix, it's been a long day :)
> 
> Now I am thinking that maybe the better idea in the first place was to split the `on-pr` workflow into two workflows:
> 
> * `on-pr-check` containing `check-all-green` job which will always be run on `pull_request` and `workflow_dispatch`
> * `on-pr` containing the other jobs which will have the `paths-ignore` parameter.
> 
> @nsumrakTT @vmilosevic what do you think?

tt-mlir has nightly uplift job that is work_dispatch with mlir_override set to latest mlir git hash in order to check if it works with frontends.
I think we should fix this issue and unblock other tasks and then we could discuss further changes.

vmilosevic (approved):

mrakitaTT:
> tt-mlir has nightly uplift job that is work_dispatch with mlir_override set to latest mlir git hash in order to check if it works with frontends.

I am aware of this but I couldn't figure out yesterday how your previous change broke the uplift flow nor how this change fixes it. Today with a fresh brain I've realized that the git commands that are used in `check-files` job are working only for pull requests :) It probably seemed obvious to you, but I would still suggest writing a slightly more informative PR descriptions if possible :)

mrakitaTT (approved):

Issue #301, opened 2025-03-04
Incorrect boolean tensor creation


mmanzoorTT:
Boolean tensor is created in `tests/infra/utils.py::random_tensor` with 

```
        elif jnp.issubdtype(dtype_converted, jnp.bool):
            # Generate random tensor of 0s and 1s and interpret is as a bool tensor.
            return jax.random.randint(
                key=prng_key,
                shape=shape,
                dtype=jnp.int32,
                minval=0,
                maxval=1,
            ).astype(dtype_converted)
```

However, the maximum value is not inclusive while creating random numbers. As a result, this `randint` function will always generate `False` instead of generating both `False` or `True`.

PR #303, opened 2025-03-04
Update maxVal while creating boolean inputs

mmanzoorTT:
### Ticket
closes https://github.com/tenstorrent/tt-xla/issues/301

### Problem description
Random input is created using minVal=0 and maxVal=1. However, the maximum value is not inclusive while creating random numbers. As a result, this randint function will always generate `False` instead of generating either `False` or `True`.

### What's changed
Update boolean random tensor creation process

### Checklist


mrakitaTT (approved):
Nice catch, thank you for fixing this @mmanzoorTT!

PR #313, opened 2025-03-06
Update fail reasons 2025 03 06

### LINKED ISSUE #308: Unsupported data type DataType::INT32
Tracking issue for the metal error message `Unsupported data type DataType::INT32`.
It should already be fixed in metal, we're just waiting for the fix to be uplifted.

sgligorijevicTT:
Another day another update to fail reasons
I am happy to report that 9 new models now pass runtime and fail only because of accuracy

This PR also fixes a minor oversight in tests for CLIP

### Ticket
closes #308

### DIFF
--- a/tests/jax/models/clip/tester.py
+++ b/tests/jax/models/clip/tester.py
@@ -23,9 +23,7 @@ class FlaxCLIPTester(ModelTester):
 
     # @override
     def _get_model(self) -> FlaxPreTrainedModel:
-        from_pt = False
-        if self._model_name == "openai/clip-vit-large-patch14-336":
-            from_pt = True
+        from_pt = self._model_name == "openai/clip-vit-large-patch14-336"
 
         return FlaxCLIPModel.from_pretrained(self._model_name, from_pt=from_pt)
 
--- a/tests/jax/models/clip/tester.py
+++ b/tests/jax/models/clip/tester.py
@@ -23,7 +23,11 @@ class FlaxCLIPTester(ModelTester):
 
     # @override
     def _get_model(self) -> FlaxPreTrainedModel:
-        return FlaxCLIPModel.from_pretrained(self._model_name)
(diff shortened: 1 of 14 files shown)


sgligorijevicTT:
FYI @kmitrovicTT I added an `accuracy_fail()` marker function analogously to `runtime_fail()` we have. I will hold off on merging this until you change the dashboard to process it too.

tests/utils.py
```diff
@@ -17,6 +17,10 @@ def runtime_fail(reason: str) -> str:
     return f"Runtime failed: {reason}"
 
 
+def accuracy_fail(reason: str) -> str:
+    return f"Accuracy failed: {reason}"
+
+
```
mrakitaTT, line 23: We (and other frontends) are currently in the process of changing this infra so that bringup status is recorded as property instead of being parsed from the xfail message, so these will be removed soon. In order to not break the dashboard in the meantine, can you please revert this and use `runtime_fail` instead.

tests/jax/models/clip/tester.py
```diff
@@ -23,7 +23,11 @@ def __init__(
 
     # @override
     def _get_model(self) -> FlaxPreTrainedModel:
-        return FlaxCLIPModel.from_pretrained(self._model_name)
+        from_pt = False
+        if self._model_name == "openai/clip-vit-large-patch14-336":
+            from_pt = True
```
mrakitaTT, line 28: `from_pt = self._model_name == "openai/clip-vit-large-patch14-336"`

tests/utils.py
sgligorijevicTT: Good to know. Reverting.

mrakitaTT (approved):

mrakitaTT:
I'll proceed to merge this so it can be picked up by tonight's nightly for the dashboard :)

PR #326, opened 2025-03-09
Manual uplift 2025 03 09

sgligorijevicTT:
### What's changed
I removed code that was setting a value for an option that no longer exists in tt-mlir

Test failure reasons will be updated in a followup PR


mrakitaTT (approved):
Thank you for catching this!

PR #337, opened 2025-03-11
Add pillow as a dependency to fix BEiT and CLIP tests

sgligorijevicTT:
### Problem description
Preprocessor classes depend on the pillow library and error out if it is not installed.
I didn't catch this before as apparently I had pillow installed in my env.

### What's changed
~~I removed the use of preprocessor classes to avoid taking an extra dependency.~~
PR has been repurposed for adding pillow as a dependency

### Checklist

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -195,3 +195,4 @@ The following dependencies are utilized by this project but are not explicitly
 distributed as part of the software:
 
 - xla/pjrt/c/pjrt_c_api.h - https://github.com/openxla/xla/blob/main/LICENSE
+- pillow - Custom License (https://github.com/python-pillow/Pillow/blob/main/LICENSE)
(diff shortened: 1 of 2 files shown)


mrakitaTT:
I think we should just add `pillow` to our `requirements.txt`. It is okay as long as we are not loading images from disk or internet.

The reason being is that certain models might require specific preprocessing in order to normalize the input, so if we have to do it manually we risk providing inputs in incorrect format for the model. Better to apply the preprocessing that is already built with the model.

mrakitaTT:
Let's also add pillow to the LICENSE at the bottom:
```
- pillow - Custom License (https://github.com/python-pillow/Pillow/blob/main/LICENSE)
```

mrakitaTT (approved):

PR #338, opened 2025-03-11
Added multichip support for fully manual sharding

ajakovljevicTT:
Adding PJRT support for fully manual sharding, with one test for eltwise ops on multichip architectures. 

Currently in draft until the tt-mlir changes go through.

High level overview:

- We now have information on the sharding from the StableHLO graph
- That information is used to create MultiDevice runtime input tensors, with strategies (as defined by tt-mlir) corresponding to those shardings. 
- The input tensors are then passed to the tt runtime `submit()` function.
- Now, the return tensors of the function can be sharded or replicated tensor, or a single device tensor
- The single device tensors goes through the same path as before
- If the tensor is just replicated, we feed to the xla backend the pointer to the tensor on the first device, as that is how it expects the input
- If it is sharded, we feed the invidual shards per device. 

- For now, please ignore the changes to the tt-mlir hash, as there are some changes waiting to be approved and merged on that side.
- Tests will be enabled in a future PR.

### DIFF
--- a/inc/common/pjrt_implementation/buffer_instance.h
+++ b/inc/common/pjrt_implementation/buffer_instance.h
@@ -25,11 +25,13 @@ class BufferInstance {
 public:
   BufferInstance(DeviceInstance &device, tt::runtime::Tensor &tensor,
                  const std::vector<std::uint32_t> &shape,
-                 const std::vector<std::uint32_t> &stride);
+                 const std::vector<std::uint32_t> &stride,
+                 std::pair<tt::target::DataType, size_t> tt_buffer_type);
 
   BufferInstance(DeviceInstance &device, tt::runtime::Tensor &tensor,
                  const std::vector<std::uint32_t> &shape,
                  const std::vector<std::uint32_t> &stride,
+                 std::pair<tt::target::DataType, size_t> tt_buffer_type,
                  std::shared_ptr<void> host_buffer_ptr);
   BufferInstance(DeviceInstance &device);
   ~BufferInstance();
@@ -58,10 +60,15 @@ public:
                             EventInstance **done_event);
 
   const int64_t *dims() { return dims_.data(); }
+  const std::vector<std::uint32_t> get_shape();
   size_t num_dims() { return dims_.size(); }
   void setType(PJRT_Buffer_Type Type) { DataType = Type; }
   std::optional<PJRT_Buffer_Type> getType() { return DataType; }
-
+  std::shared_ptr<void> get_host_buffer_ptr() { return host_buffer_ptr_; }
+  std::vector<std::uint32_t> get_stride() { return stride_; }
+  std::pair<tt::target::DataType, size_t> get_tt_buffer_type() {
+    return tt_buffer_type_;
+  }
   // Get the data type for a tensor through runtime if DataType is not set.
   PJRT_Buffer_Type getRuntimeType();
 
@@ -80,6 +87,7 @@ private:
   std::vector<int64_t> dims_;
   std::vector<std::uint32_t> stride_;
   tt::runtime::Tensor tensor_;
+  std::pair<tt::target::DataType, size_t> tt_buffer_type_;
 
   std::vector<int64_t> minor_to_major_;
   std::vector<int64_t> tile_dims_;
--- a/inc/common/pjrt_implementation/buffer_instance.h
+++ b/inc/common/pjrt_implementation/buffer_instance.h
@@ -42,7 +42,8 @@ public:
   static void BindApi(PJRT_Api *api);
 
   // iree_hal_buffer_view_t* buffer_view() { return buffer_view_.get(); }
-  DeviceInstance &device() const { return device_; }
+  DeviceInstance &device() { return device_; }
+  const DeviceInstance &device() const { return device_; }
   tt_pjrt_status AsyncDeallocate();
   tt_pjrt_status Delete();
   bool is_deleted() { return is_deleted_; }
@@ -59,14 +60,18 @@ public:
   tt_pjrt_status CopyToHost(void *dst, size_t dst_size,
                             EventInstance **done_event);
 
-  const int64_t *dims() { return dims_.data(); }
-  const std::vector<std::uint32_t> get_shape();
+  const int64_t *getRawDimensions() { return dims_.data(); }
+  const std::vector<std::uint32_t> getDimensions() const {
+    return std::vector<std::uint32_t>(dims_.begin(), dims_.end());
+  }
   size_t num_dims() { return dims_.size(); }
   void setType(PJRT_Buffer_Type Type) { DataType = Type; }
   std::optional<PJRT_Buffer_Type> getType() { return DataType; }
-  std::shared_ptr<void> get_host_buffer_ptr() { return host_buffer_ptr_; }
-  std::vector<std::uint32_t> get_stride() { return stride_; }
-  std::pair<tt::target::DataType, size_t> get_tt_buffer_type() {
+  const std::shared_ptr<void> get_host_buffer_ptr() const {
+    return host_buffer_ptr_;
+  }
+  const std::vector<std::uint32_t> get_stride() const { return stride_; }
+  std::pair<tt::target::DataType, size_t> get_tt_buffer_type() const {
     return tt_buffer_type_;
   }
   // Get the data type for a tensor through runtime if DataType is not set.
--- a/inc/common/pjrt_implementation/buffer_instance.h
+++ b/inc/common/pjrt_implementation/buffer_instance.h
@@ -42,7 +42,7 @@ public:
   static void BindApi(PJRT_Api *api);
 
   // iree_hal_buffer_view_t* buffer_view() { return buffer_view_.get(); }
-  DeviceInstance &device() { return device_; }
+  DeviceInstance &device() const { return device_; }
   tt_pjrt_status AsyncDeallocate();
   tt_pjrt_status Delete();
   bool is_deleted() { return is_deleted_; }
@@ -72,7 +72,7 @@ public:
   // Get the data type for a tensor through runtime if DataType is not set.
   PJRT_Buffer_Type getRuntimeType();
 
-  int unique_id() { return unique_id_; }
+  int unique_id() const { return unique_id_; }
 
 private:
   static int id_counter_;
--- a/inc/common/pjrt_implementation/buffer_instance.h
+++ b/inc/common/pjrt_implementation/buffer_instance.h
@@ -61,7 +61,7 @@ public:
                             EventInstance **done_event);
 
   const int64_t *getRawDimensions() { return dims_.data(); }
-  const std::vector<std::uint32_t> getDimensions() const {
+  std::vector<std::uint32_t> getDimensions() const {
     return std::vector<std::uint32_t>(dims_.begin(), dims_.end());
   }
   size_t num_dims() { return dims_.size(); }
--- a/inc/common/pjrt_implementation/buffer_instance.h
+++ b/inc/common/pjrt_implementation/buffer_instance.h
@@ -67,10 +67,10 @@ public:
   size_t num_dims() { return dims_.size(); }
   void setType(PJRT_Buffer_Type Type) { DataType = Type; }
   std::optional<PJRT_Buffer_Type> getType() { return DataType; }
-  const std::shared_ptr<void> get_host_buffer_ptr() const {
+  const std::shared_ptr<void> &get_host_buffer_ptr() const {
     return host_buffer_ptr_;
   }
-  const std::vector<std::uint32_t> get_stride() const { return stride_; }
+  const std::vector<std::uint32_t> &get_stride() const { return stride_; }
   std::pair<tt::target::DataType, size_t> get_tt_buffer_type() const {
     return tt_buffer_type_;
   }
--- a/inc/common/pjrt_implementation/loaded_executable_instance.h
+++ b/inc/common/pjrt_implementation/loaded_executable_instance.h
@@ -67,9 +67,10 @@ private:
   size_t num_devices_to_utilize_;
   std::vector<ResidentExecutable> resident_executables_;
 
-  std::unordered_map<std::string, std::string> getStrategyMapFromSharding(
+  tt_pjrt_status fillStrategyMapFromSharding(
       const mlir::tt::sharding_utils::MeshSharding &meshSharding,
-      size_t num_devices);
+      size_t num_devices,
+      std::unordered_map<std::string, std::string> &strategy);
 
   tt::runtime::Tensor getTensorFromStrategy(
       const std::unordered_map<std::string, std::string> &strategy,
@@ -78,7 +79,7 @@ private:
   std::vector<std::uint32_t> getOuputShape(size_t index, size_t num_devices);
 
   // Returns is output on the specified index replicated.
-  bool isOutputReplicated(size_t index);
+  bool isOutputReplicated(size_t index) const;
 
   void fillPJRTOutputLists(
       const std::vector<std::vector<tt::runtime::Tensor>> &rt_outputs,
@@ -89,6 +90,17 @@ private:
   getDeviceIds(PJRT_Buffer *const *const *argument_lists,
                const std::vector<DeviceInstance *> &addressable_devices,
                size_t num_args, size_t num_devices);
+
+  // Given an output list, return the number of outputs of an executable.
+  size_t getNumberOfOutputs(const std::vector<std::vector<tt::runtime::Tensor>>
+                                &rt_outputs_list) const;
+
+  // Return a tensor representing an output on a particular device with a
+  // particular index.
+  tt::runtime::Tensor
+  getOuputTensor(size_t device_index, size_t output_index,
+                 const std::vector<std::vector<tt::runtime::Tensor>>
+                     &rt_outputs_list) const;
 };
 
 } // namespace tt::pjrt
--- a/inc/common/pjrt_implementation/loaded_executable_instance.h
+++ b/inc/common/pjrt_implementation/loaded_executable_instance.h
@@ -80,10 +80,10 @@ private:
   // Returns is output on the specified index replicated.
   bool isOutputReplicated(size_t index);
 
-  void
-  fillPJRTOutputLists(std::vector<std::vector<tt::runtime::Tensor>> &rt_outputs,
-                      const std::vector<tt::runtime::TensorDesc> &output_specs,
-                      size_t num_devices, PJRT_Buffer **const *output_lists);
+  void fillPJRTOutputLists(
+      const std::vector<std::vector<tt::runtime::Tensor>> &rt_outputs,
+      const std::vector<tt::runtime::TensorDesc> &output_specs,
+      size_t num_devices, PJRT_Buffer **const *output_lists);
 
   std::vector<int>
   getDeviceIds(PJRT_Buffer *const *const *argument_lists,
--- a/inc/common/pjrt_implementation/loaded_executable_instance.h
+++ b/inc/common/pjrt_implementation/loaded_executable_instance.h
@@ -81,10 +81,14 @@ private:
   bool isOutputReplicated(size_t index);
 
   void
-  fillPJRTOutputLists(PJRT_Buffer **const *output_lists,
-                      std::vector<std::vector<tt::runtime::Tensor>> &rt_outputs,
+  fillPJRTOutputLists(std::vector<std::vector<tt::runtime::Tensor>> &rt_outputs,
                       const std::vector<tt::runtime::TensorDesc> &output_specs,
-                      size_t num_devices);
+                      size_t num_devices, PJRT_Buffer **const *output_lists);
+
+  std::vector<int>
+  getDeviceIds(PJRT_Buffer *const *const *argument_lists,
+               const std::vector<DeviceInstance *> &addressable_devices,
+               size_t num_args, size_t num_devices);
 };
 
 } // namespace tt::pjrt
--- a/inc/common/pjrt_implementation/loaded_executable_instance.h
+++ b/inc/common/pjrt_implementation/loaded_executable_instance.h
@@ -67,25 +67,34 @@ private:
   size_t num_devices_to_utilize_;
   std::vector<ResidentExecutable> resident_executables_;
 
+  // Fills a std::map<std::string, std::string>, represeting the strategy of
+  // multichip tensor creation (replidated, sharded etc.) from the meshSharding
+  // object.
   tt_pjrt_status fillStrategyMapFromSharding(
       const mlir::tt::sharding_utils::MeshSharding &meshSharding,
       size_t num_devices,
       std::unordered_map<std::string, std::string> &strategy);
 
+  // Returns a runtime tensor given a creation strategy and a vestor of pointers
+  // to data.
   tt::runtime::Tensor getTensorFromStrategy(
       const std::unordered_map<std::string, std::string> &strategy,
       BufferInstance *buffer, std::vector<std::shared_ptr<void>> &data);
 
+  // Returns the shape of the output on the specified index.
   std::vector<std::uint32_t> getOuputShape(size_t index, size_t num_devices);
 
   // Returns is output on the specified index replicated.
   bool isOutputReplicated(size_t index) const;
 
+  // Fills the output lists of the PJRT API with the outputs of tt runtime
+  // execution.
   void fillPJRTOutputLists(
       const std::vector<std::vector<tt::runtime::Tensor>> &rt_outputs,
       const std::vector<tt::runtime::TensorDesc> &output_specs,
       size_t num_devices, PJRT_Buffer **const *output_lists);
 
+  // Gets the device ids from the addressable devices.
   std::vector<int>
   getDeviceIds(PJRT_Buffer *const *const *argument_lists,
                const std::vector<DeviceInstance *> &addressable_devices,
--- a/inc/common/pjrt_implementation/loaded_executable_instance.h
+++ b/inc/common/pjrt_implementation/loaded_executable_instance.h
@@ -34,6 +34,7 @@ public:
       : client_(client), image_(image),
         addressable_devices_(addressable_devices),
         num_devices_to_utilize_(num_devices_to_utilize) {}
+
   ~LoadedExecutableInstance();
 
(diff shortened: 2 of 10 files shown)


inc/common/pjrt_implementation/loaded_executable_instance.h
```diff
@@ -65,6 +66,25 @@ class LoadedExecutableInstance {
...
+  std::vector<std::uint32_t> getOuputShape(size_t index, size_t num_devices);
+
+  // Returns is output on the specified index replicated.
+  bool isOutputReplicated(size_t index);
+
+  void
+  fillPJRTOutputLists(PJRT_Buffer **const *output_lists,
+                      std::vector<std::vector<tt::runtime::Tensor>> &rt_outputs,
+                      const std::vector<tt::runtime::TensorDesc> &output_specs,
+                      size_t num_devices);
```
mrakitaTT, line 87: Input-only arguments should be ordered before the output or in-out arguments, see more here: https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

inc/common/pjrt_implementation/buffer_instance.h
```diff
@@ -25,11 +25,13 @@ class BufferInstance {
 public:
   BufferInstance(DeviceInstance &device, tt::runtime::Tensor &tensor,
                  const std::vector<std::uint32_t> &shape,
-                 const std::vector<std::uint32_t> &stride);
+                 const std::vector<std::uint32_t> &stride,
+                 std::pair<tt::target::DataType, size_t> tt_buffer_type);
```
mrakitaTT, line 29: This should be a structure with named fields instead of passing `std::pair<tt::target::DataType, size_t>` around, but since this is something that we have to fix across the project (stemming from `src/common/pjrt_implementation/utils.cc`) let's just open an issue and come back to it later. It's also quite confusing in this class having `DataType` field which is actually `PJRT_Buffer_Type`, this class should be cleaned up a bit.

tests/infra/device_runner.py
```diff
@@ -97,9 +97,7 @@ def _put_sharded_tensor_on_multichip_device(
         TODO: This can be omitted when we find a way to get sharding information from the StableHLO
         code back to jax through a protobuf (issue #227).
         """
```
mrakitaTT, line 99: comment needs to be updated now

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -10,13 +10,17 @@
...
+#include <string>
+#include <unordered_map>
 
 #include "common/pjrt_implementation/buffer_instance.h"
 #include "common/pjrt_implementation/client_instance.h"
 #include "common/pjrt_implementation/error_instance.h"
 #include "common/pjrt_implementation/utils.h"
 
+// tt-mlir includes
+#include "ttmlir/Dialect/TT/IR/TTOpsTypes.h"
```
mrakitaTT, line 22: this group should be above tt-xla includes

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -78,63 +82,184 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
   // from the VHLO module.
   assert(args->num_devices == num_devices_to_utilize_);
 
-  int dev_index = 0;
   const tt::runtime::Binary &binary = image_->get_binary();
+  size_t num_devices = args->num_devices;
```
mrakitaTT, line 86: Put before the above assert and then use `num_devices` instead of `args->num_devices` in the assert, or just simply use `args->num_devices` everywhere

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -78,63 +82,184 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
 
   std::vector<tt::runtime::Tensor> rt_inputs;
```
mrakitaTT, line 88: Declare down below before the loop that fills it

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -78,63 +82,184 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
-  rt_inputs.reserve(args->num_args);
 
-  std::unordered_set<int> device_ids;
+  std::vector<int> device_ids;
 
-  for (size_t i = 0; i < args->num_args; ++i) {
-    BufferInstance *buffer =
-        BufferInstance::Unwrap(args->argument_lists[dev_index][i]);
-    rt_inputs.emplace_back(buffer->getTensor());
+  for (size_t i = 0; args->num_args && i < num_devices; i++) {
```
mrakitaTT, line 92: This loop can be extracted into separate function `collectDeviceIds`

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -78,63 +82,184 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+    BufferInstance *buffer = BufferInstance::Unwrap(args->argument_lists[i][0]);
```
mrakitaTT, line 93: Should be declared `const`

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -78,63 +82,184 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
     int64_t buffer_device_id =
         buffer->device().device_description()->getDeviceId();
-    device_ids.insert(buffer_device_id);
+    device_ids.push_back(buffer_device_id);
     DLOG_F(INFO, "Runtime input id: %d", buffer->unique_id());
```
mrakitaTT, line 97: `INFO` seems too high for something so verbose, do you ever find this log useful?

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -78,63 +82,184 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+  int size_inputs = rt_inputs.size();
   std::vector<tt::runtime::Tensor> rt_outputs =
       tt::runtime::submit(device, binary, 0, rt_inputs);
+  std::vector<tt::runtime::TensorDesc> output_specs =
+      binary.getProgramOutputs(0);
+  std::vector<std::vector<tt::runtime::Tensor>> rt_outputs_list(num_devices);
 
-  assert(rt_outputs.size() == image_->get_num_outputs());
+  for (size_t i = 0; i < output_specs.size(); ++i) {
+    std::vector<tt::runtime::Tensor> unitlized_output =
```
mrakitaTT, line 130: typo in `unitlized_output`

mrakitaTT (commented):
I want to review changes in `src/common/pjrt_implementation/loaded_executable_instance.cc` a bit more tomorrow

src/common/pjrt_implementation/buffer_instance.cc
```diff
@@ -224,4 +227,13 @@ PJRT_Buffer_Type BufferInstance::getRuntimeType() {
   return tt::pjrt::utils::convertElementTypeToBufferType(Type);
 }
 
+const std::vector<std::uint32_t> BufferInstance::get_shape() {
```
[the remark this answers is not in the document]
ajakovljevicTT: The information is already saved in `dims_` but the issue is that it is represented as a vector of int_64, while the usage of buffer shapes requires vectors of int_32, so this function basically just does that translation. If I added another member variable, there would be two member variables that have the same information, just different types, which I'm not sure we should do.

src/common/pjrt_implementation/loaded_executable_instance.cc
ajakovljevicTT: Not really, it was there when we first inherited the tt-xla repo, so no one actually changed it. Will put it down to DEBUG.

inc/common/pjrt_implementation/buffer_instance.h
ajakovljevicTT: Noted! Here's the issue https://github.com/tenstorrent/tt-xla/issues/349

ajakovljevicTT:
Note: This PR is currently blocked on these two tt-mlir PRs:
https://github.com/tenstorrent/tt-mlir/pull/2460
https://github.com/tenstorrent/tt-mlir/pull/2386

ajakovljevicTT:
> Note: This PR is currently blocked on these two `tt-mlir` PRs: [tenstorrent/tt-mlir#2460](https://github.com/tenstorrent/tt-mlir/pull/2460) [tenstorrent/tt-mlir#2386](https://github.com/tenstorrent/tt-mlir/pull/2386)

PR have been merged, no blockers on the tt-mlir side.

src/common/pjrt_implementation/buffer_instance.cc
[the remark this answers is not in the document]
mrakitaTT: Oh I see. Since we need pointer to data in order to pass dimensions to `PJRT_Buffer_Dimensions` and `PJRT_Buffer_UnpaddedDimensions`, that gives priority to storing dimensions in `int64_t` format. Then I would make these changes:
- Change the `BufferInstance` constructor to construct the `dims_` by `dims_(shape.begin(), shape.end())`
- Rename `dims()` to `getRawDimensions()`
- Rename the `get_shape` to `getDimensions()` and define it inline in the header file like:
```
std::vector<std::uint32_t> getDimensions() const { return std::vector<std::uint32_t>(dims_.begin(), dims_.end()); }
```

src/common/pjrt_implementation/loaded_executable_instance.cc
mrakitaTT: still there :)

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -76,65 +86,196 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
 
   // Check that the number of devices matches the number of devices counted
   // from the VHLO module.
-  assert(args->num_devices == num_devices_to_utilize_);
+  size_t num_devices = args->num_devices;
+  assert(num_devices == num_devices_to_utilize_);
```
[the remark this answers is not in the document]
mrakitaTT: Similarly below we should check if `output_specs.size() == num_devices`

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -76,65 +86,196 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+  mlir::tt::MeshShardType meshType = outputSharding.getShardType();
+  return meshType == mlir::tt::MeshShardType::Replicate;
+}
+
+void LoadedExecutableInstance::fillPJRTOutputLists(
+    std::vector<std::vector<tt::runtime::Tensor>> &rt_outputs_list,
+    const std::vector<tt::runtime::TensorDesc> &output_specs,
+    size_t num_devices, PJRT_Buffer **const *output_lists) {
+  std::vector<bool> is_replicated;
+  size_t num_outputs = rt_outputs_list[0].size();
```
mrakitaTT, line 229: We should assert here that `rt_outputs_list.size() == output_specs.size() == num_devices`. Assert is okay here because at this point these conditions have already been checked, so if it happens it is not an error from the external factors but the bug in our code.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -76,65 +86,196 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+  for (size_t i = 0; i < rt_outputs_list.size(); i++) {
+    num_outputs = std::max(num_outputs, rt_outputs_list[i].size());
+  }
+  for (size_t i = 0; i < output_specs.size(); ++i) {
+    is_replicated.push_back(isOutputReplicated(i));
+  }
+  for (int k = 0; k < num_devices; k++) {
+    for (size_t i = 0; i < num_outputs; ++i) {
+      // If the output is replicated, we just put the same tensor in all the
+      // devices.
```
mrakitaTT, line 239: You should also mention that this is what PJRT expects

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -76,65 +86,196 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+          isOutputReplicated(i) ? rt_outputs_list[0][i] : rt_outputs_list[k][i];
+      std::vector<std::uint32_t> output_shape = getOuputShape(i, num_devices);
+      std::pair<tt::target::DataType, size_t> type_pair = {
+          output_specs[i].dataType, output_specs[i].itemsize};
+      auto result_buffer = std::make_unique<BufferInstance>(
+          *this->addressable_devices_[k], output_tensor, output_shape,
+          output_specs[i].stride, type_pair);
+      result_buffer->setType(tt::pjrt::utils::convertElementTypeToBufferType(
+          output_specs[i].dataType));
+      DLOG_F(DEBUG, "Runtime output id: %d", result_buffer->unique_id());
```
mrakitaTT, line 250: I think you can remove this log too, we'll add it back if we see a need

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -76,65 +86,196 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+}
+
+void LoadedExecutableInstance::fillPJRTOutputLists(
+    std::vector<std::vector<tt::runtime::Tensor>> &rt_outputs_list,
+    const std::vector<tt::runtime::TensorDesc> &output_specs,
+    size_t num_devices, PJRT_Buffer **const *output_lists) {
+  std::vector<bool> is_replicated;
+  size_t num_outputs = rt_outputs_list[0].size();
+  for (size_t i = 0; i < rt_outputs_list.size(); i++) {
+    num_outputs = std::max(num_outputs, rt_outputs_list[i].size());
```
mrakitaTT, line 231: Hm this confuses me, how can we have different number of outputs per device? And how is it okay for the double loop below to iterate over maximal number and index into `rt_outputs_list`?

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -76,65 +86,196 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+  }
+  for (size_t i = 0; i < output_specs.size(); ++i) {
+    is_replicated.push_back(isOutputReplicated(i));
+  }
+  for (int k = 0; k < num_devices; k++) {
+    for (size_t i = 0; i < num_outputs; ++i) {
```
mrakitaTT, line 237: Use better index names

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -76,65 +86,196 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+    if (shardShape.size() == 2) {
+      strategy["strategy"] = "shard_2d";
+      strategy["mesh_shape_y"] = std::to_string(shardShape[0]);
+      strategy["mesh_shape_x"] = std::to_string(shardShape[1]);
+    } else if (shardShape.size() == 1) {
+      strategy["strategy"] = "shard";
+      strategy["shard_dim"] = "0";
+    } else {
+      DLOG_F(ERROR, "Invalid mesh sharding type");
+      return strategy;
```
mrakitaTT, line 170: If you transfer this function into tt-mlir, this should throw an error. Below too. Otherwise, we would return `tt_pjrt_status::kInternal` from here.

src/common/pjrt_implementation/loaded_executable_instance.cc
ajakovljevicTT: This is not always true, when the output is "replicated". Then output_specs.size() and rt_outputs_list.size() will be 1, and the number of devices can be larger than 1.

src/common/pjrt_implementation/loaded_executable_instance.cc
ajakovljevicTT: This is if the output is replicated., then only the first output list will have the normal size, and the rest will be 0.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -76,65 +86,196 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
 
-  tt::runtime::Device device = tt::runtime::openDevice(device_ids_vector);
+  std::vector<int> device_ids = getDeviceIds(
+      args->argument_lists, addressable_devices_, args->num_args, num_devices);
 
+  tt::runtime::Device device = tt::runtime::openDevice(device_ids);
+  std::vector<tt::runtime::Tensor> input_tensors;
+  int size_inputs = rt_inputs.size();
+  auto workaround_env =
+      tt::runtime::workaround::Env::get(true, true, false, true);
```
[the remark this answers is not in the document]
ajakovljevicTT: Currently, the multichip support in the tt-mlir is only available if we turn on the workaround flag `toLayoutAPIAssumeSingleChip` to false. This is how that is done, I will add a comment explaining it in more detail.

inc/common/pjrt_implementation/buffer_instance.h
```diff
@@ -40,7 +42,7 @@ class BufferInstance {
   static void BindApi(PJRT_Api *api);
 
   // iree_hal_buffer_view_t* buffer_view() { return buffer_view_.get(); }
-  DeviceInstance &device() { return device_; }
+  DeviceInstance &device() const { return device_; }
```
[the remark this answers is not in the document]
ajakovljevicTT: I actually needed both, thanks for this!

tests/jax/multichip/manual/unary_eltwise.py
```diff
@@ -13,23 +13,21 @@
 @pytest.mark.parametrize(
     ("x_shape", "mesh_shape", "axis_names"), [((256, 256), (1, 2), ("x", "y"))]
```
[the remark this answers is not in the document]
ajakovljevicTT: Copied from some other tests, will change.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -76,65 +86,196 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+      args->device_complete_events[i] = *(new EventInstance());
   }
 
   tt::runtime::closeDevice(device);
 
   return tt_pjrt_status::kSuccess;
 }
 
+std::unordered_map<std::string, std::string>
+LoadedExecutableInstance::getStrategyMapFromSharding(
```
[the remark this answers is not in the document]
ajakovljevicTT: I agree, however, I think it might be best to leave it for now to push through the PR. I will make a separate `tt-mlir` PR which adds this, and change tt-xla when this is merged.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -76,65 +86,196 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+    DLOG_F(ERROR, "Invalid mesh sharding type");
+    return strategy;
+  }
+  return strategy;
+}
+
+tt::runtime::Tensor LoadedExecutableInstance::getTensorFromStrategy(
+    const std::unordered_map<std::string, std::string> &strategy,
+    BufferInstance *buffer, std::vector<std::shared_ptr<void>> &data) {
+  if (strategy.at("strategy") == "identity") {
```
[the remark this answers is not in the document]
ajakovljevicTT: Sure, here's the issue: https://github.com/tenstorrent/tt-mlir/issues/2513

inc/common/pjrt_implementation/buffer_instance.h
```diff
@@ -57,15 +60,24 @@ class BufferInstance {
...
+    return std::vector<std::uint32_t>(dims_.begin(), dims_.end());
+  }
   size_t num_dims() { return dims_.size(); }
   void setType(PJRT_Buffer_Type Type) { DataType = Type; }
   std::optional<PJRT_Buffer_Type> getType() { return DataType; }
-
+  const std::shared_ptr<void> get_host_buffer_ptr() const {
+    return host_buffer_ptr_;
+  }
+  const std::vector<std::uint32_t> get_stride() const { return stride_; }
```
mrakitaTT, line 73: When you are not returning reference to the object there is no point in doing `const std::shared_ptr<void>` or `const std::vector<std::uint32_t>`

inc/common/pjrt_implementation/loaded_executable_instance.h
mrakitaTT, line 1: Please add header comments to all functions that you've added that are missing one

src/common/pjrt_implementation/buffer_instance.cc
```diff
@@ -166,7 +165,7 @@ BufferInstance::GetMemoryLayout(PJRT_Buffer_GetMemoryLayout_Args *args) {
   tile_dim_sizes_[0] = rank;
   tile_dims_.resize(rank);
   for (size_t i = 0; i < rank; i++) {
-    tile_dims_[i] = dims()[i];
+    tile_dims_[i] = getRawDimensions()[i];
```
mrakitaTT, line 168: Can simply use `dims_` here

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -76,65 +82,235 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+      buffer =
+          BufferInstance::Unwrap(args->argument_lists[device_index][arg_num]);
+      data.push_back(buffer->get_host_buffer_ptr());
+    }
+    std::unordered_map<std::string, std::string> strategy;
+    if (fillStrategyMapFromSharding(image_->getInputSharding(arg_num),
+                                    num_devices,
+                                    strategy) != tt_pjrt_status::kSuccess) {
+      return tt_pjrt_status::kInternal;
+    }
```
mrakitaTT, line 108: It would be better to do:
```
tt_pjrt_status fill_status = fillStrategyMapFromSharding(image_->getInputSharding(arg_num), num_devices, strategy);
if (fill_status != tt_pjrt_status:: kSuccess) {
  return fill_status;
}
```
that way potentially `fillStrategyMapFromSharding` can return a different status than `kInternal`.

src/common/pjrt_implementation/loaded_executable_instance.cc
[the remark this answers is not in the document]
mrakitaTT: Remove the `auto workaround_env` since it is not used, and add a link to the issue on our side which tracks removing this env variable.

src/common/pjrt_implementation/loaded_executable_instance.cc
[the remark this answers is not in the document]
mrakitaTT: Please add a TODO comment above this function to explain that we are using naked strings because strategy is not appropriately defined in `tt::runtime` as some struct/class with enum fields, but instead simple map is used. And link that issue in the comment.

src/common/pjrt_implementation/loaded_executable_instance.cc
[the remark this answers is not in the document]
mrakitaTT: Sure, please just add a TODO comment above this function to transfer it to tt-mlir, and put the link to the issue in the comment.

src/common/pjrt_implementation/loaded_executable_instance.cc
mrakitaTT: Oh, okay that complicates the assert but I think we should still have some checks that these expected conditions are met depending if the output is replicated or not.

ajakovljevicTT: I think the logic for these asserts would be too complex if we have both non-replicated and replicated outputs, however, one thing that should always be true is `num_outputs == output_specs.size()` so I added that as an assert.

inc/common/pjrt_implementation/buffer_instance.h
```diff
@@ -57,15 +60,24 @@ class BufferInstance {
   tt_pjrt_status CopyToHost(void *dst, size_t dst_size,
                             EventInstance **done_event);
 
-  const int64_t *dims() { return dims_.data(); }
+  const int64_t *getRawDimensions() { return dims_.data(); }
+  const std::vector<std::uint32_t> getDimensions() const {
```
mrakitaTT, line 64: Sorry I missed this one `const std::vector<std::uint32_t>`, const not needed here

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -76,65 +82,241 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+  return tt_pjrt_status::kSuccess;
+}
 
-    tt::runtime::Tensor untilized_output_tensor =
-        tt::runtime::toHost(rt_outputs[i], /*untilize=*/true)[0];
-    auto result_buffer = std::make_unique<BufferInstance>(
-        *this->addressable_devices_[dev_index], untilized_output_tensor,
-        image_->get_output_shape(i), image_->get_output_stride(i));
-    tt::runtime::deallocateTensor(rt_outputs[i], /*force=*/true);
+// We are using std::maps with strings as that is the way it is definet in the
```
mrakitaTT, line 203: nit: typo in `definet`

mrakitaTT: Please also put `TODO:` at the beginning of the comment so we don't forget to remove the comment and to easier locate all todos in the code

mrakitaTT (approved):
Just a few small unresolved comments left :)

Issue #333, opened 2025-03-10
Change CI tests to run only multichip tests on n300 and other tests on n150


mrakitaTT:
Change CI tests to run only multichip tests on n300 and other tests on n150. There is not that much benefit to compiler from running non-multichip tests on both n150 and n300.

PR #382, opened 2025-03-20
Small refactor of tests and adding different multichip modes of testing

ajakovljevicTT:
Refactored all tests into single_chip and multi_chip directories, additionaly added a flag which switches between automatic, fully automatic and manual sharding on multidevice. Enabled multidevice testing in CI.

Fixes https://github.com/tenstorrent/tt-xla/issues/333


.github/workflows/build-and-test.yml
```diff
@@ -253,10 +253,12 @@ jobs:
...
-        pytest ./tests/jax \
-               -m "${{ inputs.test_mark }}" \
-               --junitxml=${{ steps.strings.outputs.test_report_path }} \
-               2>&1 | tee pytest.log
+        if [[ "${{ matrix.build.pytest_mark }}" == *"n300"* ]]; then
+          export XLA_FLAGS="--xla_force_host_platform_device_count=2"
+        fi
+        pytest ./tests/jax -m "${{ matrix.build.pytest_mark }}" \
+        --junitxml=${{ steps.strings.outputs.test_report_path }} \
+        2>&1 | tee pytest.log
```
mrakitaTT, line 261: I would keep the indentation that was before for the pytest command, looked nicer

tests/infra/device_runner.py
```diff
@@ -20,6 +20,11 @@ class DeviceRunner:
 
     # ---------- Public methods ----------
 
+    @staticmethod
+    def run_on_unspecified_device(workload: Workload) -> Tensor:
```
mrakitaTT, line 24: One of the reasons we've introduced `DeviceRunner` is to be very specific on which device we run things, to avoid bugs we had where things were run on cpu instead of tt hw. Having function named like this goes against that principle and is kinda confusing :)

I would suggest using `run_on_multichip_device` function, and providing argument like `create_input_shardings`, so depending on that argument you would do `_put_multichip_workload_on_device` or not. That also simplifies the `test` function in `MultichipTester`:
```
device_res = DeviceRunner.run_on_multichip_device(compiled_device_workload, self._uses_device_put(multichip_mode));
```

tests/infra/device_runner.py
```diff
@@ -20,6 +20,11 @@ class DeviceRunner:
...
+        """Runs `workload` on unspecified device."""
+        return DeviceRunner.execute()
```
mrakitaTT, line 26: I guess you wanted to use `workload.execute()` here, but no tests hit this path currently so it wasn't caught :)

.github/workflows/build-and-test.yml
```diff
@@ -253,10 +253,12 @@ jobs:
       run: |
         export LD_LIBRARY_PATH="/opt/ttmlir-toolchain/lib/:${{ steps.strings.outputs.install-output-dir }}/lib:${LD_LIBRARY_PATH}"
         source venv/activate
-        pytest ./tests/jax \
-               -m "${{ inputs.test_mark }}" \
-               --junitxml=${{ steps.strings.outputs.test_report_path }} \
-               2>&1 | tee pytest.log
+        if [[ "${{ matrix.build.pytest_mark }}" == *"n300"* ]]; then
+          export XLA_FLAGS="--xla_force_host_platform_device_count=2"
+        fi
```
mrakitaTT, line 258: If this env variable gets left behind that means we have to be careful about the order of the runs in this pipeline. Maybe save the env state before running pytest command and then restore it later?

mrakitaTT: Btw I don't understand why is this needed now when we run n300 tests separately, why doesn't `simulate_cpu_mesh` decorator do the job? You've said that before you had an issue when the first test that gets run is singlechip, then jax initializes the backend without this flag and then multichip tests don't work. But now we run only multichip tests with one pytest command, so I don't understand how is this needed now?

tests/infra/multichip_tester.py
```diff
@@ -16,6 +17,12 @@
 from .workload import Workload
 
 
+class MultichipMode(Enum):
+    FULLY_MANUAL = 1
+    MANUAL = 2
+    AUTOMATIC = 3
```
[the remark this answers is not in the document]
ajakovljevicTT: I have added comments regarding which mode is which, but have left the names to align with the jax naming convention. Let me know what you think!

tests/infra/multichip_tester.py
```diff
@@ -139,6 +167,7 @@ def run_multichip_test_with_random_inputs(
     in_specs: Sequence[jax.sharding.PartitionSpec],
     out_specs: jax.sharding.PartitionSpec,
     use_shardy: bool,
```
kmitrovicTT, line 169: Missed it last time, but we don't use this flag. Furthermore, it looks confusing now since `use_shardy` is directly dependent on `multichip_mode`. That is: one of them is redundant.

tests/infra/multichip_tester.py
```diff
@@ -103,25 +116,40 @@ def test_with_random_inputs(
...
-            out_specs=self.out_specs,
+        module_sharded = (
+            shard_map(
+                executable,
+                mesh=self.device_mesh,
+                in_specs=self.in_specs,
+                out_specs=self.out_specs,
+            )
+            if self._uses_shard_map(multichip_mode)
+            else executable
```
kmitrovicTT, line 152: I have some difficulties understanding this part. What does this multichip mode dictate exactly? What do it's values represent? What does it mean to use `shard_map` vs `device_put` etc? I need some more context. Probably should be explain in `MultichipMode`s docstring.

tests/jax/multi_chip/n300/graphs/tensor_parallel/test_dot_psum.py
```diff
@@ -6,25 +6,55 @@
...
-@pytest.mark.skip(reason=failed_fe_compilation("Multichip still in development"))
-def test_psum(
+@pytest.mark.parametrize(
+    "multichip_mode",
+    [
+        MultichipMode.FULLY_MANUAL,
+        MultichipMode.MANUAL,
+        MultichipMode.AUTOMATIC,
+    ],
+)
```
kmitrovicTT, line 43: Parametrizing `use_shardy` and this is probably redundant as far as I understood the code.

.github/workflows/build-and-test.yml
```diff
@@ -253,10 +253,17 @@ jobs:
...
+          pytest ./tests/jax/multi_chip/n300  \
+                 -m "${{ matrix.build.pytest_mark }}" \
+                 --junitxml=${{ steps.strings.outputs.test_report_path }} \
+                 2>&1 | tee pytest.log
+        else
+          pytest ./tests/jax/single_chip/ \
+                 -m "${{ inputs.test_mark }}" \
+                 --junitxml=${{ steps.strings.outputs.test_report_path }} \
+                 2>&1 | tee pytest.log
+        fi
```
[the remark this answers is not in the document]
kmitrovicTT: Once again i back to noting that `matrix.build.pytest_mark == inputs.test_mark` if I am not mistaken.

.github/workflows/build-and-test.yml
```diff
@@ -155,16 +155,16 @@ jobs:
       fail-fast: false
       matrix:
         build: [
-          {runs-on: n150, name: "run"},
-          {runs-on: n300, name: "run"},
+          {runs-on: n150, name: "run", pytest_mark: "${{ inputs.test_mark }} and n150"},
+          {runs-on: n300, name: "run", pytest_mark: "${{ inputs.test_mark }} and n300"},
```
kmitrovicTT, line 159: What is this trying to accomplish? What does this `and` statement do?

kmitrovicTT (approved):

.github/workflows/build-and-test.yml
```diff
@@ -155,16 +155,16 @@ jobs:
...
         ]
 
     runs-on:
       - in-service
       - ${{ matrix.build.runs-on }}
 
     # Keep this name in sync with the fetch-job-id step
-    name: "build-and-run-tests ${{ inputs.test_mark }} (${{ matrix.build.runs-on }}, ${{ matrix.build.name }})"
+    name: "build-and-run-tests ${{ inputs.test_mark }} (${{ matrix.build.runs-on }}, ${{ matrix.build.name }}, ${{ matrix.build.pytest_mark }})"
```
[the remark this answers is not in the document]
vmilosevic: It should work like this:

inputs.test_mark 
will have value 'push' or 'nightly', indication which test we want to run.

matrix.build.pytest_mark
will have value 'push and 150' or 'push and 300'
depending on where it runs, to select test only for that HW, while still applying push/nightly selection

.github/workflows/build-and-test.yml
```diff
@@ -253,10 +253,17 @@ jobs:
...
+          pytest ./tests/jax/multi_chip/n300  \
+                 -m "${{ inputs.test_mark }}" \
+                 --junitxml=${{ steps.strings.outputs.test_report_path }} \
+                 2>&1 | tee pytest.log
+        else
+          pytest ./tests/jax/single_chip/ \
+                 -m "${{ inputs.test_mark }}" \
+                 --junitxml=${{ steps.strings.outputs.test_report_path }} \
+                 2>&1 | tee pytest.log
+        fi
```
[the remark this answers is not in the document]
nsumrakTT: As alternative you could add parameter dir to matrix and use ${{ matrix.build.dir }} instead of if.
Eg.
```
build: [
          {runs-on: n150, name: "run", dir: "./tests/jax/multi_chip/n300"},
          {runs-on: n300, name: "run", dir: "./tests/jax/single_chip/"},
        ]
 and then:
```
` pytest  ${{ matrix.build.dir }} $TEST_DIR -m "${{ inputs.test_mark }}"   ....`

nsumrakTT (approved):

tests/infra/device_runner.py
```diff
@@ -26,12 +26,17 @@ def run_on_tt_device(workload: Workload, device_num: int = 0) -> Tensor:
         return DeviceRunner._run_on_device(workload, DeviceType.TT, device_num)
 
     @staticmethod
-    def run_on_multichip_device(multichip_workload: MultichipWorkload) -> Tensor:
+    def run_on_multichip_device(
+        multichip_workload: MultichipWorkload, put_on_multichip_device: bool
+    ) -> Tensor:
         """Runs `workload` on a multichip device."""
```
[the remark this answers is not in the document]
kmitrovicTT: Isn't resolved.

kmitrovicTT (approved):

tests/infra/multichip_tester.py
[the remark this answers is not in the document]
mrakitaTT: I don't know I don't think JAX really has the convention, they do mention "automatic" in their docs where you just shard the inputs and let compiler take the wheel, but then they have many of these "manual" options where you take the wheel. I would rather be more explicit in our enum names so we can avoid the confusion over these manual options. Also, `MultichipMode` is kinda wrong name for the enum, because people can associate it with data/tensor/pipeline parallelism options. I am more in favour of something like this:
```
class ShardingMode(Enum):
    INPUTS = 1
    MODULE = 2
    INPUTS_AND_MODULE = 3
```

Issue #397, opened 2025-03-24
Add tests for BigBird Model


ctr-pmuruganTT:
There is an existing Flax implementation `FlaxBigBirdForCausalLM ` for BigBird model in huggingface transformers library, [here](https://huggingface.co/docs/transformers/en/model_doc/big_bird#transformers.FlaxBigBirdForCausalLM) is the documentation with the example code.

There are 3 variants with official model parameters published on huggingface:
-  bigbird_roberta_base  : https://huggingface.co/google/bigbird-roberta-base
-  bigbird_roberta_large : https://huggingface.co/google/bigbird-roberta-large
-  bigbird_base_trivia_itc :https://huggingface.co/google/bigbird-base-trivia-itc

Add tests for this model

PR #398, opened 2025-03-24
Add tests for Bigbird model

ctr-pmuruganTT:
### Ticket
closes https://github.com/tenstorrent/tt-xla/issues/397

### Problem description
Add tests for BigBird model

### What's changed
* a tester class is added to test all 3 variants
* separate test classes are written for
   - bigbird_roberta_base
   - bigbird_roberta_large
   - bigbird_base_trivia_itc
   
### Checklist

The logs for testing these files are attached

[bigbird_base_trivia_itc.log](https://github.com/user-attachments/files/19416340/bigbird_base_trivia_itc.log)
[bigbird_roberta_base_full_attention.log](https://github.com/user-attachments/files/19416341/bigbird_roberta_base_full_attention.log)
[bigbird_roberta_largee_full_attention.log](https://github.com/user-attachments/files/19416342/bigbird_roberta_largee_full_attention.log)


tests/jax/models/bigbird/tester.py
```diff
@@ -0,0 +1,46 @@
...
+        comparison_config: ComparisonConfig = ComparisonConfig(),
+        run_mode: RunMode = RunMode.INFERENCE,
+    ) -> None:
+        self._model_name = model_name
+        super().__init__(comparison_config, run_mode)
+
+    # @override
+    def _get_model(self) -> FlaxPreTrainedModel:
+        return FlaxBigBirdForCausalLM.from_pretrained(
+            self._model_name, attention_type="original_full"
```
sgligorijevicTT, line 27: Is there a reason to override the `attention_type`? As far as I understand, the difference between BigBird and other models is in the attention.

ctr-pmuruganTT: When I used the default attention type, I am getting an error as 
`ValueError: Sequence length must be multiple of block size, but sequence length is 8, while block size is 64.
`. I tried updating the blocksize and sequence length but faced some other issues.
Then I referred to this [blog](https://huggingface.co/blog/big-bird#using-bigbird-with-%F0%9F%A4%97transformers) , where they have used attention_type = "original_full". With this attention, model ran without any error. So I have used this here

ctr-pmuruganTT: In [HuggingFace blog](https://huggingface.co/docs/transformers/en/model_doc/big_bird#usage-tips) for BigBird, it is given as to use `original_full` when sequence length < 1024. When the input is too long, we can then use `block_sparse` attention

mrakitaTT: In the blog post they say:
> By setting attention_type to `original_full`, BigBird will be relying on the full attention of n^2 complexity. **This way BigBird is 99.9 % similar to BERT.**

So if we used `original_full` we wouldn't gain much with testing both this and BERT models. @ctr-pmuruganTT can you please try to use some long sequence (> 1024 words) and to use`block_sparse` attention? You can use some online word generator to generate sequence with exact number of words.

mrakitaTT:
Since the `google/bigbird-base-trivia-itc` is just RoBERTa base with a `BigBirdForQuestionAnsweringHead` on its top, I would suggest removing the `tests/jax/models/bigbird/roberta_base/` test and making these changes:
- Rename the `base_trivia_itc` dir to `roberta_base`, and also names of its tester and tests. In `MODEL_NAME` change the variant name to `roberta_base` and model task to `ModelTask.NLP_QA`
- Remove the `_get_model` and `_get_input_activations` implementations from the `tests/jax/models/bigbird/tester.py`, and override it in each variant tester separately (make a new class in each tester that extends the `BigBirdTester` class and overrides these two methods):
  - In tester for `roberta_large` you can copy the original implementations.
  - In tester for `roberta_base` use the `FlaxBigBirdForQuestionAnswering` in `_get_model` method instead of the `FlaxBigBirdForCausalLM` in order to utilize the `BigBirdForQuestionAnsweringHead`. You can see example [here](https://huggingface.co/docs/transformers/en/model_doc/big_bird#transformers.FlaxBigBirdForQuestionAnswering.__call__.example) how to prepare the input for that class in the `_get_input_activations` method.

If you have any doubts please let me know.

ctr-pmuruganTT:
Updates made after the comments :
1. Changed the current bigbird/base_trivia_itc test to test QA instead of CausalLM
2. Deleted the bigbird/roberta_base test
3. Renamed the bigbird/base_trivia_itc folder/test/classes to roberta_base
4. Added a statement with > 1024 words to test with `block_sparse` in `test_bigbird_roberta_large` file

tests/jax/single_chip/models/bigbird/tester.py
```diff
@@ -0,0 +1,31 @@
...
+from typing import Dict, Sequence
+
+from infra import ComparisonConfig, ModelTester, RunMode
+from jaxtyping import PyTree
+
+
+class BigBirdTester(ModelTester):
+    """Tester for BigBird Model variants with a language modeling head on top."""
+
+    def __init__(
```
mrakitaTT, line 14: You can add a `model_name` parameter here and then do `self._model_name = model_name`, instead of doing in all child testers above

tests/jax/single_chip/models/bigbird/tester.py
```diff
@@ -0,0 +1,31 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Dict, Sequence
+
+from infra import ComparisonConfig, ModelTester, RunMode
+from jaxtyping import PyTree
+
+
+class BigBirdTester(ModelTester):
+    """Tester for BigBird Model variants with a language modeling head on top."""
```
mrakitaTT, line 12: I would remove ` with a language modeling head on top` part now because one of the testers is testing QA and other LM

mrakitaTT (approved):

tests/jax/single_chip/models/bigbird/roberta_base/test_bigbird_roberta_base.py
```diff
@@ -0,0 +1,102 @@
...
+)
+
+
+class BigBirdBaseTester(BigBirdTester):
+    def __init__(
+        self,
+        comparison_config: ComparisonConfig = ComparisonConfig(),
+        run_mode: RunMode = RunMode.INFERENCE,
+    ) -> None:
+        super().__init__(comparison_config, run_mode)
```
mrakitaTT, line 44: You have to pass `model_name` here

mrakitaTT (approved):

Issue #374, opened 2025-03-19
Move `fillStrategyMapFromSharding` to `tt-mlir`


ajakovljevicTT:
Move the `fillStrategyMapFromSharding` function into `tt-mlir` as its logic and objects that it handles are limited to `tt-mlir`.

PR #401, opened 2025-03-24
Moved function `fillStrategyMapFromSharding` to tt-mlir

ajakovljevicTT:
Moved `fillStrategyMapFromSharding` from tt-xla to tt-mlir, also uplifted `tt-mlir` to the newest version.

Fixes https://github.com/tenstorrent/tt-xla/issues/374

### DIFF
--- a/src/common/pjrt_implementation/loaded_executable_instance.cc
+++ b/src/common/pjrt_implementation/loaded_executable_instance.cc
@@ -15,11 +15,10 @@
 #include <unordered_map>
 
 // tt-mlir includes
-#include "tt/runtime/workarounds.h"
-#include "ttmlir/Dialect/TT/IR/TTOpsTypes.h"
-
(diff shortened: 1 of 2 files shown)


src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -18,6 +18,9 @@
 #include "tt/runtime/workarounds.h"
 #include "ttmlir/Dialect/TT/IR/TTOpsTypes.h"
 
+#define TTMLIR_ENABLE_STABLEHLO 1
```
mrakitaTT, line 21: This should be safe to define at the beginning of the group of includes from tt-mlir, so you can also move the `ShardingUtils.h` include to the group above

mrakitaTT (approved):

PR #416, opened 2025-03-27
Fixing nightly runs

ajakovljevicTT:
There was an issue running nightly tests when there were no `nightly` or `model_test` marks, as pytest would mark these runs as failed. WIth this PR, now setting what runs will be triggered on which machines, and on which directories the pytest command will be ran, through arguments to the `build-and-test.yaml`


nsumrakTT (approved):

.github/workflows/on-nightly.yml
```diff
@@ -19,6 +24,10 @@ jobs:
     if: always()  # This ensures the job runs regardless of success or failure of `nightly_tests`
     with:
       test_mark: 'model_test'
+      machine_and_dir: |
+        [
+          {"runs-on": "n150", "name": "run", "dir": "./tests/jax/single_chip"},
```
kmitrovicTT, line 29: Machine, dir and something else in the middle. 
Whatever "run" is, since that's the only value for `"name"` we pass, do we need to pass it or can you just hardcode it in main yml?

.github/workflows/build-and-test.yml
```diff
@@ -26,6 +35,15 @@ on:
         required: false
         default: 'push'
         type: string
+      machine_and_dir:
+        description: 'Machine and directory to run tests on'
+        required: true
```
kmitrovicTT, line 40: Why is it required here and not above? 

Is `required: true` going to assert if no `machine_and_dir` is provided from particular job? If so, I'd remove default value.

kmitrovicTT (approved):
How did we manage to leave out these models that you moved to `single_chip`?

.github/workflows/on-nightly.yml
nsumrakTT: You can, it is not used anymore. It is a key in job name so you might consider using it with more sense which might be useful if you add more test runs as job name has to be unique.

mrakitaTT (approved):

.github/workflows/on-push.yml
```diff
@@ -22,3 +22,9 @@ jobs:
     needs: [pre-commit, spdx]
     uses: ./.github/workflows/build-and-test.yml
     secrets: inherit
+    with:
```
mrakitaTT, line 25: Hm, we don't set `push` mark here, seems like a bug

mrakitaTT: Oh it's the default mark, that's why this worked. I think it would be better to non-default it and make it required to set, and set it here

mrakitaTT: Actually it will cause problems with manual dispatch, nvm forget this

.github/workflows/on-nightly.yml
mrakitaTT: @ajakovljevicTT sorry I know you just made the change to hardcode the "run", but I think we should revert it and leave the `build.name` parameter. As @nsumrakTT said it might be useful if we for example want to run different dir on the same hw architecture, we would pass different build name.

PR #418, opened 2025-03-27
Changed the inputs to the build-and-test.yml 

ajakovljevicTT:
Changed the input structure of the `build-and-test.yml` to include name (currently only set to run), to differentiate between different builds during testing.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -15,8 +15,12 @@ on:
         options:
           - push
           - nightly
-      machine_and_dir:
-        description: 'Machine and directory to run tests on'
+      build_options:
+        description: >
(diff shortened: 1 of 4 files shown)


.github/workflows/build-and-test.yml
mrakitaTT, line 1: I would also rename the `machine_and_dir` input parameter to something like `build_options`, and say in description of that parameter something like `Build options for each run, like machine and directory to run tests on and the name of the run which can be used to differentiate unique runs` to explain how the name can be used.

mrakitaTT (approved):

mrakitaTT (approved):

kmitrovicTT (approved):

PR #423, opened 2025-03-28
Adding tests for multichip psum

ajakovljevicTT:
With the workaround fix in tt-mlir for incorrect `reduce_scatter` (https://github.com/tenstorrent/tt-mlir/issues/2626), we can now use psum in our jax models. This PR adds a test for that, only for fully_manual (using both device_put() and shard_map()), as we dont have compiler support for other types of sharding.


tests/infra/multichip_tester.py
```diff
@@ -115,10 +121,28 @@ def test_with_random_inputs(
     # ---------- Private methods ----------
 
     def _compile_for_cpu(
-        self, executable: Callable, static_argnames: Sequence[str] = None
+        self,
+        executable: Callable,
+        multichip_mode: ShardingMode,
```
mrakitaTT, line 126: `multichip_mode` should be renamed to `sharding_mode`, check in other places too

tests/jax/multi_chip/n300/ops/data_parallel/batch_sharded/test_psum.py
```diff
@@ -0,0 +1,74 @@
...
+        ShardingMode.INPUTS_AND_MODULE,
+        pytest.param(
+            ShardingMode.MODULE,
+            marks=pytest.mark.xfail(
+                reason=failed_fe_compilation(
+                    "Cannot get sharding information through the protobuf "
+                    "(https://github.com/tenstorrent/tt-xla/issues/277)"
+                )
+            ),
+        ),
```
mrakitaTT, line 47: What about `ShardingMode.INPUTS`?

mrakitaTT (approved):

tests/jax/multi_chip/n300/ops/data_parallel/batch_sharded/test_psum.py
ajakovljevicTT: `psum` cannot be used when there is no `shard_map` (which is the case for ShardingMode.INPUTS), since there is no axis naming in that case. The recommended path in that case is jax.sum(), and it expects the compiler to do the translation to the ccl op. However, we don't have support for that in tt-mlir, and the test would look completely different.

mrakitaTT: Oh I see. It could be used with pmap axis, but then we wouldn't shard the inputs so it would again be a different test.

PR #457, opened 2025-04-03
Adding an llmbox to the CI

ajakovljevicTT:
Adding an llmbox with some associated tests to our CI.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -21,6 +21,7 @@ on:
           - machine
           - directory to run tests on
           - the name of the run which can be used to differentiate unique runs
+          - container device options
         required: true
         type: string
   workflow_call:
@@ -181,7 +182,7 @@ jobs:
 
     container:
       image: ${{ needs.build-image.outputs.docker-image }}
-      options: --device /dev/tenstorrent/0
+      options: ${{ matrix.build.device-options }}
       volumes:
         - /dev/hugepages:/dev/hugepages
         - /dev/hugepages-1G:/dev/hugepages-1G
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -182,7 +182,7 @@ jobs:
 
     container:
       image: ${{ needs.build-image.outputs.docker-image }}
-      options: ${{ matrix.build.device-options }}
+      options: ${{ matrix.build.container-options }}
       volumes:
         - /dev/hugepages:/dev/hugepages
         - /dev/hugepages-1G:/dev/hugepages-1G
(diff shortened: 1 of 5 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -21,6 +21,7 @@ on:
           - machine
           - directory to run tests on
           - the name of the run which can be used to differentiate unique runs
+          - container device options
```
mrakitaTT, line 24: Should this also be added to `build_options` of `workflow_call` below?

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -120,12 +120,9 @@ LoadedExecutableInstance::Execute(PJRT_LoadedExecutable_Execute_Args *args) {
       args->argument_lists, addressable_devices_, args->num_args, num_devices);
 
   tt::runtime::MeshDeviceOptions options;
-  // Set the mesh offset to the minimum device id, so that we can run on the
-  // mesh starting with the minimum device id offset.
-  options.meshOffset = {0, static_cast<std::uint32_t>(*std::min_element(
-                               device_ids.begin(), device_ids.end()))};
```
mrakitaTT, line 126: What's the reason behind this deletion? Does this mean each client will now try to open all devices?

mrakitaTT (approved):

.github/workflows/on-nightly.yml
```diff
@@ -13,8 +13,9 @@ jobs:
       test_mark: 'nightly'
       build_options: |
         [
-          {"runs-on": "n150", "name": "run", "dir": "./tests/jax/single_chip"},
-          {"runs-on": "n300", "name": "run", "dir": "./tests/jax/multi_chip/n300"}
+          {"runs-on": "n150", "name": "run", "dir": "./tests/jax/single_chip", "container-options": "--device /dev/tenstorrent/0"},
+          {"runs-on": "n300", "name": "run", "dir": "./tests/jax/multi_chip/n300", "container-options": "--device /dev/tenstorrent/0"},
+          {"runs-on": "llmbox", "name": "run", "dir": "./tests/jax/multi_chip/llmbox", "container-options": "--device /dev/tenstorrent/0 --device /dev/tenstorrent/1 --device /dev/tenstorrent/2 --device /dev/tenstorrent/3"}
```
vmilosevic, line 18: Can we try using "--device /dev/tenstorrent" It should give us all cards that are available, it should work for both n150 n300 and llmbox. We need to test this, i'm not sure if it will work

src/common/pjrt_implementation/loaded_executable_instance.cc
ajakovljevicTT: The tests were failing on the llmbox because the ids of the chips were not ordered (due to the llmbox specific configuration), in debugging that, I found out that we can specifically set the ids of devices to be opened, so i substituted the code above (which was just a workaround) with this solution (the line below).

.github/workflows/on-nightly.yml
ajakovljevicTT: Noted, will try

Issue #434, opened 2025-03-31
Not passing in the attention mask breaks attention


sgligorijevicTT:
When the attention mask is not passed into transformers models, a code path with `jnp.ones_like` is hit. Lowering of `ones_like` involves an add operation with broadcasting, which hits a metal bug. As a result, all zeros mask is generated and attention layer returns wrong results. This ruins the accuracy of all transformer models.
A workaround is to pass in the attention mask.

Should *hopefully* be fixed by https://github.com/tenstorrent/tt-metal/pull/17160 . Update: the relevant PR was merged and then reverted, I am tracking what's happening in metal and will link the new PR as soon as it is opened.
Metal already has multiple issues opened regarding broken broadcasting in binary ops and is actively working on binary_ng.

PR #462, opened 2025-04-04
Refactor tests to pass everything from the tokenizer in LM tests

sgligorijevicTT:
### Ticket
N/A

### Problem description
There is a bug in metal which is triggered on the code path for generating the default attention mask for transformers. We want to work around it for now, to make debugging accuracy issues easier.

### What's changed
We no longer take just the `input_ids` from the tokenizer, but rather pass on everything to the model.
I also changed the `return_tensors` to `jax` inside the call to the tokenizer. It's cleaner to work with jax arrays from the get go and it will be more convenient if we ever want to send batched inputs to our tests.

### Checklist
- [ ] Update atol values in fail reasons - some values did improve while others didn't, I want to double check this before merging.

### DIFF
--- a/tests/jax/single_chip/models/albert_v2/base/test_albert_base.py
+++ b/tests/jax/single_chip/models/albert_v2/base/test_albert_base.py
@@ -53,7 +53,7 @@ def training_tester() -> AlbertV2Tester:
 )
 @pytest.mark.xfail(
     reason=incorrect_result(
-        "Atol comparison failed. Calculated: atol=131036.078125. Required: atol=0.16 "
+        "Atol comparison failed. Calculated: atol=34.321659088134766. Required: atol=0.16 "
         "https://github.com/tenstorrent/tt-xla/issues/379"
     )
 )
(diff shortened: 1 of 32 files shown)


kmitrovicTT (approved):
> We want to work around it for now

Does that mean this is something that should be reverted once metal bug is resolved? Or should we open issue which tracks that metal bug?

sgligorijevicTT:
> > We want to work around it for now
> > Does that mean this is something that should be reverted once metal bug is resolved? Or should we open issue which tracks that metal bug?

Regarding tracking: on our side the tracking issue is #434 and I linked a relevant metal PR there. Although looking back, it is not clear that this PR and that issue are related. I will update the issue description.
Regarding reverting: both ways of using the model are equally valid. For training you always want to pass in the mask, for inference it doesn't matter and there is code in the wild that doesn't do that. I think it's fine to leave it like this, as long as we have at least one other test that doesn't pass in the mask, and thorough tests for broadcasting.

mrakitaTT:
Nice! I like this because this is what examples in huggingface library are doing, they pass the whole inputs pytree. Thanks for making these changes. Do you plan to update the xfail messages in this PR with updated accuracy numbers?

mrakitaTT (approved):

sgligorijevicTT:
> Nice! I like this because this is what examples in huggingface library are doing, they pass the whole inputs pytree. Thanks for making these changes. Do you plan to update the xfail messages in this PR with updated accuracy numbers?

Yes, that is the plan. I got side tracked by investigating why some accuracy numbers didn't budge. 
Commit with fresh atol numbers is coming up shortly.

sgligorijevicTT:
😓 Much fewer models than I have expected have shown improvements. And models where I increased the sequence length above 1 show much worse accuracy(although in a way this is a good thing, we want to know if something is wrong).

Issue #380, opened 2025-03-20
Add tests for RoFormer model


ctr-pmuruganTT:
There is an existing Flax implementation `FlaxRoFormerForMaskedLM ` for RoFormer model in huggingface transformers library, [here](https://huggingface.co/docs/transformers/en/model_doc/roformer#transformers.FlaxRoFormerForMaskedLM) is the documentation with the example code.

There are 4 variants with official model parameters published on huggingface:
-  roformer_chinese_base  : https://huggingface.co/junnyu/roformer_chinese_base
-  roformer_chinese_small : https://huggingface.co/junnyu/roformer_chinese_small
-  roformer_chinese_char_base : https://huggingface.co/junnyu/roformer_chinese_char_base
-  roformer_chinese_char_small : https://huggingface.co/junnyu/roformer_chinese_char_small

Add tests for this model

ctr-pmuruganTT:
Roformer model is unblocked now as https://github.com/huggingface/transformers/issues/36854 is resolved.

PR #474, opened 2025-04-08
Add tests for RoFormer model

ctr-pmuruganTT:
### Ticket
Closes https://github.com/tenstorrent/tt-xla/issues/380

### Problem description
Add tests for Roformer model

### What's changed
A tester class is added to test 3 variants of Roformer models

### Checklist

Logs are attached : 

[chinese_base.log](https://github.com/user-attachments/files/19644259/chinese_base.log)
[chinese_char_small.log](https://github.com/user-attachments/files/19644262/chinese_char_small.log)
[chinese_small.log](https://github.com/user-attachments/files/19644263/chinese_small.log)

### DIFF
--- a/tests/jax/single_chip/models/roformer/chinese_base/test_roformer_chinese_base.py
+++ b/tests/jax/single_chip/models/roformer/chinese_base/test_roformer_chinese_base.py
@@ -1,0 +1,72 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+from infra import Framework, RunMode
+
+from tests.utils import (
+    BringupStatus,
+    Category,
+    ModelGroup,
+    ModelSource,
+    ModelTask,
+    build_model_name,
+    failed_ttmlir_compilation,
+)
+
+from ..tester import RoFormerTester
+
+MODEL_PATH = "junnyu/roformer_chinese_base"
+MODEL_NAME = build_model_name(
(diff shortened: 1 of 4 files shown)


tests/jax/single_chip/models/roformer/tester.py
```diff
@@ -0,0 +1,49 @@
...
+from transformers import (
+    RoFormerTokenizer,
+    FlaxRoFormerForMaskedLM,
+    FlaxPreTrainedModel,
+)
+from jaxtyping import PyTree
+
+
+class RoformerTester(ModelTester):
+    """Tester for Roformer Models for language Modeling task."""
```
mrakitaTT, line 18: Nit: Since they used name "RoFormer" in the paper, let's call the tester `RoFormerTester` and make update in description from `Roformer` to `RoFormer`

tests/jax/single_chip/models/roformer/chinese_base/test_roformer_chinese_base.py
mrakitaTT, line 1: Seems like the only difference in configs between `chinese_base` and `chinese_char_base` is in `max_position_embeddings` and `vocab_size`, due to the fact that `chinese_char_base` works on character tokenization level so it has bigger vocabulary. `embedding_size` defaults to `hidden_size` when omitted, so it is the same between both. I think this is not big enough difference to have both variants, so I would leave only `chinese_base`.

`chinese_small` and `chinese_char_small` have more substantial differences, so I am okay with leaving both.

mrakitaTT (approved):

PR #475, opened 2025-04-08
Added data-parallel mnist training script that uses manual sharding.

vladimirjovanovicTT:
### Problem description
Added data-parallel mnist training script that uses manual sharding.

### What's changed
It is set up to use MSE and argmax on host, similar to what is currently being done on tt device, so that it can be easily moved to tt. It generates a StableHLO that tt-mlir should be able to ingest. Training works in a multi-gpu environment.

### DIFF
--- a/examples/mnist_train_manual_shard.py
+++ b/examples/mnist_train_manual_shard.py
@@ -1,0 +1,315 @@
+import jax
+import jax.numpy as jnp
+from jax import random
+from jax.sharding import Mesh, PartitionSpec, NamedSharding
+from jax.experimental import mesh_utils
+from jax.experimental import shard_map
+import jax.lax as lax
+import torchvision
+import numpy as np
+
+class NetConfig:
+    def __init__(self, input_size=784, hidden_size=128, output_size=10):
+        self.input_size = input_size
+        self.hidden_size = hidden_size
+        self.output_size = output_size
+
+class TrainingConfig:
+    def __init__(self, num_epochs=10, batch_size=128, learning_rate=0.001):
+        self.num_epochs = num_epochs
+        self.batch_size = batch_size
+        self.learning_rate = learning_rate
+
+class LoggerConfig:
+    def __init__(self, log_every_n_steps=10):
+        self.log_every_n_steps = log_every_n_steps
+
+class ExperimentConfig:
+    def __init__(self, net_config=None, training_config=None, logger_config=None):
+        self.net_config = net_config or NetConfig()
+        self.training_config = training_config or TrainingConfig()
+        self.logger_config = logger_config or LoggerConfig()
+
+def init_device():
+    #placeholder for TT device
+    pass
+
+def export_to_stablehlo(forward_fn, params, *args):
+    fwd_jit = jax.jit(forward_fn)
+    fwd_lowered = fwd_jit.lower(params, *args)
+
+    fwd_stablehlo = fwd_lowered.compile()
+    print(fwd_lowered.as_text())
+
+def load_mnist():
+    mnist = {
+        "train": torchvision.datasets.MNIST("./data", train=True, download=True),
+        "test": torchvision.datasets.MNIST("./data", train=False, download=True),
+    }
+
+    ds = {}
+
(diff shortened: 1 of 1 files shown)


examples/mnist_train_manual_shard.py
```diff
@@ -0,0 +1,375 @@
...
+        @jax.jit
+        def training_step(params, x_batch, y_batch, lr):
+            return shard_map.shard_map(
+                lambda p, x, y, lr: update(p, x, y, lr),
+                mesh=mesh,
+                in_specs=(
+                    PartitionSpec(None),
+                    PartitionSpec("dp"),
+                    PartitionSpec("dp"),
+                    PartitionSpec(),
```
umalesTT, line 366: nit/question: Why are we using both ```PartitionSpec(None)``` and ```PartitionSpec()```? Maybe it is better to use only one for clarity purpose, as they both mean the same - replicate input across all devices?

examples/mnist_train_manual_shard.py
```diff
@@ -0,0 +1,375 @@
...
+class LoggerConfig:
+    def __init__(self, log_every_n_steps=10):
+        self.log_every_n_steps = log_every_n_steps
+
+
+class ExperimentConfig:
+    def __init__(self, net_config=None, training_config=None, logger_config=None):
+        self.net_config = net_config or NetConfig()
+        self.training_config = training_config or TrainingConfig()
+        self.logger_config = logger_config or LoggerConfig()
```
kmitrovicTT, line 38: Dataclasses would be convenient to use here. Check out `tests/infra/comparison.py`.

examples/mnist_train_manual_shard.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+def init_device():
+    # placeholder for TT device
+    pass
```
kmitrovicTT, line 43: What's gonna happen to this?

examples/mnist_train_manual_shard.py
```diff
@@ -0,0 +1,375 @@
...
+                    PartitionSpec(),
+                ),
+                out_specs=(PartitionSpec(None), PartitionSpec()),
+            )(params, x_batch, y_batch, lr)
+
+        print("StableHLO representation of training_step function:")
+        lowered = training_step.lower(params, x_batch, y_batch, learning_rate)
+        print(lowered.as_text())
+    else:
+        train_mnist()
```
kmitrovicTT, line 375: Why is `else` branch one call and `if` contains a bunch of logic with nested function definitions? Can we extract `if` block into a function?

examples/mnist_train_manual_shard.py
kmitrovicTT: @ajakovljevicTT  maybe you have some suggestions on how to use `PartitionSpec` here

examples/mnist_train_manual_shard.py
```diff
@@ -0,0 +1,375 @@
...
+    x_train, y_train, x_val, y_val, x_test, y_test = load_mnist()
+    key = random.PRNGKey(0)
+
+    train_mlp(
+        x_train, y_train, x_val, y_val, x_test, y_test, key, config, param_sharding
+    )
+
+
+if __name__ == "__main__":
+    import sys
```
kmitrovicTT, line 334: Import on top with all other imports

examples/mnist_train_manual_shard.py
```diff
@@ -0,0 +1,375 @@
...
+
+            if (i + 1) % logger_config.log_every_n_steps == 0:
+                avg_loss = batch_loss_accum / logger_config.log_every_n_steps
+                print(f"Epoch {epoch}, Batch {i +1}, Loss: {avg_loss}")
+                batch_loss_accum = 0.0
+
+        val_loss = evaluate(params, x_val_host, y_val_host, mesh, data_sharding)
+        print(f"Epoch {epoch}, Validation Loss: {val_loss}")
+
+    test_loss, test_accuracy = evaluate(
```
umalesTT, line 255: ``` evaluate() ``` returns only ```avg_loss``` so this will fail.

PR #481, opened 2025-04-09
Fix community contributed builds

nsumrakTT:
### Ticket
slack req

### Problem description
Community contribution build fails

### What's changed
Add continue on error for showing test report as comment so community contributed builds with lower privileges can succeed

### Checklist
Will test it after it is merged

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -46,84 +46,9 @@ on:
 permissions:
   packages: write
   checks: write
-  pull-requests: write
+  # pull-requests: write
 
 jobs:
-  # build-ttxla:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -46,84 +46,9 @@ on:
 permissions:
   packages: write
   checks: write
-  pull-requests: write
+  # pull-requests: write
```
mrakitaTT, line 49: Please don't forget to revert the commented-out code before merge :)

nsumrakTT: You're right, I should delete this.

PR #492, opened 2025-04-11
Replace model_name to model_path in tester files

ctr-pmuruganTT:
### Problem description

Update the constructor argument name from `model_name` to `model_path`  in tester files of the models, as it is actually denoting the path of the model

### What's changed

`model_name` is replaced with `model_path` in tester files of models

### DIFF
--- a/tests/jax/single_chip/models/albert_v2/tester.py
+++ b/tests/jax/single_chip/models/albert_v2/tester.py
@@ -15,20 +15,20 @@ class AlbertV2Tester(ModelTester):
 
     def __init__(
         self,
-        model_name: str,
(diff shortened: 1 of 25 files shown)


mrakitaTT (approved):
Please just try running all these models locally to make sure we don't get surprised in next nightly :) for example if some model had some hidden dependency on `self._model_name`

ctr-pmuruganTT:
I have tested all the tests in model directory and checked that there are no failures with the changes made
`===== 11 passed, 108 skipped, 42 xfailed, 2 warnings in 1300.16s (0:21:40) =====`

Issue #339, opened 2025-03-11
Change logical and, or, not and xor tests to use bool instead of int32


kmitrovicTT:

kmitrovicTT:
See https://github.com/tenstorrent/tt-xla/pull/499/files#r2044214196

PR #499, opened 2025-04-14
Added missing tests for atan2, reverse, bitwise and/or/not/xor ops

kmitrovicTT:
All of them were successfully added some time ago in tt-mlir, some were blocked by issues, now that they all work I am adding torch tests.


sgligorijevicTT (approved):

tests/jax/single_chip/ops/test_and.py
```diff
@@ -33,3 +33,27 @@ def logical_and(a: jax.Array, b: jax.Array) -> jax.Array:
     lhs = random_tensor(shape, jnp.int32, minval=0, maxval=2, random_seed=3)
     rhs = random_tensor(shape, jnp.int32, minval=0, maxval=2, random_seed=6)
```
kmitrovicTT, line 34: I wanted to fix  https://github.com/tenstorrent/tt-xla/issues/339 as well, but using `jnp.bool` here leads to a segfault.

However, using `int32` like this and restricting values to `[0, 1]` will result in envoking `stablehlo.and` function which works with `i1` tensors, so it basically does manifest its boolean/logical version.

PR #519, opened 2025-04-17
Small changes to PJRT in order to enable SPMD torch xla execution.

ajakovljevicTT:
Torch xla spmd execution produces slightly different stablehlo code than jax, which means some of the functions for inferring the mesh and the number of devices have to be changed. These are smaller changes which will not change the behaviour of PJRT for jax, but will unblock torch xla spmd execution, along with some necessary tt-mlir changes, which are in development by @wooseokTT

### DIFF
--- a/src/common/module_builder.cc
+++ b/src/common/module_builder.cc
@@ -7,6 +7,7 @@
 
 // c++ standard library includes
 #include <cstdlib>
+#include <numeric>
 #include <optional>
 
 // loguru includes
@@ -97,8 +98,6 @@ ModuleBuilder::buildModule(const std::string_view &code,
     return m_status;
   }
 
-  collectNumDevicesToUtilize(mlir_module);
-
   convertFromVHLOToSHLO(mlir_module);
   if (!tt_pjrt_status_is_ok(m_status)) {
     return m_status;
@@ -114,6 +113,7 @@ ModuleBuilder::buildModule(const std::string_view &code,
   }
 
   collectMeshShape(mlir_module);
+  collectNumDevicesToUtilize(mlir_module);
 
   convertFromTTIRToTTNN(system_descriptor_path, mlir_module);
   if (!tt_pjrt_status_is_ok(m_status)) {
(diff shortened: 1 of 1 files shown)


src/common/module_builder.cc
```diff
@@ -114,6 +113,7 @@ ModuleBuilder::buildModule(const std::string_view &code,
   }
 
   collectMeshShape(mlir_module);
+  collectNumDevicesToUtilize(mlir_module);
```
mrakitaTT, line 116: Since you've moved the place where this function is called, could you please move its declaration in the .h file and definition below to be after the `collectMeshShape` function (actually after `estimateMeshShape` after you rebase my changes)

src/common/module_builder.cc
```diff
@@ -156,9 +156,12 @@ void ModuleBuilder::collectNumDevicesToUtilize(
     m_num_devices_to_utilize =
         num_partitions_attr.getInt() * num_replicas_attr.getInt();
   } else {
-    m_num_devices_to_utilize = 1;
-    DLOG_F(WARNING, "mhlo.num_partitions, mhlo.num_replicas not found, using "
-                    "default number of devices: 1");
+    // mhlo.num_partitions and mhlo.num_replicas are not populated by torch_xla,
+    // so we infer the number of devices from mesh shape.
```
mrakitaTT, line 160: Nit: "from **the** mesh shape"

src/common/module_builder.cc
```diff
@@ -156,9 +156,12 @@ void ModuleBuilder::collectNumDevicesToUtilize(
...
+    DLOG_F(WARNING, "Num replicas and num partitions are not set, inferring "
+                    "the number of devices from mesh shape");
+    m_num_devices_to_utilize = std::accumulate(
+        m_mesh_shape.begin(), m_mesh_shape.end(), 1, std::multiplies<>());
```
mrakitaTT, line 164: Since I've changed this function a lot, I would recommend that you make these changes to adapt:
```
if (!num_partitions_attr && !num_replicas_attr) {
    // When both mhlo.num_partitions and mhlo.num_replicas are not populated
    // (torch_xla doesn't populate them), we estimate the number of devices from
    // the mesh shape.
    DLOG_F(WARNING, "Num replicas and num partitions are not set, inferring "
                    "the number of devices from mesh shape");
    m_num_devices_to_utilize = std::accumulate(
        m_mesh_shape.begin(), m_mesh_shape.end(), 1, std::multiplies<>());
  } else {
    // If at least one mhlo parameter is populated we assume the default value
    // of the other one.
    m_num_devices_to_utilize = m_num_partitions * m_num_replicas;
  }
```

kmitrovicTT (approved):

mrakitaTT (approved):

PR #527, opened 2025-04-22
Store binary as shared_ptr, add mutable ref accessor

vwellsTT:
### Problem description
This tt-mlir PR: https://github.com/tenstorrent/tt-mlir/pull/2756  makes it so `submit()` takes a reference to a binary instead of a value.  We need a new accessor s.t. we can pass a mutable ref into submit, and we want to share the same binary object throughout code, because `submit()` can modify the object's cache now.

### What's changed
non-const ref accessor for binary so that we can use it with submit(), changed binary to be stored + passed as `std::shared_ptr` to ensure the same underlying object is used everywhere

### Checklist

### DIFF
--- a/inc/common/pjrt_implementation/executable_image.h
+++ b/inc/common/pjrt_implementation/executable_image.h
@@ -34,7 +34,7 @@ public:
   // Creates new executable image instance from the information given by the
   // compiler.
   static std::shared_ptr<ExecutableImage> createInstance(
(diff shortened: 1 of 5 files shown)


mrakitaTT:
Vincent handled this in tt-mlir https://github.com/tenstorrent/tt-mlir/pull/2756 so closing this now

PR #531, opened 2025-04-23
Skip blenderbot tests

sgligorijevicTT:
It was segfaulting in CI


mrakitaTT (changes requested):
I've investigated what happened, let's wait with this skip, @ctr-pmuruganTT will try to convert weights to bf16

PR #534, opened 2025-04-23
Test for SBERT

mmaksimovicTT:
Added a test for the SBERT model.

### DIFF
--- a/tests/jax/single_chip/models/sbert/model_implementation.py
+++ b/tests/jax/single_chip/models/sbert/model_implementation.py
@@ -1,0 +1,48 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+from flax import linen as nn
+import jax.numpy as jnp
+from transformers import FlaxBertModel, BertConfig
+
+
+class FlaxSentenceTransformerBERT(nn.Module):
+    config: BertConfig
+
+    @nn.compact
+    def __call__(
+        self,
+        input_ids: jnp.ndarray,
+        attention_mask: jnp.ndarray,
+        token_type_ids: jnp.ndarray = None,
+        normalize: bool = True,
+        include_prompt: bool = True,
+        prompt_length: int = 0,
+    ) -> jnp.ndarray:
+        if not include_prompt and prompt_length > 0:
+            attention_mask = attention_mask.at[:, :prompt_length].set(0)
+
+        model = FlaxBertModel(self.config, dtype=jnp.float32)
+        output = model(
+            input_ids=input_ids,
+            attention_mask=attention_mask,
+            token_type_ids=token_type_ids,
+        )
+        token_embeddings = output.last_hidden_state
+
+        # Mean pooling
+        input_mask_expanded = jnp.expand_dims(attention_mask, axis=-1)
+        sum_embeddings = jnp.sum(token_embeddings * input_mask_expanded, axis=1)
+        sum_mask = jnp.clip(jnp.sum(input_mask_expanded, axis=1), a_min=1e-9)
+        embeddings = sum_embeddings / sum_mask
+
+        if normalize:
+            norm = jnp.linalg.norm(embeddings, axis=1, keepdims=True)
+            embeddings = embeddings / norm
+
+        return self.cosine_similarity(embeddings[0], embeddings[1])
+
+    def cosine_similarity(self, a: jnp.ndarray, b: jnp.ndarray) -> jnp.ndarray:
+        return jnp.dot(a, b) / jnp.maximum(
+            jnp.linalg.norm(a) * jnp.linalg.norm(b), 1e-9
+        )
--- a/tests/jax/single_chip/models/sbert/model_implementation.py
+++ b/tests/jax/single_chip/models/sbert/model_implementation.py
@@ -40,9 +40,4 @@ class FlaxSentenceTransformerBERT(nn.Module):
             norm = jnp.linalg.norm(embeddings, axis=1, keepdims=True)
             embeddings = embeddings / norm
 
-        return self.cosine_similarity(embeddings[0], embeddings[1])
-
-    def cosine_similarity(self, a: jnp.ndarray, b: jnp.ndarray) -> jnp.ndarray:
-        return jnp.dot(a, b) / jnp.maximum(
-            jnp.linalg.norm(a) * jnp.linalg.norm(b), 1e-9
-        )
+        return embeddings
--- a/tests/jax/single_chip/models/sbert/test_sbert.py
+++ b/tests/jax/single_chip/models/sbert/test_sbert.py
@@ -1,11 +1,10 @@
-# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
 #
 # SPDX-License-Identifier: Apache-2.0
 
 from typing import Dict
 
 import jax
-import jax.numpy as jnp
 import pytest
 from flax import linen as nn
 from transformers import AutoTokenizer, FlaxBertModel, BertConfig
@@ -36,7 +35,7 @@ MODEL_NAME = build_model_name(
 
 
 class FlaxSBERTTester(ModelTester):
-    """Tester for a SentenceTransformer BERT model (Flax version)"""
+    """Tester for a SentenceTransformer BERT model"""
 
     def __init__(
         self,
@@ -59,13 +58,13 @@ class FlaxSBERTTester(ModelTester):
     # @override
     def _get_input_activations(self) -> Dict[str, jax.Array]:
         tokenizer = AutoTokenizer.from_pretrained(MODEL_PATH)
-        encoded = tokenizer(
+        inputs = tokenizer(
             ["Bu örnek bir cümle", "Her cümle vektöre çevriliyor"],
             padding=True,
             truncation=True,
             return_tensors="jax",
         )
-        return encoded
+        return inputs
 
     # @override
     def _get_forward_method_kwargs(self) -> Dict[str, jax.Array]:
@@ -85,7 +84,7 @@ class FlaxSBERTTester(ModelTester):
 
     # @override
     def _get_static_argnames(self):
-        return ["normalize", "include_prompt", "prompt_length"]
+        return ["normalize", "include_prompt"]
 
 
 # ----- Fixtures -----
--- a/tests/jax/single_chip/models/sbert/test_sbert.py
+++ b/tests/jax/single_chip/models/sbert/test_sbert.py
@@ -1,0 +1,135 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Dict
+
+import jax
+import jax.numpy as jnp
+import pytest
+from flax import linen as nn
+from transformers import AutoTokenizer, FlaxBertModel, BertConfig
+
+from infra import ComparisonConfig, Framework, ModelTester, RunMode
+from tests.utils import (
+    BringupStatus,
+    Category,
+    ModelGroup,
+    ModelSource,
+    ModelTask,
+    build_model_name,
+    failed_fe_compilation,
+)
+
+from .model_implementation import FlaxSentenceTransformerBERT
+
+MODEL_PATH = "emrecan/bert-base-turkish-cased-mean-nli-stsb-tr"
+MODEL_NAME = build_model_name(
+    Framework.JAX,
+    "sentence_transformer",
+    None,
+    ModelTask.NLP_TEXT_CLS,
+    ModelSource.HUGGING_FACE,
+)
+
+# ----- Tester -----
+
+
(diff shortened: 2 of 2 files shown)


tests/jax/single_chip/models/sbert/test_sbert.py
```diff
@@ -0,0 +1,134 @@
...
+
+    # @override
+    def _get_forward_method_name(self) -> str:
+        return "apply"
+
+    # @override
+    def _get_input_activations(self) -> Dict[str, jax.Array]:
+        tokenizer = AutoTokenizer.from_pretrained(MODEL_PATH)
+        inputs = tokenizer(
+            ["Bu örnek bir cümle", "Her cümle vektöre çevriliyor"],
```
kmitrovicTT, line 62: Leave a comment that this model was trained on Turkish sentences, just so it's clear.

kmitrovicTT (approved):

tests/jax/single_chip/models/sbert/model_implementation.py
```diff
@@ -0,0 +1,43 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+from flax import linen as nn
+import jax.numpy as jnp
+from transformers import FlaxBertModel, BertConfig
+
+
+class FlaxSentenceTransformerBERT(nn.Module):
```
mrakitaTT, line 9: As documented [here](https://github.com/UKPLab/sentence-transformers/blob/master/docs/sentence_transformer/usage/custom_models.rst#structure-of-sentence-transformer-models), Sentence Transformer model is formed as a combination of these components:
- Pretrained Transformer model
- Pooling layer
- Dense layer (optional)
- Normalize layer (optional)

You can find which components model uses by checking the `modules.json` file in the model HF package. For example in SBERT variant Dense layer and normalization is not done, can be seen [here](https://huggingface.co/emrecan/bert-base-turkish-cased-mean-nli-stsb-tr/blob/main/modules.json) (there is no dense or normalization modules listed).

Since you've already added implementation for some of the things that are not needed for the SBERT variant (like prompt masking and normalization), I would keep them and rename this class to `FlaxSentenceTransformer` class, and then create a separate `FlaxSentenceTransformerBERT` class which extends it and initializes things according to SBERT specification.

Here is how that could look like:
```
class PoolingMode(Enum):
    # Use the first token (CLS token) as text representations.
    CLS_TOKEN = "cls_token"

    # Use max in each dimension over all tokens.
    MAX_TOKENS = "max_tokens"

    # Perform mean-pooling.
    MEAN_TOKENS = "mean_tokens"

    # Perform mean-pooling, but divide by sqrt(input_length).
    MEAN_SQRT_LEN_TOKENS = "mean_sqrt_len_tokens"

    # Perform (position) weighted mean pooling (https://arxiv.org/abs/2202.08904).
    WEIGHTEDMEAN_TOKENS = "weightedmean_tokens"

    # Perform last token pooling (https://arxiv.org/abs/2201.10005).
    POOLING_MODE_LASTTOKEN = "pooling_mode_lasttoken"


class FlaxSentenceTransformer(nn.Module):
    """
    Sentence Transformer model is formed as a combination of these components:
        - Pretrained Transformer model
        - Pooling layer
        - Dense layer (optional)
        - Normalize layer (optional)
    """

    transformer: nn.Module
    pooling_mode: PoolingMode
    dense_layer: nn.Module = None
    normalize: bool = False

    # Some models (INSTRUCTOR, NV-Embed) exclude the prompt from the mean pooling step,
    # such that it's only used in the Transformer blocks.
    pooling_include_prompt: bool = True
    prompt_length: int = 0

    def get_params(self) -> PyTree:
        return self.transformer.params

    @nn.compact
    def __call__(
        self,
        input_ids: jnp.ndarray,
        params: dict,
        attention_mask: jnp.ndarray,
        token_type_ids: jnp.ndarray = None,
    ) -> jnp.ndarray:
        transformer_outputs = self.transformer(
            input_ids=input_ids,
            params=params,
            attention_mask=attention_mask,
            token_type_ids=token_type_ids,
        )
        token_embeddings = transformer_outputs.last_hidden_state

        # Mean pooling
        if not self.pooling_include_prompt and self.prompt_length > 0:
            attention_mask = attention_mask.at[:, :self.prompt_length].set(0)
        assert self.pooling_mode == PoolingMode.MEAN_TOKENS, "Only mean pooling supported for now"
        if self.pooling_mode == PoolingMode.MEAN_TOKENS:
            input_mask_expanded = jnp.expand_dims(attention_mask, axis=-1)
            sum_embeddings = jnp.sum(token_embeddings * input_mask_expanded, axis=1)
            sum_mask = jnp.clip(jnp.sum(input_mask_expanded, axis=1), a_min=1e-9)
            embeddings = sum_embeddings / sum_mask

        # TODO Add support for dense layer.
        assert self.dense_layer == None, "Dense layer is not supported yet"

        if self.normalize:
            norm = jnp.linalg.norm(embeddings, axis=1, keepdims=True)
            embeddings = embeddings / norm

        return embeddings


class FlaxSentenceTransformerBERT(nn.Module):
    def __init__(
        self,
        pretrained_model: nn.Module,
    ) -> None:
        super().__init__(
            transformer = pretrained_model,
            pooling_mode = PoolingMode.MEAN_TOKENS,
            )
```

tests/jax/single_chip/models/sbert/model_implementation.py
```diff
@@ -0,0 +1,43 @@
...
+        self,
+        input_ids: jnp.ndarray,
+        attention_mask: jnp.ndarray,
+        token_type_ids: jnp.ndarray = None,
+        normalize: bool = True,
+        include_prompt: bool = True,
+        prompt_length: int = 0,
+    ) -> jnp.ndarray:
+        if not include_prompt and prompt_length > 0:
+            attention_mask = attention_mask.at[:, :prompt_length].set(0)
```
mrakitaTT, line 23: As mentioned [here](https://github.com/UKPLab/sentence-transformers/blob/dd8a8d7d4d5660e7b2b69e8a1ece8202417c0069/examples/sentence_transformer/training/prompts/README.md?plain=1#L91) and [here](https://github.com/UKPLab/sentence-transformers/blob/dd8a8d7d4d5660e7b2b69e8a1ece8202417c0069/sentence_transformers/models/Pooling.py#L37), this is done only for the pooling component. In the suggested code I've provided in the comment above I've moved it to happen right before the pooling.

tests/jax/single_chip/models/sbert/model_implementation.py
```diff
@@ -0,0 +1,43 @@
...
+
+        model = FlaxBertModel(self.config, dtype=jnp.float32)
```
mrakitaTT, line 25: This initializes model with the random parameters, which is probably why you are getting the `Failed to legalize operation 'stablehlo.shift_right_logical'` error, because the `__call__` is executed on device and parameters initialization uses RNG, and we do not support RNG operations on device yet.

You should use the loaded pretrained BERT model instead. In the suggested code I've provided in the comment above I've added `pretrained_model` parameter to the `FlaxSentenceTransformerBERT ` class, you should set that parameter from the tester in `_get_model` function, I'll leave the comment there too.

tests/jax/single_chip/models/sbert/model_implementation.py
```diff
@@ -0,0 +1,43 @@
...
+        output = model(
```
mrakitaTT, line 26: You need to pass the loaded model params here, otherwise it will use `self.params` which is set only for the randomly initialized params like when model is created from the config, like you did above.

tests/jax/single_chip/models/sbert/test_sbert.py
```diff
@@ -0,0 +1,137 @@
...
+        super().__init__(comparison_config, run_mode)
+
+    # @override
+    def _get_model(self) -> nn.Module:
+        config = BertConfig.from_pretrained(MODEL_PATH)
+        return FlaxSentenceTransformerBERT(config=config)
+
+    # @override
+    def _get_forward_method_name(self) -> str:
+        return "apply"
```
mrakitaTT, line 56: Can be removed, our `FlaxSentenceTransformer` class is using `__call__` method which is the default name on model tester.

tests/jax/single_chip/models/sbert/test_sbert.py
```diff
@@ -0,0 +1,137 @@
...
+        comparison_config: ComparisonConfig = ComparisonConfig(),
+        run_mode: RunMode = RunMode.INFERENCE,
+    ) -> None:
+        self._model_path = model_path
+        super().__init__(comparison_config, run_mode)
+
+    # @override
+    def _get_model(self) -> nn.Module:
+        config = BertConfig.from_pretrained(MODEL_PATH)
+        return FlaxSentenceTransformerBERT(config=config)
```
mrakitaTT, line 52: This now has to be changed to load the pretrained BERT model and pass it to the `FlaxSentenceTransformerBERT` constructor. Also should use `self._model_path` instead of `MODEL_PATH`.

Like this:
```
pretrained_model = FlaxBertModel.from_pretrained(self._model_path, from_pt=True)
return FlaxSentenceTransformerBERT(pretrained_model)
```

tests/jax/single_chip/models/sbert/test_sbert.py
```diff
@@ -0,0 +1,137 @@
...
+            return_tensors="jax",
+        )
+        return inputs
+
+    # @override
+    def _get_forward_method_kwargs(self) -> Dict[str, jax.Array]:
+        pretrained_model = FlaxBertModel.from_pretrained(MODEL_PATH, from_pt=True)
+        params = pretrained_model.params
+
+        inputs = self._get_input_activations()
```
mrakitaTT, line 77: These lines can now be removed, and below you will use:
```
return {
    "params": self._input_parameters,
    "input_ids": self._input_activations["input_ids"],
    "attention_mask": self._input_activations["attention_mask"],
    "token_type_ids": self._input_activations.get("token_type_ids"),
}
```

tests/jax/single_chip/models/sbert/test_sbert.py
```diff
@@ -0,0 +1,137 @@
...
+            "attention_mask": inputs["attention_mask"],
+            "token_type_ids": inputs.get("token_type_ids"),
+            "normalize": True,
+            "include_prompt": True,
+            "prompt_length": 0,
+        }
+
+    # @override
+    def _get_static_argnames(self):
+        return ["normalize", "include_prompt"]
```
mrakitaTT, line 90: Not needed anymore

tests/jax/single_chip/models/sbert/test_sbert.py
```diff
@@ -0,0 +1,137 @@
...
+    model_name=MODEL_NAME,
+    model_group=ModelGroup.RED,
+    run_mode=RunMode.INFERENCE,
+    bringup_status=BringupStatus.FAILED_FE_COMPILATION,
+)
+@pytest.mark.xfail(
+    reason=failed_fe_compilation(
+        "Failed to legalize operation 'stablehlo.shift_right_logical'. "
+        "https://github.com/tenstorrent/tt-xla/issues/417"
+    )
```
mrakitaTT, line 122: Should be updated now

tests/jax/single_chip/models/sbert/test_sbert.py
```diff
@@ -0,0 +1,137 @@
...
+        run_mode: RunMode = RunMode.INFERENCE,
+    ) -> None:
+        self._model_path = model_path
+        super().__init__(comparison_config, run_mode)
+
+    # @override
+    def _get_model(self) -> nn.Module:
+        config = BertConfig.from_pretrained(MODEL_PATH)
+        return FlaxSentenceTransformerBERT(config=config)
+
```
mrakitaTT, line 53: You now have to override the `def _get_input_parameters(self) -> PyTree:` method, and add this implementation:
```
# @override
def _get_input_parameters(self) -> PyTree:
    return self._model.get_params()
```

tests/jax/single_chip/models/sbert/test_sbert.py
```diff
@@ -0,0 +1,137 @@
...
+        tokenizer = AutoTokenizer.from_pretrained(MODEL_PATH)
+
+        # The model weights used in this test were trained on Turkish sentence pairs.
+        # These sentences below match the training data distribution.
+        inputs = tokenizer(
+            ["Bu örnek bir cümle", "Her cümle vektöre çevriliyor"],
+            padding=True,
+            truncation=True,
+            return_tensors="jax",
+        )
```
mrakitaTT, line 69: In Sentence Transformer they are calling tokenizer with specific configuration, as can be seen [here](https://github.com/UKPLab/sentence-transformers/blob/dd8a8d7d4d5660e7b2b69e8a1ece8202417c0069/sentence_transformers/models/Transformer.py#L507). They are using `truncation="longest_first"` and `max_length=self.max_seq_length`. Parameter `max_seq_length` is loaded from the HF model package file `sentence_bert_config.json` [here](https://huggingface.co/emrecan/bert-base-turkish-cased-mean-nli-stsb-tr/blob/main/sentence_bert_config.json). Ideally we would parse this info from the config file, but for now we can just hardcode the value into the constant here before the tokenizer call, with a todo comment to do it later.

Issue #468, opened 2025-04-05
Add support for PJRT_Memory


mrakitaTT:
`PJRT_Memory` structure and related API is currently not supported in our PJRT implementation. Luckily our plugin works without it, but we should investigate how it is used in clients and support it if needed.

mrakitaTT:
From PJRT [documentation](https://openxla.org/xla/pjrt/cpp_api_overview#pjrtmemoryspace):
>Memory spaces can be used to describe a location of memory. These can either be unpinned, and are free to live anywhere but be accessible from a device, or they can be pinned and must live on a specific device.
>
>Memory spaces know their associated buffers of data, and the devices (plural) that a memory space is associated with, as well as the client it is a part of.

More API documentation [here](https://github.com/openxla/xla/blob/3e448cf9e86775a37ec5f7d3c69dfb20e0c760df/xla/pjrt/pjrt_client.h#L72).

LPanosTT:
@ajakovljevicTT The branch I've been working on is experimental for the most part. I would start from scratch and use my branch as reference if you need. I wouldn't assume my implementations are entirely correct/by the book. I was mostly testing the waters, doing some discovery.

PR #536, opened 2025-04-24
Implementing the PJRT_Memory 

ajakovljevicTT:
Closes https://github.com/tenstorrent/tt-xla/issues/468

Implements the API functions necessary for PJRT_Memory support, needed for running `torch_xla` through `tt_xla`.


src/common/pjrt_implementation/client_instance.cc
```diff
@@ -119,6 +129,28 @@ tt_pjrt_status ClientInstance::populateDevices() {
   return tt_pjrt_status::kSuccess;
 }
 
+tt_pjrt_status ClientInstance::populateMemories() {
+  DLOG_F(LOG_DEBUG, "ClientInstance::PopulateMemories");
+
+  std::unique_ptr<MemoryInstance> host_memory = MemoryInstance::createInstance(
+      m_addressable_devices_raw, /*id=*/0, "tt_host");
```
kmitrovicTT, line 136: Arbitrary string. I'd declare it as static class member.

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -119,6 +129,28 @@ tt_pjrt_status ClientInstance::populateDevices() {
...
+  m_addressable_memories_raw.push_back(host_memory.get());
+  m_addressable_memories.push_back(std::move(host_memory));
+  for (size_t i = 0; i < m_devices.size(); ++i) {
+    m_devices[i]->addAddressableMemory(m_addressable_memories_raw[0]);
+    std::vector<DeviceInstance *> single_addressable_device = {
+        m_addressable_devices_raw[i]};
+    std::unique_ptr<MemoryInstance> device_memory =
+        MemoryInstance::createInstance(single_addressable_device, /*id=*/i + 1,
+                                       "tt_device");
```
kmitrovicTT, line 145: Same for this one.

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -168,7 +200,8 @@ ClientInstance::compileMlirProgram(const PJRT_Program *mlir_program,
 
   std::unique_ptr<LoadedExecutableInstance> executable =
       LoadedExecutableInstance::createInstance(executable_image,
-                                               std::move(addressable_devices));
+                                               std::move(addressable_devices),
+                                               m_addressable_memories[0].get());
```
kmitrovicTT, line 204: Maybe we should create `getHostMemory` method to retrieve this. Or even store special attribute `m_host_memory`.

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -119,6 +129,28 @@ tt_pjrt_status ClientInstance::populateDevices() {
   return tt_pjrt_status::kSuccess;
 }
 
+tt_pjrt_status ClientInstance::populateMemories() {
+  DLOG_F(LOG_DEBUG, "ClientInstance::PopulateMemories");
+
+  std::unique_ptr<MemoryInstance> host_memory = MemoryInstance::createInstance(
+      m_addressable_devices_raw, /*id=*/0, "tt_host");
+  m_addressable_memories_raw.push_back(host_memory.get());
+  m_addressable_memories.push_back(std::move(host_memory));
```
kmitrovicTT, line 138: We could store `m_host_memory` here, but also keep it in `m_addressable_memories(_raw)`. Just to be explicit about it, this way it is implicit that first memory in array represents host memory.

kmitrovicTT: Empty line before `for`.

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -348,10 +379,18 @@ onBufferFromHostBuffer(PJRT_Client_BufferFromHostBuffer_Args *args) {
                 .release();
   }
 
+  DeviceInstance *device_instance =
+      args->memory == nullptr
+          ? DeviceInstance::unwrap(args->device)
+          : MemoryInstance::Unwrap(args->memory)->getAddressableByDevices()[0];
```
kmitrovicTT, line 385: Why is it hardcoded to 0? Should we leave a TODO or something?

src/common/pjrt_implementation/executable_image.cc
```diff
@@ -10,10 +10,13 @@
 
 #include "common/pjrt_implementation/executable_image.h"
 
+// tt-xla includes
 #include "common/pjrt_implementation/data_type_utils.h"
 
 namespace tt::pjrt {
 
+const std::string ExecutableImage::host_memory_kind_name = "tt_host";
```
kmitrovicTT, line 18: Oh you already have it here. This is what I meant by one of the previous comments. Define the same thing for `tt_device` and put it somewhere where it is reachable in `ClientInstance::populateMemories` as well.

third_party/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-set(TT_MLIR_VERSION "ddb9b71e5aaa032986754972f51b36bb07666f20")
+set(TT_MLIR_VERSION "3a64bff6239ed549ca7824cf3d88793bbe606b72")
```
kmitrovicTT: I don't see this commit on main. Maybe you meant to use `eb448969aaaa0152b570882849c892eeadcedc16` (https://github.com/tenstorrent/tt-mlir/commit/eb448969aaaa0152b570882849c892eeadcedc16)

kmitrovicTT (changes requested):

third_party/CMakeLists.txt
ajakovljevicTT: Noted, now left the tt-mlir hash unchanged, since no new tt-mlir functionality is needed for our tt-xla tests. For tt-torch, we need to uplift tt-mlir since https://github.com/tenstorrent/tt-mlir/pull/2975 was merged.

kmitrovicTT (approved):

mrakitaTT (commented):
First round of comments, will continue reviewing tomorrow :)

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -446,13 +449,8 @@ PJRT_Error *
...
 
-  tt_pjrt_status status;
-  if (args->execute_device) {
-    DLOG_F(ERROR, "Executing on a specific single device is not supported");
-    status = tt_pjrt_status::kUnimplemented;
-  } else {
-    status = LoadedExecutableInstance::unwrap(args->executable)->execute(args);
-  }
+  tt_pjrt_status status =
+      LoadedExecutableInstance::unwrap(args->executable)->execute(args);
```
mrakitaTT, line 453: What happened here?

src/common/pjrt_implementation/stubs.inc
```diff
@@ -46,10 +41,9 @@
   // https://github.com/tenstorrent/tt-xla/issues/467
   _STUB(PJRT_Buffer_CopyToDevice);
 
-  // PJRT_Memory is not yet supported in our implementation.
-  // https://github.com/tenstorrent/tt-xla/issues/468
+  // PJRT_Buffer_CopyToMemory is not used in jax or torch client, but should be implemented.
```
mrakitaTT, line 44: `... in JAX or PyTorch client, ...`

src/common/pjrt_implementation/memory_instance.cc
```diff
@@ -0,0 +1,129 @@
...
+    make_unique_enabler(std::vector<DeviceInstance *> &addressable_by_devices,
+                        size_t id, std::string memory_kind)
+        : MemoryInstance(addressable_by_devices, id, memory_kind) {}
+  };
+  return std::make_unique<make_unique_enabler>(addressable_by_devices, id,
+                                               memory_kind);
+}
+
+void MemoryInstance::bindApi(PJRT_Api *api) {
+  DLOG_F(LOG_DEBUG, "DeviceInstance::BindApi");
```
mrakitaTT, line 38: Remove log

src/common/pjrt_implementation/memory_instance.cc
```diff
@@ -0,0 +1,129 @@
...
+}
+
+DeviceInstance *MemoryInstance::getDevice() {
+  if (m_memory_kind == host_memory_kind_name) {
+    DLOG_F(WARNING,
+           "MemoryInstance::getDevice: Host memory does not have a device.");
+
+    return nullptr;
+  }
+  assert(m_addressable_by_devices.size() == 1 &&
```
mrakitaTT, line 65: add blank line above

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -348,10 +378,18 @@ onBufferFromHostBuffer(PJRT_Client_BufferFromHostBuffer_Args *args) {
                 .release();
   }
 
+  DeviceInstance *device_instance =
+      args->memory == nullptr
+          ? DeviceInstance::unwrap(args->device)
+          : MemoryInstance::unwrap(args->memory)->getDevice();
+
+  MemoryInstance *memory_instance =
+      args->memory == nullptr ? nullptr : MemoryInstance::unwrap(args->memory);
```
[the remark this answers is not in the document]
mrakitaTT: Also, after initializing the `memory_instance`, you should add a check to see if `memory_instance` kind is host memory, in which case we should log error saying that we only support creating on-device buffers and return error with `tt_pjrt_status::kUnimplemented`

src/common/pjrt_implementation/memory_instance.cc
```diff
@@ -0,0 +1,129 @@
...
+  return m_addressable_by_devices[0];
+}
+
+namespace internal {
+
+PJRT_Error *onMemoryId(PJRT_Memory_Id_Args *args) {
+  DLOG_F(LOG_DEBUG, "MemoryInstance::PJRT_Memory_Id");
+  MemoryInstance *memory_instance = MemoryInstance::unwrap(args->memory);
+  args->id = memory_instance->getId();
+  return nullptr;
```
mrakitaTT, line 78: Add blank lines after the log and before return, in this and other functions below

src/common/pjrt_implementation/memory_instance.cc
```diff
@@ -0,0 +1,129 @@
...
+
+  return m_addressable_by_devices[0];
+}
+
+namespace internal {
+
+PJRT_Error *onMemoryId(PJRT_Memory_Id_Args *args) {
+  DLOG_F(LOG_DEBUG, "MemoryInstance::PJRT_Memory_Id");
+  MemoryInstance *memory_instance = MemoryInstance::unwrap(args->memory);
+  args->id = memory_instance->getId();
```
mrakitaTT, line 77: Can just do `args->id = MemoryInstance::unwrap(args->memory)->getId();`

src/common/pjrt_implementation/memory_instance.cc
```diff
@@ -0,0 +1,129 @@
...
+  args->kind = memory_instance->getMemoryKind().data();
+  args->kind_size = memory_instance->getMemoryKind().size();
+  return nullptr;
+}
+
+PJRT_Error *onMemoryKindId(PJRT_Memory_Kind_Id_Args *args) {
+  DLOG_F(LOG_DEBUG, "MemoryInstance::PJRT_Memory_Kind_Id");
+  MemoryInstance *memory_instance = MemoryInstance::unwrap(args->memory);
+  args->kind_id =
+      memory_instance->getMemoryKind() == MemoryInstance::host_memory_kind_name
```
mrakitaTT, line 93: I don't like that we have to compare strings every time to check the memory kind. I think we should change the `MemoryInstance` to accept `is_host_memory` parameter, and then based on that the constructor can set the proper memory kind and can cache this parameter in member variable, and then you can expose the `isHostMemory()` function which you call here and in other places.

src/common/pjrt_implementation/memory_instance.cc
```diff
@@ -0,0 +1,129 @@
...
+  return nullptr;
+}
+
+PJRT_Error *
+onMemoryAddressableByDevices(PJRT_Memory_AddressableByDevices_Args *args) {
+  DLOG_F(LOG_DEBUG, "MemoryInstance::PJRT_Memory_AddressableByDevices");
+  args->num_devices =
+      MemoryInstance::unwrap(args->memory)->getAddressableByDevices().size();
+  const std::vector<DeviceInstance *> &addressable_by_devices =
+      MemoryInstance::unwrap(args->memory)->getAddressableByDevices();
```
mrakitaTT, line 121: First unwrap the memory instance into variable, to avoid unwrapping it twice

src/common/pjrt_implementation/memory_instance.cc
```diff
@@ -0,0 +1,129 @@
...
+  args->devices = const_cast<PJRT_Device **>(
+      reinterpret_cast<PJRT_Device *const *>(addressable_by_devices.data()));
```
mrakitaTT, line 123: These casts shouldn't be necessary, at least the `const_cast`

src/common/pjrt_implementation/loaded_executable_instance.cc
ajakovljevicTT: Torch_xla in its multiprocess configuration actually sets this parameter, along with putting the buffer on the device.

mrakitaTT: Oh, interesting, maybe we should then put some check in `openDevices` function when `args->execute_device` is not null, to make sure that `device_ids.size() == 1 && device_ids[0] == args->execute_device`?

inc/common/pjrt_implementation/memory_instance.h
```diff
@@ -0,0 +1,116 @@
...
+                 size_t id, bool is_host_memory);
+
+  // List of devices that can access this memory.
+  std::vector<DeviceInstance *> m_addressable_by_devices;
+
+  // String representing the kind of memory, can be 'tt_host' or 'tt_device'.
+  std::string m_memory_kind;
+
+  // Id of the memory.
+  size_t m_id;
```
mrakitaTT, line 85: In `PJRT_Memory_Id_Args` type of id is `int`

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -348,10 +382,21 @@ onBufferFromHostBuffer(PJRT_Client_BufferFromHostBuffer_Args *args) {
                 .release();
   }
 
+  MemoryInstance *memory_instance =
+      args->memory == nullptr ? nullptr : MemoryInstance::unwrap(args->memory);
```
mrakitaTT, line 386: You can just do `MemoryInstance *memory_instance = MemoryInstance::unwrap(args->memory);`, it should be safe to cast nullptr to `MemoryInstance *`

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -348,10 +382,21 @@ onBufferFromHostBuffer(PJRT_Client_BufferFromHostBuffer_Args *args) {
...
+
+  DeviceInstance *device_instance = args->memory == nullptr
```
mrakitaTT, line 388: You can do `memory_instance ?` instead of `args->memory == nullptr` (just reverse the order of asignments below)

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -348,10 +382,21 @@ onBufferFromHostBuffer(PJRT_Client_BufferFromHostBuffer_Args *args) {
...
-      BufferInstance::createInputBufferInstance(
-          args->type, args->dims, args->num_dims,
-          DeviceInstance::unwrap(args->device));
+      BufferInstance::createInputBufferInstance(args->type, args->dims,
+                                                args->num_dims, device_instance,
+                                                memory_instance);
+  if (memory_instance->isHostMemory()) {
+    DLOG_F(ERROR, "We only support creating buffers on device memory");
+    return *ErrorInstance::makeError(tt_pjrt_status::kUnimplemented).release();
+  }
```
mrakitaTT, line 399: Should also check if `memory_instance` is nullptr, can do just `if (memory_instance && memory_instance->isHostMemory())`. And move this check above, after initializing `memory_instance` variable.

src/common/pjrt_implementation/device_instance.cc
```diff
@@ -33,6 +34,8 @@ void DeviceInstance::bindApi(PJRT_Api *api) {
   api->PJRT_Device_GetDescription = internal::onDeviceGetDescription;
   api->PJRT_Device_IsAddressable = internal::onDeviceIsAddressable;
   api->PJRT_Device_LocalHardwareId = internal::onDeviceLocalHardwareId;
+  api->PJRT_Device_DefaultMemory = internal::onDeviceDefaultMemory;
+  api->PJRT_Device_AddressableMemories = internal::onDeviceAddressableMemories;
```
mrakitaTT, line 38: Update order here too

src/common/pjrt_implementation/memory_instance.cc
```diff
@@ -0,0 +1,140 @@
...
+
+PJRT_Error *
+onMemoryAddressableByDevices(PJRT_Memory_AddressableByDevices_Args *args) {
+  DLOG_F(LOG_DEBUG, "MemoryInstance::PJRT_Memory_AddressableByDevices");
+
+  const MemoryInstance *memory_instance = MemoryInstance::unwrap(args->memory);
+
+  args->num_devices = memory_instance->getAddressableByDevices().size();
+  const std::vector<DeviceInstance *> &addressable_by_devices =
+      memory_instance->getAddressableByDevices();
```
mrakitaTT, line 131: Move this below `memory_instance` variable initialization, and use for both `args->num_devices` and `args->devices` assignment. Also, you should assign the `args->devices` first, and then `args->num_devices`, because that's the order in which they are declared in `PJRT_Memory_AddressableByDevices_Args`.

inc/common/pjrt_implementation/memory_instance.h
```diff
@@ -0,0 +1,116 @@
...
+    return reinterpret_cast<MemoryInstance *>(memory);
+  }
+
+  // Gets the list of devices that can address this memory.
+  const std::vector<DeviceInstance *> &getAddressableByDevices() const {
+    return m_addressable_by_devices;
+  }
+
+  // Checks if the memory is host memory.
+  bool isHostMemory() const { return m_memory_kind == c_host_memory_kind_name; }
```
mrakitaTT, line 52: This is still doing the string comparison, which we should avoid. You can have the `m_is_host_memory` private variable, which you set in constructor based on `is_host_memory` parameter, and use here.

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -119,6 +129,31 @@ tt_pjrt_status ClientInstance::populateDevices() {
...
+
+  for (size_t i = 0; i < m_devices.size(); ++i) {
+    // Adding host memory to device addressable memories.
+    m_devices[i]->addAddressableMemory(m_addressable_host_memory.get());
+
+    // Adding device memory to device addressable memories.
+    std::vector<DeviceInstance *> single_addressable_device = {
+        m_addressable_devices_raw[i]};
+    std::unique_ptr<MemoryInstance> device_memory =
+        MemoryInstance::createInstance(single_addressable_device, /*id=*/i + 1,
```
mrakitaTT, line 146: One thought that just occurred to me is that it might be useful to have the memory id matching the device id, for easier debugging. We could have two creation functions for `MemoryInstance`, `createHostMemoryInstance` which sets `m_is_host_memory` to `true` and `m_id` to `-1`, and `createDeviceMemoryInstance` which takes `local_device_id` and sets `m_id` to that and `m_is_host_memory` to `false`.

On the other hand, memory instance is connected with device instance, so we might not have a need to deduce this connection from ids. I am just leaving this comment as a potential idea in case we see this need in the future, but consider it no-op for now.

mrakitaTT (approved):

Issue #526, opened 2025-04-22
[Multi-device] Fix issue in estimating output shape


wooseokTT:
Current output shape estimation may need to be updated in determining the output shape. 

Traceback (most recent call last):
  File "/proj_sw/user_dev/wooseoklee/tt-xla/./examples/test.py", line 46, in <module>
    mesh_test(input)
  File "/proj_sw/user_dev/wooseoklee/tt-xla/./examples/test.py", line 36, in mesh_test
    print(res)
  File "/proj_sw/user_dev/wooseoklee/tt-xla/venv/lib/python3.10/site-packages/jax/_src/array.py", line 283, in __str__
    return str(self._value)
  File "/proj_sw/user_dev/wooseoklee/tt-xla/venv/lib/python3.10/site-packages/jax/_src/profiler.py", line 333, in wrapper
    return func(*args, **kwargs)
  File "/proj_sw/user_dev/wooseoklee/tt-xla/venv/lib/python3.10/site-packages/jax/_src/array.py", line 648, in _value
    npy_value[ind] = self._arrays[i]._single_device_array_to_np_array()
ValueError: could not broadcast input array from shape (1024,128) into shape (128,1024)

-----------------------------------------------------------------------------

import os
import jax
import jax.numpy as jnp
import jax._src.xla_bridge as xb
from jax.sharding import Mesh, PartitionSpec, NamedSharding
from jax.sharding import PartitionSpec as P
from jax.experimental import mesh_utils
from jax.experimental.shard_map import shard_map
from functools import partial

def initializePJRT():
  path = os.path.join(os.path.dirname(__file__), "../build/src/tt/pjrt_plugin_tt.so")
  if not os.path.exists(path):
    raise FileNotFoundError(f"Could not find tt_pjrt C API plugin at {path}, have you compiled the project?")
  plugin = xb.register_plugin('tt', priority=10, library_path=path, options=None)
  jax.config.update("jax_platforms", "cpu,tt")
  jax.config.update("jax_use_shardy_partitioner", True)

def mesh_test(value):
  device_tt = jax.devices('tt')
  devices = mesh_utils.create_device_mesh((1, 8), device_tt)
  mesh = Mesh(devices=devices[:4], axis_names=('x', 'y'))

  input_sharding = NamedSharding(mesh, P('y', None))
  output_sharding = NamedSharding(mesh, P('y', None))

  value = jax.device_put(value, input_sharding)
[...]

PR #538, opened 2025-04-24
Fixing the calculation of the output shape for multidevice runs

ajakovljevicTT:
Currently, we reliad on the `sharding_strategy` map to calculate what is the output of the individual output shards of our program. This can be wrong for shardings of tensors of rank bigger than 2, and incorrect for some usages in `torchax`. Switched to using `shard_shape` from `sharding_utils::MeshSharding` to calculate the output shard shape.

Closes https://github.com/tenstorrent/tt-xla/issues/526

### DIFF
--- a/src/common/pjrt_implementation/loaded_executable_instance.cc
+++ b/src/common/pjrt_implementation/loaded_executable_instance.cc
@@ -323,8 +323,7 @@ void LoadedExecutableInstance::fillPJRTOutputLists(
     for (size_t output_index = 0; output_index < num_outputs; ++output_index) {
       tt::runtime::Tensor output_tensor =
           untilized_output_tensors[output_index][device_index];
-      std::vector<std::uint32_t> output_shape =
-          getOutputShape(output_index, num_devices);
+      std::vector<std::uint32_t> output_shape = getOutputShape(output_index);
 
       std::unique_ptr<BufferInstance> output_buffer =
           BufferInstance::createOutputBufferInstance(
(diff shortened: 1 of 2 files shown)


ajakovljevicTT:
cc: @wooseokTT

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -341,35 +340,31 @@ void LoadedExecutableInstance::fillPJRTOutputLists(
...
       m_executable_image->getOutputSharding(output_index);
 
-  mlir::FailureOr<std::unordered_map<std::string, std::string>>
-      shardingStrategy =
-          mlir::tt::sharding_utils::MeshSharding::fillStrategyMapFromSharding(
-              outputSharding, num_devices);
-  if (mlir::failed(shardingStrategy)) {
-    DLOG_F(WARNING, "No valid output sharding, returning the original shape");
+  if (outputSharding.getShardType() == mlir::tt::MeshShardType::Identity) {
+    DLOG_F(WARNING, "No output sharding, returning the original shape");
```
mrakitaTT, line 350: Would it be better to say "Output is replicated, returning the original shape"? Btw why do we even log this, and with a warning, this seems like something that can expectedly happen no?

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -341,35 +340,31 @@ void LoadedExecutableInstance::fillPJRTOutputLists(
...
-    assert(outputShape[0] % std::stoi(shardingStrategy->at("mesh_shape_y")) ==
-           0);
-    assert(outputShape[1] % std::stoi(shardingStrategy->at("mesh_shape_x")) ==
-           0);
-    outputShape[0] =
-        outputShape[0] / std::stoi(shardingStrategy->at("mesh_shape_y"));
-    outputShape[1] =
-        outputShape[1] / std::stoi(shardingStrategy->at("mesh_shape_x"));
+  llvm::ArrayRef<int64_t> shard_shape = outputSharding.getShardShape();
+  if (shard_shape.size() != outputShape.size()) {
```
mrakitaTT, line 355: If this is always expected, we should check it in `module_builder` and return compilation error if it happens, and then just assert here instead of this `if`.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -341,35 +340,31 @@ void LoadedExecutableInstance::fillPJRTOutputLists(
...
+    DLOG_F(ERROR,
+           "Output sharding shape (%zu) doesn't match the output shape (%zu)",
+           shard_shape.size(), outputShape.size());
+    return outputShape;
+  }
+
+  for (size_t i = 0; i < outputShape.size(); ++i) {
+    if (shard_shape[i] != 1) {
+      assert(outputShape[i] % shard_shape[i] == 0 &&
```
mrakitaTT, line 364: This should also be checked in `module_builder`, beside this assert. That way compilation will error out, we will find this more timely, and we will not rely on runtime failing with the inappropriately divided output shape.

src/common/pjrt_implementation/loaded_executable_instance.cc
ajakovljevicTT: You are correct, changed.

src/common/pjrt_implementation/loaded_executable_instance.cc
ajakovljevicTT: We only get the outputShapes when we already have the flatbuffer, in the creation of the `ExecutableImage()`, i can put the check after that, but for me it seems that the checks are then spread out over the code. What do you think?

ajakovljevicTT: Added an overreaching check in the `ModuleBuilder` for all of these asserts, please review.

src/common/pjrt_implementation/loaded_executable_instance.cc
ajakovljevicTT: Added an overreaching check in the ModuleBuilder for all of these asserts, please review.

src/common/module_builder.cc
```diff
@@ -565,6 +565,50 @@ void ModuleBuilder::verifyCreatedFlatbufferBinary() {
...
+
+  checkOutputShardingShapes();
+}
+
+void ModuleBuilder::checkOutputShardingShapes() {
+  // Assuming only one program per flatbuffer for now.
+  std::uint32_t program_index = 0;
+  std::vector<tt::runtime::TensorDesc> output_specs =
+      m_flatbuffer_binary.getProgramOutputs(program_index);
+  size_t num_outputs = output_specs.size();
```
mrakitaTT, line 577: This is already calculated in `verifyCreatedFlatbufferBinary`, should be passed in to this function

mrakitaTT (approved):

src/common/module_builder.cc
```diff
@@ -565,6 +565,46 @@ void ModuleBuilder::verifyCreatedFlatbufferBinary() {
     m_status = tt_pjrt_status::kInternal;
     return;
   }
+
+  checkOutputShardingShapes(output_specs, num_outputs);
+}
+
+void ModuleBuilder::checkOutputShardingShapes(
+    const std::vector<tt::runtime::TensorDesc> &output_specs,
+    size_t num_outputs) {
```
mrakitaTT, line 574: Sorry for nitpicking, but since `num_outputs` is used only once in this function, we don't have to pass it as a parameter, you can use `output_specs.size()` directly in the for loop below

mrakitaTT (approved):

PR #550, opened 2025-04-29
Add AlexNet model multichip implementation, example script and tests

mrakitaTT:
### Problem description
Adding AlexNet model multichip implementation, example script and tests.

### What's changed
- Added AlexNet model multichip implementation, utilizing data parallelism in feature extraction layers, and model parallelism in classifier layers, following the [original paper](https://arxiv.org/abs/1404.5997)
- Added example script for running the model on TT devices
- Added n300 (1x2 mesh) and llmbox (1x4 and 1x8 meshes) tests for alexnet multichip implementation
- Implemented `MultichipModelTester` in test infra
- Refactored `ModelTester` a bit to cache necessary workload inputs and added default behaviour for Flax linen and HF `FlaxPreTrainedModel`
- Removed now unnecessary code from model tests

### Checklist

### DIFF
--- a/examples/alexnet_multichip.py
+++ b/examples/alexnet_multichip.py
@@ -1,0 +1,130 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+from typing import Sequence
+
+import jax
+import jax._src.xla_bridge as xb
+from flax import linen as nn
+from jax.experimental.shard_map import shard_map
+from jax.sharding import NamedSharding
+from jax.sharding import PartitionSpec as P
+
+# TODO Move this to TT models repo.
+from tests.jax.multi_chip.n300.models.tensor_parallel.alexnet.model_implementation import (
+    AlexNetMultichipModel,
+)
+
+# Allocating enough CPU devices so we can create various meshes depending on which TT
+# device this example is running. Can't be set to exact number of TT devices because
+# after calling `jax.devices` function this config update doesn't work anymore.
+jax.config.update("jax_num_cpu_devices", 8)
+
+# Change if you want to use shardy.
+jax.config.update("jax_use_shardy_partitioner", False)
+
+
+def register_pjrt_plugin():
+    """Registers TT PJRT plugin."""
+
+    plugin_path = os.path.join(
+        os.path.dirname(__file__), "../build/src/tt/pjrt_plugin_tt.so"
+    )
+    if not os.path.exists(plugin_path):
+        raise FileNotFoundError(f"Could not find TT PJRT plugin at {plugin_path}")
+
+    xb.register_plugin("tt", library_path=plugin_path)
+    jax.config.update("jax_platforms", "cpu,tt")
+
+
+def generate_inputs_on_cpu(prng_key: jax.Array) -> Sequence[jax.Array]:
+    """Generates inputs on CPU."""
+
+    with jax.default_device(jax.devices("cpu")[0]):
+        return jax.random.randint(
+            key=prng_key,
+            # B, H, W, C
+            shape=(32, 224, 224, 3),
+            # In the original paper inputs are normalized with individual channel
+            # values learned from training set.
+            minval=-128,
+            maxval=128,
+        )
+
+
+def initialize_parameters(
+    model, inputs_specs, cpu_inputs, params_specs, device_mesh, prng_key
+):
+    """Initializes model parameters. Currently we don't support RNG on device, so inputs
+    are initialized on CPU and moved to device."""
+
+    # Initializing parameters on CPU.
+    cpu_mesh = jax.make_mesh(
+        device_mesh.axis_sizes, device_mesh.axis_names, devices=jax.devices("cpu")
+    )
+    init_fn = shard_map(
+        model.init, cpu_mesh, in_specs=(None, inputs_specs), out_specs=params_specs
+    )
+    params = init_fn(prng_key, cpu_inputs)
+
+    # Moving parameters to device.
+    params = jax.tree.map(
+        lambda spec, param: jax.device_put(param, NamedSharding(device_mesh, spec)),
+        params_specs,
+        params,
+        is_leaf=lambda x: isinstance(x, nn.Partitioned) or isinstance(x, P),
+    )
+
+    return params
+
+
+register_pjrt_plugin()
+
+axis_name = "X"
+# Currently we support only 2D mesh with shardy enabled.
+device_mesh = jax.make_mesh(
+    (1, len(jax.devices("tt"))), ("Y", axis_name), devices=jax.devices("tt")
+)
+
+model = AlexNetMultichipModel(
+    axis_name=axis_name, num_devices=len(jax.devices("tt")), train_mode=False
+)
+
+prng_key = jax.random.PRNGKey(23)
+
+# Sharding data on batch axis since data parallelism is utilized for the convolutional
+# layers.
+inputs_specs = P(axis_name)
+cpu_inputs = generate_inputs_on_cpu(prng_key)
+
+# Have to use shard_map because CCL ops need a mapped axis for tracing to work.
+params_specs = nn.get_partition_spec(
+    jax.eval_shape(
+        shard_map(
+            model.init, device_mesh, in_specs=(None, inputs_specs), out_specs=P()
+        ),
+        prng_key,
+        cpu_inputs,
+    )
+)
+params = initialize_parameters(
+    model, inputs_specs, cpu_inputs, params_specs, device_mesh, prng_key
+)
+
+# Now we can move inputs to device, needed them on CPU to initialize the parameters.
+device_inputs = jax.device_put(cpu_inputs, NamedSharding(device_mesh, inputs_specs))
+
+compiled_apply = jax.jit(
+    shard_map(
+        model.apply,
+        device_mesh,
+        in_specs=(params_specs, inputs_specs),
+        out_specs=P(),
+        check_rep=False,
+    )
+)
+results = compiled_apply(params, device_inputs)
+
+print(results)
--- a/examples/alexnet_multichip.py
+++ b/examples/alexnet_multichip.py
@@ -23,7 +23,7 @@ from tests.jax.multi_chip.n300.models.tensor_parallel.alexnet.model_implementati
 jax.config.update("jax_num_cpu_devices", 8)
 
 # Change if you want to use shardy.
-jax.config.update("jax_use_shardy_partitioner", True)
+jax.config.update("jax_use_shardy_partitioner", False)
 
 
 def register_pjrt_plugin():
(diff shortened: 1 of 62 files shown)


nsumrakTT (approved):

tests/infra/multichip_utils.py
```diff
@@ -39,8 +47,41 @@ def enable_shardy(use_shardy: bool):
...
+def make_partition_spec(axis_names: tuple) -> PartitionSpec:
     """
     Returns a PartitionSpec object for the given `axis_names`.
     """
-    return jax.sharding.PartitionSpec(*axis_names)
+    return PartitionSpec(*axis_names)
+
+
+def make_flax_linen_parameters_partition_specs(
+    model: linen.Module, cpu_mesh: Mesh, inputs_specs: PartitionSpec, cpu_inputs: Tensor
```
ajakovljevicTT, line 58: should be `device_mesh`

tests/jax/multi_chip/n300/models/tensor_parallel/alexnet/model_implementation.py
```diff
@@ -0,0 +1,145 @@
...
+            kernel_size=(3, 3),
+            strides=(1, 1),
+            padding=1,
+            param_dtype=self.param_dtype,
+        )(x)
+        x = nn.relu(x)
+
+        # Fourth feature extraction layer
+        x = nn.Conv(
+            features=256,
```
sgligorijevicTT, line 71: You probably meant to put 384 here.
"One weird trick for parallelizing convolutional neural networks" page 5 footnote 1: "In detail, the single-column model has 64, 192, 384, 384, 256 filters in the five convolutional layers, respectively."

tests/infra/multichip_utils.py
mrakitaTT: I'm purposely doing this on CPU by passing `cpu_mesh` and `cpu_inputs`, because we can't initialize parameters on device, and also because device is not needed for this partition specs calculation

sgligorijevicTT (approved):
There are two tiny mismatches between the cited paper and the model code but otherwise lgtm.

jmcgrathTT (approved):

tests/infra/device_runner.py
```diff
@@ -158,6 +142,30 @@ def _put_multichip_workload_on_device(
...
+                lambda spec, param: jax.device_put(
+                    param, NamedSharding(device_mesh, spec)
+                ),
+                partition_spec,
+                arg,
+                is_leaf=lambda x: (
+                    isinstance(x, linen.Partitioned) or isinstance(x, PartitionSpec)
+                ),
+            )
+        return arg
```
kmitrovicTT, line 167: I'd put `return arg` in `else` branch, kinda seems weird that there is an `if` and an `elif` branch, but no `else`.
Or lose `elif` in favor of another `if`.

tests/infra/device_runner.py
```diff
@@ -158,6 +142,30 @@ def _put_multichip_workload_on_device(
             in_specs=multichip_workload.in_specs,
         )
 
+    @staticmethod
+    def _put_sharded_arg_on_multichip_device(
+        arg: Any, device_mesh: Mesh, partition_spec: PartitionSpec | PyTree
```
kmitrovicTT, line 147: How can `partition_spec` be a `Pytree`?

tests/infra/device_runner.py
```diff
@@ -121,34 +113,26 @@ def _put_multichip_workload_on_device(
...
-                    multichip_workload.device_mesh,
-                    multichip_workload.in_specs[spec_index],
-                )
+        for key, arg in multichip_workload.kwargs.items():
+            device_arg = DeviceRunner._put_sharded_arg_on_multichip_device(
+                arg,
+                multichip_workload.device_mesh,
+                multichip_workload.in_specs[spec_index],
+            )
+            if device_arg is not arg:
```
kmitrovicTT, line 133: This statement seems weird, maybe a short comment `# If arg can't be put on device we increment...`

tests/infra/multichip_model_tester.py
```diff
@@ -0,0 +1,148 @@
...
+            compiled_cpu_workload, ShardingMode.INPUTS_AND_MODULE
+        )
+
+        self._compare(tt_res, cpu_res)
+
+    def _compile_for_device(
+        self, device_mesh: Mesh, forward_method_arg_specs: tuple[PartitionSpec | PyTree]
+    ) -> Callable:
+        """JIT-compiles model's forward pass into optimized kernels for the given device
+        mesh."""
```
kmitrovicTT, line 134:
```
"""
JIT-compiles model's forward pass into optimized kernels for the given device mesh.
"""
```

tests/jax/single_chip/models/example_model/only_kwargs/test_example_model_only_kwargs.py
```diff
@@ -31,6 +31,10 @@ def _get_input_activations(self) -> Sequence[jax.Array]:
         act = jax.numpy.ones(act_shape)
         return [act]
 
+    # @override
+    def _get_input_parameters(self) -> PyTree:
+        return ()
```
kmitrovicTT, line 36: You could return `()` by default instead of raising an exception in `ModelTester._get_input_parameters`. That plus the `FlaxPreTrainedModel` default behaviour seems to cover a lot of cases. You can override it anyway if you need to in derived class.

kmitrovicTT (approved):
All good as far as changes to infra and single chip models are concerned. Minor comments below. 

Also, I can see a lot of special cases involving jax is starting to pile up. This might mean more trouble when trying to support torch. Just a note.

tests/infra/device_connector.py
```diff
@@ -122,14 +104,17 @@ def __init__(self) -> None:
 
         self._initialized = False
 
+        # Allocating enough CPU devices so we can create various meshes depending on
+        # which TT device tests are running. Can't be set to exact number of TT devices
+        # because after calling `jax.devices` function this config update doesn't work
+        # anymore.
+        jax.config.update("jax_num_cpu_devices", 8)
```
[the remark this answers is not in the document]
ajakovljevicTT: Also take a look if the `tests/infra/device_connector.py::simulate_cpu_mesh` function helps.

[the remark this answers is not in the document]
ajakovljevicTT: I agree with Kristijan, is there a specific reason that we are moving from the `_simulate_multiple_cpu_devices` function and the setup as it was?

tests/infra/device_runner.py
ajakovljevicTT: I might be reading something wrong, but does this support scalar parameters? Ones that correspond to:
```
            if not isinstance(arg, Tensor):
                args_on_device.append(arg)
```
in the old function. Kristijan's suggestion of the else branch will cover this case I think.

tests/infra/multichip_model_tester.py
```diff
@@ -0,0 +1,148 @@
...
+
+        compiled_cpu_workload = MultichipWorkload(
+            self._compile_for_device(self.cpu_mesh, forward_method_arg_specs),
+            self._workload.args,
+            self._workload.kwargs,
+            device_mesh=self.cpu_mesh,
+            in_specs=forward_method_arg_specs,
+        )
+        cpu_res = DeviceRunner.run_on_multichip_device(
+            compiled_cpu_workload, ShardingMode.INPUTS_AND_MODULE
```
ajakovljevicTT, line 125: Same as above.

tests/infra/multichip_utils.py
ajakovljevicTT: Could we make this more specific in the function test/comments? It was not clear to me, and I don't like that the only clue is the name of one variable.

tests/infra/multichip_utils.py
```diff
@@ -39,8 +47,41 @@ def enable_shardy(use_shardy: bool):
...
+    # Have to use shard_map because CCL ops need a mapped axis for tracing to work.
+    return linen.get_partition_spec(
+        jax.eval_shape(
+            shard_map(
+                model.init,
+                cpu_mesh,
+                in_specs=(None, inputs_specs),
+                out_specs=PartitionSpec(),
+            ),
+            jax.random.PRNGKey(21),
```
ajakovljevicTT, line 70: Not related to this specific PR, is there a guidance on what seed we should use/should be using? There is 21,27, i think I also saw something else. It might be good to make seeds the same as for the single chip models, as it helped me reproduce same inputs.

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -383,15 +383,36 @@ onBufferFromHostBuffer(PJRT_Client_BufferFromHostBuffer_Args *args) {
...
+
+  if (memory_instance) {
+    if (device_instance && device_instance != memory_instance->getDevice()) {
+      DLOG_F(ERROR, "Device set in `device` arg is different from the memory "
+                    "space device set in `memory` arg");
+      return *ErrorInstance::makeError(tt_pjrt_status::kInvalidArgument)
+                  .release();
+    }
+    device_instance = memory_instance->getDevice();
+  } else {
```
ajakovljevicTT, line 396: Maybe we could a comment here echoing what is being said in the `pjrt_c_api`? 
```
  // If nullptr, host data will be copied to `device`, otherwise we copy data to
  // `memory`.
  PJRT_Memory *memory;
```

tests/infra/multichip_utils.py
mrakitaTT: Sure, updated function name and header comment

tests/jax/multi_chip/n300/models/tensor_parallel/alexnet/model_implementation.py
mrakitaTT: Unfortunately architecture descriptions in both alexnet papers differ in some subtle ways from the actual implementation they used, you can see [here](https://github.com/akrizhevsky/cuda-convnet2/blob/master/layers/layers-imagenet-2gpu-data.cfg#L81), and also in PyTorch implementation [here](https://docs.pytorch.org/vision/main/_modules/torchvision/models/alexnet.html).

tests/jax/multi_chip/n300/models/tensor_parallel/alexnet/model_implementation.py
```diff
@@ -0,0 +1,145 @@
...
+            param_dtype=self.param_dtype,
+            kernel_init=dense_kernel_init,
+            bias_init=dense_bias_init,
+        )(x)
+
+        # Gather results from all devices
+        x = jax.lax.all_gather(x, self.axis_name, axis=1, tiled=True)
+
+        # Calculate probabilities
+        x = nn.softmax(x)
```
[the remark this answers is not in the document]
mrakitaTT: They mention that they used it because it is more performant, but then in the footnote they say that with only 1000 classes this doesn't play a role. In practice I've seen everyone use classic categorical cross entropy loss with softmax, and even them used that in the actual source [code](https://github.com/akrizhevsky/cuda-convnet2/blob/master/layers/layers-imagenet-2gpu-data.cfg#L145).

examples/alexnet_multichip.py
```diff
@@ -0,0 +1,130 @@
...
+# Change if you want to use shardy.
+jax.config.update("jax_use_shardy_partitioner", False)
+
+
+def register_pjrt_plugin():
+    """Registers TT PJRT plugin."""
+
+    plugin_path = os.path.join(
+        os.path.dirname(__file__), "../build/src/tt/pjrt_plugin_tt.so"
+    )
```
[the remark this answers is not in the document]
mrakitaTT: But then we will rely on running this script strictly from the repo root right? What if someone does `cd examples && python alexnet_multichip.py`

tests/infra/device_connector.py
```diff
@@ -138,7 +123,10 @@ def _number_of_devices(self, device_type: DeviceType) -> int:
 
     def _supported_devices(self) -> Sequence[DeviceType]:
         """Returns list of supported device types."""
-        # TODO support GPU
+        # The order here is important, JAX will respect that order to choose the default
+        # device. That way we don't need to use `with jax.default_device` to generate
+        # inputs on CPU etc. It also makes a difference in how JAX puts sharded tensors
+        # on device (https://github.com/tenstorrent/tt-xla/issues/542).
```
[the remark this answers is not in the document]
mrakitaTT: Haha yeah, but we should investigate this further though because our users probably won't use `jax.config.update`, or not with that exact order.

tests/infra/device_connector.py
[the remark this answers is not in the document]
mrakitaTT: >is there a specific reason that we are moving from the _simulate_multiple_cpu_devices function and the setup as it was?

Yes, because it doesn't work :) After you use `_simulate_multiple_cpu_devices` once it won't work anymore with different number of devices, so using it could lead to bugs. Also, if any of the previous tests used `jax.devices()` it will already initialize XLA devices, so `_simulate_multiple_cpu_devices` won't work at all.

Honestly I am surprised you guys did't like this, I thought it is pretty neat :) With this one liner we can now use any cpu mesh in tests, without need for any decorator functions and having to separate tests into different CI runs, etc. None of our tests should depend on the number of CPU devices, so it doesn't matter which number we set here as long as it is high enough (>= max number of tt devices). I thought about putting 1024 here but I wasn't sure if XLA has some limitations regarding this, so I've just put 8 since it is enough for now, we can raise it later.

tests/infra/device_runner.py
mrakitaTT: There is no difference in functionality with Kristijan's suggestion, only one extra line of code :D

This:
```
if (a):
    return 1
elif (b):
    return 2
return 3
```

is the same as:
```
if (a):
    return 1
elif (b):
    return 2
else:
    return 3
```

Both are commonly used, but yes as Kristijan pointed out `elif` can be replaced with only `if` in the first version. Will do that.

tests/infra/device_runner.py
mrakitaTT: When arg is a `PyTree` its partition spec is also a `PyTree`

tests/infra/device_runner.py
mrakitaTT: Yeah, I was not happy with this at all, in c++ I would pass `spec_index` by reference to the `_put_sharded_arg_on_multichip_device` so it can increment it, but in Python I can't pass int by reference so I had to either do this or put `spec_index` in the list and pass it to `_put_sharded_arg_on_multichip_device`... I will leave the comment as you've said. Btw we increment when we **can** put on device.

tests/jax/single_chip/models/example_model/only_kwargs/test_example_model_only_kwargs.py
mrakitaTT: Not really, `()` is legit only for the example model. We wan't to make sure user doesn't forget to implement this method, I think it is better to raise the exception in non-default case.

tests/infra/multichip_model_tester.py
```diff
@@ -0,0 +1,148 @@
...
+
+        compiled_tt_workload = MultichipWorkload(
+            self._compile_for_device(self.tt_mesh, forward_method_arg_specs),
+            self._workload.args,
+            self._workload.kwargs,
+            device_mesh=self.tt_mesh,
+            in_specs=forward_method_arg_specs,
+        )
+        tt_res = DeviceRunner.run_on_multichip_device(
+            compiled_tt_workload, ShardingMode.INPUTS_AND_MODULE
```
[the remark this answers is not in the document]
mrakitaTT: You are right, but in order to support other sharding modes I would have to make changes in some other places too, for example not to call the `shard_map` during compile. I would rather leave that for later once we have some working case of the model in other modes and do it in PR for that model.

tests/infra/multichip_utils.py
mrakitaTT: In general I think we shouldn't align seeds between different model tests, because that gives better chance that we will encounter some specific edge case. But in this case you are right, it might make sense to use the same seed between singlechip/multichip tests of the same model. I've extracted seed into common variable in alexnet tests.

tests/infra/multichip_model_tester.py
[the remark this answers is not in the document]
ajakovljevicTT: Noted, can you leave a TODO here?

mrakitaTT:
Nightly [passed](https://github.com/tenstorrent/tt-xla/actions/runs/15103443381)

tests/infra/multichip_model_tester.py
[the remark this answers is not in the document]
mrakitaTT: Sure, added

ajakovljevicTT (approved):

kmitrovicTT (approved):

PR #553, opened 2025-04-29
Test running on shared runners

vmilosevic:
### Ticket
N/A

### Problem description
Test running on shared runners

### What's changed
Run on shared runners

### Checklist

### DIFF
--- a/.github/workflows/on-nightly.yml
+++ b/.github/workflows/on-nightly.yml
@@ -13,8 +13,8 @@ jobs:
       test_mark: 'nightly'
       build_options: |
         [
-          {"runs-on": "n150", "name": "run", "dir": "./tests/jax/single_chip"},
-          {"runs-on": "n300", "name": "run", "dir": "./tests/jax/multi_chip/n300"},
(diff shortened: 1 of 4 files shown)


.github/workflows/on-nightly.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       test_mark: 'model_test'
       build_options: |
         [
-          {"runs-on": "n150", "name": "run", "dir": "./tests/jax/single_chip"},
+          {"runs-on": "tt-beta-ubuntu-2204-n150-large-stable", "name": "run", "dir": "./tests/jax/single_chip"},
```
mrakitaTT, line 31: Just curious, how much RAM do these `large-stable` shared runners have? Am I correct in assuming that these runners are "shared" only in terms of sharing machines across teams, but they are still running only one job at a time, i.e. they don't "share" resources between different jobs? I'd be worried because of RAM memory if two jobs could share one machine, for example if machine has multiple cards and each card gets allocated by another job but they still share the host RAM.

PR #573, opened 2025-05-09
Added support for torch in test infrastructure

kmitrovicTT:
### Ticket
/

### Problem description
We needed a way to support writing Torch models in our tests. Thanks to [torch_xla](https://github.com/pytorch/xla) Torch can hook to our custom PJRT plugin, as we did with JAX. What was left is to accommodate test infra to support different paths for jax and torch.

Result of this effort is that writing torch tests now looks exactly the same as it was for jax. You only need to write a model in torch, write custom tester by inheriting `ModelTester`, override some methods and that's it. Tester will automatically detect which framework model was written in and take care of all the complexity of compiling it, running it on device and comparing the results with CPU.

### What's changed
- Due to the need to support both jax and torch and different ways of doing stuff in those frameworks, and the need to leave the API towards test writer unchanged, I decided to leave `ModelTester` as one class and have many different helper classes which tester uses kinda like [strategies](https://refactoring.guru/design-patterns/strategy) on how to do certain things. Strategy classes are meant to encapsulate a certain part of the process (like connecting to a device, running a workload on it, comparing the results, etc), and are thus kept small. They all have a factory class, base class and `Jax` and `Torch` subclasses. Here is the new folder structure for infra:
```
tests/infra
|-- __init__.py
|-- adapters
|   |-- __init__.py
|   |-- framework_adapter.py
|   |-- framework_adapter_factory.py
|   |-- jax_adapter.py
|   `-- torch_adapter.py
|-- comparators
|   |-- __init__.py
|   |-- comparator.py
|   |-- comparator_factory.py
|   |-- comparison_config.py
|   |-- jax_comparator.py
|   `-- torch_comparator.py
|-- connectors
|   |-- __init__.py
|   |-- device_connector.py
|   |-- device_connector_factory.py
|   |-- jax_device_connector.py
|   `-- torch_device_connector.py
|-- runners
|   |-- __init__.py
|   |-- device_runner.py
|   |-- device_runner_factory.py
|   |-- jax_device_runner.py
|   |-- torch_device_runner.py
|   `-- utils.py
|-- testers
|   |-- __init__.py
|   |-- base_tester.py
|   |-- graph_tester.py
|   |-- model_tester.py
|   |-- multichip_tester.py
|   `-- op_tester.py
`-- utilities
    |-- __init__.py
    |-- multichip_utils.py
    |-- types.py
    |-- utils.py
    `-- workloads
        |-- __init__.py
        |-- jax_workload.py
        |-- torch_workload.py
        |-- workload.py
        `-- workload_factory.py
```
- `adapters` take care of configuring model for inference or training and compiling it. The main difference encapsulated here is how the compilation is done. 
- `comparators` encapsulate differences in comparing tensors in torch and jax (simply put, in case of jax tensors, we have to run `jax` commands, in case of torch tensors we have to run `torch` commands, so this class captures that in order not to have a bunch of dual util functions like `compare_pcc_torch` and `compare_pcc_jax`)
- `connectors` take care of connecting to TT device and CPU. `DeviceConnector` class is devised as thread-safe singleton with simple public `connect*` methods. All the complexity of registering PJRT plugin and implementing the `connect*` methods is left to subclasses.
  - One benefit of this refactoring is that I got rid of global `device_connector` instance
- `runners` encapsulate running something on some device. That "something" I call `Workload` (it itself is encapsulated in `tests/infra/utilities/workloads`). Runners hide away the complexity of sending workload to device and executing it there. Take note that no support for multichip has been added to torch runner.
- `testers` remain more or less the same, they are just grouped together and refactored to accommodate new way of doing things.
  - However, due to the inheritance chain `BaseTester -> OpTester -> GraphTester` and `BaseTester -> ModelTester` the initialization of testers has certain steps to follow in order to trigger PJRT plugin registration at the right time. All of that is hidden neatly in those classes.
  - Strategy factories require explicit `Framework` to be passed to them in order to know which strategy to create
  - `OpTester` and `GraphTester` have need added `framework` parameter which needs to be passed explicitly when calling `run_op_test...` or `run_graph_test...`. It is set to `JAX` by default, since in foreseeable future we won't be adding torch op and graph tests, just model
  - `ModelTester` automatically detects in which framework the model it tests was written, so no new parameters there. How user writes Torch tests remains exactly the same as it was with Jax tests
- Changed `requirements.txt` to install very specific version of `torch`, `torch_xla` and `torchvision`. If we let `pip` install default versions, compatibility issues arise. These versions that i set are not that old, we won't need to update them for a while.
- Added two examples in `examples/torch` on how to register our PJRT plugin and write a simple torch model and run it on our device

### Checklist


tests/infra/comparators/torch_comparator.py
```diff
@@ -0,0 +1,112 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+from jax.tree import map as tree_map
+from jaxtyping import PyTree
+from runners.utils import run_on_cpu
+from torch.utils._pytree import tree_flatten, tree_map
+from utilities.types import Framework, PyTree
```
mrakitaTT, line 10: You are also including JAX PyTree on line 7, and using JAX utility functions for pytrees below in comparison functions. I'm not sure how this works, do JAX functions work with PyTorch pytrees? Do PyTorch models even return pytrees?

tests/infra/adapters/framework_adapter.py
```diff
@@ -0,0 +1,48 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from abc import ABC, abstractmethod
+
+from utilities.types import Model
+from utilities.workloads import Workload
+
+
+class FrameworkAdapter(ABC):
```
mrakitaTT, line 11: I think we should create `JaxModelTester` and `TorchModelTester`, instead of having these adaptor classes, because all they currently have are things related to models, and I don't think we will be able to generilise model testers that easily so it would be better to have separate classes for them and have more flexibility. Especially for multichip model testers. Let's chat more tomorrow.

examples/jax/simple_regression.py
mrakitaTT, line 1: Move `alexnet_multichip.py` here too

tests/infra/connectors/jax_device_connector.py
```diff
@@ -0,0 +1,87 @@
...
+                # Check if the wheel-installed jax plugin exists.
+                plugin_spec = importlib.util.find_spec("jax_plugins.pjrt_plugin_tt")
+
+                if plugin_spec is not None:
+                    # Wheel-installed plugin is present, it will self-register on demand.
+                    return
+
+            # No wheel plugin found, fall back to local build.
+            # Check if plugin is already loaded, otherwise double registration error
+            # willoccur.
```
mrakitaTT, line 65: will occur*

mrakitaTT:
Btw please update PR description before merge to reflect new changes

mrakitaTT (approved):

tests/torch/single_chip/models/mnist/cnn/dropout/test_mnist_cnn_dropout.py
```diff
@@ -0,0 +1,73 @@
...
+@pytest.mark.model_test
+@pytest.mark.record_test_properties(
+    category=Category.MODEL_TEST,
+    model_name=MODEL_NAME,
+    model_group=ModelGroup.GENERALITY,
+    run_mode=RunMode.INFERENCE,
+    bringup_status=BringupStatus.PASSED,
+)
+@pytest.mark.xfail(
+    reason=failed_runtime("Runtime error: Unsupported PJRT buffer data type")
```
mrakitaTT, line 57: `failed_fe_compilation`

tests/torch/single_chip/models/mnist/cnn/nodropout/test_mnist_cnn_nodropout.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@pytest.mark.push
+@pytest.mark.model_test
+@pytest.mark.record_test_properties(
+    category=Category.MODEL_TEST,
+    model_name=MODEL_NAME,
+    model_group=ModelGroup.GENERALITY,
+    run_mode=RunMode.INFERENCE,
+    bringup_status=BringupStatus.PASSED,
```
mrakitaTT, line 54: `FAILED_TTMLIR_COMPILATION`

PR #583, opened 2025-05-12
Add Mixtral-8x7b tensor parallel implementation

Alexa2706:
# Draft PR: Tensor Parallelized Mixtral-8x7b Model in JAX

## Current Status
This is a draft PR to get early feedback on my approach to implementing tensor parallelism for Mixtral-8x7b. The implementation is not yet complete, and I'm seeking guidance on the architectural decisions before proceeding with full implementation. The single device model should be working, even though I've only tested it with a maximum of 3 hidden layers due to hardware limitations. The model is a vanilla implementation with SDPA attention and caching.

## Proposed Implementation
I'm planning to implement tensor parallelism using the following approach:
- Focus on tensor parallelism in the Attention and MLP components
- For the MLP part, I'm considering either: (1) assigning 1 device per expert, or (2) tensor-parallelizing the weights across devices
- Target mesh shape: 1x8 
- The implementation follows Mixtral's 8x7b architecture with caching mechanism and prepare_input adapted from Mistral's Flax implementation

## Key Technical Questions
1. Which approach is preferred for parallelizing the MLP experts?
2. Are there specific optimizations recommended for the 1x8 mesh topology?
3. A key challenge I'm facing is with the Hugging Face weights conversion - I need guidance on how to properly shard these weights when verifying model accuracy. This is one of the main reasons for this early PR.

## Next Steps
After receiving feedback, I plan to:
1. Finalize the implementation based on guidance
2. Complete performance testing across target mesh shapes 
3. Verify GSM8K score parity with single-device implementation
4. Complete documentation


staylorTT:
The team is taking a look and will provide feedback soon. Thank you @Alexa2706 for the contribution.

mrakitaTT:
Hi @Alexa2706, thank you for working on this contribution!

I will be reviewing your changes and providing feedback over the coming days, but let me now first answer your questions to unblock you.

### 1. Which approach is preferred for parallelizing the MLP experts?

Let's review some different strategies here.

This is what they said in the Mixtral [paper](https://arxiv.org/pdf/2401.04088):
>MoE layers can be run efficiently on single GPUs with high performance specialized kernels. For
example, Megablocks [13] casts the feed-forward network (FFN) operations of the MoE layer as large
sparse matrix multiplications, significantly enhancing the execution speed and naturally handling
cases where different experts get a variable number of tokens assigned to them. Moreover, the
MoE layer can be distributed to multiple GPUs through standard Model Parallelism techniques, and
through a particular kind of partitioning strategy called Expert Parallelism (EP) [28]. During the MoE
layer’s execution, tokens meant to be processed by a specific expert are routed to the corresponding
GPU for processing, and the expert’s output is returned to the original token location. Note that EP
introduces challenges in load balancing, as it is essential to distribute the workload evenly across the
GPUs to prevent overloading individual GPUs or hitting computational bottlenecks.

#### Sparse approaches

Megablocks uses custom GPU kernels to enable efficient sparse matrix multiplications, so that is not something that is useful to us and I would skip all approaches that utilize sparsity.

#### Model Parallelism vs Expert Parallelism

[Here](https://nvidia.github.io/TensorRT-LLM/advanced/expert-parallelism.html#tensor-parallel-vs-expert-parallel) is a good visual explaining the difference between the two. In practice Expert Parallelism works better than Model Parallelism which is why you will find that most production platforms utilize it (or some combinations of it together with data and model parallelism, if you have a larger mesh than the number of devices), and is why most of the scientific literature is focused on optimizing the Expert Parallelism.

[Here](https://docs.nvidia.com/megatron-core/developer-guide/latest/api-guide/moe.html#performance-best-practice) is some more insight from Nvidia:
>Prefer EP over TP for the expert layer when possible:
>  - TP saves more memory than EP, but EP can achieve better GEMM efficiency and less communication overhead than TP.
>  - In practice, EP8TP1 is better than EP4TP2 for 8x7B.

To give you some explanation for these insights: If you visualize how you would split the tokens and experts across devices, you will find that in TP you have the 1/N tokens and N * 1/N expert tensors on each device, so you do N computations of 1/N tokens with 1/N expert tensor, resulting in N times lower peak memory than in EP where you would do 1 computation of 1/N tokens with 1 whole expert tensor. Unfortunately that also leads to having to do N sequential computations, which is slower in practice than one larger computation because of dispatching overheads. Additionaly it leads to having to merge the result of N computations before doing the all-gather across devices, leading to some more overhead.

The main challenge with EP, as mentioned in quote from the Mixtral paper above, is in load balancing tokens across experts, and there are many papers exploring this topic. [Here](https://developer.nvidia.com/blog/applying-mixture-of-experts-in-llm-architectures/#experimental_results) you can find some visuals from Nvidia showing this:
>These experiments show that despite the load-balancing algorithm, there could still be large distributional imbalances, potentially affecting inference inefficiency, as some experts finish their work early while others are overloaded. This is an interesting area of active research.

Since this is a pretty complicated topic and active area of research, we don't expect you to implement any load balancing technique or any of the advanced EP techniques for this bounty. Feel free to just implement the simplest EP from the Mixtral paper.

One thing to note, for EP to be efficient, is that all other layers should work in data parallel mode, and then you do all-gather before/after the expert layers. We don't require you to implement this optimization, but if you are interested to try it would be really great. Here is a good visual of how that works (from [this](https://dl.acm.org/doi/pdf/10.1145/3503221.3508418) paper):
Screenshot 2025-05-16 at 00 33 27

One more thing to note, we expected (but unfortunately missed to mention in the bounty description) the model to be sharded using [manual parallelism](https://docs.jax.dev/en/latest/notebooks/shard_map.html) in JAX, because that gives you much greater flexibility to shard the parts of the models exactly as you need. [Here](https://github.com/tenstorrent/tt-xla/blob/6c97e1bbe6317e7c0506df535cc3d0923a636c95/tests/jax/multi_chip/n300/models/tensor_parallel/alexnet/model_implementation.py) you can see our example implementation of AlexNet model where CNN layers are using data parallelism, then we all-gather and do tensor parallelism in MLP layers. That is something very similar to what you need here. [Here](https://github.com/tenstorrent/tt-xla/blob/6c97e1bbe6317e7c0506df535cc3d0923a636c95/examples/alexnet_multichip.py) is how that model is compiled with the `shard_map`.

Additionally, manual parallelism (using `shard_map`) is the only supported mode for our compiler currently, because it is the only mode from JAX to get all CCL ops emitted in the graph. JAX automatic parallelism (i.e. "let the compiler take the wheel") is using the XLA compiler under the hood to figure out the CCLs, but we do not use their compiler. We are fully aware that we haven't properly communicated this and I will help you to convert your model implementation to be using the `shard_map`, simply in order to be able to test this model on our stack.

### 2. Are there specific optimizations recommended for the 1x8 mesh topology?

Not really since in the paper they use exactly 8 experts, so every device will have one expert and it is pretty straight forward. In case of 2x8 mesh for example you would be able to mix Model Parallelism and Data Parallelism strategies with Expert Parallelism. Also, to quote Nvidia again:
>In practice, EP8TP1 is better than EP4TP2 for 8x7B.

Btw, it would be great if your implementation could handle meshes with less than 8 devices, for example 1x4 mesh where two experts are residing on each device.

### 3. A key challenge I'm facing is with the Hugging Face weights conversion - I need guidance on how to properly shard these weights when verifying model accuracy. This is one of the main reasons for this early PR.

Could you provide some more details about your struggles there? In case you want to verify your implementation correctness, you don't need to load the HF weights, you can just initialize all weights with ones and then compare the outputs with single-device mesh `(1,)`, should get the same results. Then when you've verified that your implementation is working correctly, you shard the HF weights so that each expert layer ends up on its own device, and all other weights are replicated. Let me know with more details so I can help.

Alexa2706:
Thanks for the feedback  @mrakitaTT !
> One thing to note, for EP to be efficient, is that all other layers should work in data parallel mode, and then you do all-gather before/after the expert layers. We don't require you to implement this optimization, but if you are interested to try it would be really great. Here is a good visual of how that works (from [this](https://dl.acm.org/doi/pdf/10.1145/3503221.3508418) paper):

Since the bounty said not to implement DP should I just try to do TP on attention and just do EP on MLP, both with manual parallelism (trying both 1x4 and 1x8 meshes)?

mrakitaTT:
>Since the bounty said not to implement DP should I just try to do TP on attention and just do EP on MLP, both with manual parallelism (trying both 1x4 and 1x8 meshes)?

The main reason we emphasized in bounty description that TP is required insted of DP, is because it is pretty easy to parallelise models in fully DP mode, you just split your data on the batch axis and replicate the single-chip model to each device and then run each batch part independently, so we were interested to see solutions utilizing TP in layers which would benefit from it. That doesn't prohibit using DP in other layers, the only requirement is to have TP (or EP in this case) in some layers at least. We definitely have to re-word our bounty descriptions better :)

The ideal solution we would like to see is DP + EP, where you shard data on the batch axis and run all layers in DP mode, except the MLP experts which you run in EP mode. If you also implement some kind of TP for attention blocks, for example EP by putting each attention head to separate device, that is a big plus, but not a requirement of course.

tests/jax/models/mixtral/multichip/multichipmixtral.py
```diff
@@ -0,0 +1,618 @@
...
+        routing_weights = jax.nn.softmax(router_logits, axis = 1)
+        routing_weights, selected_experts = lax.top_k(routing_weights, self.top_k)
+        routing_weights /= jnp.sum(routing_weights, axis = -1, keepdims = True)
+
+        routing_weights = routing_weights.astype(hidden_states.dtype)
+        final_hidden_states = jnp.zeros(
+            (batch_size * seq_len, hid_dim), dtype=hidden_states.dtype)
+
+        for expert_idx in range(self.num_experts):
+            expert_layer = self._get_expert(expert_idx)
```
mrakitaTT, line 105: Unfortunately I think that this will work in parallel fashion only in eager mode, with `jax.jit` it will just unroll these calls in sequential manner :( And even in eager mode it depends on the backend, if the backend supports asynchronous execute.

In order to invoke experts in parallel you would have to run this with `pmap`. Unfortunately `pmap` and `shard_map` do not combine. I will try to investigate how other JAX MoE implementations do this, I invite you to do so too, but AFAIK all of the open sourced ones are single chip implementations...

Now I'm thinking on going back on my previous advice of implementing EP, and simply just doing TP here where you shard each expert across all devices and execute them in a sequential manner utilizing all devices for each. What do you think?

tests/jax/models/mixtral/test/multi_vs_single.py
```diff
@@ -0,0 +1,210 @@
...
+        outputs = model(
+            input_ids=x,
+            attention_mask=mask,
+            past_key_values=cache,
+            position_ids=pos,
+            init_cache=True,
+        )
+        return outputs.logits, outputs.past_key_values
+    
+    sharded_forward = shard_map(
```
mrakitaTT, line 118: I think you will have to use [nnx.shard_map](https://flax.readthedocs.io/en/latest/api_reference/flax.nnx/transforms.html#flax.nnx.shard_map) in order to provide sharding specs for the model parameters. You can get those specs by wrapping your layer parameters with [nnx.with_partitioning](https://flax.readthedocs.io/en/latest/guides/flax_gspmd.html#define-a-model-with-specified-sharding) and then using [nnx.get_partition_spec](https://flax.readthedocs.io/en/latest/api_reference/flax.nnx/spmd.html#flax.nnx.get_partition_spec) to get the specs from the model, which you pass into `nnx.shard_map`.

I did something similar with Flax Linen [here](https://github.com/tenstorrent/tt-xla/blob/main/examples/alexnet_multichip.py#L104).

tests/jax/models/mixtral/test/multi_vs_single.py
```diff
@@ -0,0 +1,210 @@
...
+        forward_pass,
+        device_mesh,
+        in_specs=(inputs_spec, P("X", None), cache_specs, P("X", None)), 
+        out_specs=(P("X", None, None), cache_specs),  # logits and updated cache
+        check_rep=False,
+    )
+    
+    jit_forward = jax.jit(sharded_forward, donate_argnums=(2, ))
```
mrakitaTT, line 126: Will probably also need to use [nnx.jit](https://flax.readthedocs.io/en/latest/api_reference/flax.nnx/transforms.html#flax.nnx.jit)

tests/jax/models/mixtral/test/multi_vs_single.py
```diff
@@ -0,0 +1,210 @@
...
+
+    out = model.generate(
+        input_ids = input_data,
+        attention_mask = extended_attention_mask,
+        max_new_tokens = max_len - seq_len 
+    )
+    return out
+
+
+def run_multi_chip(input_data, attention_mask, max_len):
```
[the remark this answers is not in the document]
Alexa2706: So then I will be jitting the whole ```generate``` function? I was afraid of getting OOM issues with that?

tests/jax/models/mixtral/test/hf_vs_single.py
```diff
@@ -0,0 +1,101 @@
...
+    
+    batch_size, seq_len = input_ids.shape
+    input_ids = jnp.array(input_ids)
+    attention_mask = jnp.array(attention_mask)
+    flax_model = make_model(config, model)
+    
+    for i in range(config.num_hidden_layers):
+        flax_model.model.layers[i].attn.cached_key = jnp.zeros((batch_size, max_len, 8, 128), dtype = jnp.float32)
+        flax_model.model.layers[i].attn.cached_value = jnp.zeros((batch_size, max_len, 8, 128), dtype = jnp.float32)
+        flax_model.model.layers[i].attn.cache_index = jnp.array(0, dtype = jnp.int32)
```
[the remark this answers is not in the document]
Alexa2706: That's true but I am using ```make_model```  with the HF model and in ``` multi_vs_single.py``` both models are implemented in jax so there is no need for that function, should i then rename the function or maybe change its purpose?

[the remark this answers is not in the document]
Alexa2706: I made a new function called ```prepare_inputs_for_generation``` inside ```flaxmixtral.py``` i think this will be the best solution!

tests/jax/multi_chip/bounties/mixtral_8x7b/multichip/multichipmixtral.py
```diff
@@ -0,0 +1,862 @@
...
+            outputs = sharded_model(
+                input_ids,
+                attention_mask,
+                position_ids,
+                past_key_values,
+            )
+            outputs = outputs.raw_value
+            return outputs.logits, outputs.past_key_values
+
+        sharded_forward = nnx.shard_map(
```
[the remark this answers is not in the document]
mrakitaTT: Mayor difference when using `shard_map` is that functions inside it need to work with dimensions of one shard, i.e. you have to assume that you are working with a chunk of the tensor that one device will see, which is why for example I had to scale the number of neurons in linear layers by the number of devices [here](https://github.com/tenstorrent/tt-xla/blob/main/tests/jax/multi_chip/n300/models/tensor_parallel/alexnet/model_implementation.py#L116). The fact that you didn't have to do this and multi_vs_single test still works for you adds to my suspicion above that the model doesn't actually get sharded. As mentioned in another comment above, could you please try to print the shardings during the runtime and verify this?

tests/jax/multi_chip/bounties/mixtral_8x7b/multichip/multichipmixtral.py
```diff
@@ -0,0 +1,862 @@
...
+            attentions=all_self_attns,
+            router_logits=all_router_logits,
+            past_key_values=past_key_values,
+        )
+
+
+class FlaxMixtralForCausalLM(nnx.Module):
+    """Mixtral model with a language modeling head implemented in NNX."""
+
+    def __init__(self, config, dtype=jnp.float32, param_dtype=jnp.float32, rngs=None):
```
mrakitaTT, line 604: This will initialize model with random parameters. Would be also cool if you could add functionality to load the multichip model from HF parameters with shardings applied. You can find example [here](https://github.com/tenstorrent/tt-xla/blob/main/tests/jax/multi_chip/bounties/mistral_small/mistral_nnx/model.py#L736).

tests/jax/multi_chip/bounties/mixtral_8x7b/multichip/multichipmixtral.py
```diff
@@ -0,0 +1,935 @@
...
+        return FlaxMoeCausalLMOutput(
+            logits=logits,
+            hidden_states=outputs.hidden_states,
+            attentions=outputs.attentions,
+            router_logits=outputs.router_logits,
+            past_key_values=past_key_values,
+        )
+
+    @staticmethod
+    def load_hf_params(state, pt_model, config) -> nnx.state:
```
mrakitaTT, line 725: You have similar code in `convert_weights.py`, maybe you could make it so both single chip and multi chip models can reuse it, instead of having duplicated logic in two places.

mrakitaTT (approved):
Model implementation and parallelisation LGTM, thank you for also adding loading of HF weights for multichip model and corresponding tests!

@tapspatel once you run the tests on your side and verify we have everything we need  then this can be merged, otherwise please work with @Alexa2706 to cover what's missing.

PR #620, opened 2025-05-22
Change tensor memory output kinds to 'tt_device'

ajakovljevicTT:
Currently, we have set all output memory kinds to be 'tt_host', as we put the output tensors on host after the execution. However, jax expects the output buffer memory kind to be 'device' if the buffer is sharded, leading to a mismatch and fails in the jax library. This PR fixes that.


src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -337,7 +337,8 @@ void LoadedExecutableInstance::fillPJRTOutputLists(
       std::unique_ptr<BufferInstance> output_buffer =
           BufferInstance::createOutputBufferInstance(
               output_tensor, std::move(output_shape),
-              m_addressable_devices[device_index], m_host_memory);
```
mrakitaTT, line 340: `m_host_memory` doesn't seem needed anymore, can be removed from the class and constructor

PR #625, opened 2025-05-23
Mistral small implementation for testing tensor parallelism

yiding:
### Ticket
https://github.com/tenstorrent/tt-forge/issues/56

### Problem description
Add a Tensor Parallel JAX Mistral-small Model to TT-xla Model Demos

### What's changed
This pr adds a bespoke implementation of mistral inference code using Flax NNX.

Tensor parallelism is achieved by adding logical axis annotations to the weight initializers, which are then applied as sharding constraints to the loaded parameter arrays. This should serve to constrain jax jit to generate tensor parallel code.

### Checklist

### DIFF
--- a/tests/jax/multi_chip/bounties/mistral_small/README.md
+++ b/tests/jax/multi_chip/bounties/mistral_small/README.md
@@ -1,0 +1,61 @@
+mistral_nnx is a Flax NNX implementation of Mistral models.
+
+Copied from https://github.com/yiding/mistral-nnx
+
+## Support
+
+The model has been tested with mistral-small and ministral-8b. Note that
+Ministral-8b uses sliding window, which is not supported, so long sequences may
+not behave correctly.
+
+### Supported Features
+
+Attention: Grouped Query Attention (GQA), but implementation should also support MHA.
+
+Sliding window: not supported.
+
+KV Cache: Very rudimentary KV Cache, supports batch-size 1.
+
+## Usage
+
+### Config
+
+The model uses configuration from the model repo from Huggingface Hub.
+
+### Weights
+
+Model supports loading the safetensor weights from huggingface hub. Weights
+loaded this way need to be transposed or reshaped before usage. There's
+additional code to save the transformed weights into an orbax checkpoint, which
+loads much faster.
+
+### Inference
+
+MistralModel supports basic forward pass using `MistralModel.__call__()`, which
+takes an array of a batch of tokens `(batch_size, seq_len)`.
+
+It also supports KV cache with batch-size 1 using `MistralModel.decode()`.
+`generate.py` contains code to setup and run inference with kv cache.
+
+### Parallelism
+
+The model parameter initializers has been annotated with Flax [local axis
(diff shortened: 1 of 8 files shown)


yiding:
### Question regarding MultichipModelTester

I haven't attempted to hook this up with the MultichipModelTester test harness as I'm not sure if it supports executing in cpu-only mode without tt hardware (which I don't have). I also ran into an issue attempting to load the xla plugin with a FATAL in native code (possibly because I don't have any attached TT hardware). 

MultichipModelTester also doesn't seem to have support for NNX models, I can wrap the model in a linen wrapper, or add support for NNX models to MultichipModelTester, provided there's a way for me to actually run and test it. And of course, if it's in scope of the original bounty task.


### Alternative approaches (namely, using HF Flax mistral model)

The huggingface transformers flax mistral model doesn't support Grouped Query Attention (used by mistral-small). The change to support this is not difficult. Sharding annotations can also be applied to the hf flax model without major code changes. The model loading code that exist in the HF Flax model is a bit slower. Finally, the code seems to have received little attention since initial implementation a year ago.

I found the code review process for changes to the code a bit lackluster though, so I don't know how long it would take to get changes pushed into the transformers library.

mrakitaTT:
Hi @yiding, thank you for working on this contribution!

>Question regarding MultichipModelTester

No need to hook up your model with our test infra, we will do that once your PR is merged. It is enough to test your implementation on CPU using your own test scripts, and we will also run them on our side once your PR is ready to verify that the produced SHLO matches expected multichip behavior. We usually advise people to compare their multichip implementation with the singlechip implementation that they either find online or implement themselves, which you already did by comparing with the HF implementation, so cudos :)

>Alternative approaches (namely, using HF Flax mistral model)

Yup, we've been told that HF transformers library FLAX models are on life support and they also don't accept any non-trivial contributions anymore, so this is definitely not the path we want to take. In case when your changes could be elegantly applied on top of their implementation by extending their classes and overriding them where needed, we could merge those changes in our repo. But that's not necessary since you've already implemented everything yourself, and as you say your implementation is more complete/performant.

yiding:
I ran this on kaggle w/ TPUv3-8 and it seems to be sharding the weights and the KV cache properly (though the one-token-at-a-time generation performance was pretty abysmal @ around 10t/s, I didn't try very hard to optimize this). Forward pass does 2000 t/s using batch size 4. Details in the [example notebook](https://www.kaggle.com/code/yidingjia/mistral-nnx-test)

Model is saved using orbax, hf tokenizer and model config are reused. There's a script for generating this from hf, or model can just be loaded from hf directly.
 
Pre-converted model: https://www.kaggle.com/models/yidingjia/mistral-small-24b/

The code here is copied over from my repo that packages the code as a proper python package @ https://github.com/yiding/mistral-nnx, not sure if you want to take on the external-repo dependency or not, so I just copied relevant files and a few of the tests over.

mrakitaTT:
Btw please make sure to run the [pre-commit](https://github.com/tenstorrent/tt-xla/blob/main/docs/src/getting_started.md#pre-commit) on this branch just to make sure it doesn't fail in CI.

yiding:
pre-commit checks pass.

PR #630, opened 2025-05-26
Add a graph test for gelu

sgligorijevicTT:
### What's changed
I had to write this test anyway so I might as well commit it

### Checklist

### DIFF
--- a/tests/jax/single_chip/graphs/test_gelu.py
+++ b/tests/jax/single_chip/graphs/test_gelu.py
@@ -1,0 +1,29 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import jax
+import pytest
+from infra import run_graph_test_with_random_inputs
+from tests.utils import incorrect_result
+from tests.utils import Category
+
+import flax.linen as nn
+
+
+@pytest.mark.push
+@pytest.mark.nightly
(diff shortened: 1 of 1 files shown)


tests/jax/single_chip/graphs/test_gelu.py
```diff
@@ -0,0 +1,30 @@
...
+@pytest.mark.record_test_properties(category=Category.GRAPH_TEST)
+@pytest.mark.parametrize(
+    "x_shape",
+    [
+        (32, 32),
+        (64, 64),
+    ],
+)
+@pytest.mark.parametrize("approximate", [False, True])
+@pytest.mark.xfail(reason=incorrect_result("Bad accuracy"))
```
mrakitaTT, line 25: Interesting that they both fail with bad accuracy, does `where` appear in approximate version too?

mrakitaTT (approved):

tests/jax/single_chip/graphs/test_gelu.py
sgligorijevicTT: `where` does not appear in that case. In this case I suspect the problem lies with `tanh`, as it is known to be inaccurate. I already have a MLIR PR that fixes tanh as it is also blocking albert, https://github.com/tenstorrent/tt-mlir/pull/3546 . Do note that approximate=True case has much milder inaccuracy(only fails allclose instead of outputting infs or nans), I should have separated the fail messages for those two cases. I will do that before merging.

kmitrovicTT (approved):

Issue #598, opened 2025-05-14
Add Support for Logistic Operation


ctr-lelanchelianTT:
The Logistic operation is currently not supported in the TT-XLA stack. We aim to implement a test for the logistic operation

**Operation Semantics:**
`
result[i] = 1 / (1 + exp(-input[i]))
`

**Where:**

- `input`: The tensor containing values to apply the logistic (sigmoid) function on.
- `result`: A tensor with the same shape as `input`, where each element is the result of applying the logistic function.

**Attributes**

No additional attributes are required for this operation.

**Inputs**

- `input`: A tensor of type `float16`, `bfloat16`, `float32`, or `float64`.

**Outputs**

- `result`: A tensor of the same shape and type as `input`, containing the elementwise logistic function values.

**Reference:**
https://openxla.org/stablehlo/spec#logistic

PR #637, opened 2025-05-28
add test for logistic op

ctr-lelanchelianTT:
### Ticket
closes https://github.com/tenstorrent/tt-xla/issues/598

### Problem description
Add test for logistic operations

### What's changed

Modified the test files to test the `jax.lax.logistic`

### Checklist

I have attached the logs below
[test_logistic.log](https://github.com/user-attachments/files/20476387/test_logistic.log)

### DIFF
--- a/tests/jax/single_chip/ops/test_logistic.py
+++ b/tests/jax/single_chip/ops/test_logistic.py
@@ -2,4 +2,25 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-# TODO add tests for `stablehlo.logistic`.
+import jax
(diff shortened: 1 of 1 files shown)


mrakitaTT:
@ctr-lelanchelianTT looking at the generated shlo graph, it seems that JAX is decomposing `jax.lax.logistic` into `division(1, addition(1, exp(-x)))` ops instead of lowering to `stablehlo.logistic` op directly. Also, in the stablehlo spec it says that `stablehlo.logistic` op is going to be deprecated and is unused by frameworks. So I think we should remove the test for this op completely.

PR #641, opened 2025-05-28
Added some more model info in model tests reports

kmitrovicTT:
### Problem description
We needed some more info about model under test in `tags` dict in pytest report in our reporting infra to display in superset charts.

### What's changed
- [x] Created new util data structure `ModelInfo` which wraps all important model test properties
- [x] Updated `conftest.py` to parse passed model info properly
- [ ] Updated all model tests to use new way of reporting

Here are differences between new and old report

image

### Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,6 +4,8 @@
 
 import pytest
 
+from .utils import ModelInfo
+
 
 def pytest_configure(config: pytest.Config):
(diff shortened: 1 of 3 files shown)


kmitrovicTT:
Putting this on hold since there are ongoing discussions on unifying reporting for all frontends.

PR #645, opened 2025-05-29
Release artifact generation and workflow speed up

vvukomanTT:
### Ticket
/

### Problem description
Tt-xla didn't have a setup for releasing a wheel, since the setup was created recently we want to generate it in the workflow.

### What's changed
`on pr`, `on push` and `on nightly` workflows now generate a `xla-whl-release` artifact.

Additionaly: 
In the `on pr`, `on push` and `on nightly` workflows, tt-xla is now built only once and then reused in all the test runs.
Workflows are now faster because of it.

### Checklist
- [ ] On PR workflow is passing

#### Future improvement
This PR addresses build artifact reuse and wheel generation.
There will be a PR for a cleaner refactor of the `on nightly` workflow with some other improvements.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -85,7 +85,6 @@ jobs:
 
   build-ttxla:
     needs: build-image
-    if: ${{ inputs.test_mark == 'push' || inputs.test_mark == 'nightly' }}
     timeout-minutes: 120
     runs-on: ubuntu-latest
     name: "Build tt-xla"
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -82,11 +82,76 @@ jobs:
...
+
+    # Build project
+    - name: Run tt-xla build
+      uses: ./.github/actions/build-tt-xla-action
+      with:
+        build-output-dir: ${{ steps.strings.outputs.build-output-dir }}
+        work-dir: ${{ steps.strings.outputs.work-dir }}
+        enable_artifact_upload: ${{ inputs.test_mark == 'push' || inputs.test_mark == 'nightly' }}
+
+    - name: Verify tt-mlir SHA override
```
nsumrakTT, line 133: Shouldn't override happen before build? (I see it was not before, but still...)

nsumrakTT (approved):

vvukomanTT:
> @vvukomanTT have you tried running tt-xla tests after installing wheel inside repo? Does this PR do exactly that?

Good question. It does actually, all of the tests in the `on-pr`, `on-push` and `on-nigthly` workflows are ran against the wheel.

PR #675, opened 2025-06-05
Add a Tensor Parallel JAX Gemma3-27b Model to TT-xla Model Demos

lanchongyizu:
### Ticket
https://github.com/tenstorrent/tt-forge/issues/57

### Problem description

### What's changed

### Checklist
- [x] Add an open source single-device model implementation
- [ ] Add extension to the model implementation to support Tensor Parallelism
- [ ] Verify that single-device version works on [GSM8K](https://github.com/openai/grade-school-math)
- [ ] Verify that tensor parallel version works on [GSM8K](https://github.com/openai/grade-school-math)

### DIFF
--- a/tests/jax/multi_chip/bounties/gemma3/gemma/__init__.py
+++ b/tests/jax/multi_chip/bounties/gemma3/gemma/__init__.py
@@ -1,0 +1,27 @@
+# Copyright 2025 DeepMind Technologies Limited.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""gemma API."""
+
+# A new PyPI release will be pushed every time `__version__` is increased.
+# When changing this, also update the CHANGELOG.md.
+__version__ = '3.0.2'
+
+
+def __getattr__(name: str):  # pylint: disable=invalid-name
+  """Catches `import gemma as gm` errors."""
+  del name
+  raise AttributeError(
+      'Please use "from gemma import gm", NOT "import gemma as gm".'
+  )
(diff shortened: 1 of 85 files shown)


lanchongyizu:
@mrakitaTT OK, I will move the gemma folder into the third_party directory. I was occupied with other tasks over the past few days, but I will now resume work on this issue. Thanks.

cguerreroTT:
Hi @lanchongyizu, just checking in since we haven't heard from you in a few weeks.  Please provide a quick status update if you're still working on this.

tapspatel (approved):
Fantastic PR. Clean code and nicely organized with good comments.

I verified both the HF and jax model implementations, and verified the execution of both single chip and multi-chip model variants.

```
python3 test/hf_vs_single.py
python3 test/hf_vs_multi.py 
```

Couple of comments (by no means blocking PR): 
- the current implementation does not use jax.jit directly(although the multi-device version uses nnx.shard_map variation)
- you can extend this model into a 2x4 config with batch of 2 and tensor parallel of 4 to model our llmbox configuration (8 chips in 2x4 mesh). As stated above, calling jit compile will allow compiler to better automatically solve the graph and insert collectives/reshards for you if needed.
- the current implementation uses 41 gb of ram for single chip and 70 gb of ram for multi-chip
- by default, jax uses GSPMD style partitioning (string based IR's) but you can use shardy partitioner instead. We support both today in tenstorrent compiler but jax is also moving to shardy as default in short term.
- you can't jit __call__ functions so some tweaks of the model will need to be made in order to support this (example, you can use lambda and pass in "self" instead).

```
jax.config.update("jax_use_shardy_partitioner", True)
```

I've pasted the shardy graphs down below for our team's reference. If you want a quick and dirty way to get the stablehlo graph, you can do something like this
```
for _ in range(max_new_tokens):
            # Get logits and updated cache
            fwd_jit = jax.jit(
                lambda next_token, cache: self(
                    input_ids=next_token,
                    cache=cache,
                    use_cache=True,
                    deterministic=True,
                )
            )

            fwd_lowered = fwd_jit.lower(next_token, cache)
            fwd_stablehlo = fwd_lowered.compiler_ir(dialect='stablehlo')
            print(fwd_stablehlo.operation.get_asm(large_elements_limit=0))
```

[single_chip.txt](https://github.com/user-attachments/files/22913581/single_chip.txt)


I verified the graphs in our tt-mlir backend compiler and there should be nothing blocking or ops remaining to be supported to run this e2e.

mrakitaTT (approved):

PR #694, opened 2025-06-16
Added Xfail marker to Bloom and Vit models

ctr-lelanchelianTT:
This PR temporarily sets pccconfig to false in order to investigate the cause of test failures observed in the CI/CD pipeline for the Bloom and ViT models.


mrakitaTT (approved):
@ctr-lelanchelianTT Thank you for fixing this, could you please just change the title of this PR so we don't get confusing message in commit history? :)

PR #704, opened 2025-06-17
Update fail reasons

sgligorijevicTT:
### What's changed
Albert xlarge is now passing, so marked it as such.
Resnet models are no longer crashing so I unskipped them.

### Checklist


sgligorijevicTT:
Nightly passed

mrakitaTT:
Interesting that only certain variants of resnet are having a bad accuracy

PR #723, opened 2025-06-25
Adding infrastructure for ingesting compile_options through a flatbuffer

ajakovljevicTT:
We can pass `compile_options` to the `jax.jit()` function as a python dictiionary, which will get propagated through a protobuffer to PJRT. For pecific meaning of those values in the protobuf we would need the schema from the xla repository. However, we can easily see what the custom dictionary values are, and use those in our compiler.


inc/common/pjrt_implementation/client_instance.h
```diff
@@ -71,8 +74,14 @@ class ClientInstance {
...
 
   // Compiles given mlir program.
-  tt_pjrt_status compileMlirProgram(const PJRT_Program *mlir_program,
-                                    LoadedExecutableInstance **out_executable);
+  tt_pjrt_status compileMlirProgram(
+      const PJRT_Program *mlir_program,
+      LoadedExecutableInstance **out_executable,
+      const std::unordered_map<std::string, std::string> &compile_options);
+
+  static std::optional<std::unordered_map<std::string, std::string>>
```
mrakitaTT, line 82: Add header comment

inc/common/pjrt_implementation/client_instance.h
```diff
@@ -122,6 +131,10 @@ class ClientInstance {
   // TODO: Remove once tt-mlir supports passing the system descriptor object to
   // TTIR to TTNN backend pipeline.
   std::string m_cached_system_descriptor_path;
+
+  static std::unordered_map<std::string, std::string>
+  ExtractCustomProtobufFields(
```
mrakitaTT, line 136: Should start with lowercase. Also, add header comment

mrakitaTT (approved):

PR #740, opened 2025-06-30
XFail Albert xlarge to unblock nightly

sgligorijevicTT:
### Ticket
proper fix tracked in #739

### Problem description
Albert xlarge PCC regressed from >0.99 to 0.98 something

### What's changed
Marked the model in question as xfail. Proper investigation pending.

### Checklist


sgligorijevicTT:
Per offline discussion, we are going to xfail this model instead

Issue #751, opened 2025-07-02
Gather Op test lacks test cases


sdjukicTT:
The Gather Op test isn't versatile because the default min and max values for generating random inputs are 0 and 1. When the indices input is cast to an integer, it becomes a tensor of zeros.

PR #752, opened 2025-07-02
Fix Gather Op Test

sdjukicTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/751

### Problem description
Gather Op test wasn't versatile.

### What's changed
Added shapes that appear in our models and increased the maximum value for test inputs, so that the indices tensor doesn't become all zeros when cast to integers.

### Checklist

### DIFF
--- a/tests/jax/single_chip/ops/test_gather.py
+++ b/tests/jax/single_chip/ops/test_gather.py
@@ -10,6 +10,16 @@ from infra import run_op_test_with_random_inputs
 from utils import Category
 
 
+def conditionally_skip(start_index_map: tuple):
+    """
+    Helper function which checks test input combinations and xfails if necessary.
+
+    Extracted here in order not to pollute the test function.
+    """
+    if len(start_index_map) > 1:
+        pytest.xfail("Test cases that index multiple dimensions need integer inputs")
+
+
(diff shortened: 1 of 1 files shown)


tests/jax/single_chip/ops/test_gather.py
```diff
@@ -18,30 +18,43 @@
...
+    "data_shape, indices_shape, offset_dims, start_index_map, slice_sizes",
     [
-        ((32, 32), (16, 1)),
-        ((64, 64), (32, 1)),
+        ((2048, 1, 200), (1, 6, 1), (2, 3), (0,), (1, 1, 200)),
+        ((32000, 3200), (1, 6, 1), (2,), (0,), (1, 3200)),
+        ((8, 26, 26, 192), (1,), (0, 1, 2), (3,), (8, 26, 26, 1)),
+        ((8, 54, 54, 64), (1,), (0, 1, 2), (3,), (8, 54, 54, 1)),
+        # Fail because of ttnn.embedding bug:
+        # ((732,12),(3880,1),(1,),(0,),(1,12,)),
```
sgligorijevicTT, line 28: We probably want to mark these as xfail instead

PR #806, opened 2025-07-10
FIx the test for squeezebert

sgligorijevicTT:
### Problem description
A prior refactor seems to have broken the test for Squeezebert

### What's changed
Fixed the test and marked the model as passing.

### Checklist


tests/jax/single_chip/models/squeezebert/test_squeezebert.py
```diff
@@ -64,6 +64,9 @@ def _get_input_parameters(self) -> PyTree:
 
         return self._model.init_from_pytorch_statedict(state_dict, dtype=jnp.bfloat16)
 
+    def _get_forward_method_args(self) -> list:
```
sgligorijevicTT, line 67: I could have moved some stuff around to make the default behavior from the tester class work for us, but it was quicker to just override this method as empty to make existing code do what we want

PR #810, opened 2025-07-11
Add torch test for Xception model

ctr-pmuruganTT:
### Problem description
Add torch tests for Xception model from tt-forge-models repo

### Checklist

PFA logs for reference:

[xception.log](https://github.com/user-attachments/files/21179775/xception.log)


tests/torch/single_chip/models/xception/test_xception.py
```diff
@@ -0,0 +1,64 @@
...
+)
+
+from .tester import XceptionTester
+
+VARIANT_NAME = "xception"
+
+MODEL_NAME = build_model_name(
+    Framework.TORCH,
+    "xception",
+    "base",
```
sgligorijevicTT, line 24: I can see that in tt-forge-models the variant is listed as base, but the code actually fetches xception65, can you please change the variant here to `65`?

sgligorijevicTT (approved):

PR #826, opened 2025-07-14
Enable ttmlir runtime debugging

### LINKED ISSUE #827: Enable ttmlir runtime debugging
tt-mlir uses `TT_RUNTIME_DEBUG` flag to enable runtime logging. Add this flag with tt-mlir build.

tt-mlir runtime debugging can be enable using
```
cmake -G Ninja -B build -DTT_RUNTIME_DEBUG=ON
```
tt-mlir runtime debug supports following debug levels.
```
TRACE, DEBUG, INFO, WARNING, ERROR, FATAL
```
and can be set as
```
export TTMLIR_RUNTIME_LOGGER_LEVEL=DEBUG
```

mmanzoorTT:
### Ticket
closes #827 

### Problem description
Add support for ttmlir runtime debugging/logging

### What's changed
Update CMakeLists.txt to include `TT_RUNTIME_DEBUG` flag with tt-mlir build.

### Checklist

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -29,6 +29,8 @@ else()
     set(TTMLIR_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/tt-mlir")
     set(TTMLIR_BUILD_DIR "${TTMLIR_SOURCE_DIR}/build")
     set(TTMLIR_LIB_DIR ${TTMLIR_SOURCE_DIR}/install)
+    set(TT_RUNTIME_DEBUG OFF CACHE BOOL "Enable ttmlir runtime debugging")
+    message(STATUS "ttmlir runtime debugging is set to: ${TT_RUNTIME_DEBUG}")
 
     ExternalProject_Add(
         tt-mlir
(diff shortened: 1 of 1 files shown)


mmanzoorTT:
We can also automatically enable ttmlir runtime debugging/logging with Debug and RelWithDebInfo build types. Thoughts?

third_party/CMakeLists.txt
```diff
@@ -29,6 +29,8 @@ else()
     set(TTMLIR_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/tt-mlir")
     set(TTMLIR_BUILD_DIR "${TTMLIR_SOURCE_DIR}/build")
     set(TTMLIR_LIB_DIR ${TTMLIR_SOURCE_DIR}/install)
+    set(TT_RUNTIME_DEBUG OFF CACHE BOOL "Enable ttmlir runtime debugging")
+    message(STATUS "ttmlir runtime debugging is set to: ${TT_RUNTIME_DEBUG}")
```
mrakitaTT, line 33: Since we are not logging values for other build options, I am wondering if there is a value in logging this one either, because it will be quite obvious during runtime if `TT_RUNTIME_DEBUG` was `ON` :) What do you think?

mrakitaTT (approved):

mrakitaTT:
>We can also automatically enable ttmlir runtime debugging/logging with Debug and RelWithDebInfo build types. Thoughts?

I wouldn't do that by default for any tt-xla build type because those runtime debug logs per op are pretty noisy and would hinder debugging of PJRT outside of runtime. On the other hand, for tt-mlir Debug build type it makes more sense, and when tt-mlir is built in Debug mode it will automatically enable `TT_RUNTIME_DEBUG` anyways, so we should provide option in tt-xla to configure tt-mlir in Debug mode.

third_party/CMakeLists.txt
mmanzoorTT: This option is added to pass through `tt-torch` build option. `tt-torch` uses runtime debug logs to verify intermediate results for op-by-op flow (if enabled).

PR #847, opened 2025-07-17
Add torch tests for autoencoder linear

ctr-pmuruganTT:
### Problem description
Add torch tests for `autoencoder linear` model from tt-forge-models  repo

### Checklist

PFA logs for reference:

[autoencoder_linear.log](https://github.com/user-attachments/files/21282975/autoencoder_linear.log)


requirements.txt
```diff
@@ -11,6 +11,7 @@ jaxtyping
 librosa
 lit
 loguru
+matplotlib
```
mrakitaTT, line 14: These tt-forge-models dependencies are blowing up our requirements. Can you please change the model loader in tt-forge-models to use `pillow` to post-process the image instead of matplotlib? Then you will be able to remove this dependency. Every model loader there uses different image processing library 😅 We should converge on a single one.

ctr-pmuruganTT: Modified the model loader in tt-forge-model to use `pillow` and removed the matplotlib dependency

PR #869, opened 2025-07-21
Changing serialization to flatbuffer

ajakovljevicTT:
In line with the needs of `tt-forge`, changing the python serialization function to output the flatbuffer instead of ttir code.


sdjukicTT (approved):
LGTM

PR #925, opened 2025-08-01
Dump MLIR mechanism in ModuleBuilder - tt-xla

ddilbazTT:
### Ticket
None

### Problem description

This is a proof of concept. 

We want a way to dump MLIR from models into a file, which is already supported in tt-torch through torch-mlir. For torch-xla path, I think it's a little more complicated. In this implementation, I am using environment variables to control 
- if and at which stage we will dump MLIR files (DUMP_MLIR)
- model name to save the file for (MODEL_NAME)

This might be a hacky way, but it allows us to be able to save MLIR files through tt-torch as well. Here is a demo of tt-torch: https://github.com/tenstorrent/tt-torch/compare/main...ddilbaz/dumpMLIR Simply, tt-torch will be setting these environment variables through the existing ModelTester/ CompilerConfig framework.

In the eager mode tests though, model name and dump mlir need to be explicitly set.

I tested that this works through tt-torch. Using tt-torch environment variables as we used to is enough. 

### What's changed
- Added MLIR dumping functions to module builder

### Checklist

### DIFF
--- a/src/common/module_builder.cc
+++ b/src/common/module_builder.cc
@@ -7,6 +7,8 @@
 
 // c++ standard library includes
 #include <cstdlib>
+#include <filesystem>
+#include <fstream>
 #include <numeric>
 #include <optional>
 
@@ -16,6 +18,7 @@
 // llvm includes
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/LogicalResult.h"
+#include "llvm/Support/raw_ostream.h"
 
 // llvm mlir includes
 #include "mlir/Dialect/Arith/IR/Arith.h"
@@ -89,6 +92,9 @@ ModuleBuilder::ModuleBuilder()
   // https://github.com/tenstorrent/tt-xla/issues/355
   m_context->allowUnregisteredDialects();
   m_context->appendDialectRegistry(registry);
+
+  m_model_name = getModelName();
+  InitializeMLIRFolder();
 }
 
 tt_pjrt_status ModuleBuilder::buildModule(
@@ -187,6 +193,7 @@ void ModuleBuilder::convertFromVHLOToSHLO(
 
   DLOG_F(LOG_DEBUG, "SHLO Module:");
   printModule(mlir_module);
+  dumpModuleToFile(mlir_module, ModuleType::STABLEHLO);
 }
 
 void ModuleBuilder::collectInputShardings(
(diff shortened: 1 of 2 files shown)


src/common/module_builder.cc
```diff
@@ -709,4 +718,115 @@ std::optional<mlir::sdy::MeshOp> ModuleBuilder::getFirstShardyMeshOp(
...
+      } else if (c == '_' || c == '-') {
+        // Keep underscores and hyphens as they're common in model names
+        cleaned_name += c;
+      }
+    }
+    return cleaned_name;
+  }
+  LOG_F(INFO,
+        "MODEL_NAME environment variable is not set. Using default name.");
+  return "default_model"; // Provide a default name if the env var isn't set
```
mmanzoorTT, line 743: LLVM/MLIR coding style: Use early exits to simplify the code https://llvm.org/docs/CodingStandards.html#id40
```suggestion
  if (!model_name_env) {
    LOG_F(INFO,
        "MODEL_NAME environment variable is not set. Using default name.");
    return "default_model"; // Provide a default name if the env var isn't set
  }

  LOG_F(INFO, "Model name from environment: %s", model_name_env);
  std::string model_name = std::string(model_name_env);

  std::string cleaned_name;
  cleaned_name.reserve(std::string(model_name).size());
  for (char c : model_name) {
    if (std::isalnum(c)) {
      // Keep alphanumeric characters, convert to lowercase
      cleaned_name += std::tolower(c);
    } else if (c == '_' || c == '-') {
      // Keep underscores and hyphens as they're common in model names
      cleaned_name += c;
    }
  }
  return cleaned_name;
```

src/common/module_builder.cc
```diff
@@ -709,4 +718,115 @@ std::optional<mlir::sdy::MeshOp> ModuleBuilder::getFirstShardyMeshOp(
...
+}
+
+void ModuleBuilder::InitializeMLIRFolder() {
+  std::filesystem::path mlir_folder_path =
+      std::filesystem::current_path() / "model_mlir";
```
hshahTT, line 748: I think this should be a class-level constant instead (something like `kMlirDumpBaseDir`).

Then you can change L812 within this file from:
```C++
std::filesystem::path file_path =
      std::filesystem::current_path() / "model_mlir" / file_name;
```
to
```C++
std::filesystem::path file_path = kMlirDumpBaseDir / file_name;
```

Also, does `std::filesystem::current_path()` change based on where you run the base tt-xla program from? If so, we shouldn't use it, and instead should hard code a path relative to the base tt-xla folder instead

hshahTT (changes requested):

Issue #941, opened 2025-08-06
Resnet produces maxpool_2d which is not supported in tt-metal with optimizer turned on


ajakovljevicTT:
When turning on the optimizer through tt-xla, the `tests/jax/single_chip/models/resnet_v1_5/resnet_26/test_resnet_26.py` model fails with `Only BFLOAT16 supported for now` then running maxpoll_2d op. Needs further investigation on why it happens. CC: @odjuricicTT

PR #942, opened 2025-08-06
Adding optimizer enabling through compile options

ajakovljevicTT:
Adding a flag to enable optimizer through compile options in case of jax models. Set up `ResNetTester` with optimizer as an example, however it currently fails in runtime - https://github.com/tenstorrent/tt-xla/issues/941


src/common/module_builder.cc
```diff
@@ -102,6 +102,8 @@ tt_pjrt_status ModuleBuilder::buildModule(
 
   m_status = tt_pjrt_status::kSuccess;
 
+  collectIsOptimized(compile_options);
```
mrakitaTT, line 105: Let's create a struct `CompileOptions` which will contain fields for all compile options we will use, for now we will have only `optimizer_enabled` field. You can add this struct in `module_builder.h` for now, I am working in another PR to move compiler code into separate folder so I will then create a separate file for compile options and move it there.

Then you can have a member variable `m_compile_options` in `ModuleBuilder`, you can rename this function to `parseCompileOptions` which will populate the `m_compile_options`. Member variable `m_use_optimizer` should be removed, and `m_compile_options` used instead when populating `TTIRToTTNNBackendPipelineOptions` below.

src/common/module_builder.cc
```diff
@@ -189,6 +191,15 @@ void ModuleBuilder::convertFromVHLOToSHLO(
   printModule(mlir_module);
 }
 
+void ModuleBuilder::collectIsOptimized(
+    const std::unordered_map<std::string, std::string> &compile_options) {
+  if (compile_options.find("optimize") != compile_options.end()) {
+    m_use_optimizer = (compile_options.at("optimize") == "True");
+  } else {
+    m_use_optimizer = false;
+  }
```
mrakitaTT, line 200: Every bool compile option will probably use this code, so we can add a static function in `CompileOptions` called `static bool parseBoolOption(const std::unordered_map<std::string, std::string> &compile_options, const std::string& option_name)` and move this code there.

tests/infra/testers/single_chip/model/jax_model_tester.py
```diff
@@ -183,21 +185,25 @@ def _configure_model_for_training(model: Model) -> None:
         model.train()
 
     # @override
-    def _compile(self, workload: Workload) -> Workload:
+    def _compile(self, workload: Workload, on_tt_device: bool) -> Workload:
```
mrakitaTT, line 188: This function now doesn't make sense since it has a different logic per flag so it's not anymore a common logic for both compiles, so I would remove it and just implement `_compile_for_cpu` and `_compile_for_tt_device` differently.

tests/infra/testers/single_chip/model/jax_model_tester.py
```diff
@@ -185,25 +185,22 @@ def _configure_model_for_training(model: Model) -> None:
         model.train()
 
     # @override
-    def _compile(self, workload: Workload, on_tt_device: bool) -> Workload:
+    def _compile_for_tt_device(self, workload: Workload) -> Workload:
```
mrakitaTT, line 188: Switch the order of `_compile_for_tt_device` and `_compile_for_cpu` definitions, because in that order they are called from `_test_inference` function in model tester.

src/common/module_builder.h
```diff
@@ -32,6 +32,16 @@
 
 namespace tt::pjrt {
 
+// POD struct all compile options our plugin recognizes.
```
mrakitaTT, line 35: I'd change to: "POD struct containing various options used to customize module compilation."

src/common/module_builder.h
```diff
@@ -32,6 +32,16 @@
...
+struct CompileOptions {
+  // Enables the ttmlir optimizer,
+  // ie the optimization passes and memory layout analysis.
```
mrakitaTT, line 38: Premature line break, also `i.e.`

src/common/module_builder.h
```diff
@@ -32,6 +32,16 @@
...
+  bool use_optimizer = false;
```
mrakitaTT, line 39: I like `enable_optimizer` more than `use_optimizer`

src/common/module_builder.cc
```diff
@@ -65,6 +65,28 @@ namespace tt::pjrt {
 
 const std::string ModuleBuilder::c_mlir_format_name = "mlir";
 
+bool parseBoolOption(
```
mrakitaTT, line 68: I like the changes you've made, but since now `CompileOptions` has more code in it, I would move it right away into separate files `compile_options.h/cpp`. Add them to `src/common` similar to `module_builder` and I will move it in my following change into separate directory.

This function should be in the `internal` namespace, declared in .h file and defined in .cpp file, so we can access it from unit tests later.

src/common/module_builder.h
sgligorijevicTT: I like that phrasing better, I'll change it.

mrakitaTT (approved):
Please resolve conflicts and push new iteration

third_party/CMakeLists.txt
```diff
@@ -58,6 +58,7 @@ else()
           -DTTMLIR_ENABLE_OPMODEL=${TTMLIR_ENABLE_OPMODEL}
           -DCMAKE_INSTALL_PREFIX=${TTMLIR_LIB_DIR}
           -DTT_RUNTIME_ENABLE_PERF_TRACE=${TTMLIR_ENABLE_PERF_TRACE}
+          -DTTMLIR_ENABLE_OPMODEL=ON
```
pilkicTT, line 61: should we remove the existing cmake option `TTMLIR_ENABLE_OPMODEL`?

sgligorijevicTT: Good catch, yeah we should

PR #944, opened 2025-08-06
Fixed large model tests and reduce_scatter tests

ajakovljevicTT:
Now, with a workaround for 2D reduce_scatter being merged into tt-mlir main (https://github.com/tenstorrent/tt-mlir/pull/4109), we can enable the corresponding tests. Additionally, due to the low disk space of the shared runners, the nightly tests were flaky, so dividing the large model tests into more runners to ease the memory load.


sgligorijevicTT (approved):

.github/workflows/test.yml
```diff
@@ -28,10 +28,10 @@ on:
           [
             { "runs-on": "n150",        "name": "run_jax",               "dir": "./tests/jax/single_chip",   "extra_test_marks": "not large" },
             { "runs-on": "n150",        "name": "run_torch",             "dir": "./tests/torch/single_chip"},
-            { "runs-on": "n150",        "name": "run_large_jax_models",  "dir": "./tests/jax/single_chip",   "extra_test_marks": "large", "use-shared-runners": "true" },
+            { "runs-on": "n150",        "name": "run_large_jax_models",  "dir": "./tests/jax/single_chip",   "extra_test_marks": "large", "use-shared-runners": "true", "test_group_cnt": 5 },
```
nsumrakTT, line 31: "test_group_cnt": 5 this will split tests into 5 groups and execute only 1st one.
Is this what you wanted to achieve?

.github/workflows/on-nightly.yml
```diff
@@ -101,6 +102,26 @@ jobs:
...
+                "name": "run_torch",
+                "dir": "./tests/torch/single_chip",
+                "use-shared-runners": "true"
+              },
+              {
+                "runs-on": "p150b",
+                "name": "run_large_jax_models",
+                "dir": "./tests/jax/single_chip",
+                "extra_test_marks": "large",
+                "use-shared-runners": "true"
```
vvukomanTT, line 124: suggestion: Addition of not used object fields

Any field other than `runs-on` and `name` does nothing in this context, so they should be omitted (I know that they were set for some objects already but let's please fix that as well).

vvukomanTT (approved):
Great job, thanks 👍

kmabeeTT:
Should this have been updated? 

```
default_test_group_cnt=3
```

Nightly run via cron doesn't take inputs, so the change from default 4 to 5 groups doesn't have an effect on cron'd nightly runs (it used 3).

vvukomanTT:
> Should this have been updated?
> 
> ```
> default_test_group_cnt=3
> ```
> 
> Nightly run via cron doesn't take inputs, so the change from default 4 to 5 groups doesn't have an effect on cron'd nightly runs (it used 3).

I presume this PR was for introducing parallelization to test suites that didn't have it before, not necessarily making the already parallelized test suites execute on more machines.

sgligorijevicTT:
> > Should this have been updated?
> > ```
> > default_test_group_cnt=3
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Nightly run via cron doesn't take inputs, so the change from default 4 to 5 groups doesn't have an effect on cron'd nightly runs (it used 3).
> 
> I presume this PR was for introducing parallelization to test suites that didn't have it before, not necessarily making the already parallelized test suites execute on more machines.

I think it was supposed to split up the already parallelized tests into more groups, therefore I think we should have updated that param

PR #947, opened 2025-08-06
Serialize IRs alongside flatbuffers

### LINKED ISSUE #932: Add mlir modules serialization to tt-xla
Currently, we only support serialisation of the flatbuffer through or `ttxla_tools` API. This is not enough information for the tt-forge users, and as such we need to include serialisation of mlir modules in the StableHLO, TTIR and TTNN dialects. For this, we need create a custom structure (preferably using protobuf), which we can then embed in `SerializedExecutableInstance` for serialization. On the tt_xla side, we can set multiple functions which will then parse this custom structures and output the corresponding mlir code or flatbuffer into a file.

sgligorijevicTT:
### Ticket
closes #932 

### Problem description
Perf team needs TTIR and TTNN MLIR code to be serialized alongside flatbuffer binaries.

### What's changed
The compile logic was changed to store MLIR code as necessary and pass it on, all the way to SerializedExecutableInstance class.
Serialization format was changed from just the fb to a custom format that incorporates the needed IRs alongside the fb.
Python side helper function was refactored to work with the new requirements.

### Checklist

### DIFF
--- a/inc/common/pjrt_implementation/serialized_executable_instance.h
+++ b/inc/common/pjrt_implementation/serialized_executable_instance.h
@@ -8,11 +8,11 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 // https://llvm.org/LICENSE.txt
 
-// PJRT C API includes
-#include "xla/pjrt/c/pjrt_c_api.h"
-
 // c++ standard library includes
 #include <cstddef>
+
+// PJRT C API includes
+#include "xla/pjrt/c/pjrt_c_api.h"
 
 // tt-mlir includes
 #include "tt/runtime/runtime.h"
--- a/src/common/pjrt_implementation/module_builder/module_builder.cc
+++ b/src/common/pjrt_implementation/module_builder/module_builder.cc
@@ -495,6 +495,7 @@ void ModuleBuilder::convertFromSHLOToTTIR(
   }
 
   m_ttir_mlir = getMlirCode(mlir_module);
+
   DLOG_F(LOG_DEBUG, "TTIR Module:");
   printModule(mlir_module);
 }
@@ -616,6 +617,7 @@ void ModuleBuilder::convertFromTTIRToTTNN(
   }
 
   m_ttnn_mlir = getMlirCode(mlir_module);
+
   DLOG_F(LOG_DEBUG, "TTNN Module:");
   printModule(mlir_module);
 }


src/common/pjrt_implementation/client_instance.cc
```diff
@@ -184,8 +184,10 @@ tt_pjrt_status ClientInstance::compileMlirProgram(
   std::shared_ptr<ExecutableImage> executable_image =
       ExecutableImage::createInstance(
           m_module_builder->getFlatbufferBinary(),
-          std::move(original_mlir_code), std::move(executable_name),
-          m_module_builder->getNumPartitions(),
+          std::move(original_mlir_code),
+          std::move(m_module_builder->getTTIRMlirCode()),
```
sgligorijevicTT, line 188: One questionable design decision I made: I made the string getters return a non const reference and move from that reference, leaving the ModuleBuilder with empty strings after constructing an ExecutableImage.
Arguably the interface of buildModule could be refactored to return something like std::expected or StatusOr from compile, instead of returning just the status code and storing results as it's state.

examples/pytorch/serialization_example.py
```diff
@@ -0,0 +1,79 @@
...
+from ttxla_tools.torch import parse_from_cache_to_disk
+
+# --------------------------------
+# Plugin registration
+# --------------------------------
+os.environ["PJRT_DEVICE"] = "TT"
+os.environ["XLA_STABLEHLO_COMPILE"] = "1"
+
+
+def find_plugin_path():
```
ajakovljevicTT, line 22: I think we don't need this anymore (https://github.com/tenstorrent/tt-xla/pull/1008)

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -144,6 +146,8 @@ tt_pjrt_status ModuleBuilder::buildModule(
     return m_status;
   }
 
+  m_ttnn_mlir = collectMlirModule(mlir_module);
```
ajakovljevicTT, line 149: Should this be done twice?

src/common/pjrt_implementation/serialized_executable_instance.cc
```diff
@@ -13,6 +13,7 @@
 // tt-xla includes
 #include "common/pjrt_implementation/buffer_instance.h"
 #include "common/pjrt_implementation/error_instance.h"
+#include <string>
```
ajakovljevicTT, line 16: add:
`// c++ standard library includes` and a newline

src/common/pjrt_implementation/serialized_executable_instance.cc
```diff
@@ -30,7 +31,62 @@ SerializedExecutableInstance::SerializedExecutableInstance(
...
+
+  m_payload.resize(total_size);
+  std::byte *data_ptr = m_payload.data();
+  // Write magic string.
+  std::string magic_string = "TTSERv00";
+  std::memcpy(data_ptr, magic_string.data(), magic_string.size());
+  data_ptr += magic_string.size();
+  // Offsets are calculated from the start of the data section,
+  // ie we expect TTIR offset to be 0 almost always.
+  uint64_t ttir_offset = 0;
```
ajakovljevicTT, line 59: You can define a class/struct `SerializationHeader`, which can be populated and then written directly.

src/common/pjrt_implementation/serialized_executable_instance.cc
```diff
@@ -30,7 +31,62 @@ SerializedExecutableInstance::SerializedExecutableInstance(
...
+  std::memcpy(data_ptr, &ttnn_offset, sizeof(ttnn_offset));
+  data_ptr += sizeof(ttnn_offset);
+  std::memcpy(data_ptr, &ttnn_size, sizeof(ttnn_size));
+  data_ptr += sizeof(ttnn_size);
+  std::memcpy(data_ptr, &fb_offset, sizeof(fb_offset));
+  data_ptr += sizeof(fb_offset);
+  std::memcpy(data_ptr, &fb_size, sizeof(fb_size));
+  data_ptr += sizeof(fb_size);
+  // Write TTIR code.
+  std::memcpy(data_ptr, ttir_code.data(), ttir_size);
```
ajakovljevicTT, line 79: Use some sort a helper function for this pattern:
```
    static void write_bytes_and_advance_pointer(std::byte*& dst, const void* src, size_t size) {
        std::memcpy(dst, src, size);
        dst += size;
    }
```

ttxla_tools/torch.py
```diff
@@ -0,0 +1,94 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+PyTorch-specific functions for ttxla_tools.
+"""
+
+import os
+from .ttxla_tools import parse_executable
```
ajakovljevicTT, line 10: nit: newline here

ttxla_tools/ttxla_tools.py
```diff
@@ -3,66 +3,51 @@
...
-        os.makedirs(dirname, exist_ok=True)
-    with open(binary_file_path, "wb") as f:
-        f.write(flatbuffer_binary)
+    Returns:
+        tuple: (ttir_mlir, ttnn_mlir, flatbuffer_binary)
+    """
+    executable_io = io.BytesIO(executable)
+
+    # Verify magic number
+    magic = executable_io.read(8)
```
ajakovljevicTT, line 29: use MAGIC_SIZE=8 instead of the pure constant

src/common/pjrt_implementation/client_instance.cc
ajakovljevicTT: Is there a specific reason for that decision? Is the copying of the strings too slow?

src/common/pjrt_implementation/executable_image.cc
```diff
@@ -13,12 +13,14 @@
 // tt-xla includes
 #include "common/pjrt_implementation/data_type_utils.h"
 #include "common/pjrt_implementation/memory_instance.h"
+#include <string>
```
ajakovljevicTT, line 16: Add newline and `// c++ standard library includes`

examples/pytorch/serialization_example.py
sgligorijevicTT: This was a leftover from rebasing, will fix

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -207,6 +211,15 @@ void ModuleBuilder::runFrontendSHLOPipeline(
   printModule(mlir_module);
 }
 
+std::string ModuleBuilder::collectMlirModule(
```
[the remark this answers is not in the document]
sgligorijevicTT: Duplicating a function just to change where the output is stored seems like an antipattern to me?

src/common/pjrt_implementation/serialized_executable_instance.cc
```diff
@@ -26,11 +29,72 @@ SerializedExecutableInstance::createInstance(
...
       executable_image->getFlatbufferBinary();
-  flatbuffer_binary.storeToMemory(m_serialized_flatbuffer);
+  const std::string &ttir_code = executable_image->getTTIRMlirCode();
+  const std::string &ttnn_code = executable_image->getTTNNMlirCode();
+
+  std::vector<std::byte> flatbuffer_data;
+  // TODO(stefan): We could avoid double copy if storeToMemory took a span.
+  flatbuffer_binary.storeToMemory(flatbuffer_data);
+
+  const size_t body_size =
```
ajakovljevicTT, line 81: Make the `SerializationHeader` here and put this into `get_total_size` function, which will take the header as input.

src/common/pjrt_implementation/serialized_executable_instance.cc
```diff
@@ -26,11 +29,72 @@ SerializedExecutableInstance::createInstance(
...
+std::byte *write_raw(std::byte *dst, const void *src, size_t size) {
+  std::memcpy(dst, src, size);
+  return dst + size;
+}
+
+template <typename T> std::byte *write(std::byte *dst, const T &src) {
+  return write_raw(dst, &src, sizeof(src));
+}
+
+struct SerializationHeader {
```
ajakovljevicTT, line 41: I think this should be in the .h file

ttxla_tools/ttxla_tools.py
```diff
@@ -3,66 +3,51 @@
 # SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
ajakovljevicTT: Oh I see, makes sense! Shouldn't then the jax.py and torch.py functions be in the `__init__.py `file of `ttxla_tools`?

src/common/pjrt_implementation/module_builder/module_builder.cc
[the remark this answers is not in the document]
ajakovljevicTT: On second thought, you are right, and I'm guessing some of your next refactors will make this point moot anyways. Could you just then change the name to `get_mlir_code()` since all of the other `collect...` functions have the different semantics. I think that name fits the function more as well, as we are returning a string object.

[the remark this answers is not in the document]
sgligorijevicTT: Works for me

ttxla_tools/ttxla_tools.py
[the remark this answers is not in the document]
sgligorijevicTT: I would rather avoid re-exporting framework specific functions into the common module because of possible name clashes. However I will re-export the jax and torch submodules from the `__init__.py`, as Claude reminded me of a gotcha with discoverability.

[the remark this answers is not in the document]
sgligorijevicTT: You might have overlooked that the function currently in ttxla_tools is intended to be public. I have one niche usecase* for it from torch. If we go with that restructure, I could reexport it in tt_torch for now and we can see later if something else needs changing. Otherwise, sounds reasonable.

*use being that if you run multiple graphs trough torchxla with persistent caching on, serialization code currently in tt_torch errors out because of multiple cache entires; in that case the users can implement their own logic to pick the right cache entry, load it, and pass it to `parse_executable`

inc/common/pjrt_implementation/serialized_executable_instance.h
```diff
@@ -18,12 +18,47 @@
...
+  SerializationHeader(const std::string &ttir_code,
+                      const std::string &ttnn_code,
+                      const std::vector<std::byte> &flatbuffer_data)
+      : ttir_offset(0), ttir_size(ttir_code.size()),
+        ttnn_offset(ttir_offset + ttir_size), ttnn_size(ttnn_code.size()),
+        fb_offset(ttnn_offset + ttnn_size), fb_size(flatbuffer_data.size()) {
+    std::memcpy(magic, "TTSERv00", sizeof(magic));
+  }
+
+  size_t getHeaderSize() const { return sizeof(magic) + 3 * sizeof(uint64_t); }
```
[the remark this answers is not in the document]
sgligorijevicTT: Right, this is confusing. I was using Header to refer to magic+sizes, but elsewhere I also referred to just the sizes section as header. I will clean this up.
And this function is indeed incorrect, a `* 2 *` got lost while extracting this into a separate function. Will fix.
As for offsets, I would prefer to keep them, as 1) they make extracting just one part easier 2) they offer added flexibility to the format. I can image a case where some third party wishes to extend this format in a fully backwards compatible fashion, with gaps being allowed between data subsections they can easily write a serializer that doesn't pack data contiguously and exploit the gaps to stuff their own data.

python_package/tt_jax/serialization.py
```diff
@@ -7,17 +7,20 @@
 from jax.experimental import serialize_executable
 import os
 import pickle
+from ttxla_tools import parse_executable
 
 
-def serialize_function_to_binary(func, binary_file_path, *args, **kwargs):
+def serialize_function(func, *args, **kwargs):
```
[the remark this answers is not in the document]
sgligorijevicTT: Those are better names and I don't have a better idea. I will adopt those exact names.

Issue #950, opened 2025-08-07
[JAX Model] Update Beit tests to use ModelLoader from tt-forge-models


ctr-lelanchelianTT:
JAX model for Beit is added in tt-forge-models repo.

XLA Tests forBeit has to be updated to use ModelLoader from which load_model & load_inputs function can be used

PR #951, opened 2025-08-07
update beit model to use model loader from tt-forge repoistory

ctr-lelanchelianTT:
### Ticket 
close https://github.com/tenstorrent/tt-xla/issues/950

### Problem description
Update JAX tests of Beit to use ModelLoader

### What's changed
2 variants of Beit is modified to use ModelLoader from tt-forge-models

- The Base & Large  variants are passing if we make `do_rescale=False` ,previously we was getting atol error
- If we not making  `do_rescale=False` we are getting very bad pcc
- I am using HF_dataset instead of random image,I tried using IMAGENET-1k but since its gated I am using cat-image dataset
- For now I am raising the PR with the default config of the model

### Checklist

I have attached the logs for your reference while keepint 'do_rescale=False':
[test_beit_base_without_xfail.log](https://github.com/user-attachments/files/21660786/test_beit_base_without_xfail.log)
[test_beit_large_without_xfail.log](https://github.com/user-attachments/files/21660788/test_beit_large_without_xfail.log)

I have attached the log with default config:
[test_beit_base_with_xfail.log](https://github.com/user-attachments/files/21677761/test_beit_base_with_xfail.log)
[test_beit_base_without_xfail.log](https://github.com/user-attachments/files/21677768/test_beit_base_without_xfail.log)
[test_beit_large_with_xfail.log](https://github.com/user-attachments/files/21677771/test_beit_large_with_xfail.log)
[test_beit_large_without_xfail.log](https://github.com/user-attachments/files/21677774/test_beit_large_without_xfail.log)


mrakitaTT:
`do_rescale=False` was necessary before because we were creating random input image which was already in range `[0, 1]`. Now you are loading image from HF datasets which will be in range `[0, 255]` so we need `do_rescale=True` which is default value of `do_rescale` parameter for `BeitImageProcessor`.

ctr-lelanchelianTT:
> `do_rescale=False` was necessary before because we were creating random input image which was already in range `[0, 1]`. Now you are loading image from HF datasets which will be in range `[0, 255]` so we need `do_rescale=True` which is default value of `do_rescale` parameter for `BeitImageProcessor`. Tests were already failing before with bad PCC, it was just the message that was outdated saying it fails with bad ATOL, now we don't assert on ATOL anymore so it fails on PCC metric.
> 
> Let's just update the xfail reasons and re-run on_pr CI.

Sure

ctr-lelanchelianTT:
> Seems like submodule update might be needed to find beit model in tt-forge-models

I am facing conflict while updating submodule

PR #954, opened 2025-08-08
DRAFT use tt-torch backend for torch tests

AleksKnezevic:
### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist

### DIFF
--- a/tests/infra/runners/torch_device_runner.py
+++ b/tests/infra/runners/torch_device_runner.py
@@ -24,12 +24,15 @@ class TorchDeviceRunner(DeviceRunner):
         self, workload: Workload, device_type: DeviceType, device_num: int = 0
     ) -> Tensor:
         device = self._device_connector.connect_device(device_type, device_num)
-        device_workload = self._put_on_device(workload, device=device)
+
+        # commenting this out as the "tt" torch.compile backend will place/remove weights/inputs
+        # from device automatically
+        # device_workload = self._put_on_device(workload, device=device)
 
         # TODO this context manager disables gradient calculation to save memory. We
         # will need to enable it for training.
         with torch.no_grad():
-            return device_workload.execute()
+            return workload.execute()
 
     # @override
     def _put_tensors_on_device(
--- a/tests/infra/testers/single_chip/model/torch_model_tester.py
+++ b/tests/infra/testers/single_chip/model/torch_model_tester.py
@@ -12,6 +12,9 @@ from infra.workloads import TorchWorkload, Workload, WorkloadFactory
 
 from .model_tester import ModelTester, RunMode
 
+# Registers PJRT plugin to torch-xla and registers "tt" backend for torch.compile
+import tt_torch
+
 
 class TorchModelTester(ModelTester):
     """
@@ -104,7 +107,7 @@ class TorchModelTester(ModelTester):
     # @override
     def _compile_for_tt_device(self, workload: Workload) -> Workload:
         """Compiles `workload` for TT device."""
-        return self._compile_for_backend(workload, backend="openxla")
+        return self._compile_for_backend(workload, backend="tt")
 
     def _compile_for_backend(self, workload: Workload, backend: str) -> Workload:
         """JIT-compiles model into optimized kernels."""
--- a/tests/torch/single_chip/models/resnet/test_resnet.py
+++ b/tests/torch/single_chip/models/resnet/test_resnet.py
@@ -52,12 +52,12 @@ def training_tester() -> ResnetTester:
     run_mode=RunMode.INFERENCE,
     bringup_status=BringupStatus.FAILED_TTMLIR_COMPILATION,
 )
-@pytest.mark.xfail(
-    reason=failed_ttmlir_compilation(
-        " Error: torch_xla/csrc/aten_xla_bridge.cpp:110 : Check failed: xtensor "
-        "https://github.com/tenstorrent/tt-xla/issues/795"
-    )
-)
+# @pytest.mark.xfail(
+#     reason=failed_ttmlir_compilation(
+#         " Error: torch_xla/csrc/aten_xla_bridge.cpp:110 : Check failed: xtensor "
+#         "https://github.com/tenstorrent/tt-xla/issues/795"
+#     )
+# )
 def test_torch_resnet_inference(inference_tester: ResnetTester):
     inference_tester.test()
 
--- a/tt_torch/__init__.py
+++ b/tt_torch/__init__.py
@@ -1,0 +1,62 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Process start method must be "forkserver" for op-by-op compilation for the following reasons:
+#   - The backend will hang if attempting to retrieve the device descriptor from a forked process if there are multiple chips
+#   - torch tensors cannot be made contiguous in a forked process
+import multiprocessing as mp
+import torch
+
+if mp.get_start_method() != "forkserver":
+    mp.set_start_method("forkserver", force=True)
+
+import os
+import importlib.util
+
+# find the tt-metal directory, it can either be in the venv if installed from a wheel or in the third_party source tree
+package_name = "tt-metal"
+spec = importlib.util.find_spec(package_name)
+if spec is not None:
+    tt_metal_home = os.path.abspath(spec.submodule_search_locations[0])
+    os.environ["TT_METAL_HOME"] = tt_metal_home
+
+# Import these modules so backends are registered ("tt", and "tt-experimental")
+import tt_torch.backend.backend
+
+from torch_xla.experimental import plugins
+
+
+class TTPjrtPlugin(plugins.DevicePlugin):
+    def library_path(self):
+        # This is where the pjrt plugin will be located if you've built and installed from source
+        direct_build_install_path = os.path.join(
+            os.path.dirname(__file__), "../install/lib/pjrt_plugin_tt.so"
+        )
+        if os.path.exists(direct_build_install_path):
+            return direct_build_install_path
+
+        # This is where the pjrt plugin will be located if you've installed the tt-torch wheel into a virtual environment
+        env_path = os.path.join(os.path.dirname(__file__), "../../../pjrt_plugin_tt.so")
+        if os.path.exists(env_path):
+            return env_path
+
+        # This is where the pjrt plugin will be located if you've only built and installed the wheel - but you're running your code with the root of the source tree (CI does this)
+        source_path = os.path.join(
+            os.path.dirname(__file__), "../env/venv/lib/pjrt_plugin_tt.so"
+        )
+        if os.path.exists(source_path):
+            return source_path
+
+        tt_xla_build_path = os.path.join(
+            os.path.dirname(__file__), "../build/src/tt/pjrt_plugin_tt.so"
+        )
+        if os.path.exists(tt_xla_build_path):
+            return tt_xla_build_path
+
+        assert False, "Could not find pjrt_plugin_tt.so"
+
+
+plugins.register_plugin("TT", TTPjrtPlugin())
+os.environ["XLA_STABLEHLO_COMPILE"] = "1"
+os.environ["PJRT_DEVICE"] = "TT"
--- a/tt_torch/backend/backend.py
+++ b/tt_torch/backend/backend.py
@@ -1,0 +1,189 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import torch
+import time
+import operator
+
+from .decompositions import (
+    CUSTOM_DECOMPOSITION_TABLE,
+)
+import os
+import tempfile
+import multiprocessing as mp
+import re
+import pickle
+import faulthandler
+import collections
+from .passes import (
+    bypass_redundant_getitem,
+    bypass_dtype_promotion,
+    bypass_redundant_cast,
+    rectify_buffer_inplace_copy,
+    run_shape_prop,
+    constant_fold,
+)
+
+from torch.export.graph_signature import InputKind
+from torch._dynamo import register_backend
+
+from tt_torch.tools.utils import (
+    CompilerConfig,
+    CompileDepth,
+    Op,
+    OpCompilationStatus,
+    calculate_atol,
+    calculate_pcc,
+)
+
+import torch_xla
+import torch_xla.core.xla_model as xm
+
+
+def bypass_assert_tensor_metadata(gm):
+    for node in gm.graph.nodes:
+        if (
+            node.op == "call_function"
+            and node.target == torch.ops.aten._assert_tensor_metadata.default
+        ):
+            gm.graph.erase_node(node)
+    return gm
+
+
+def xla_pass_pipeline(gm, example_inputs, compiler_config):
+    decompositions = torch._decomp.core_aten_decompositions()
+    decompositions.update(CUSTOM_DECOMPOSITION_TABLE)
+    compiled_graph = (
+        torch.export.export_for_training(gm, tuple(example_inputs), strict=False)
+        .run_decompositions(decompositions)
+        .module()
+    )
+
+    compiled_graph = bypass_dtype_promotion(compiled_graph, compiler_config)
+    run_shape_prop(compiled_graph, example_inputs)
+    compiled_graph = bypass_redundant_cast(compiled_graph)
+
+    if compiler_config.enable_consteval:
+        compiled_graph = constant_fold(compiled_graph)
+    elif compiler_config.consteval_parameters:
+        raise Exception("consteval_parameters is enabled but enable_consteval is not")
+
+    compiled_graph = bypass_redundant_getitem(compiled_graph)
+    compiled_graph = rectify_buffer_inplace_copy(compiled_graph)
+    compiled_graph = bypass_assert_tensor_metadata(compiled_graph)
+    program = torch.export.export(compiled_graph, tuple(example_inputs), strict=False)
+
+    return program
+
+
+class XLAExecutor:
+    def __init__(self, program, compiler_config):
+        self.program = program
+        self.compiler_config = compiler_config
+        self.arg_type_map_str = None
+
+        self.inputs = []
+        self.user_input_indices = []
+        for idx, input_spec in enumerate(self.program._graph_signature.input_specs):
+            if input_spec.kind == InputKind.USER_INPUT:
+                self.inputs.append(None)
+                self.user_input_indices.append(idx)
+            else:
+                self.inputs.append(self.program.state_dict[input_spec.target].to("xla"))
+
+    def push_tensors_to_device(self, inputs, device):
+        if hasattr(inputs, "to"):
+            if device not in [inputs.device, inputs.device.type]:
+                return inputs.to(device)
+            else:
+                return inputs
+        elif isinstance(
+            inputs, dict
+        ):  # transformers input/output objects are subclasses of dict, however we still wish to return the same wrapper object
+            return type(inputs)(
+                **{k: self.push_tensors_to_device(v, device) for k, v in inputs.items()}
+            )
+        elif isinstance(inputs, collections.abc.Sequence):
+            return type(inputs)(
+                [self.push_tensors_to_device(i, device) for i in inputs]
+            )
+        elif hasattr(inputs, "key_cache") or hasattr(inputs, "value_cache"):
+            if hasattr(inputs, "key_cache"):
+                inputs.key_cache = self.push_tensors_to_device(inputs.key_cache, device)
+            if hasattr(inputs, "value_cache"):
+                inputs.value_cache = self.push_tensors_to_device(
+                    inputs.value_cache, device
+                )
+            return inputs
+        else:
+            return inputs
+
+    def generate_arg_type_map_str(self, output_object):
+        hlo_input_ids, _ = torch_xla._XLAC._get_tensors_xla_device_data_node(
+            output_object
+        )
+
+        # xm.get_stablehlo(output_object) gives a graph with just as many inputs as in hlo_input_ids
+
+        hlo_input_positions = [id - min(hlo_input_ids) for id in hlo_input_ids]
+
+        def get_kind_str(kind):
+            if kind == InputKind.USER_INPUT:
+                return "input"
+            elif kind == InputKind.PARAMETER:
+                return "parameter"
+            else:
+                return "constant"
+
+        arg_types = []
+        output_args = [o.arg for o in self.program.graph_signature.output_specs]
+        for idx in range(len(hlo_input_positions)):
+            if hlo_input_positions[idx] < len(self.program.graph_signature.input_specs):
+                in_spec = self.program.graph_signature.input_specs[
+                    hlo_input_positions[idx]
+                ]
+
+                # If an input is passed right through to the output, it will not be
+                # captured as an argument
+                if in_spec.arg in output_args:
+                    continue
+
+                arg_types.append(get_kind_str(in_spec.kind))
+            else:
+                arg_types.append("constant")
+
+        self.arg_type_map_str = "main=" + ",".join(arg_types)
+
+    def __call__(self, *args):
+        args = self.push_tensors_to_device(args, "xla")
+        inputs = self.inputs
+        for idx in range(len(args)):
+            inputs[self.user_input_indices[idx]] = args[idx]
+
+        output = self.program.graph_module(*inputs)
+
+        if self.compiler_config.arg_type_map_override:
+            if self.arg_type_map_str is None:
+                self.generate_arg_type_map_str(output)
+            if os.environ.get("ARG_TYPE_MAP_OVERRIDE") != self.arg_type_map_str:
+                os.environ["ARG_TYPE_MAP_OVERRIDE"] = self.arg_type_map_str
+
+        xm.mark_step()
+        if self.compiler_config.push_outputs_to_cpu:
+            return self.push_tensors_to_device(output, "cpu")
+        return output
+
+    def __del__(self):
+        # Remove the arg type map override environment variable
+        os.environ.pop("ARG_TYPE_MAP_OVERRIDE", None)
+
+
+@register_backend(name="tt")
+def xla_backend(gm, example_inputs, options: CompilerConfig = None):
+    breakpoint()
+    compiler_config = options
+    if compiler_config is None:
+        compiler_config = CompilerConfig()
+
+    program = xla_pass_pipeline(gm, example_inputs, compiler_config)
+    return XLAExecutor(program, compiler_config)
--- a/tt_torch/backend/decompositions.py
+++ b/tt_torch/backend/decompositions.py
@@ -1,0 +1,441 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+from typing import Callable, Dict, List, Optional, Sequence, Union
+
+import contextlib
+import threading
+
+import torch
+from torch._decomp import get_decompositions, remove_decompositions
+
+DecompositionTable = Dict[torch._ops.OperatorBase, Callable]
+DecompositionOpsList = Sequence[
+    Union[torch._ops.OperatorBase, torch._ops.OpOverloadPacket]
+]
+
+# Manages "scopes" for decompositions used. Each unique scope is an attribute on
+# the _decomp_local. If the attribute is missing, then the default
+# decompositions are used. The scope "aot" is used for all AOT cases.
+_decomp_local = threading.local()
+
+
+def _get_decomp_stack(scope: str) -> List[DecompositionTable]:
+    try:
+        return getattr(_decomp_local, scope)
+    except AttributeError:
+        stack: List[DecompositionTable] = []
+        setattr(_decomp_local, scope, stack)
+        return stack
+
+
+def _current(scope: str) -> DecompositionTable:
+    """Gets the current decomposition table (which may be the default)."""
+    stack = _get_decomp_stack(scope)
+    if stack:
+        return dict(stack[-1])
+    else:
+        return dict(CUSTOM_DECOMPOSITION_TABLE)
+
+
+@contextlib.contextmanager
+def _extend_context_manager(
+    scope: str,
+    *,
+    from_current: bool = True,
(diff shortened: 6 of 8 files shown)


tt_torch/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+import torch
+
+if mp.get_start_method() != "forkserver":
+    mp.set_start_method("forkserver", force=True)
+
+import os
+import importlib.util
+
+# find the tt-metal directory, it can either be in the venv if installed from a wheel or in the third_party source tree
+package_name = "tt-metal"
```
AleksKnezevic, line 18: The loading of tt-metal and the .so paths probably need to change here, no?

tt_torch/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+            os.path.dirname(__file__), "../build/src/tt/pjrt_plugin_tt.so"
+        )
+        if os.path.exists(tt_xla_build_path):
+            return tt_xla_build_path
+
+        assert False, "Could not find pjrt_plugin_tt.so"
+
+
+plugins.register_plugin("TT", TTPjrtPlugin())
+os.environ["XLA_STABLEHLO_COMPILE"] = "1"
```
AleksKnezevic, line 61: Should these be in the activate script?

tt_torch/backend/backend.py
```diff
@@ -0,0 +1,189 @@
...
+                inputs.key_cache = self.push_tensors_to_device(inputs.key_cache, device)
+            if hasattr(inputs, "value_cache"):
+                inputs.value_cache = self.push_tensors_to_device(
+                    inputs.value_cache, device
+                )
+            return inputs
+        else:
+            return inputs
+
+    def generate_arg_type_map_str(self, output_object):
```
AleksKnezevic, line 121: This is no longer needed, right?

tt_torch/backend/backend.py
```diff
@@ -0,0 +1,189 @@
...
+        return output
+
+    def __del__(self):
+        # Remove the arg type map override environment variable
+        os.environ.pop("ARG_TYPE_MAP_OVERRIDE", None)
+
+
+@register_backend(name="tt")
+def xla_backend(gm, example_inputs, options: CompilerConfig = None):
+    breakpoint()
```
AleksKnezevic, line 183: remove.

tt_torch/backend/passes.py
```diff
@@ -0,0 +1,690 @@
...
+def constant_fold(gm):
+    gm = const_fold.split_const_subgraphs(gm)
+    gc.collect()
+    gm.run_folding()
+
+    gm.graph.eliminate_dead_code()
+    return gm
+
+
+def node_to_device(node, device_map, verbose=False):
```
AleksKnezevic, line 137: you can remove this for now, it can be re-introduced when we add the multi-chip graph.

tt_torch/tools/utils.py
```diff
@@ -0,0 +1,956 @@
...
+The CompileDepth's below represent the different stages of the compilation
+pipeline.
+
+tt-torch has two entrypoints it can compile from: PyTorch and ONNX. At the
+beginning of the compile flow these entrypoints follow different paths, but
+converge early in the compilation pipeline. The flow is as follows:
+
+                                 PyTorch nn.Module
+                                         |
+     ONNX ModelProto               Torch FX Graph
```
AleksKnezevic, line 30: This should be modified.

tt_torch/tools/utils.py
```diff
@@ -0,0 +1,956 @@
...
+class CompileDepth(Enum):
+    EXECUTE_OP_BY_OP = 5
+    EXECUTE = 6
+
+
+class OpCompilationStatus(IntEnum):
+    NOT_STARTED = 0
+    CREATED_GRAPH = 1
+    CONVERTED_TO_TORCH_IR = 2
+    CONVERTED_TO_TORCH_BACKEND_IR = 3
```
AleksKnezevic, line 54: we don't have all of these any more, right, it's either Compiled or not?

tt_torch/tools/utils.py
```diff
@@ -0,0 +1,956 @@
...
+    CONVERTED_TO_STABLE_HLO = 4
+    CONVERTED_TO_TTIR = 5
+    CONVERTED_TO_TTNN = 6
+    EXECUTED = 7
+
+
+class OpByOpBackend(Enum):
```
AleksKnezevic, line 61: There will only be one op-by-op backend.

tt_torch/tools/utils.py
```diff
@@ -0,0 +1,956 @@
...
+    TORCH = 1
+    STABLEHLO = 2
+
+
+class IOType(Enum):
```
AleksKnezevic, line 66: please remove all multi-chip stuff from this PR.

tt_torch/tools/utils.py
```diff
@@ -0,0 +1,956 @@
...
+        self.json = ""
+        self.binary = ""
+        self.runtime_stack_dump = ""
+        self.compilation_status = OpCompilationStatus.NOT_STARTED
+        self.parsed_stable_hlo_ops = False
+        self.parsed_ttnn_ops = False
+        self.pcc = None
+        self.atol = None
+
+    def parse_json(self):
```
AleksKnezevic, line 162: don't think any of this op-by-op support stuff is needed, it will all be picked up from tt-mlir.

tt_torch/backend/backend.py
LPanosTT: My argument type change isn’t merged yet. Nor does that method of populating them have a thumbs up from the tt-xla crew

tt_torch/tools/utils.py
LPanosTT: You are right

tt_torch/tools/utils.py
```diff
@@ -0,0 +1,956 @@
...
+    @property
+    def consteval_parameters(self):
+        return self._consteval_parameters
+
+    @consteval_parameters.setter
+    def consteval_parameters(self, value):
+        self._consteval_parameters = value
+        self.post_init()
+
+    def apply_environment_overrides(self):
```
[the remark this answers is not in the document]
LPanosTT: I actually don't think any of these are valid

tests/conftest.py
```diff
@@ -316,3 +317,10 @@ def monkeypatch_import(request):
         patch_config.patch()
 
     yield
+
+
+@pytest.fixture(autouse=True)
+def run_around_tests():
+    torch.manual_seed(0)
+    yield
+    torch._dynamo.reset()
```
sshonTT, line 326: could you also add `gc.collect()` to here?

tests/infra/testers/single_chip/model/model_tester.py
```diff
@@ -120,6 +120,8 @@ def _test_inference(self) -> None:
         cpu_res = self._run_on_cpu(compiled_cpu_workload)
 
         compiled_device_workload = self._compile_for_tt_device(self._workload)
+
+        # If the framework is Torch, we do not want to explicitly place the workload on device as that is done by the "tt" backend executor.
```
nvukobratTT, line 124: TODO: Checkout

tests/infra/testers/single_chip/model/torch_model_tester.py
```diff
@@ -9,9 +9,13 @@
 from infra.comparators import ComparisonConfig
 from infra.utilities import Framework, Model
 from infra.workloads import TorchWorkload, Workload, WorkloadFactory
+from tt_torch.tools.utils import CompilerConfig
 
 from .model_tester import ModelTester, RunMode
 
+# Registers PJRT plugin to torch-xla and registers "tt" backend for torch.compile
+import tt_torch
```
nvukobratTT, line 17: Potentially we don't need this one after this PR gets merged:
- https://github.com/tenstorrent/tt-xla/pull/1008

tests/torch/single_chip/models/resnet/test_resnet.py
```diff
@@ -52,12 +52,12 @@ def training_tester() -> ResnetTester:
     run_mode=RunMode.INFERENCE,
     bringup_status=BringupStatus.FAILED_TTMLIR_COMPILATION,
 )
-@pytest.mark.xfail(
-    reason=failed_ttmlir_compilation(
-        " Error: torch_xla/csrc/aten_xla_bridge.cpp:110 : Check failed: xtensor "
-        "https://github.com/tenstorrent/tt-xla/issues/795"
-    )
-)
+# @pytest.mark.xfail(
```
nvukobratTT, line 55: Can we remove these comments? Is there any value in keeping them?

tests/torch/single_chip/models/resnet/tester.py
```diff
@@ -16,8 +17,11 @@ def __init__(
         comparison_config: ComparisonConfig = ComparisonConfig(),
         run_mode: RunMode = RunMode.INFERENCE,
     ) -> None:
+        cc = CompilerConfig()
+        cc.enable_consteval = True
```
nvukobratTT, line 21: Isn't MLIR consteval pass working for this model? 

Can you elaborate on when FE-based consteval should be used next to the MLIR one?

tests/torch/single_chip/test_torch_xla_basic.py
```diff
@@ -0,0 +1,409 @@
...
+    input_x = torch.randn(32, 32, dtype=torch.bfloat16)
+
+    model = MM()
+    golden = model(input_x)
+
+    model = torch.compile(model, backend="tt")
+
+    output = model(input_x)
+    comparator = TorchComparator(
+        ComparisonConfig(
```
nvukobratTT, line 37: Do we need to propagate comparator configurations? Seems like these (or similar) atol and pcc values are already set by default.

tests/torch/single_chip/test_torch_xla_basic.py
```diff
@@ -0,0 +1,409 @@
...
+            pcc=PccConfig(enabled=False, required_pcc=0.99),
+        )
+    )
+    comparator.compare(output, golden)
+
+
+@pytest.mark.parametrize("op", eltwise_unary_ops)
+def test_eltwise_unary_eager(op):
+    if op is torch.erf:
+        pytest.skip(
```
nvukobratTT, line 262: What about just failing this one? Once it's supported, it'll be shown on the test report as passing (if strict mode is enabled; if not, we should).

nvukobratTT: Also, instead of comments, I see greater value in just putting the issue link. This way, we can more easily group failing tests/models, check out the latest state with more details, organize, and prioritize through Superset, etc.

tt_torch/__init__.py
```diff
@@ -0,0 +1,62 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
nvukobratTT, line 1: You will not need this kind of initialization (whole file) anymore after wheel cleanup. Please check this PR for more details:
- https://github.com/tenstorrent/tt-xla/pull/1008

nvukobratTT: Please sync with @pilkicTT for more details and potential further enhancments if needed :))

tt_torch/backend/backend.py
```diff
@@ -0,0 +1,159 @@
...
+
+from tt_torch.tools.utils import CompilerConfig
+
+from torch.utils._pytree import tree_map
+
+import torch_xla
+import torch_xla.core.xla_model as xm
+
+
+def bypass_assert_tensor_metadata(gm):
```
nvukobratTT, line 38: Can we reorganize the function position within this file, and follow some standard Python recommendations? E.g., keeping top-level function at the beginning of the file, then follow with sub-function definitions next, etc. 

This will make this code more readable and easier to follow, without the need to jump through the whole file to find different references.

nvukobratTT, line 38: Let's add Python docs to each of these functions. For starters, it'll be good enough just to briefly describe the purpose of the function. In this and similar cases, the function name isn't enough to get details like:
- What is the purpose?
- Why?
- Is this a workaround or should it be further improved/fixed somewhere else in the code
- Etc.

tt_torch/backend/backend.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+# This function runs a series of passes on a torch GraphModule.
+# The passes here may be necesarry (depending on the model) to
+# convert a GraphModule into a form which tt-mlir can compile/execute.
+def torch_pass_pipeline(gm, example_inputs, compiler_config):
+    decompositions = torch._decomp.core_aten_decompositions()
+    decompositions.update(CUSTOM_DECOMPOSITION_TABLE)
+    compiled_graph = (
+        torch.export.export_for_training(gm, tuple(example_inputs), strict=False)
```
nvukobratTT, line 55: Can we add a comment on why `export_for_training` is used, instead of the regular `export`? E.g. brief overview and ref if this is the reason: https://docs.pytorch.org/docs/stable/export.html#export-for-training-and-inference

tt_torch/backend/backend.py
```diff
@@ -0,0 +1,159 @@
...
+
+    compiled_graph = bypass_redundant_getitem(compiled_graph)
+    compiled_graph = rectify_buffer_inplace_copy(compiled_graph)
+    compiled_graph = bypass_assert_tensor_metadata(compiled_graph)
+    program = torch.export.export(compiled_graph, tuple(example_inputs), strict=False)
+
+    return program
+
+
+class XLAExecutor:
```
nvukobratTT, line 77: Please provide more details on how the executor works.

tt_torch/backend/backend.py
```diff
@@ -0,0 +1,159 @@
...
+    def __init__(self, program, compiler_config):
+        self.program = program
+        self.compiler_config = compiler_config
+        self.arg_type_map_str = None
+
+        self.inputs = []
+        self.user_input_indices = []
```
nvukobratTT, line 84: How would this one work if we didn't track the indices for specific user inputs? I.e., is it possible to separate user inputs for everything else, and make this part easier to understand and follow through executor?

tt_torch/backend/backend.py
```diff
@@ -0,0 +1,159 @@
...
+        for idx, input_spec in enumerate(self.program._graph_signature.input_specs):
+            if input_spec.kind == InputKind.USER_INPUT:
+                self.inputs.append(None)
+                self.user_input_indices.append(idx)
+            else:
+                self.inputs.append(self.program.state_dict[input_spec.target].to("xla"))
```
nvukobratTT, line 90: Why are we marking parameters and constants to be pushed to the device, while not doing the same for user inputs? Current limitation or by design?

tt_torch/backend/backend.py
nvukobratTT: @LPanosTT, do you have a reference to another PR that tackles this function?

tt_torch/backend/decompositions.py
```diff
@@ -0,0 +1,441 @@
...
+    Union[torch._ops.OperatorBase, torch._ops.OpOverloadPacket]
+]
+
+# Manages "scopes" for decompositions used. Each unique scope is an attribute on
+# the _decomp_local. If the attribute is missing, then the default
+# decompositions are used. The scope "aot" is used for all AOT cases.
+_decomp_local = threading.local()
+
+
+def _get_decomp_stack(scope: str) -> List[DecompositionTable]:
```
nvukobratTT, line 23: Similar comment for function organization as stated for the example above.

tt_torch/backend/decompositions.py
```diff
@@ -0,0 +1,441 @@
...
+):
+    table: DecompositionTable
+    if from_current:
+        table = dict(_current(scope))
+    else:
+        table = {}
+    if add_ops:
+        table.update(get_decompositions(add_ops))
+    if remove_ops:
+        remove_decompositions(table, remove_ops)  # type: ignore
```
nvukobratTT, line 57: Which ops are removed from decomposition? Do we have an example for it?

tt_torch/backend/decompositions.py
```diff
@@ -0,0 +1,441 @@
...
+    weights = torch.where(
+        torch.logical_and(sample_f >= -0.5, sample_f <= input_size - 0.5),
+        weights,
+        0,
+    )
+    weights = weights.squeeze()
+    return weights.to(dtype)
+
+
+def upsample_linear(
```
nvukobratTT, line 110: For these decompositions, it would be valuable to add more details, like:
- Why was it added
- Is this something that we should transfer to MLIR
-a  Is this a workaround that should be fixed on torch-xla
- Etc. 

With this transfer, we'll lose git history for each of them on tt-xla, so it'll be good to have more context around each of these decompositions.

tt_torch/backend/passes.py
```diff
@@ -0,0 +1,89 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import torch
+import gc
+from torch.fx.experimental import const_fold
+
+
+def run_shape_prop(gm, example_inputs):
```
nvukobratTT, line 9: Can you add documentation details for each of these passes? Once you do, I'll take a more details look into each of them :))

tt_torch/tools/utils.py
```diff
@@ -0,0 +1,641 @@
...
+                      TTIR Dialect
+                           |
+                      TTNN Dialect
+                           |
+                  Flatbuffer Executable
+"""
+
+
+class CompileDepth(Enum):
+    EXECUTE_OP_BY_OP = 5
```
nvukobratTT, line 42: Do we need to port the op_by_op logic? This PR should take care and enable similar logic on IR level (both TTIR and sHLO):
- https://github.com/tenstorrent/tt-xla/issues/1002

Therefore, do we need to add it here as well?

tt_torch/tools/utils.py
```diff
@@ -0,0 +1,641 @@
...
+    EXECUTE = 6
+
+
+class CompilerConfig:
+    def __init__(self):
+        self.compile_depth = CompileDepth.EXECUTE
```
nvukobratTT, line 48: Are we certain that we need all of these compiler configurations?

tt_torch/tools/utils.py
```diff
@@ -0,0 +1,641 @@
...
+    def model_name(self):
+        return self._model_name
+
+    @model_name.setter
+    def model_name(self, value):
+        self._model_name = value
+        if value and (self.save_mlir_override or self.dump_binary):
+            self.cleanup_old_files()
+
+    def cleanup_old_files(self):
```
nvukobratTT, line 82: Please provide function documentation for this file as well :))

nvukobratTT (changes requested):
General note. I would prefer if this PR were split into multiple ones, each targeting a specific domain. 

It's harder to follow the changes and which one is related to which part of the problem/solution this way. For future reference, I would recommend doing smaller, meaningful iterations at a time, instead of pushing a few notable changes as part of a single PR.

tests/infra/runners/device_runner.py
```diff
@@ -25,17 +25,35 @@ def run_on_cpu(self, workload: Workload) -> Tensor:
...
+        workload: Workload,
+        device_type: DeviceType,
+        device_num: int = 0,
+        explicitly_place_on_device: bool = True,
     ) -> Tensor:
         """Orchestrates workload execution by connecting to device, transferring data, and running."""
 
         device = self._device_connector.connect_device(device_type, device_num)
-        device_workload = self._put_on_device(workload, device=device)
+        device_workload = (
```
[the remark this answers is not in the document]
LPanosTT: The "tt" backend will place the module and inputs on device automatically. If it's done beforehand, it can potentially cause issues in the `torch_pass_pipeline`. 

As for the mnist test, it is not necessary. But this requires explanation.

torch-xla traces all eager execution into a graph. If you decide to `torch.compile` with a different backend (or not at all), as long as when you **run** the model, the underlying execution are calls to `aten` ops, torch-xla will trace them. So, you _could_ use the `openxla` backend, or probably the `inductor` backend for that matter. The `tt` backend exists to tweak the torch GraphModule **on host** before sending it to device to ensure that it is compilable by tt-mlir.

tests/torch/single_chip/test_torch_xla_basic.py
LPanosTT: It seems like the default `AtolConfig` is 0.16, so I'll keep my explicit one. The PCC default is 0.99 though, so that is not necessary.

tests/infra/runners/device_runner.py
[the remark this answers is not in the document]
AleksKnezevic: The xla backend also does some behind the scenes cpu fallback and is not suitable.

tt_torch/backend/backend.py
LPanosTT: The user inputs are unknown at compile time, aside from the "example_inputs". The "example_inputs" are typically the set of inputs the user provides on the first call to the model. However, for future calls to the model we will not know what they are, so we must push them do device upon `__call__`.

tt_torch/tools/utils.py
LPanosTT: @AleksKnezevic what is the status of this, and do we wish to use the CompilerConfig to signal when to use the SHLO op by op flow?

tt_torch/tools/utils.py
LPanosTT: The majority of these are used to store information about the execution of a model via the old torch op-by-op flow in tt-torch. They are unused in the additions made in this PR. However, it is not clear to me that they will remain unused in the upcoming SHLO op-by-op mode.

tt_torch/backend/passes.py
LPanosTT: Done

tt_torch/backend/backend.py
LPanosTT: Added

tt_torch/backend/decompositions.py
LPanosTT: I've added explanations for these at the bottom of the file where the decompositions are registered.

tt_torch/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+
+# find the tt-metal directory, it can either be in the venv if installed from a wheel or in the third_party source tree
+package_name = "tt-metal"
+spec = importlib.util.find_spec(package_name)
+if spec is not None:
+    tt_metal_home = os.path.abspath(spec.submodule_search_locations[0])
+    os.environ["TT_METAL_HOME"] = tt_metal_home
+
+# Import these modules so backends are registered ("tt", and "tt-experimental")
+import tt_torch.backend.backend
```
mrakitaTT, line 25: Will have to update the `python_package/torch_plugin_tt/__init__.py` now to call this to register the "tt" backend.

mrakitaTT: Btw what is "tt-experimental"?

mrakitaTT (commented):
I'm still going over the logic in `tt_torch/backend` files but I won't block this PR if @nvukobratTT's team signs off on these. They are neatly encapsulated without affecting rest of the codebase so even if I have some comments on those we can follow up later.

LPanosTT:
@mrakitaTT 
>What if the user does put tensors onto device before compiling for "tt" backend? It seems like a tricky situation, on one hand it is great that we can do it automatically, but on the other it is not great if it will cause issues if done manually. Can we check if they are already on device and not put them automatically in that case?

We do some host-side constant folding in the "tt" backend if it is enabled in the compiler config. Occasionally this contains operations which are not supported in tt-mlir. If the weights are already on device by the time we compile, then the constant folding will actually be traced and we will attempt to compile and run that on device, which then of course causes an error.

We should aim to eliminate this in the future and allow the user to compile a module on device, in fact the "openxla" backend is intended to be used that way. We can move towards this in the future. We may already be close as this PR was merged recently: https://github.com/tenstorrent/tt-mlir/pull/4283

tt_torch/__init__.py
LPanosTT: @mrakitaTT "tt-experimental" is the name of the torch-xla based backend in the tt-torch repo since it is not the main flow there. It was merged recently. It seems it will remain that way or be removed in tt-torch since we're moving that functionality here to be the main torch flow in tt-xla.

tt_torch/tools/utils.py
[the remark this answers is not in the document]
LPanosTT: I will weed as many of these out as I can right now.

tests/conftest.py
LPanosTT: Removed

tests/torch/single_chip/test_torch_xla_basic.py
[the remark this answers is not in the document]
LPanosTT: It looks like the op tester will `torch.compile` the model and then run that compiled model on both cpu and device. This is an issue because a model that is compiled on the "tt" backend **will** be run on the device regardless. I will open an issue for this.

tt_torch/tools/utils.py
nvukobratTT: If we agree to remove torch op-by-op, I'm more in favor of cleaning this one up as well. Especially as we're now adding more compile options in the related PR (linked above; bfp8 support).

nvukobratTT (approved):
Left few more comments.

@LPanosTT, there are few minor ones that are related to mostly to deliting parts of code/logic. Let's determine if those are needed before we merge this PR. 

Regarding other comments, most of them can be addressed after we get this PR merged as well. That said, I'm approving this PR to unblock other workloads, so when you make few additional correction, I'm OK with merging this one :))

tests/infra/runners/device_runner.py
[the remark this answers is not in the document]
LPanosTT: This does not seem to be how torch.compile is truly intended to be used. For example, the "openxla" backend examples show a model which is already on device being compiled: https://docs.pytorch.org/xla/master/torch_compile.html

1. The reason we compile a host module, then push to device in the "tt" backend is so that we can perform constant folding. 
2. The reason we want to do constant folding in the fronted is to take care of a few ops which we cannot compile (do not remember which off the top of my head). 
3. The reason we cannot compile constant folding on a model which is already on device is because it will then try to compile and execute the ops which we wanted to constant-fold away in the first place.
4. So, the solution must be to `torch.compile` the host module, constant fold that host module, _then_ push to device.

As I stated at the beginning of this comment, we should not need to do this. Moving forward it should be a goal of ours to push the model to device, and then `torch.compile` it. And we should remove the host-side constant folding entirely by making whatever changes to the stack required. At the moment, we do not have the support for this though.

tests/infra/testers/single_chip/model/torch_model_tester.py
```diff
@@ -93,7 +93,7 @@ def _compile(self, workload: Workload) -> Workload:
     # @override
     def _compile_for_tt_device(self, workload: Workload) -> Workload:
         """Compiles `workload` for TT device."""
-        return self._compile_for_backend(workload, backend="openxla")
+        return self._compile_for_backend(workload, backend="tt")
```
mrakitaTT, line 96: Should also change the op tester https://github.com/tenstorrent/tt-xla/blob/main/tests/infra/testers/single_chip/op/op_tester.py#L62 (line 84 too)

To avoid this change in two places in the future we should make a common `_compile()` function which both `_compile_for_tt_device` and `_compile_for_cpu` call and pass `compiler_options` dict to it (`_compile_for_cpu` would pass an empty dict).

tests/infra/workloads/workload.py
```diff
@@ -58,8 +61,22 @@ def is_torch(self) -> bool:
...
+        # Since the model output will be an arbitrary collection of tensors,
+        # use tree_map to move all tensors to CPU.
+        if result is not None and self.framework == Framework.TORCH:
+            # NOTE: Each call to .to('cpu') will halt until that tensor's data
+            # has been populated. That is to say, the program has finished compiling and executing.
+            # However, once the program has finished executing, all outputs will be ready, so in reality
+            # the halt will only be for the first tensor, and the rest will be ready immediately.
+            result = tree_map(
+                lambda x: x.to("cpu") if isinstance(x, torch.Tensor) else x, result
+            )
```
mrakitaTT, line 80: This shouldn't be necessary since these results are passed to `TorchComparator` whose comparation functions all have `@run_on_cpu(Framework.TORCH)` decorator which will move all inputs to CPU. `JaxComparator` does the same thing.

tests/torch/single_chip/test_torch_xla_basic.py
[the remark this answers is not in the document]
mrakitaTT: If you don't plan to attend to this in this PR, please add a TODO comment in this file explaining that these will be moved, with the link to that issue you've opened. I would also update the issue title since op and graph testers are now able to use "tt" backend.

ttxla_tools/ttxla_tools.py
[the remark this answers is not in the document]
LPanosTT: I can add it, should I rename the folder and the file? Or just one or the other.

tests/infra/workloads/workload.py
LPanosTT: It is necessary. The comparison functions attempt to compute `torch.allclose` between the golden and on-device output. It errors out every time. I will see if there is a simple fix in the comparator logic.

LPanosTT: Ah, `workload.args` are not necessarily a list of tensors, it could be any pytree, I will fix `_safely_put_workload_on_device` to handle this.

tests/conftest.py
```diff
@@ -316,3 +317,11 @@ def monkeypatch_import(request):
         patch_config.patch()
 
     yield
+
+
+@pytest.fixture(autouse=True)
+def run_around_tests():
```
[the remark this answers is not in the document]
mrakitaTT: Sure, no worries, please just leave a TODO comment here with the link to that issue.

mrakitaTT (approved):
🚀

tests/infra/runners/torch_device_runner.py
```diff
@@ -37,17 +38,14 @@ def _safely_put_workload_on_device(
...
-                args_on_device.append(arg)
+                return x
 
-        for key, value in workload.kwargs.items():
-            try:
-                kwargs_on_device[key] = value.to(device)
-            except:
-                kwargs_on_device[key] = value
+        args_on_device = tree_map(attempt_to_device, workload.args)
+        kwargs_on_device = tree_map(attempt_to_device, workload.kwargs)
```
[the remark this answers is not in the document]
kmabeeTT: This change from 55e31c533bf38b194626512837546b543f4c79c8 is swallowing tt-metal asserts and causes them to seen as segfaults now which is not great (kills pytest), opened issue : https://github.com/tenstorrent/tt-xla/issues/1259

PR #985, opened 2025-08-13
Infra workload class merge

vzeljkovicTT:
### What's changed

- Merged `Workload` classes into infra
- Removed `Workload Factory`

**Motivation**
There is no longer a need to separate JAX and Torch workloads.
Both require a model and an executable at certain points, so with these changes we simplify the infra.


mrakitaTT:
And please a more detailed PR description to explain why you are doing this, i.e. that we will use both model and executable in the future for both JAX/Torch workloads etc. :)

mrakitaTT (approved):

PR #1032, opened 2025-08-18
Switch all card tests (except llmbox) to CIv1

vvukomanTT:
### Ticket
/

### Problem description
Low disk on CIv2 runners was causing test failures.

### What's changed
Switch all tests (except llmbox) to CIv1.
Note: In the future CIv2 should support runners with more disk, we will switch to those then.

### Checklist


vvukomanTT:
> @vvukomanTT, do we know what causes disk space issues?

tt-xla tests are downloading and using more disk than the CIv2 runners have available (100GB).

PR #1102, opened 2025-08-22
Dgolubovic/enable lower df

dgolubovicTT:
### What's changed
 - Add CompilerConfig and propagate it to compile methods in tester classes
 - Add new compile option enable_bfp8_conversion that enables block fp8 in MLIR. For this to work TTIR graph has to be in bfloat16. Final graph will have input and output nodes in bfloat16 and everything else in bfp8. Essentially adding type casts at the beginning and in the end of the graph, while all intermediate results are in bfp8. This bfloat16 wrapping is done to enable comparisom with golden, because block formats are TT hardware specific, and user should provide and get tensors of common dtype.
 - Make all compile methods void - they change workload object inplace so they don't have to return it. Change is in all tester classes: OpTester, ModelTester, JaxModelTester, JaxMultichipModelTester, TorchModelTester, JaxMultichipOpTester
- Add lower dataformat tests (bfloat16 and bfp8) for jax ops that have float inputs.

### DIFF
--- a/tests/infra/testers/multichip/model/jax_multichip_model_tester.py
+++ b/tests/infra/testers/multichip/model/jax_multichip_model_tester.py
@@ -45,6 +45,7 @@ class JaxMultichipModelTester(JaxModelTester, ABC):
         axis_names: tuple,
         comparison_config: ComparisonConfig = ComparisonConfig(),
         run_mode: RunMode = RunMode.INFERENCE,
+        compiler_config: CompilerConfig = None,
     ) -> None:
         self._mesh_shape = mesh_shape
         self._axis_names = axis_names
@@ -59,7 +60,7 @@ class JaxMultichipModelTester(JaxModelTester, ABC):
         self._input_parameters_partition_specs: PyTree = None
         self._input_parameters: PyTree = None
 
-        super().__init__(comparison_config, run_mode)
+        super().__init__(comparison_config, run_mode, compiler_config)
 
     # @override
     def _initialize_components(self) -> None:
@@ -118,10 +119,12 @@ class JaxMultichipModelTester(JaxModelTester, ABC):
         )
         output_sharding = NamedSharding(workload.device_mesh, workload.out_spec)
 
+        compile_options = self._compiler_config.to_jax_compiler_options()
         workload.executable = jax.jit(
             module_sharded_executable,
             out_shardings=output_sharding,
             static_argnames=workload.static_argnames,
+            compiler_options=compile_options,
         )
 
     def _create_multichip_workload(
--- a/tests/infra/testers/multichip/model/jax_multichip_model_tester.py
+++ b/tests/infra/testers/multichip/model/jax_multichip_model_tester.py
@@ -90,17 +90,17 @@ class JaxMultichipModelTester(JaxModelTester, ABC):
         )
 
     # @override
-    def _compile_for_cpu(self, workload: Workload) -> Workload:
+    def _compile_for_cpu(self, workload: Workload) -> None:
         """Compiles `workload` for CPU."""
-        return self._compile(self._create_multichip_workload(self._cpu_mesh))
+        self._compile(self._create_multichip_workload(self._cpu_mesh))
 
     # @override
-    def _compile_for_tt_device(self, workload: Workload) -> Workload:
+    def _compile_for_tt_device(self, workload: Workload) -> None:
         """Compiles `workload` for TT device."""
-        return self._compile(self._create_multichip_workload(self._device_mesh))
+        self._compile(self._create_multichip_workload(self._device_mesh))
 
     # @override
-    def _compile(self, workload: Workload) -> Workload:
+    def _compile(self, workload: Workload) -> None:
         """
         Sets up `workload.executable` for just-in-time compile and execution.
 
@@ -123,7 +123,6 @@ class JaxMultichipModelTester(JaxModelTester, ABC):
             out_shardings=output_sharding,
             static_argnames=workload.static_argnames,
         )
-        return workload
 
     def _create_multichip_workload(
         self, mesh: jax.sharding.Mesh
--- a/tests/infra/testers/multichip/op/jax_multichip_op_tester.py
+++ b/tests/infra/testers/multichip/op/jax_multichip_op_tester.py
@@ -8,6 +8,7 @@ from typing import Callable, Sequence
 
 import jax
 from infra.comparators import ComparisonConfig
+from infra.compiler_config import CompilerConfig
 from infra.connectors import JaxDeviceConnector
 from infra.runners import JaxDeviceRunner
 from infra.utilities import (
@@ -51,7 +52,11 @@ class JaxMultichipOpTester(BaseTester):
         mesh_shape: tuple,
         axis_names: tuple,
         comparison_config: ComparisonConfig = ComparisonConfig(),
+        compiler_config: CompilerConfig = None,
     ) -> None:
+        if compiler_config is None:
+            compiler_config = CompilerConfig()
+        self._compiler_config = compiler_config
         self._in_specs = in_specs
         self._out_spec = out_spec
         self._mesh_shape = mesh_shape
@@ -142,10 +147,12 @@ class JaxMultichipOpTester(BaseTester):
         )
         output_sharding = NamedSharding(workload.device_mesh, workload.out_spec)
 
+        compile_options = self._compiler_config.to_jax_compiler_options()
         workload.executable = jax.jit(
             module_sharded_executable,
             out_shardings=output_sharding,
             static_argnames=workload.static_argnames,
+            compiler_options=compile_options,
         )
 
     # --- Convenience wrappers ---
@@ -186,6 +193,7 @@ def run_jax_multichip_op_test_with_random_inputs(
     minval: float = 0.0,
     maxval: float = 1.0,
     comparison_config: ComparisonConfig = ComparisonConfig(),
+    compiler_config: CompilerConfig = None,
 ) -> None:
     """
     Tests an input executable with random inputs in range [`minval`, `maxval`) by
@@ -195,7 +203,7 @@ def run_jax_multichip_op_test_with_random_inputs(
     """
     with enable_shardy(use_shardy):
         tester = JaxMultichipOpTester(
-            in_specs, out_specs, mesh_shape, axis_names, comparison_config
+            in_specs, out_specs, mesh_shape, axis_names, comparison_config, compiler_config
         )
         tester.test_with_random_inputs(
             executable, input_shapes, sharding_mode, minval, maxval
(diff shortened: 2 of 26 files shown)


tests/infra/testers/single_chip/model/jax_model_tester.py
```diff
@@ -40,14 +41,13 @@ def __init__(
         self,
         comparison_config: ComparisonConfig = ComparisonConfig(),
         run_mode: RunMode = RunMode.INFERENCE,
-        use_optimizer: bool = False,
+        compiler_config: CompilerConfig = None,
```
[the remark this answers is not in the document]
dgolubovicTT: Ofc, good catch...

tests/infra/testers/single_chip/op/op_tester.py
```diff
@@ -100,12 +141,16 @@ def run_op_test_with_random_inputs(
     input_shapes: Sequence[tuple],
     minval: float = 0.0,
     maxval: float = 1.0,
+    dtype: str | DTypeLike | torch.dtype = "float32",
     comparison_config: ComparisonConfig = ComparisonConfig(),
     framework: Framework = Framework.JAX,
+    compiler_config: CompilerConfig = None,
```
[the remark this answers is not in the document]
mrakitaTT: I am aware of that gotcha but I didn't see the case here where we will edit `compiler_config` inside these functions, similar like `comparison_config` that is something that is either setup by tests or left as a default, so in that case this python quirk even plays into our advantage because it will use one default comparison config object in all tests. It would be a red flag to me if someone wants to introduce some logic which changes default config in these places so I would be mindful to also remind them about this quirk, but I understand that you want to put protections in place to be 100% sure no one introduces such bug, so I don't have hard opinions on this except that we should be consistent, so either cover the `comparison_config` too or don't cover both :)

tests/infra/testers/compiler_config.py
```diff
@@ -0,0 +1,39 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from dataclasses import dataclass
```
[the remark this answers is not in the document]
dgolubovicTT: I agree. @mrakitaTT is this ok with you?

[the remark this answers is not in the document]
mrakitaTT: This class is used in our testing infra to facilitate passing compiler options from tests through testers up to the corresponding compile calls from JAX/TorchXLA API. I am a bit torn on whether we should expose it to our users, so let me share my thinking and I would love to hear your opinion. We should probably also include other people in this discussion, good topic for Monday sync :)

**On one hand**, I don't think we should expose it to our users and require them to use our custom API to compile their module from JAX/TorchXLA. Users of these frameworks are already familiar with passing compiler options as string flags into existing framework API, and requiring them to import our custom library just to do that would be an overhead. It would also go against the XLA principle of native plugin integration where users can use the existing framework API and seemlesly switch different HW backends, which is probably why they went with the plain strings map instead of having some compile config structure. I don't think we should go "against the grain" and require users to change behaviour they are already used to.

For example here is how they are setting up XLA compile options in JAX (options documented [here](https://docs.jax.dev/en/latest/xla_flags.html) and [here](https://openxla.org/xla/flags_guidance)):
```
jax.jit(model.generate, compiler_options={"xla_enable_async_all_gather": true}
``` 

JAX config options are setup in the similar way (options documented [here](https://docs.jax.dev/en/latest/config_options.html)):
```
jax.config.update("jax_enable_x64", True)
```

and here is example from TorchXLA:
```
torch_xla.compile(model.generate, custom_compile_options={"xla_enable_async_all_gather": true}
```

**On the other hand**, I definitely agree with you that it would be cleaner to have these flags documented in the code and passed through the structure instead of through custom map of strings documented on the website. There is also a versioning issue, if the flags are documented on the website we need to keep the old flags documented and marked as deprecated from version X.Y. If the flags are consumed through the structure shipped with plugin, then old flags are simply not going to be present in newers versions of plugin. But then again, users will still be able to use the native framework API to set the compile options, and **will** expect us to have these flags documented on the website, so we would have to do both. Then there is a problem with experimental flags that we want to use in our tests but are not yet ready to be exposed to external users.

Idk, I am leaning on the side of not exposing this to our users and just documenting the flags we support in our docs, and leaving users to use them as they are already used to through native framework API. I definitely recommend going with this option first (leaving this PR as it currently is), and then having a wider discussion to agree if we want to expose the `CompilerConfig` structure too.

tests/infra/testers/multichip/model/jax_multichip_model_tester.py
```diff
@@ -59,13 +61,23 @@ def __init__(
...
 
+    # @override
+    def _initialize_workload(self) -> None:
+        # call parent method first
+        super()._initialize_workload()
+        # then replace single chip workload with multichip one
+        self.tt_device_multichip_workload = self._create_multichip_workload(
+            self._device_mesh
+        )
+        self.cpu_workload = self._create_multichip_workload(self._cpu_mesh)
```
mrakitaTT, line 79: Would call this `cpu_multichip_workload` to be consistent with `tt_device_multichip_workload` (it is a multi-chip workload).

tests/infra/testers/multichip/model/jax_multichip_model_tester.py
```diff
@@ -90,17 +102,31 @@ def _get_cpu_device_mesh(self) -> jax.sharding.Mesh:
...
     # @override
-    def _compile_for_tt_device(self, workload: Workload) -> Workload:
+    def _compile_for_tt_device(self, workload: JaxMultichipWorkload) -> None:
         """Compiles `workload` for TT device."""
-        return self._compile(self._create_multichip_workload(self._device_mesh))
+        self._compile(workload, cpu_device=False)
 
     # @override
-    def _compile(self, workload: Workload) -> Workload:
+    def _compile(self, workload: Workload, cpu_device: bool = False) -> None:
```
mrakitaTT, line 129: This is not an override anymore (not present in the base class), and is now present only in tester classes where there is a common compile logic for TT and CPU. Because here it isn't the case anymore (hence the `cpu_device` flag you added) there is no point having this function at all. I would leave just `_compile_for_cpu` and `_compile_for_tt_device ` functions, and move the corresponding logic from the if/else branches below to there.

mrakitaTT: I see that you also did a similar thing in `JaxMultichipOpTester`, please fix there too. Also, we can now put `_compile_for_cpu` and `_compile_for_tt_device` into the `BaseTester` class since all testers use them.

dgolubovicTT: Makes sense!

dgolubovicTT:
> On one hand, I don't think we should expose it to our users and require them to use our custom API to compile their module from JAX/TorchXLA. Users of these frameworks are already familiar with passing compiler options as string flags into existing framework API, and requiring them to import our custom library just to do that would be an overhead. It would also go against the XLA principle of native plugin integration where users can use the existing framework API and seemlesly switch different HW backends, which is probably why they went with the plain strings map instead of having some compile config structure. I don't think we should go "against the grain" and require users to change behaviour they are already used to.
> 
> For example here is how they are setting up XLA compile options in JAX (options documented [here](https://docs.jax.dev/en/latest/xla_flags.html) and [here](https://openxla.org/xla/flags_guidance)):

Thanks for additional info. After looking into flags guidance, it makes more sense to keep this config only for testing for now. `jax.jit` requires string -> string mapping for `compiler_options` and we transform compiler_config object to this format. torch compile will have some other way of passing configs that we will integrate into.

One of the core benefits of tt-xla over tt-forge-fe is seamless integration of our compiler to user workflow. Therefore, I think that exposing CompilerConfig violates the whole idea of XLA, as you said above.

tests/infra/testers/multichip/model/jax_multichip_model_tester.py
```diff
@@ -90,20 +101,49 @@ def _get_cpu_device_mesh(self) -> jax.sharding.Mesh:
...
-        return self._compile(self._create_multichip_workload(self._device_mesh))
+        self._compile_for_tt_device(self.tt_device_multichip_workload)
+        tt_res = self._run_on_tt_device(self.tt_device_multichip_workload)
+
+        self._compare(tt_res, cpu_res)
 
     # @override
-    def _compile(self, workload: Workload) -> Workload:
-        """
+    def _compile_for_cpu(self, workload: Workload) -> None:
```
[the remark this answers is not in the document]
dgolubovicTT: Yeah, code copying is the reason I opted for the first implementation (with cpu flag as input). This one looks cleaner. Assuming that 

>     def _compile_for_cpu(self, workload: Workload, compile_options: Mapping[str, str]) -> None:
        assert isinstance(workload, JaxMultichipWorkload)
        module_sharded_executable = shard_map(
            workload.executable,
            mesh=workload.device_mesh,
            in_specs=workload.in_specs,
            out_specs=workload.out_spec,
            # For some reason this check doesn't like replicated outputs.
            check_rep=False,
        )
        output_sharding = NamedSharding(workload.device_mesh, workload.out_spec)

        compile_options = self._compiler_config.to_jax_compiler_options()
        workload.executable = jax.jit(
            module_sharded_executable,
            out_shardings=output_sharding,
            static_argnames=workload.static_argnames,
            compiler_options=compile_options,
        )

You ment `def _compile`

tests/infra/testers/single_chip/model/torch_model_tester.py
```diff
@@ -79,25 +81,24 @@ def _get_forward_method_kwargs(self) -> Mapping[str, Any]:
...
 
     # @override
-    def _compile_for_cpu(self, workload: Workload) -> Workload:
+    def _compile_for_cpu(self, workload: Workload) -> None:
         """Compiles `workload` for CPU."""
-        return self._compile(workload)
+        self._compile(workload)
 
-    def _compile(self, workload: Workload) -> Workload:
+    def _compile(self, workload: Workload) -> None:
```
mrakitaTT, line 88: No need for this function anymore, you can just move call `self._compile_for_backend(workload, backend="inductor")` to `def _compile_for_cpu`

tests/infra/testers/multichip/op/jax_multichip_op_tester.py
```diff
@@ -112,20 +117,33 @@ def test(
...
 
         self._comparator.compare(device_res, cpu_res)
 
     # @override
-    def _compile(self, workload: Workload) -> Workload:
+    def _compile_for_cpu(self, workload: Workload) -> None:
+        assert isinstance(workload, JaxMultichipWorkload)
+
+        # Compile options are not used for CPU compilation since they are TT backend specific.
+        self._compile(workload, compile_options={})
```
mrakitaTT, line 134: `compiler_options` instead of `compile_options`, this is why CI is failing :)

mrakitaTT (approved):

PR #1288, opened 2025-09-04
Dump per-pass IR with LOGGER_LEVEL=VERBOSE

jameszianxuTT:
### Ticket
None

### Problem description
There is no way to dump all IR between all compiler passes, which is useful when a minor compiler pass fails (i.e something other than a dialect-dialect conversion). For example, if a TTIRFusion pass fails, the current LOGGER_LEVEL=DEBUG will only show the TTIR before any passes are run, which is hard to debug.

### What's changed
Add higher LOGGER_LEVEL=VERBOSE which will verbosely dump IR at each compiler pass using mlir::PassManager::enableIRPrinting. This also dumps all DEBUG logs. 

### Checklist

### DIFF
--- a/inc/common/pjrt_implementation/module_builder/module_builder.h
+++ b/inc/common/pjrt_implementation/module_builder/module_builder.h
@@ -170,7 +170,7 @@ private:
   // Prints module to console for debug purposes.
   static void printModule(mlir::OwningOpRef<mlir::ModuleOp> &mlir_module);
 
-  // Enable printing of IR between passes if the log level for debug purposes
+  // Enables IR printing between passes with VERBOSE or higher logger level.
   static void enableVerboseIRPrinting(mlir::PassManager &pm);
 
   // Checks if a particular type is scalar.
--- a/inc/common/pjrt_implementation/module_builder/module_builder.h
+++ b/inc/common/pjrt_implementation/module_builder/module_builder.h
@@ -17,6 +17,7 @@
 #include "mlir/IR/MLIRContext.h"
 #include "mlir/IR/OwningOpRef.h"
 #include "mlir/IR/Value.h"
+#include "mlir/Pass/PassManager.h"
 
 // PJRT C API includes
 #include "xla/pjrt/c/pjrt_c_api.h"
@@ -168,6 +169,9 @@ private:
 
   // Prints module to console for debug purposes.
   static void printModule(mlir::OwningOpRef<mlir::ModuleOp> &mlir_module);
+
+  // Enable printing of IR between passes if the log level for debug purposes
+  static void enableVerboseIRPrinting(mlir::PassManager &pm);
 
   // Checks if a particular type is scalar.
   bool isScalarType(mlir::Type type);
--- a/src/common/pjrt_implementation/module_builder/module_builder.cc
+++ b/src/common/pjrt_implementation/module_builder/module_builder.cc
@@ -174,6 +174,8 @@ void ModuleBuilder::convertFromVHLOToSHLO(
 
   mlir::stablehlo::createStablehloDeserializePipeline(vhlo_to_shlo_pm);
 
+  enableVerboseIRPrinting(vhlo_to_shlo_pm);
+
   if (mlir::failed(vhlo_to_shlo_pm.run(mlir_module.get()))) {
     DLOG_F(ERROR, "Failed to convert from VHLO to SHLO module");
     m_status = tt_pjrt_status::kInternal;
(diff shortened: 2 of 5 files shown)


jameszianxuTT:
This technique is lifted wholesale from tt-torch's TT_TORCH_IR_LOG_LEVEL=DEBUG [[ref in tt-torch](https://github.com/search?q=repo%3Atenstorrent%2Ftt-torch+enableIRPrinting&type=code)]

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -186,6 +186,12 @@ void ModuleBuilder::convertFromVHLOToSHLO(
   // Detect Shardy by looking at the meshes attribute in module.
   if (isUsingShardy(mlir_module)) {
     mlir::PassManager shardy_pm(mlir_module.get()->getName());
+
+    if (loguru::g_stderr_verbosity >= LOG_DEBUG_IR) {
+      shardy_pm.getContext()->disableMultithreading();
+      shardy_pm.enableIRPrinting();
+    }
```
mrakitaTT, line 193: You can create a function similar to `ModuleBuilder::printModule` (and place its declaration/definition near it) which takes a `mlir::PassManager` as an input and does this, so you can call it in all these places.

```
void ModuleBuilder::enableIntermediateIRPrinting(
    mlir::PassManager &pass_manager) {
  // Enabled only with VERBOSE logging level.
  if (loguru::g_stderr_verbosity < LOG_VERBOSE) {
    return;
  }

  pass_manager.getContext()->disableMultithreading();
  pass_manager.enableIRPrinting();
}
```

mrakitaTT: And please also add comment above the `pass_manager.getContext()->disableMultithreading();` line explaining why you are disabling multithreading.

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -798,5 +814,4 @@ ModuleBuilder::createArgumentTypeMap(
 
   return argTypesMap;
 }
-
```
mrakitaTT, line 801: Nit: revert empty line

src/common/pjrt_implementation/module_builder/module_builder.cc
mrakitaTT: Btw there is also a `PassManager` created in `runCompilerStableHLOPipeline`, it might be useful to call this there too.

mrakitaTT (commented):
Thank you for adding this @jameszianxuTT!

src/common/status.h
```diff
@@ -11,6 +11,7 @@
 // Loguru doesn't have debug named verbosity and the next verbosity after INFO
 // is called `Verbosity_1`, so defining this to avoid doing `DLOG_F(1, ...)`.
 #define LOG_DEBUG 1
+#define LOG_DEBUG_IR 2
```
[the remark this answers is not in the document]
mrakitaTT: Your understanding is correct, I just didn't explain my point well :) What I meant is that if we for example wanted to add some verbose logs that print the parameters of each PJRT API call, and if we did it by adding another log level for that like `#define LOG_DEBUG_API 3`, then it wouldn't be obvious to devs by searching the codebase for `LOG_DEBUG_API` that by using that log level they would also get the intermediate IR printed. If we just simply add general `LOG_VERBOSE` level and use them for both, then it will be more obvious what it covers. But your understanding is correct that it wouldn't decouple those two loggings, which I'm totally fine with, I don't think we need to separate them and control them with different flags. If we later decide that some things are more verbose than the other, we can simply add some `LOG_ULTRA_VERBOSE` flag :)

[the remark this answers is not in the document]
mrakitaTT: So when I said "couple" I simply meant naming wise, we should keep standard log level names like INFO/DEBUG/VERBOSE.

src/common/pjrt_implementation/module_builder/module_builder.cc
jameszianxuTT: Done!

inc/common/pjrt_implementation/module_builder/module_builder.h
```diff
@@ -169,6 +170,9 @@ class ModuleBuilder {
   // Prints module to console for debug purposes.
   static void printModule(mlir::OwningOpRef<mlir::ModuleOp> &mlir_module);
 
+  // Enable printing of IR between passes if the log level for debug purposes
```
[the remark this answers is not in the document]
mrakitaTT: Also I would replace "if the log level for debug purposes" with "when logging level is verbose or higher."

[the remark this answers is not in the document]
jameszianxuTT: Thanks for the advice, I'll fix that typo and look through the styleguide

Issue #1003, opened 2025-08-13
Pipe compile options though torch-xla


AleksKnezevic:

mrakitaTT:
FYI for example from jax tests which adds compile options parsing in module builder and enables optimizer https://github.com/tenstorrent/tt-xla/pull/942

odjuricicTT:
@sshonTT Help me understand what is the status on this? Are we blocked on the issue to the upstream PyTorch-XLA repo (https://github.com/pytorch/xla/issues/9555)?

AleksKnezevic:
@sshonTT, that's perfect, the reason we maintain the fork is so we're not blocked by upstreams.

sshonTT:
https://github.com/tenstorrent/pytorch-xla/pull/5
 has been merged, but it looks like we can’t use it yet because it’s incompatible with JAX, which needs to be at version 0.7.0. I will add a PR when it is ready.

nvukobratTT:
@sshonTT Jax has been upgraded to v0.7.1: 
- https://github.com/tenstorrent/tt-xla/pull/1284\

Do we have any other blockers to push this one further? Also, can you link a PR to an issue? I see that issue is in "In Review" state, so I assume you have a PR ready.

sshonTT:
@nvukobratTT I am waiting for review on it.

PR #1321, opened 2025-09-08
Add support for parsing custom compiler options in PJRT backend compile

sshonTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/1003)

### Problem description
In torch pass, there was no way to passing user's compile options to pjrt backend.

### What's changed
User can pass compiler options using torch_xla.set_compile_options(opt) before compile model. For example,
```
CompilerOptions option
option.enable_optimizer = True
torch_xla. set_custom_compile_options(option)
torch.compile(model, ....)
```
Or, if torch_xla.compile is used
```
CompilerOptions option
option.enable_optimizer = True
torch_xla.compile(model, custom_compile_options=option)
```

Note that torch.compile(model, options=options) is not supported yet

### Checklist

### DIFF
--- a/src/common/pjrt_implementation/client_instance.cc
+++ b/src/common/pjrt_implementation/client_instance.cc
@@ -216,9 +216,9 @@ tt_pjrt_status ClientInstance::compileMlirProgram(
   return tt_pjrt_status::kSuccess;
 }
 
-std::unordered_map<std::string, std::string>
-ClientInstance::getCompileOptions(const char *compile_options_data,
-                                  size_t compile_options_size) {
+tt_pjrt_status ClientInstance::getCompileOptions(
+    const char *compile_options_data, size_t compile_options_size,
+    std::unordered_map<std::string, std::string> &out_compile_options) {
 
   google::protobuf::io::CodedInputStream cis(
       reinterpret_cast<const uint8_t *>(compile_options_data),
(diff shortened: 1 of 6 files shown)


src/common/pjrt_implementation/client_instance.cc
```diff
@@ -239,73 +239,81 @@ std::unordered_map<std::string, std::string>
...
-        if (value_field.type() ==
-            google::protobuf::UnknownField::TYPE_LENGTH_DELIMITED) {
-          value = value_field.length_delimited();
+      const std::string &bytes = field.length_delimited();
+
+      google::protobuf::io::CodedInputStream cis(
+          reinterpret_cast<const uint8_t *>(bytes.data()), bytes.size());
+
+      google::protobuf::UnknownFieldSet map_entry_fields;
+      if (!map_entry_fields.MergeFromCodedStream(&cis)) {
```
nvukobratTT, line 263: What happens if this one is false? How does it impact the compile option?

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -239,73 +239,81 @@ std::unordered_map<std::string, std::string>
...
+        continue;
+      }
+
+      std::string key;
+      std::string value;
+
+      for (int j = 0; j < map_entry_fields.field_count(); ++j) {
+        const auto &entry_field = map_entry_fields.field(j);
+
+        if (entry_field.number() == 1 &&
```
nvukobratTT, line 273: Can we add a description for each case to make the code easier to read?

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -239,73 +239,81 @@ std::unordered_map<std::string, std::string>
...
+          google::protobuf::io::CodedInputStream override_stream(
+              reinterpret_cast<const uint8_t *>(override_bytes.data()),
+              override_bytes.size());
+
+          google::protobuf::UnknownFieldSet value_fields;
+          if (value_fields.MergeFromCodedStream(&override_stream)) {
+            for (int k = 0; k < value_fields.field_count(); ++k) {
+              const auto &value_field = value_fields.field(k);
+              switch (value_field.number()) {
+              case 1:
```
nvukobratTT, line 291: Can we get other cases as well? Let's add an assertion if this is a possibility.

nvukobratTT (commented):
General comment, can we add a few e2e tests to confirm this? E.g., a small op test that will iterate over different compile option configurations.

src/common/pjrt_implementation/client_instance.cc
sshonTT: Good catch. If MergeFromCodedStream returns false, that means the map entry couldn’t be parsed. In the current logic we skip that entry, which effectively ignores that particular override and falls back to default behavior for that key. I make this explicit by adding a warning.

src/common/pjrt_implementation/client_instance.cc
sshonTT: Agreed. I’ll annotate each branch and replace raw field numbers with named constants to clarify intent. Also added what they are parsing using example.

src/common/pjrt_implementation/client_instance.cc
sshonTT: Added original message proto in comment, so once proto is changed, this line should be lifted also. Added default for raise error.

mrakitaTT (commented):
Sorry for delay in review! You've force pushed a branch during my review so some comments may be out of sync, please let me know if any of the comments is unclear.

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -216,95 +216,131 @@ tt_pjrt_status ClientInstance::compileMlirProgram(
...
-          // simple key-value pair of strings, so we skip it for now.
-          continue;
+        const std::string &override_bytes = entry_field.length_delimited();
+        google::protobuf::io::CodedInputStream override_stream(
+            reinterpret_cast<const uint8_t *>(override_bytes.data()),
+            override_bytes.size());
+
+        google::protobuf::UnknownFieldSet value_fields;
+        if (!value_fields.MergeFromCodedStream(&override_stream)) {
+          DLOG_F(ERROR, "Failed to parse value fields");
```
mrakitaTT, line 294: I would say something like `Failed to parse the map entry field value from the custom compile options protobuf data`

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -216,95 +216,131 @@ tt_pjrt_status ClientInstance::compileMlirProgram(
...
+          return tt_pjrt_status::kInternal;
         }
-        const google::protobuf::UnknownField &value_field =
-            custom_nested_set.field(0);
-        if (value_field.type() ==
-            google::protobuf::UnknownField::TYPE_LENGTH_DELIMITED) {
-          value = value_field.length_delimited();
+        for (int k = 0; k < value_fields.field_count(); ++k) {
```
mrakitaTT, line 297: Nit: Add a blank line above the for loop

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -216,95 +216,131 @@ tt_pjrt_status ClientInstance::compileMlirProgram(
...
+          const google::protobuf::UnknownField &value_field =
+              value_fields.field(k);
+          // https://github.com/openxla/xla/blob/main/xla/pjrt/proto/compile_options.proto#L151C1-L158C2
+          // Field numbers and types for OptionOverrideProto
+          // message OptionOverrideProto {
+          //   oneof value {
```
mrakitaTT, line 303: Since we expect that value will be `oneof` you should add a check if `value_fields.field_count() == 1` and log+return the error otherwise, because that will mean that our expectation became wrong and we need to know that. Otherwise the current code will simply just take the value from the last field.

You can then also get rid of the for loop and simply just take the first field.

tests/infra/testers/single_chip/model/torch_model_tester.py
```diff
@@ -37,6 +38,12 @@ def __init__(
         self._input_activations: Dict | Sequence[Any] = None
 
         super().__init__(comparison_config, run_mode, Framework.TORCH, compiler_config)
+        # Set custom compile options if provided
+        # Use explicit API for passing compiler options
```
mrakitaTT, line 42: Nit: Add dots at the end of comments.

tests/infra/testers/single_chip/model/torch_model_tester.py
```diff
@@ -37,6 +38,12 @@ def __init__(
...
+        if compiler_config is not None:
+            torch_xla.set_custom_compile_options(
```
mrakitaTT, line 44: I've just realized this is being set on top of the whole `torch_xla` module instead of being set for the specific model compilation. That means if we want to run two models in parallel on two different cards, using different compilation config, it wouldn't be thread safe. We probably won't have such scenario in our tests soon, but it might be a use case we will need to support in the future. FYI @nvukobratTT

tests/torch/single_chip/models/resnet/test_resnet.py
mrakitaTT, line 1: Please add a `@pytest.mark.push` marker for the `test_torch_resnet_inference` test in this file, so it can be run on PR CI to verify your changes, and in general this test should be marked as push similar as the JAX resnet test since this is one of the core models we care about.

mrakitaTT (commented):
@sshonTT Thank you for attending to all comments! Left few more nitpicks and smaller comments, otherwise LGTM.

tests/infra/testers/single_chip/model/torch_model_tester.py
nvukobratTT: I see. This is a valid point. 

@sshonTT, do you have feedback from torch-xla maintainers on this approach for propagating compile options? 

I agree with @mrakitaTT that this isn't a scenario that we'll hit now, but I don't think we're that far from having a user test/use this as well.

tests/torch/single_chip/models/resnet/test_resnet.py
```diff
@@ -59,6 +61,7 @@ def training_tester() -> ResnetTester:
         "https://github.com/tenstorrent/tt-xla/issues/1384"
     )
 )
+@pytest.mark.push
```
mrakitaTT, line 64: Please move this above `@pytest.mark.model_test` since those two are related (they affect in which pipelines test will be run)

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -216,95 +216,148 @@ tt_pjrt_status ClientInstance::compileMlirProgram(
...
-      result[key] = value;
+    if (!key.empty()) {
+      out_compile_options[key] = value;
     }
   }
 
-  return result;
+  return tt_pjrt_status::kSuccess;
 }
-
```
mrakitaTT, line 307: Please return this deleted empty line so that internal namespace is separated

tests/infra/testers/single_chip/model/torch_model_tester.py
mrakitaTT: Makes sense. We should sync with torch-xla maintainers though since we want to be able to upstream everything from our fork to the main torchxla.

mrakitaTT (approved):

tests/infra/testers/single_chip/model/torch_model_tester.py
sshonTT: https://github.com/pytorch/xla/pull/9575 I have already a PR for this to upstream, and I got approved. I am waiting for running ci.

Issue #946, opened 2025-08-06
Add API to generate standalone C++ solution


svuckovicTT:
# Feature Request: Implement API for Generating Standalone C++ and Python TT-NN Solutions

## Description

Add an API that allows users to generate standalone C++ and Python TT-NN solutions from compiled models. This would provide a complete, self-contained solution that users can build and run independently.

## Background

FFE currently has a similar but more rudimentary feature that [generates cpp code](https://github.com/tenstorrent/tt-forge-fe/blob/7349ff2a0720883d0514f233465ee1643d7e3eb5/forge/test/test_api.py#L129-L150) in console. However, we need to generate a full standalone solution rather than just console output.

## Proposed API

The API should allow users to specify output directory and target language:

```python
compiled_model.generate_standalone_solution(dir="output_dir", lang=LANG.CPP)
compiled_model.generate_standalone_solution(dir="output_dir", lang=LANG.PYTHON)
```

> **Note:** This is just an example API - the exact interface is flexible and can be modified as needed.

## Implementation Details

The underlying implementation will be provided by tt-mlir using the [existing APIs](https://github.com/tenstorrent/tt-mlir/blob/c2a645a27dd554b3a97882048721961d9035cc55/tools/tt-alchemist/include/tt-alchemist/tt_alchemist_c_api.hpp#L29-L37).

## Future Improvements

The following improvements should be implemented iteratively after the initial API:

### Improvement 1: Tensor Propagation
Add the ability to propagate tensors (weights, constants, and sample input tensors) to the standalone solution.

**Implementation approach:**
- Tag each tensor in the IR with a unique name using something like `setDiscardableAttr()`
- Serialize tensors to disk with names corresponding to IR names
- Ideally serialize `ttnn::Tensor` objects directly
- tt-mlir team will provide serialization logic if it doesn't exist in ttnn lib

### Improvement 2: Enhanced Code Readability
Improve readability of generated code by attaching meaningful locations to operations.

**Implementation approach:**
- Attach locations containing stack trace, line number, and operation name
- Format similar to previous FFE/tt-torch implementations: `op_name::resnet.py::Resnet::DownsampleBlock::__forward__::16:4`
- Ensure formatting is unambiguous

## Implementation Priority

1. **Phase 1:** Create the basic API with current capabilities
2. **Phase 2:** Implement tensor propagation (Improvement 1)  
3. **Phase 3:** Enhance code readability (Improvement 2)

[...]

PR #1328, opened 2025-09-09
ModuleBuilder refactor to minimize state

sgligorijevicTT:
### Ticket
This PR is one of the prerequisites for #946 

This PR depends on my serialization branch and will be targetting it until it is merged in main, at which point this PR will be rebased and retargeted. Strictly speaking this PR does not depend on changes to serialization, but a followup PR to this one will.

### Problem description
I didn't like that `ModuleBuilder` kept around state.

### What's changed
`ModuleBuilder` was refactored to not store results of `buildModule` as state.
`ExecutableImage` was repurposed as a builder class for use inside `ModuleBuilder`.
There was minor cleanup done to data being store, namely `is_output_scalar` was removed.
Information about number of arguments is now extracted from IR and not flatbuffer. Checks were added to ensure flatbuffer values match.

### Checklist

### DIFF
--- a/inc/common/pjrt_implementation/module_builder/module_builder.h
+++ b/inc/common/pjrt_implementation/module_builder/module_builder.h
@@ -131,13 +131,11 @@ private:
                              std::vector<std::uint32_t> devices_mesh_shape);
 
   // Converts TTIR module to TTNN module.
-  tt_pjrt_status
-  convertFromTTIRToTTNN(const std::string &system_descriptor_path,
-                        mlir::OwningOpRef<mlir::ModuleOp> &mlir_module,
-                        const CompileOptions &compile_options,
-                        ClientInstance *client_instance,
-                        std::vector<std::uint32_t> devices_mesh_shape,
-                        std::string &ttnn_code);
+  tt_pjrt_status convertFromTTIRToTTNN(
+      const std::string &system_descriptor_path,
+      mlir::OwningOpRef<mlir::ModuleOp> &mlir_module,
+      const CompileOptions &compile_options, ClientInstance *client_instance,
+      std::vector<std::uint32_t> devices_mesh_shape, std::string &ttnn_code);
 
   // Creates flatbuffer binary from the built TTNN module.
   tt_pjrt_status createFlatbufferBinary(


src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -8,10 +8,13 @@
 // c++ standard library includes
 #include <cassert>
 #include <cstdlib>
+#include <memory>
 #include <numeric>
 #include <optional>
+#include <variant>
 
 // loguru includes
+#include "common/status.h"
```
mrakitaTT, line 17: not a loguru include

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -8,10 +8,13 @@
 // c++ standard library includes
 #include <cassert>
 #include <cstdlib>
+#include <memory>
 #include <numeric>
 #include <optional>
+#include <variant>
```
mrakitaTT, line 14: you are not using variant

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -8,10 +8,13 @@
 // c++ standard library includes
 #include <cassert>
 #include <cstdlib>
+#include <memory>
```
mrakitaTT, line 11: `<memory>` is already included in the .h file

inc/common/pjrt_implementation/executable_image.h
```diff
@@ -29,31 +32,35 @@
...
 
+namespace module_builder {
+class ModuleBuilder;
+}
+
 // Represents compiled image containing all the required information for its
 // execution.
 class ExecutableImage {
 
+  friend class module_builder::ModuleBuilder;
```
mrakitaTT, line 43: There is absolutely no need for this. I would make the following changes:
- Revert the constructor with fields and `createInstance` method
- Return variables back to `const`
- In `module_builder.cc` in `buildModule` function, every function call like for example `createVHLOModule` should return its result into local variable, and then at the end of `buildModule` function you should create `ExecutableImage` instance from those local variables. I know that it will require you to pass these variables around into other functions, which will cause their number of parameters to grow, but it is better than accessing private variables from the partially created executable image.
- `validate()` function can be removed and validation done in the `ExecutableImage` constructor as it previously was

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -92,79 +97,132 @@ ModuleBuilder::ModuleBuilder()
...
-  if (!tt_pjrt_status_is_ok(m_status)) {
-    return m_status;
+  tt_pjrt_status status;
+  mlir::OwningOpRef<mlir::ModuleOp> mlir_module;
+  status = createVHLOModule(mlir_code, mlir_module);
+  if (!tt_pjrt_status_is_ok(status)) {
+    return {status, nullptr};
+  }
+
+  auto original_mlir_code = std::string(mlir_code);
```
mrakitaTT, line 116: alternative: `std::string original_mlir_code(mlir_code);`

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -759,14 +865,33 @@ bool ModuleBuilder::isUsingShardyManualComputation(
...
+
+  for (size_t output_index = 0; output_index < num_outputs; ++output_index) {
+    output_memory_kinds.emplace_back(
+        MemoryInstance::c_device_memory_kind_name.c_str());
+    output_memory_kinds_sizes.emplace_back(
+        MemoryInstance::c_device_memory_kind_name.size());
+  }
+}
+
+tt_pjrt_status ModuleBuilder::getFirstShardyMeshOp(
```
mrakitaTT, line 882: Please add some vertical spacing in this function, for example:
```
  mlir::sdy::MeshOp found_mesh_op = nullptr;
  module.get().walk([&](mlir::sdy::MeshOp op) {
    found_mesh_op = op;
    return mlir::WalkResult::interrupt();
  });

  if (!found_mesh_op) {
  }

  mesh_op = found_mesh_op;

```

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -759,14 +865,33 @@ bool ModuleBuilder::isUsingShardyManualComputation(
...
+    mlir::sdy::MeshOp &mesh_op) {
+  mlir::sdy::MeshOp found_mesh_op = nullptr;
   module.get().walk([&](mlir::sdy::MeshOp op) {
-    mesh_op = op;
+    found_mesh_op = op;
     return mlir::WalkResult::interrupt();
   });
-  return mesh_op;
+  if (!found_mesh_op) {
+    return tt_pjrt_status::kInternal;
```
mrakitaTT, line 891: I don't think this function should return `tt_pjrt_status` because it is not always the error if the mesh op doesn't exist, so we should leave it to the caller places to decide what to do if the op doesn't exist and report the error, and this function should simply just try to find the mesh op and return nullopt if it doesn't exist.

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -304,57 +378,78 @@ void ModuleBuilder::collectOutputShardingsShardy(
...
+
+  return output_types;
 }
 
-void ModuleBuilder::collectOutputTypes(
+NumArgumentsResult ModuleBuilder::collectNumArguments(
     const mlir::OwningOpRef<mlir::ModuleOp> &module) {
-  m_is_output_scalar.clear();
-  m_output_data_types.clear();
+  DLOG_F(LOG_DEBUG, "ModuleBuilder::collectNumArguments");
```
mrakitaTT, line 411: You probably forgot to remove this log

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -304,57 +378,78 @@ void ModuleBuilder::collectOutputShardingsShardy(
...
 
   std::vector<mlir::func::FuncOp> publicFuncOps = getPublicFuncOps(module);
+  assert(publicFuncOps.size() == 1 && "Expected exactly one public function");
```
mrakitaTT, line 414: All of the asserts in this function are checking for something that is not a bug in our code but instead is coming from the external input, so instead of asserts we should log errors and return `tt_pjrt_status::kInternal`.

mrakitaTT: Btw I would try running nightly on this branch to verify none of the models are hitting these errors, if you already haven't

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -220,24 +284,27 @@ void ModuleBuilder::collectInputShardingsGSPMD(
...
 
   mlir::LogicalResult result =
-      createShardingsFromGSPMD(gspmd_attributes, m_input_shardings);
+      createShardingsFromGSPMD(gspmd_attributes, input_shardings);
   if (result.failed()) {
-    m_status = tt_pjrt_status::kInternal;
+    DLOG_F(ERROR, "Failed to create input shardings from GSPMD attributes");
+    return tt_pjrt_status::kInternal;
   }
+  return tt_pjrt_status::kSuccess;
```
mrakitaTT, line 292: Nit: add blank line above

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -92,79 +97,132 @@ ModuleBuilder::ModuleBuilder()
...
-  collectInputArgumentRoles(mlir_module);
-  collectOutputTypes(mlir_module);
+  NumArgumentsResult num_arguments = collectNumArguments(mlir_module);
+  size_t num_inputs = num_arguments.num_inputs;
+  size_t num_outputs = num_arguments.num_outputs;
+  std::vector<std::vector<std::uint32_t>> output_dimensions =
+      num_arguments.m_output_dimensions;
+  std::vector<size_t> output_ranks = num_arguments.m_output_ranks;
+  std::vector<std::int64_t> output_dimensions_flat =
+      num_arguments.m_output_dimensions_flat;
```
mrakitaTT, line 147: No need to pull these out to local variables, you can simply use `num_arguments.num_inputs` below instead of `num_inputs`, etc. Similar for fields of `num_devices_result` below.

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -92,79 +97,132 @@ ModuleBuilder::ModuleBuilder()
...
+  std::string executable_name = "tt_executable";
+
+  return {tt_pjrt_status::kSuccess,
+          ExecutableImage::createInstance(
+              flatbuffer, std::move(original_mlir_code), std::move(ttir_mlir),
+              std::move(ttnn_mlir), executable_name, num_inputs, num_outputs,
+              output_dimensions, output_ranks, output_dimensions_flat,
+              num_partitions, num_replicas, num_devices_to_utilize, mesh_shape,
+              input_shardings, output_shardings, output_types,
+              std::move(output_memory_kinds),
```
mrakitaTT, line 202: Why not pass other vectors above (`output_dimensions`, `output_ranks`, etc) by rvalue too?

src/common/pjrt_implementation/executable_image.cc
```diff
@@ -24,113 +25,133 @@ namespace tt::pjrt {
...
     make_shared_enabler(
         const tt::runtime::Binary &flatbuffer_binary,
         std::string &&original_mlir_code, std::string &&ttir_mlir_code,
         std::string &&ttnn_mlir_code, std::string &&executable_name,
-        size_t num_partitions, size_t num_replicas,
-        size_t num_devices_to_utilize,
+        size_t num_inputs, size_t num_outputs,
+        std::vector<std::vector<std::uint32_t>> output_dimensions,
+        std::vector<size_t> output_ranks,
+        std::vector<std::int64_t> output_dimensions_flat, size_t num_partitions,
```
mrakitaTT, line 49: You can accept these vectors by rvalue

PR #1330, opened 2025-09-09
Codegen support, v1

sgligorijevicTT:
### Ticket
Adresses core of #946 . However there will be more work left to be done

### Problem description
We want to support exporting models as C++ or Python TTNN code.

### What's changed
Support for codegen was added via tt-alchemist. This means models can be exported as either C++ or Python using compile options. Actual execution of models via C++ will come at a later date.

### Checklist


src/common/pjrt_implementation/module_builder/compile_options.cc
```diff
@@ -23,7 +24,14 @@ CompileOptions CompileOptions::parse(
...
                         .value_or(Backend::Default);
-
+  options.dump_inputs = internal::parseBoolOption(
+      compile_options,
+      "dump_inputs"); // nonexistent value is handled differently based on other
+                      // options, therefore no explicit defaulting with
+                      // value_or.
+  options.export_path =
+      internal::parseStringOption(compile_options, "export_path")
+          .value_or("codegen_solution");
```
svuckovicTT, line 34: This sets the default value of the export path? I'm thinking if it might be better to error out rather than set a default path. If we use a default path, I'm afraid of users overwriting their previous solution unintentionally, whereas having to set a path would proc the user to do it and make it less likely that they'd unintentionally overwrite. Thoughts?

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -934,4 +1052,75 @@ ModuleBuilder::createArgumentTypeMap(
...
+      }
+    }
+  }
+
+  return ""; // Not found
+}
+
+void ModuleBuilder::loadTTAlchemistFunctions() {
+  std::string so_path = findTTAlchemistLibraryPath();
+  if (!so_path.empty()) {
```
svuckovicTT, line 1081: You can invert the conditionals to make this more readable.

```cpp
if (so_path.empty()) {
    DLOG_F(WARNING, "tt-alchemist library not found in Python environment");
    m_alchemist_available = false;
    
    return;
}

if (!m_tt_alchemist_handle) {
  DLOG_F(WARNING, "Cannot load tt-alchemist library: %s", dlerror());
  m_alchemist_available = false;

  return;
}

... rest of the logic ...
```

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -934,4 +1052,75 @@ ModuleBuilder::createArgumentTypeMap(
...
+    m_tt_alchemist_handle = dlopen(so_path.c_str(), RTLD_LAZY);
```
svuckovicTT, line 1082: Before calling `dlopen()`, you'll want to do a `dlerror()` to clean any potential previous errors. Then after calling `dlopen()`, you can call `dlerror()` again to retrieve an error message to print to the user.

svuckovicTT (approved):

inc/common/pjrt_implementation/compile_options.h
```diff
@@ -57,6 +57,14 @@ struct CompileOptions {
   // Backend runtime which should be targeted for compilation and execution.
   BackendRuntime backend = BackendRuntime::TTNNFlatbuffer;
 
+  // Enables saving graph inputs to disk whenever Execute() is called.
+  // This is useful for chisel and codegen. Defaults to false on the default
+  // backend and true on the codegen backend.
+  std::optional<bool> dump_inputs = std::nullopt;
```
ajakovljevicTT, line 63: can we set it to false by default and not use optional, as with other bools?

inc/common/pjrt_implementation/module_builder/module_builder.h
```diff
@@ -283,8 +284,33 @@ class ModuleBuilder {
...
+  // Finds tt-alchemist library path using environment variables
+  std::string findTTAlchemistLibraryPath();
+
+  // Loads tt-alchemist library and function pointers
+  void loadTTAlchemistFunctions();
+
   // MLIR context handle.
   std::unique_ptr<mlir::MLIRContext> m_context;
+
+  // tt-alchemist library handle and function pointers
```
ajakovljevicTT, line 304: Give more context and comments to each of these.

inc/common/pjrt_implementation/module_builder/module_builder.h
```diff
@@ -283,8 +284,33 @@ class ModuleBuilder {
       std::vector<size_t> &&output_memory_kinds_sizes,
       CompileOptions &&compile_options);
 
+  tt_pjrt_status
```
ajakovljevicTT, line 287: Comment what these functions do

src/common/pjrt_implementation/compile_options.cc
```diff
@@ -34,7 +35,13 @@ CompileOptions CompileOptions::parse(
           .value_or(false);
   options.backend = internal::parseBackendOption(compile_options, "backend")
                         .value_or(BackendRuntime::TTNNFlatbuffer);
-
+  auto maybe_export_path =
```
ajakovljevicTT, line 38: do not use auto, but maybe its just easier to return "" instead of using optional?

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -986,8 +1028,162 @@ ModuleBuilder::buildModuleForTTNNCodegenPy(
...
     std::vector<size_t> &&output_memory_kinds_sizes,
     CompileOptions &&compile_options) {
-  DLOG_F(ERROR, "Codegen python backend is not supported yet.");
-  return {tt_pjrt_status::kUnimplemented, nullptr};
+  tt_pjrt_status status = performCodegenPy(ttir_mlir, compile_options);
+  if (!tt_pjrt_status_is_ok(status)) {
+    return {status, nullptr};
+  }
+
+  return {tt_pjrt_status::kSuccess, SOExecutableImage::createInstance(
```
ajakovljevicTT, line 1036: nit: I would separate createInstance and pari creation, it's just more readable:
```
unique_ptr< SOExecutableImage> so_executable_image = ...
return {succes, so_executableimage}
```

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -986,8 +1028,162 @@ ModuleBuilder::buildModuleForTTNNCodegenPy(
...
+    return tt_pjrt_status::kInternal;
+  }
+
+  void *instance = m_tt_alchemist_get_instance();
+  if (!instance) {
+    DLOG_F(ERROR, "Failed to get tt-alchemist instance");
+    return tt_pjrt_status::kInternal;
+  }
+
+  auto input_file = folder + "/ttir.mlir";
```
ajakovljevicTT, line 1101: do not use autos (here and in other places)

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -986,8 +1028,162 @@ ModuleBuilder::buildModuleForTTNNCodegenPy(
...
+    return tt_pjrt_status::kInternal;
+  }
+
+  return tt_pjrt_status::kSuccess;
+}
+
+std::string ModuleBuilder::findTTAlchemistLibraryPath() {
+  // HACK!
+
+  // Option 1: Use VIRTUAL_ENV if available
```
ajakovljevicTT, line 1117: Is there an option2?

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -986,8 +1028,162 @@ ModuleBuilder::buildModuleForTTNNCodegenPy(
...
+void ModuleBuilder::loadTTAlchemistFunctions() {
+  std::string so_path = findTTAlchemistLibraryPath();
+  if (so_path.empty()) {
+    DLOG_F(WARNING, "tt-alchemist library not found in Python environment");
+    m_alchemist_available = false;
+  }
+
+  dlerror(); // Clear any existing error
+  m_tt_alchemist_handle = dlopen(so_path.c_str(), RTLD_LAZY);
+  const char *dlsym_error = dlerror();
```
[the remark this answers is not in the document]
ajakovljevicTT: Currently, you can leave stuff semi initialized

inc/common/pjrt_implementation/compile_options.h
sgligorijevicTT: Dump inputs wasn't supposed to be included in this PR, seems like I made a mistake while splitting or rebasing. I will delete it completely for now as it is unused, and we can discuss further on the next PR.

Context for why I left it as optional is in the comment above. On a branch where I have dumping, I do `dump_inputs.value_or(false)` on the flatbuffer path and `dump_inputs.value_or(true)` on the SO path. I considered that cleaner than having logic in parse and writing to a variable twice, however in retrospect it is probably cleaned to keep the logic contained to once place, and parse has to contain a bit of logic anyway.

src/common/pjrt_implementation/module_builder/module_builder.cc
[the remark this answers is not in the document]
sgligorijevicTT: If it's a separate function how can I forge it's caller to return? I could make a macro, but that's ugly. FYI we have very similar type of boilerplate in buildModule. 
Besides, the iferr-return chains are pretty self contained here.

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -986,8 +1027,179 @@ ModuleBuilder::buildModuleForTTNNCodegenPy(
...
+ModuleBuilder::performCodegenCpp(std::string_view ttir_mlir,
+                                 const CompileOptions &compile_options) {
+  std::string folder = compile_options.export_path;
+  std::filesystem::create_directories(folder);
+
+  std::ofstream ttir_file(folder + "/ttir.mlir");
+  ttir_file << ttir_mlir;
+  ttir_file.close();
+
+  if (!m_tt_alchemist_handles.has_value()) {
```
mrakitaTT: I would first check this, before doing anything else in this function

PR #1331, opened 2025-09-09
Examples of codegen

sgligorijevicTT:
### Ticket
Adresses core of https://github.com/tenstorrent/tt-xla/issues/946 . However there will be more work left to be done

Depends on https://github.com/tenstorrent/tt-xla/pull/1330 .

### Problem description
We want to demonstrate how to use Codegen from the frontend.

### What's changed
Examples of how tt-alchemist can be used were added for both jax and torch.

This PR was subsequently edited to also include examples using compile options

### Checklist

### DIFF
--- a/examples/jax/codegen_via_options_example.py
+++ b/examples/jax/codegen_via_options_example.py
@@ -18,6 +18,7 @@ class Model(nnx.Module):
 
     def __call__(self, x):
         x = self.A(x)
+        x = nnx.relu(x)
         x = self.B(x)
         x = nnx.tanh(x)
         return jnp.sum(x**2)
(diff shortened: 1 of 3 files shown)


examples/pytorch/codegen__via_serialize_example.py
svuckovicTT, line 1: Double underscore

examples/pytorch/codegen__via_serialize_example.py
```diff
@@ -0,0 +1,59 @@
...
+    output_dir="model/cpp",
+    local=False,
+)
+
+# Generate Python code from the TTIR
+tt_alchemist.generate_python(
+    input_file="model/model_ttir.mlir",
+    output_dir="model/py",
+    local=False,
+)
```
svuckovicTT, line 59: Just to double-check, is this our only option for torch? Ideally, we'd avoid having `tt_alchemist` as an import, and rather add an API/option to used the codegen as a backend like in jax jit. Maybe add a function in `tt_torch`?

lnspirafirmaGPK (changes requested):
Request change

PR #1341, opened 2025-09-09
Remove "sanitize test_mark" in call-test.yml no longer needed since produce_cicd_data bug fixed

kmabeeTT:
### Ticket
https://github.com/tenstorrent/tt-github-actions/issues/78

### Problem description
 - produce_cicd_data was tripping up on artifacts with spaces in name, had to put in a temporary workaround to merge previous changes

### What's changed
- The bug got fixed in tt-github-actions today (thx Vladamir), can remove local workaround in tt-xla
- Tested locally on branch with workaround removed can see artifacts parsed and results in superset DB https://github.com/tenstorrent/tt-xla/actions/runs/17593394271/job/49979519823

### Checklist
- [x] Tested small set of affected tests on branch

### DIFF
--- a/.github/workflows/call-test.yml
+++ b/.github/workflows/call-test.yml
@@ -235,27 +235,18 @@ jobs:
                --junitxml=${{ steps.strings.outputs.test_report_path }} \
                2>&1 | tee pytest.log
 
-    # produce_cicd_data silently drops artifacts with spaces in the name, must replace spaces with underscores
-    # until https://github.com/tenstorrent/tt-github-actions/issues/78 is fixed
-    - name: Sanitize test_mark
-      id: sanitize
-      shell: bash
-      run: echo "test_mark=${TEST_MARK// /_}" >> $GITHUB_OUTPUT
-      env:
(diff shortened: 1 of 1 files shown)


jameszianxuTT (approved):
I think it's acceptable to keep this the way it is, I don't like spaces in filenames (which will end up being baked in to the artifact filename). 

It might cause other parsing issues for downstream consumers of this artifact, like the issue this workaround was originally added to fix.

However if others are in favor of removing this sanitation change, that is fine too.

vmilosevic:
> I think it's acceptable to keep this the way it is, I don't like spaces in filenames (which will end up being baked in to the artifact filename).
> 
> It might cause other parsing issues for downstream consumers of this artifact, like the issue this workaround was originally added to fix.
> 
> However if others are in favor of removing this sanitation change, that is fine too.

I like sanitized strings better, as they are safer in bash, but we can make it work with spaces too, whatever you prefer.

kmabeeTT:
kk I am going to keep things as-is in this case and close this PR. Thx guys.

PR #1348, opened 2025-09-10
Run only failed tests on retry

nsumrakTT:
### Problem description
When test job is rerun after an error, all tests are run.

### What's changed
Run only tests that failed in previous attempt

### Checklist

Push test run with one test made to assert always:
Attempt 1 (run all tests): https://github.com/tenstorrent/tt-xla/actions/runs/17618570450/attempts/1
Attempt 2 (run only failed from previous attempt): https://github.com/tenstorrent/tt-xla/actions/runs/17618570450/job/50059828156

tt_forge_models run: https://github.com/tenstorrent/tt-xla/actions/runs/17673786364/attempts/1
I've rerun job p150,1: https://github.com/tenstorrent/tt-xla/actions/runs/17673786364/job/50244825994
It run just 4 tests that previously failed.
Then, rerun of job p150,16: https://github.com/tenstorrent/tt-xla/actions/runs/17673786364/job/50245707965
Here it found run from the first attempt (because the second attempt is just one job) and rerun just 4 jobs that failed.


.github/workflows/call-test.yml
```diff
@@ -156,6 +156,47 @@ jobs:
         tar xf artifact.tar
         rm -f artifact.tar
 
+    - name: Check for second attempt
+      id: check-second-attempt
```
vmilosevic, line 160: Can we rename this to "check re-run attempt" or similar, as it handles all re-runs, not just second

Issue #1366, opened 2025-09-11
Port embedding API to plugin


AleksKnezevic:
We need to port embedding API support in the vLLM plugin that will be part of this repo #1365.

mmanzoorTT:
vLLM Embedding API update
- Pre-compilation of multiple binaries for different input sequence lengths.
- Process multiple batches; A batch contains one or more input sequences. 
- Next step is to include `attention_mask`; currently, we are relying on tt-metal default `attention_mask`

AleksKnezevic:
@LPanosTT added that part, so he can better comment.

LPanosTT:
@bgoelTT The inputs are padded to the next highest power of 2 when vllm receives them and it runs that pre-compiled version of the model. The user entering prompts doesn't need to pad anything themselves.

LPanosTT:
@bgoelTT when vllm receives the request it pads them automatically to the next power of 2 before executing the model.

mmanzoorTT:
A little bit more detail: `execute_model` function receives `SchedulerOutput` object which contains the unpadded input tokens as `list` (This input tokens contain one or more input sequences). `_update_states` function create/add requests for these input sequence(s) and then `_prepare_inputs` function generates padded input tokens which is then fed to the model in `execute_model` function. `max_num_seqs` flag is used to control the max number of sequences in a batch. 

For example (max_num_seqs=3 to process three input prompts in a batch), If we have 3 prompts of length 10, 12, and 14 then vllm will give us unpadded input (list of length 36) and we create a tensor of [1 x 64] (including padding)

AleksKnezevic:
@mmanzoorTT, can we close?

Issue #1365, opened 2025-09-11
Add vLLM plugin to tt-xla


AleksKnezevic:
Please add plugin to enable vLLM through tt-xla. This should build as a separate wheel that we release and users can pip install.

LPanosTT:
The branch `lpanos/vllm_plugin_tt` has the functionality to run the embedding model: `Qwen/Qwen3-Embedding-4B` up to 16k tokens.

The plugin contains the functionality to perform embedding requests (like the model above performs) as well as generations based on a prompt.

PR #1377, opened 2025-09-12
Edit vLLM plugin code to support TT static cache

LPanosTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/1365, https://github.com/tenstorrent/tt-xla/issues/1366

### What's changed
Modified the TPU implementation clone to support TT hardware using static cache.
- Upgraded torch, torchvision, and transformers versions to be in line with what the latest vllm release requires.
- Add opt generation test

- The original implementation (which this PR modifies) imports some functions which I have now pasted here (i.e. in `model_runner.py` I've put the `copy_kv_blocks` function.
    - This is because the original implementation was taken from vllm/main. However, in the vllm release we are using `vllm==0.10.1.1`, these functions are still located in the TPU runner. So, to keep the model runner as similar to the TPU runner we are using, I have included `copy_kv_blocks` in `model_runner.py`

- Added example scripts for serving/using vllm as a sercvice. One for Qwen3-Embedding-4B, and a chatbot example for TinyLlama-1.1B

- Updated PYTHONPATH in `venv/activate` so that we do not need to install `tt_torch` and `vllm_tt` explicitly during development.

Embedding API support
  vLLM TPU implementation does not support pooling models. 
  - Added pooling runner and pooling input batch to support pooling model inference.
  - Added tests to verify different functionalities like
    - Multiple batches with one or more inputs
    - Padded and non-padded inputs
    - Reduced embedding output dimensionality

### Checklist

### DIFF
--- a/.github/workflows/call-test.yml
+++ b/.github/workflows/call-test.yml
@@ -103,6 +103,7 @@ jobs:
           .test_durations
           pytest.ini
           python_package/requirements.txt
+          integrations/vllm_plugin/requirements-vllm-plugin.txt
           tests
           venv
 


integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1887,24 +1959,28 @@ def prepare_structured_decoding_input(
         self.grammar_bitmask_cpu.zero_()
         self.require_structured_out_cpu.zero_()
 
-        sorted_struct_requests = sorted(
```
LPanosTT, line 1890: The changes in `prepare_structured_deconding_input` are not my own, they are differences between vllm/main, and the tpu_model_runner.py as it stands in vllm release 0.10.1.1

nsumrakTT (commented):
When is vllm plugin used?
Should it be released along tt_xla wheel?
Is there any reason why it should be separate wheel from existing tt_xla wheel?

nsumrakTT:
Thanks for explanation @LPanosTT !
Regarding 1st point, I don't see how test is executed if wheel is not installed on test machine.
Regarding 2nd point (lvvm wheel should be released), @jmcgrathTT please mind this development.

@LPanosTT Please remove build-vllm-tt-release job, it is not needed.

LPanosTT:
@nsumrakTT No worries!
For point 1. I am currently attempting to get the wheel installed on the test machine for CI. For development, installation of the wheel will not be necessary as I have edited the `activate` script so that it adds the directory containing the plugins source code to the PYTHONPATH. `vllm` will be able to find it this way too.

I have removed that job now :)

jmcgrathTT:
>Regarding 2nd point (lvvm wheel should be released), @jmcgrathTT please mind this development.
Thanks @nsumrakTT, created a ticket to track here

https://github.com/tenstorrent/tt-forge/issues/444

Looks like we are good for pick up at least
image

We will need to address this Python ABI issue, but we can do that in a follow-up PR if needed.

image

It seems that the ABI issue was caused by not having a pyproject.toml if `requires-python = ">=3.11"` in this folder or setup.py with `python_requires=">=3.11",`
https://github.com/tenstorrent/tt-xla/tree/main/integrations/vllm_plugin

LPanosTT:
@jmcgrathTT Thanks for pointing this out. I'll add the line to `integrations/vllm_plugin/setup.py`.

LPanosTT:
@pilkicTT So the plugin does contain an entry point that vllm will gather and use. The reason we have a separate wheel is so that we may support our hardware without modifying the vllm source code. vLLM intends out-of-tree hardware support to be implemented this way, more on that [here](https://docs.vllm.ai/en/v0.10.1.1/design/plugin_system.html). For example, Ascend has a plugin for vLLM just like we will be making: https://github.com/vllm-project/vllm-ascend

nsumrakTT (approved):
@LPanosTT Great, thank you!
On test machines we still have repo checked out so if it is python only project and requirements are already satisfied (by tt-xla or in docker), you can also consider doing local install instead of downloading and installing wheel.

integrations/vllm_plugin/vllm_tt/worker.py
```diff
@@ -225,9 +214,15 @@ def determine_available_memory(self) -> int:
             total_memory_size = device_usage[0].total_memory
             current_mem = device_usage[0].memory_usage
         else:
-            m = xm.get_memory_info(self.device)
-            total_memory_size = m["bytes_limit"]
-            current_mem = m["bytes_used"]
+            # @LPanosTT: TODO we should find out if/how we can implement the PJRT API function(s) necesarry to exegcute xm.get_memory_info, and implement them.
```
sgligorijevicTT, line 217: @ajakovljevicTT would you know more about this?
@LPanosTT can you open an issue for this and link it here?

LPanosTT: https://github.com/tenstorrent/tt-xla/issues/1414

I've added this to the comment in the code as well.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -788,7 +868,9 @@ def _prepare_inputs(self, scheduler_output: "SchedulerOutput", start_index: int)
             query_start_loc = self.query_start_loc_cpu[
                 : self.num_reqs_max_model_len + 1
             ].to(self.device)
-            seq_lens = self.seq_lens_cpu[: self.num_reqs_max_model_len].to(self.device)
+            seq_lens = self.seq_lens_cpu[
+                : self.num_reqs_max_model_len
+            ]  # .to(self.device)
```
AleksKnezevic: why is to.device commented out? Same q below.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1351,6 +1417,7 @@ def _dummy_run(self, num_tokens: int, num_reqs: int, num_blocks: int) -> None:
             out = self.model(
                 input_ids=input_ids, positions=position_ids, inputs_embeds=inputs_embeds
             )
+            xm.mark_step()
```
AleksKnezevic: shy the sync point?

venv/requirements-dev.txt
```diff
@@ -31,9 +31,9 @@ soundfile
 tabulate
 timm
 # torchvision 0.23.0 causes torch v2.8.0 to be installed, which isn't supported by our torch_xla whl
-torchvision==0.22.0
+torchvision==0.22.1
 torchxrayvision
-transformers>=4.50.0,<4.53.0
+transformers==4.55.0
```
AleksKnezevic: @jazpurTT's changes to uplift transformers rely on torch==2.9.0. Don't make the global change here, it can go with his PR.

integrations/vllm_plugin/vllm_tt/platform.py
```diff
@@ -23,22 +23,40 @@
 
 logger = init_logger(__name__)
 
-USE_TPU_COMMONS = False
+from torch_xla import runtime as xrt
+
+
+@dataclass
+class TTConfig:
+    enable_const_eval: bool = True
```
AleksKnezevic: Can you link to the [issue](https://github.com/tenstorrent/tt-mlir/issues/3888) that would allow us to enable const-eval through vLLM.

AleksKnezevic (approved):
Overall looks great. I left a few comments to address before merging.

integrations/vllm_plugin/vllm_tt/model_runner.py
LPanosTT: Since the view op changes are not mainlined in tt-mlir, keeping the seq lens on host here is a workaround to solve the issue regarding in-place reshape as that single reshape graph is no longer compiled.

integrations/vllm_plugin/vllm_tt/pooling_runner.py
```diff
@@ -0,0 +1,1910 @@
+# SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
mmanzoorTT: We are keeping these two `runners` separate so we can work independently and simultaneously. We will unify both `runners` in subsequent PR; both `runners` will be stable enough by that time.

integrations/vllm_plugin/vllm_tt/model_runner.py
AleksKnezevic: thanks. Can you please remove the comment? If it's worthwhile placing the seq_lens on device, you can add a task to follow up.

LPanosTT: I have removed it.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1047,11 +1120,13 @@ def execute_model(
                 self.vllm_config,
                 num_tokens=scheduler_output.total_num_scheduled_tokens,
             ):
+
                 hidden_states = self.model(
                     input_ids=input_ids,
                     positions=self.position_ids,
                     inputs_embeds=inputs_embeds,
                 )
+                xm.mark_step()
```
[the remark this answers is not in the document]
LPanosTT: I am testing now.

inc/common/pjrt_implementation/compile_options.h
```diff
@@ -68,6 +68,12 @@ struct CompileOptions {
   // This is useful for chisel and codegen.
   bool dump_inputs = false;
 
+  // Path that will contain any exported artifacts.
+  // This includes: codegen solutions, graph inputs and intermediate IRs.
+  // Setting this will enable IR dumping.
```
mrakitaTT: This is comment from another compile option below, seems like a bad merge

python_package/requirements.txt
```diff
@@ -1,4 +1,5 @@
 jax==0.7.1
 jaxlib==0.7.1
-torch==2.7.0
-torch-xla@https://pypi.eng.aws.tenstorrent.com/torch-xla/torch_xla-2.9.0%2Bgit1f21467-cp311-cp311-linux_x86_64.whl
+torch==2.7.1
+requests
```
mrakitaTT: please put `requests` above `torch` (so the requirements are sorted alphabetically and also to keep `torch` and `torch-xla` close together)

mrakitaTT (approved):

jameszianxuTT (approved):

PR #1311, opened 2025-09-08
Add PJRT_Executable_Fingerprint API implementation

mstojkovicTT:
### Ticket
fixes: #495 

### What's changed
  - Implemented `PJRT_Executable_Fingerprint` API in `executable_instance.cc`
  - Added fingerprint generation to `ExecutableImage` using hash of:
    - MLIR code
    - Compile options (optimizer, bfp8_conversion)
    - tt-mlir compiler version
  - Updated `ExecutableImage` constructor to store compile options and generate fingerprint
  - Removed stub from `stubs.inc`
  - Added missing include guards to `compile_options.h`

PR #1395, opened 2025-09-15
PJRT_Executable_Fingerprint refactor

mstojkovicTT:
### Ticket
Addressing comments from #1311

### What's changed
- `m_parsed_compile_options` added to `ModuleBuilder` to store parsed options
- added `getParsedCompileOptions()` to expose parsed options
- updated `ClientInstance` to remove duplicate parsing, now using getter from `ModuleBuilder`
- `ExecutableImage` now accepts `const CompileOptions&` instead of `&&` since both `ModuleBuilder` and `ExecutableImage` need to keep copies
- Added debug log showing compile options during module building
- centralized `toString()` function inside `module_builder`, ensuring it stays in sync when new options are added
- now using `toString()` for `generateFingerprint() ` instead of creating it there

### DIFF
--- a/inc/common/pjrt_implementation/executable_image.h
+++ b/inc/common/pjrt_implementation/executable_image.h
@@ -47,7 +47,7 @@ public:
           &output_sharding,
       const std::vector<bool> &is_output_scalar,
       const std::vector<PJRT_Buffer_Type> &expected_output_data_types,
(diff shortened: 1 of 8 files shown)


mstojkovicTT:
Closing this as its outdated, gave more contex to @sgligorijevicTT

Issue #952, opened 2025-08-07
Change device opening/closing strategy


mrakitaTT:
Change device opening/closing strategy to open parent mesh of all devices when PJRT client loads devices info and close them when PJRT client is destroyed, and open submeshes for optimizer and execute. Currently devices are opened on each execute and closed afterwards, which kills performance for continuous inference/training over batches.

One thing to keep in mind is that `PJRT_LoadedExecutable_Execute` is meant to be async, so multiple executions can happen at the same time and we need to ensure that we don't open overlapping submeshes from executions or optimizer.

We also need to investigate how best to pass the submesh to optimizer, some proposed approaches can be seen in the picture:

<img width="2730" height="1516" alt="Image" src="https://github.com/user-attachments/assets/2b0a2ba3-c16a-4a50-95fe-6c3a2986ad64" />

PR #1417, opened 2025-09-17
[pjrt] reuse device if possible

### LINKED ISSUE #1415: Reuse the device across multiple executions
We are currently, opening/closing the device on each execution of a model. This is inefficient and doesn't allow us to leave tensors on the device, making subsequent executions of the same model unnecessarilly slow - since each execution will need to transfer weights from the host to device.

Implement a simple mechanism to open the device once and reuse it. Complex multi chip scenarios will be handled in separate issues/PRs. There are some problems with using sub-meshes and the reshape APIs provided by the `MeshDevice`.

pilkicTT:
We are currently opening/closing the device on each `LoadedExecutableInstance::execute()` call.

This change enables us to open the device in advance and reuse it, for example across multiple executions of the same model. This will allow us to leave tensors (weights/constants) on the device, improving the performance on subsequent executions - avoid constantly transferring tensors from host to device.

We open the device when creating the `ClientInstance`. Then when executing the models, in `LoadedExecutableInstance::execute()`, we verify the wanted mesh shape for the device. If the device mesh shape matches the target mesh shape, we simply reuse the previously opened mesh device. In case we need a different mesh shape, we will close and reopen the mesh device with wanted mesh shape.

There were issues with using sub-meshes and also with reshaping mesh device. Hence, I've decided to go with the simplest approach which works.

Issue #952
Closes #1415

### DIFF
--- a/inc/common/pjrt_implementation/client_instance.h
+++ b/inc/common/pjrt_implementation/client_instance.h
@@ -11,6 +11,7 @@
 #include "xla/pjrt/c/pjrt_c_api.h"
 
 // c++ standard library includes
+#include <cstdlib>
 #include <memory>
 #include <string>
 #include <vector>
@@ -75,6 +76,15 @@ public:
     return m_addressable_memories_raw;
   }
 
+  tt::runtime::Device getParentMesh() const { return *m_parent_mesh; }
+
+  // Returns the mesh device of the provided shape. If there is already opened
+  // mesh device within this client instance and its shape matches the provided
+  // shape, it is returned. Otherwise, we close any previously opened mesh
+  // device and open a new one with the provided shape.
+  tt::runtime::Device
+  getOrCreateMeshDevice(const std::vector<uint32_t> &target_mesh_shape);
+
   // Compiles given mlir program.
   tt_pjrt_status compileMlirProgram(
       const PJRT_Program *mlir_program,
@@ -93,6 +103,11 @@ protected:
 private:
   tt_pjrt_status populateDevices();
   tt_pjrt_status populateMemories();
+
+  // Wrapper method around `tt::runtime::openMeshDevice` that also handles
+  // setting fabric config when needed.
+  tt::runtime::Device
+  openMeshDevice(::tt::runtime::MeshDeviceOptions options = {});
 
   std::unique_ptr<Platform> platform_;
 
@@ -140,6 +155,8 @@ private:
   static std::unordered_map<std::string, std::string>
   extractCustomProtobufFields(
       const google::protobuf::UnknownFieldSet &unknown_fields);
+
+  std::optional<tt::runtime::Device> m_parent_mesh;
 };
 
 namespace internal {
--- a/src/common/pjrt_implementation/client_instance.cc
+++ b/src/common/pjrt_implementation/client_instance.cc
@@ -13,9 +13,12 @@
 // c++ standard library includes
 #include <cstddef>
 #include <filesystem>
+#include <optional>
+#include <sstream>
 
 // tt-mlir includes
 #include "tt/runtime/runtime.h"
+#include "tt/runtime/types.h"
 
 // third-party includes
 #include <google/protobuf/io/coded_stream.h>
@@ -34,7 +37,8 @@ namespace tt::pjrt {
 
 ClientInstance::ClientInstance(std::unique_ptr<Platform> platform)
     : platform_(std::move(platform)), m_system_descriptor(nullptr),
-      m_module_builder(std::make_unique<module_builder::ModuleBuilder>()) {
+      m_module_builder(std::make_unique<module_builder::ModuleBuilder>()),
+      m_parent_mesh(std::nullopt) {
   DLOG_F(LOG_DEBUG, "ClientInstance::ClientInstance");
 
   // TODO(mrakita): Add support for multi-process environment. Process index is
@@ -51,6 +55,10 @@ ClientInstance::ClientInstance(std::unique_ptr<Platform> platform)
 
 ClientInstance::~ClientInstance() {
   DLOG_F(LOG_DEBUG, "ClientInstance::~ClientInstance");
+
+  if (m_parent_mesh.has_value()) {
+    tt::runtime::closeMeshDevice(*m_parent_mesh);
+  }
 
   std::remove(m_cached_system_descriptor_path.data());
 }
@@ -130,7 +138,32 @@ tt_pjrt_status ClientInstance::populateDevices() {
     return tt_pjrt_status::kInternal;
   }
 
+  m_parent_mesh = openMeshDevice();
+
   return tt_pjrt_status::kSuccess;
+}
+
+tt::runtime::Device
+ClientInstance::openMeshDevice(::tt::runtime::MeshDeviceOptions options) {
+  auto num_devices = m_devices.size();
+  if (options.meshShape.has_value()) {
+    num_devices = static_cast<size_t>(
+        std::accumulate(options.meshShape->begin(), options.meshShape->end(), 1,
+                        std::multiplies<std::uint32_t>{}));
+  }
+
+  // NOTES:
+  // - this should probably be set automatically by the mlir runtime.
+  // - it looks like metal context is being reinitialized each time we
+  // open/close the device, so we need to set the fabric config each time
+  // (even if we always set it to the same value).
+  if (num_devices > 1) {
+    tt::runtime::setFabricConfig(tt::runtime::FabricConfig::FABRIC_1D);
+  } else {
+    tt::runtime::setFabricConfig(tt::runtime::FabricConfig::DISABLED);
+  }
+
+  return tt::runtime::openMeshDevice(options);
 }
 
 tt_pjrt_status ClientInstance::populateMemories() {
@@ -166,6 +199,7 @@ tt_pjrt_status ClientInstance::compileMlirProgram(
 
   tt_pjrt_status compile_status = m_module_builder->buildModule(
       mlir_code, m_cached_system_descriptor_path, compile_options);
+
   if (!tt_pjrt_status_is_ok(compile_status)) {
     return compile_status;
   }
@@ -206,8 +240,8 @@ tt_pjrt_status ClientInstance::compileMlirProgram(
           m_module_builder->getNumDevicesToUtilize());
 
   std::unique_ptr<LoadedExecutableInstance> executable =
-      LoadedExecutableInstance::createInstance(executable_image,
-                                               std::move(addressable_devices));
+      LoadedExecutableInstance::createInstance(
+          executable_image, std::move(addressable_devices), this);
 
   // Releasing the ownership to the PJRT API caller since the caller is
   // responsible for calling `PJRT_LoadedExecutable_Destroy` on the executable.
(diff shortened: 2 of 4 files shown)


src/common/pjrt_implementation/client_instance.cc
```diff
@@ -305,6 +339,54 @@ ClientInstance::extractCustomProtobufFields(
...
+         "reshaping mesh device - %s -> %s",
+         to_string(parent_mesh_shape).c_str(),
+         to_string(target_mesh_shape).c_str());
+
+  // NOTE: Due to some issues hit when testing, instead of using the reshape
+  // mesh API, we are closing and re-opening the device with the wanted mesh
+  // shape. This should be revisited in the future.
+  //
+  // Additionally, we are supposed to utilize sub-meshes if the target mesh
+  // shape is is contained within the already opened parent mesh. Also, in case
```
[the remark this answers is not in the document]
ajakovljevicTT: I think it should still be linked here.

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -359,6 +393,54 @@ tt_pjrt_status ClientInstance::extractCustomProtobufFields(
...
+  // we are running multiple models on different parts of the mesh (pipeline
+  // parallel). However, similar as to the case with reshape API, there were
+  // some issues when testing sub-meshes, so for now we are always closing and
+  // re-opening the whole mesh.
+  tt::runtime::closeMeshDevice(*m_parent_mesh);
+
+  auto options = tt::runtime::MeshDeviceOptions{.meshShape = target_mesh_shape};
```
ajakovljevicTT, line 439: can we set the real type of options here instead of auto?

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -359,6 +393,54 @@ tt_pjrt_status ClientInstance::extractCustomProtobufFields(
...
+  m_parent_mesh = openMeshDevice(options);
+  return *m_parent_mesh;
```
ajakovljevicTT, line 441: nit: return should be separate with a newline

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -130,9 +138,34 @@ tt_pjrt_status ClientInstance::populateDevices() {
...
+                        std::multiplies<std::uint32_t>{}));
+  }
+
+  // NOTES:
+  // - this should probably be set automatically by the mlir runtime.
+  // - it looks like metal context is being reinitialized each time we
+  // open/close the device, so we need to set the fabric config each time
+  // (even if we always set it to the same value).
+  if (num_devices > 1) {
+    tt::runtime::setFabricConfig(tt::runtime::FabricConfig::FABRIC_1D);
```
mrakitaTT, line 161: Can you please add some note regarding usage of this particular `FabricConfig` value?

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -130,9 +138,34 @@ tt_pjrt_status ClientInstance::populateDevices() {
...
+  } else {
+    tt::runtime::setFabricConfig(tt::runtime::FabricConfig::DISABLED);
+  }
+
+  return tt::runtime::openMeshDevice(options);
```
mrakitaTT, line 166: Is default behavior when `options.meshShape` is not set to open all devices? Shouldn't we set it to `[1, num_devices]`?

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -130,9 +138,34 @@ tt_pjrt_status ClientInstance::populateDevices() {
     return tt_pjrt_status::kInternal;
   }
 
+  m_parent_mesh = openMeshDevice();
+
   return tt_pjrt_status::kSuccess;
 }
 
+tt::runtime::Device
+ClientInstance::openMeshDevice(::tt::runtime::MeshDeviceOptions options) {
```
mrakitaTT, line 147: We tend to sort function definitions top->down, in order they are called, so please move this below `getOrCreateMeshDevice` definition

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -359,6 +393,56 @@ tt_pjrt_status ClientInstance::extractCustomProtobufFields(
   return tt_pjrt_status::kSuccess;
 }
 
+template <typename T> std::string to_string(const std::vector<T> vec) {
```
mrakitaTT, line 396: Looks like a good candidate for the utility function. In situations like this when the function itself doesn't warrant specific file, I create generic `utils.h/cpp` file (if it doesn't already exist) and move it there, and then once there are couple more functions there with a similar theme I extract them into separate file. For example that is how `inc/common/pjrt_implementation/data_type_utils.h` became.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -96,28 +103,50 @@ LoadedExecutableInstance::execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+  if (device_instance &&
+      !(device_ids.size() == 1 &&
+        *device_ids.begin() == device_instance->getGlobalDeviceId())) {
+    DLOG_F(ERROR, "Input buffers are placed on a different device than the one "
+                  "specified in the execute_device argument");
+    return tt_pjrt_status::kInternal;
+  }
+
+  tt::runtime::Device runtime_device =
+      m_client_instance->getOrCreateMeshDevice(devices_mesh_shape);
```
mrakitaTT, line 134: Seems like you could've just replaced last couple lines of the previously existing `LoadedExecutableInstance::openDevices` function with this line, instead of porting all of its code here :) You'll just need to update the function name, maybe it can also be `getOrCreateMeshDevice`.

inc/common/pjrt_implementation/client_instance.h
```diff
@@ -94,6 +104,11 @@ class ClientInstance {
   tt_pjrt_status populateDevices();
   tt_pjrt_status populateMemories();
 
+  // Wrapper method around `tt::runtime::openMeshDevice` that also handles
+  // setting fabric config when needed.
+  tt::runtime::Device
+  openMeshDevice(::tt::runtime::MeshDeviceOptions options = {});
```
mrakitaTT, line 110: Subjective nit: I don't like using default parameter values if there is no clear benefit, because it forces me to remember default behavior when I am reading the code, for example here I have to remember that if `openMeshDevice` is called with no options passed it will open all devices. Since you call this function with desired default behavior only in one place so there would be no save on code duplication, I would either change it to not accept default value and pass specific value at the call place, or create two separate functions - one which accepts specific value and the other called `openMeshFromAllDevices` which doesn't accept any parameter but from the function name it is obvious what it does.

Also, since you only need the mesh shape as the input to this function and not the full `MeshDeviceOptions`, I would change it to accept mesh shape as the input parameter and constructs `MeshDeviceOptions` inside.

mrakitaTT (commented):
Looks good overall, left few smaller comments/nitpicks :)

inc/common/pjrt_implementation/client_instance.h
```diff
@@ -75,6 +76,15 @@ class ClientInstance {
     return m_addressable_memories_raw;
   }
 
+  tt::runtime::Device getParentMesh() const { return *m_parent_mesh; }
```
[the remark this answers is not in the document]
mrakitaTT: Actually I've just realized this getter is not used at all, you can just remove it and add it later if you need it

inc/common/pjrt_implementation/data_type_utils.h
```diff
@@ -11,8 +11,8 @@
 // tt-mlir includes
 #include "tt/runtime/types.h"
 
-#ifndef TT_XLA_INC_COMMON_PJRT_IMPLEMENTATION_UTILS_H_
-#define TT_XLA_INC_COMMON_PJRT_IMPLEMENTATION_UTILS_H_
+#ifndef TT_XLA_INC_COMMON_PJRT_IMPLEMENTATION_DTYPE_UTILS_H_
+#define TT_XLA_INC_COMMON_PJRT_IMPLEMENTATION_DTYPE_UTILS_H_
```
mrakitaTT, line 15: Good catch!

inc/common/pjrt_implementation/utils.h
```diff
@@ -0,0 +1,25 @@
...
+
+#ifndef TT_XLA_INC_COMMON_PJRT_IMPLEMENTATION_UTILS_H_
+#define TT_XLA_INC_COMMON_PJRT_IMPLEMENTATION_UTILS_H_
+
+#include <sstream>
+#include <vector>
+
+namespace tt::pjrt::utils {
+
+template <typename T> std::string to_string(const std::vector<T> vec) {
```
mrakitaTT, line 13: Please add a header comment with the description

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -359,6 +371,71 @@ tt_pjrt_status ClientInstance::extractCustomProtobufFields(
...
+  }
+
+  std::vector<uint32_t> parent_mesh_shape =
+      tt::runtime::getMeshShape(*m_parent_mesh);
+
+  if (parent_mesh_shape == target_mesh_shape) {
+    DLOG_F(LOG_DEBUG,
+           "ClientInstance::getOrCreateMeshDevice - reusing "
+           "already opened mesh device %s",
+           utils::to_string(parent_mesh_shape).c_str());
```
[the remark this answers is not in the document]
pilkicTT: it doesn't work, well at least in the debug build :smile:

added the include... thanks!

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -130,9 +138,34 @@ tt_pjrt_status ClientInstance::populateDevices() {
...
+        std::accumulate(options.meshShape->begin(), options.meshShape->end(), 1,
+                        std::multiplies<std::uint32_t>{}));
+  }
+
+  // NOTES:
+  // - this should probably be set automatically by the mlir runtime.
+  // - it looks like metal context is being reinitialized each time we
+  // open/close the device, so we need to set the fabric config each time
+  // (even if we always set it to the same value).
+  if (num_devices > 1) {
```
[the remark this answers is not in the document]
pilkicTT: yup, we'll need submeshes... for now, only covering the basics...

src/common/pjrt_implementation/client_instance.cc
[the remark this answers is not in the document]
mrakitaTT: I think chatgpt is on to something :)

Screenshot 2025-09-26 at 04 31 29

tests/jax/single_chip/models/roformer/chinese_base/test_roformer_chinese_base.py
```diff
@@ -29,7 +30,10 @@
 
 @pytest.fixture
 def inference_tester() -> RoFormerTester:
-    return RoFormerTester(MODEL_PATH)
+    # Reduced PCC threshold - #1454
```
mrakitaTT, line 33: You should also update the `bringup_status` of `test_roformer_chinese_base_inference` below to `BringupStatus.INCORRECT_RESULT`

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -8,6 +8,7 @@
 // c++ standard library includes
 #include <cassert>
 #include <cstdlib>
+#include <memory>
```
mrakitaTT, line 11: `<memory>` is already included in the `module_builder.h`

PR #1420, opened 2025-09-18
[ci] run examples in ci

### LINKED ISSUE #137: Add test which runs our example code
We have a `examples/simple_regression.py`. We should add some test in CI to run this and verify that our public example code is not broken.

pilkicTT:
Pytest wrapper test which runs python scripts in the `examples/` directory is implemented. The test is parameterized by all collected python scripts. Example scripts are ran as subprocesses and their exit code is checked for failure.

Basic tests preset is extended to run this new test. All examples are ran on `n300` so that both single and multi-chip scripts can work. This can be expanded in the future if we need more sophisticated setups.

Side note: had to lower the number of iterations in `simple_regression.py` in order for tests not to take too long.

Closes #137


pilkicTT:
NOTE: currently the CI is failing due to one of the examples failing... Need to wait for #1413 to be merged...

vvukomanTT (approved):

nvukobratTT (approved):
Looks good! You even managed to find an example that regressed ;)

jameszianxuTT:
Just as a forward looking point, I don't think this is totally safe. We currently have the assumption that these examples don't have global, irreversible side effects on each other, so we can just run everything under `./tests/examples` under the same pytest command it it will just work. 

However, once we start adding tensor parallel examples using pytorch/xla's SPMD APIs we may run into some unexpected behaviour since this does have global, irreversible side effects causing exotic failures when mixing spmd/non-spmd pytorch/xla pytests. (i.e. https://github.com/pytorch/xla/issues/9578)

For this reason I had to separate out torch_xla tensor parallel tests from all the other tests in tt-torch.

I am currently bringing up a TP llama example so I mention this, because if such failures are encountered, I will probably need to split this `example` job into smaller, compatible groups.

jameszianxuTT (approved):

mrakitaTT (approved):

vmilosevic (approved):

PR #1428, opened 2025-09-19
Add manual dispatch for data collection

jameszianxuTT:
### Ticket
None 

### Problem description
Workflow dispatch autotriggers on a narrow set of tests. In particular it does not run on manual tests called via the Run Test pipeline, so testing of data production is more difficult than necessary. 

Also, there are occasional bugs in data collection or downstream ingestion issues that may necessitate a data collection reattempt without rerunning the target pipeline.

### What's changed
Add workflow dispatch to internal data collection pipeline to permit manual trigger.

### Checklist


jameszianxuTT:
Note - the run-name cannot be set dynamically based on whether or not this comes from a workflow dispatch, and doing so would not be trivial. 

`run-name: "Collect data for run_id ${{ github.event.workflow_run.id }} attempt ${{ github.event.workflow_run.run_attempt }}"`

When triggered from a workflow dispatch, these ${{}} variables will resolve to empty strings, which I think is acceptable... given that this workflow dispatch trigger will only be for internal testing I don't see the need to overcomplicate the yaml.

jameszianxuTT:
Thanks @jmcgrathTT. Due to small blast radius potential for this PR I will just merge this and we can revert/amend it in follow up if it introduces more issues.

jameszianxuTT:
It does work - https://github.com/tenstorrent/tt-xla/actions/runs/17868632962

PR #1442, opened 2025-09-23
Rename tt_torch to tt_torch_backend for tt-forge unified wheel packaging

jameszianxuTT:
### Ticket
https://github.com/tenstorrent/tt-torch/issues/1241

### Problem description
Python site-packages clobbering of "tt-torch" due to duplicate naming of python from tt-torch project and tt-xla project when both are installed in the same venv.

### What's changed
Temporarily rename python_package/tt_torch to tt_torch_core and update references for this release, so we can cut a tt-forge wheel with tt-torch and tt-xla coexisting.

### Checklist


jameszianxuTT:
Also @nvukobratTT, if we do rename the `tt-torch` package from the `tt-torch` repo to something else, the user will have this experience:
- pip install tt-forge wheel where tt-forge-fe, tt-torch and tt-xla projects are all installed into the same venv
- in python, `import tt_torch`  -> This resolves to tt-xla's "tt-torch" rather than the tt-torch project's `tt-torch` package
- we would tell users of tt-torch project that in order to use the tt-torch project, they will need to import something else, like `tt_torch_legacy`. And if they actually `import tt_torch` they will get a different package than the one they are expecting.

That is why the change is preferably in tt-xla rather than tt-torch. I reiterate that this is just to unblock the release and does not need to be a permanent solution... I agree it is less than ideal to jump these hoops but the ask to package all three frontends into the same wheel requires some hacks, given the partial overlap situation between tt-xla and tt-torch.

nvukobratTT:
@jameszianxuTT, can you add a bit more detail about the larger scope of change on the tt-torch side? What additions do we need to update compared to tt-xla?

jameszianxuTT:
@nvukobratTT In the tt-torch repo, we have a lot more files that import or reference the python tt_torch package.  [See these 170 files](https://github.com/search?q=repo%3Atenstorrent%2Ftt-torch+%28%22from+tt_torch%22+OR+%22import+tt_torch%22%29&type=code), compared to the 12 files changed in tt-xla. 

Also, in our tt-torch docs, we would need to tell the user - to use this project, you must not `import tt_torch`, because that will either
- fail because they only installed the tt-torch wheel or built from source
- or, if you're in the tt-forge docker or using the unified tt-forge wheel, it will import the tt_torch backend from the tt-xla project

PR #1460, opened 2025-09-24
Add BGE-M3 encode test

gengelageTT:
### Ticket
#1196 

### Problem description
We want bge_m3 encode running in CI and recorded by superset. The current TorchModelTester doesn't support inputs/outputs that aren't tensors and so this model can't use the existing test infra.

### What's changed
Added a unique test for BAAI/bge-m3 encode_single_device. This test loads the model from tt-forge-models, runs on cpu and device then computes the PCC. PCC is currently ~~0.99~~ [EDIT]: _0.98 on bh_ and asserted on so we can catch regressions. This test will run in the nightly CI and report to superset.

Why TorchModelTester is not used: 
- The inputs are sentences which cannot be moved onto device using `.to(device)`. In this test the inputs are moved onto device after tokenization by passing in the device as an argument to `encode_single_device`. This is why we use `encode_single_device` instead of `encode` (encode doesn't take in a device argument)
- Some outputs aren't torch tensors and calls like `torch.allclose()` fail to run when comparing them.
- `model.compile` doesn't support compiling `methods` but `torch.compile` does

### Checklist

### DIFF
--- a/tests/torch/single_chip/models/bge_m3/test_bge_m3_encode.py
+++ b/tests/torch/single_chip/models/bge_m3/test_bge_m3_encode.py
@@ -1,0 +1,188 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# This test does not use the TorchModelTester infrastructure as it requires sentence inputs (not tensors) that cannot be moved onto device using `.to(device)`.
+
+import torch
+import torch_xla.core.xla_model as xm
+import torch_xla.runtime as xr
+import numpy as np
+import subprocess
+import sys
+import pytest
+from infra import Framework, RunMode
+from utils import (
+    BringupStatus,
+    Category,
+    ModelGroup,
+    ModelSource,
+    ModelTask,
+    build_model_name,
+    incorrect_result,
+)
+from third_party.tt_forge_models.bge_m3.encode.pytorch.loader import (
+    ModelLoader,
+    ModelVariant,
+)
+
+VARIANT_NAME = ModelVariant.BASE
+MODEL_INFO = ModelLoader.get_model_info(VARIANT_NAME)
+
+
+def calculate_pcc(tensor1, tensor2):
+    """Calculate Pearson Correlation Coefficient between two tensors."""
+    # Convert to numpy if they're torch tensors
+    if hasattr(tensor1, "numpy"):
+        tensor1 = tensor1.numpy()
+    if hasattr(tensor2, "numpy"):
+        tensor2 = tensor2.numpy()
+
+    # Flatten tensors
+    x = tensor1.flatten()
+    y = tensor2.flatten()
+
+    # Calculate means
+    x_mean = np.mean(x)
+    y_mean = np.mean(y)
+
+    # Calculate centered vectors
+    x_centered = x - x_mean
+    y_centered = y - y_mean
+
+    # Calculate norms
+    x_norm = np.linalg.norm(x_centered)
+    y_norm = np.linalg.norm(y_centered)
+
+    # Handle edge case where norm is zero
+    if x_norm == 0 or y_norm == 0:
+        return float("nan")
+
+    # Calculate PCC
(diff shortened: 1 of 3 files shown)


gengelageTT:
Re-running test in CI here: https://github.com/tenstorrent/tt-xla/actions/runs/17990885856

tests/torch/single_chip/models/bge_m3/test_bge_m3_encode.py
```diff
@@ -0,0 +1,194 @@
...
+    from FlagEmbedding import BGEM3FlagModel
+except ImportError:
+    subprocess.check_call([sys.executable, "-m", "pip", "install", "FlagEmbedding"])
+    from FlagEmbedding import BGEM3FlagModel
+
+VARIANT_NAME = ModelVariant.BASE
+MODEL_INFO = ModelLoader.get_model_info(VARIANT_NAME)
+
+
+def calculate_pcc(tensor1, tensor2):
```
AleksKnezevic, line 39: Is it possible to reuse the calculation from: `torch_comparator.py'

AleksKnezevic (approved):
If possible, please use pcc calculation from torch_comparator, otherwise looks good.

tests/torch/single_chip/models/bge_m3/test_bge_m3_encode.py
gengelageTT: but I can try copy the calculation from that pr if we want it to match what's done in the test infra a closely as possible

AleksKnezevic: Yeah, I think it's worthwhile adding that.

gengelageTT:
pcc lowered to 0.92 which means bringup status is now `INCORRECT_RESULT`. Test is passing in CI here: https://github.com/tenstorrent/tt-xla/actions/runs/17991553896/job/51182748319
Added issue to investigate lower pcc on bh here: https://github.com/tenstorrent/tt-xla/issues/1461

tests/torch/single_chip/models/bge_m3/test_bge_m3_encode.py
```diff
@@ -0,0 +1,194 @@
...
+    compiled_model = torch.compile(model, backend="tt")
+
+    cpu_inputs = inputs
+    cpu_inputs["device"] = "cpu"
+    golden_output = compiled_model(**cpu_inputs)
+
+    # Connect the device.
+    device = xm.xla_device()
+
+    # Move inputs and model to device.
```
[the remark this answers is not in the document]
gengelageTT: I'll change this to run the not-compiled model for cpu. Normally we would move the model to device, but the `encode_single_device` function handles this internally when we pass in the device as an argument: https://github.com/FlagOpen/FlagEmbedding/blob/master/FlagEmbedding/inference/embedder/encoder_only/m3.py#L340

gengelageTT:
@AleksKnezevic I adjusted pcc calculation to more closely follow what's done in the TorchComparator. This required adding the following 2 functions:
-  `tree_map_to_torch()` turns the lexical weights dictionary into a tensor of values so that the pcc compares the outputted tensors instead of the individual items (individual items -> nan pcc). 
- `correct_golden_lexical_weights()`fixes shape mismatches between the output. Since `encode_single_device` [removes lexical weights that are 0](https://github.com/FlagOpen/FlagEmbedding/blob/master/FlagEmbedding/inference/embedder/encoder_only/m3.py#L358) (after the computation done on device) blackhole tests were failing since some values were incorrectly non-zero, leading to [shape mismatches](https://github.com/tenstorrent/tt-xla/actions/runs/18013919884/job/51253819350#step:15:181). 

@mrakitaTT I also moved `FlagEmbedding` to `venv/requirements-dev.txt` and created an issue to add support for methods/functions and non-tensor inputs: https://github.com/tenstorrent/tt-xla/issues/1471

changes are passing in CI here: https://github.com/tenstorrent/tt-xla/actions/runs/18019902210

kmabeeTT (approved):
Awesome! Let's ship it.

AleksKnezevic:
I agree!

mrakitaTT:
+1 :)

PR #1464, opened 2025-09-25
Single test 'contains' feature

nsumrakTT:
### Ticket
slack

### Problem description
Single test workflow has 'additional arguments' field.
But when used with -k option (which is the most used option in this field) with quotes (such as -k "yolo or bert") it fails because quotes are part of json and it is quite hard to escape them. 

### What's changed
Added new field 'contains' that is -k argument for pytest. Use without quotes.

### Checklist
https://github.com/tenstorrent/tt-xla/actions/runs/18004769745/job/51222760340


ndrakulicTT (approved):
Thanks for quick fix, LGTM

PR #1470, opened 2025-09-25
Remove batchnorm decomposition

odjuricicTT:
### Problem description
* Decomposing batchnorm results in a graph with a lot more ops which reduces performance.
* A better version of batchnorm decomposition (which allows for fusing of the decomposed ops) exists in TTIR , so there is not need for it in torch passes

### What's changed
* Removed batchnorm decomposition from torch passes

Note: I'm not 100% sure how this will affect batchnorm for training, but to my knowledge we are not training any models trough torch with batchnorm currently. If this becomes a problem, support for lowering training batchnorm to TTIR should be added.

### Checklist

### DIFF
--- a/python_package/tt_torch/backend/decompositions.py
+++ b/python_package/tt_torch/backend/decompositions.py
@@ -271,10 +271,6 @@ def _get_default_decomposition_ops() -> DecompositionOpsList:
         aten.masked_fill.Scalar,
         aten.t,
         aten.addmm,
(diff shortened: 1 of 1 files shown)


nvukobratTT:
@kmabeeTT FYI, this might cause some regressions on our nightly run.

odjuricicTT:
@nvukobratTT Thanks for the heads up. If this happens let me know and let's push for a solution in mlir instead.

PR #1482, opened 2025-09-26
Implemented composite ops for torch.

acolicTT:
Composite op is a mechanism that enables propagation of complex ops directly to device without decompositions.

Similar to jax implementation, this change adds support for composite ops in torch. It uses stableHLOCompositeBuilder for wrapping torch ops in composite ops, and substitutes torch native functions with composite function wrappers. This way, user does not have to change model in order to get performance improvement.

Stable HLO looks like this:
   ...
   %1 = stablehlo.reshape %0 : (tensor<1x32x32xf32>) -> tensor<32x32xf32> loc(#loc3)
   %2 = stablehlo.composite "tenstorrent.gelu" %1 {decomposition = @tenstorrent.gelu.impl} : (tensor<32x32xf32>) -> tensor<32x32xf32> loc(#loc4)
   return %2 : tensor<32x32xf32> loc(#loc)
   ...

IR (before fusing) like this:
   ...
   %0 = "ttnn.get_device"() <{mesh_offset = #ttnn<mesh_offset 0x0>, ...
   %1 = "ttnn.empty"(%0) <{dtype = #ttcore.supportedDataTypes<f32>, ...
   %2 = "ttnn.gelu"(%arg0) : (tensor<32x32xf32, #ttnn_layout>) -> tensor<32x32xf32, #ttnn_layout>
   return %2 : tensor<32x32xf32, #ttnn_layout>
   ....

### DIFF
--- a/python_package/tt_torch/composite_ops.py
+++ b/python_package/tt_torch/composite_ops.py
@@ -1,0 +1,49 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+from torch import Tensor
+from torch_xla.experimental.mark_pattern_utils import StableHLOCompositeBuilder
+
+"""
+From XLA documentation: '(Composite operation) Encapsulates an operation made up (composed) of
+other StableHLO operations, taking inputs and composite_attributes and producing results.
+The semantics of the op are implemented by the decomposition attribute. The composite op can be
+replaced with its decomposition without changing program semantics.'
+
+So, composite op is made because of high-level ops that are decomposed into dozens stable HLO
+operations. It enables custom backends to handle op however they want. Since we have a native
+support for, let's say, gelu operation in ttir, we will wrap torch gelu op into gelu composite
+op and handle it in XLA without decompostions, enabling our device to execute custom gelu implementation.
+
+Since we want to run torch models wihout modifying them, we will substitute torch operations
+(that we have a direct support for) with composite ops. This way, user will not have to change
+anything in model in order to get performance improvement.
+"""
+
+
+def composite_gelu(input: Tensor, approximate: str = "none") -> Tensor:
+    """
+    Creates composite gelu operation for torch xla using StableHLOCompositeBuilder.
+    Note that operation name must be tenstorrent.gelu[_tanh] for MLIR to handle it.
+
+    Returns a tensor.
+    """
+    tanh = approximate == "tanh"
+    name = "tenstorrent.gelu" + ("_tanh" if tanh else "")
+    attr = {"approximate": "tanh"} if tanh else None
+
+    builder = StableHLOCompositeBuilder(name=name, attr=attr)
+
+    input = builder.mark_inputs(input)
+    input = torch.nn.functional.gelu(input, approximate=approximate)
+    input = builder.mark_outputs(input)
+
+    return input
+
+
+"""
+Dictionary holding replacement composite functions for torch functions.
+"""
+replacements = {torch.nn.functional.gelu: composite_gelu}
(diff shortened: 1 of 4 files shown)


pglusacTT:
Can we add an example where we also patch the backward op? 
GELU has composite bwd that was added for jax [here](https://github.com/tenstorrent/tt-xla/blob/32ca4caf3489fcb72a0aca05cc85bacbf0f731e9/python_package/jax_plugin_tt/monkeypatch.py#L234).

python_package/tt_torch/composite_ops.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def composite_gelu(input: Tensor, approximate: str = "none") -> Tensor:
+    """
+    Creates composite gelu operation for torch xla using StableHLOCompositeBuilder.
+    Note that operation name must be tenstorrent.gelu[_tanh] for MLIR to handle it.
+
+    Returns a tensor.
+    """
+    tanh = approximate == "tanh"
```
ajakovljevicTT, line 43: nit: name this `is_tanh` or `is_approximate`

tests/torch/single_chip/ops/test_composite_ops.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+
+    class MM(torch.nn.Module):
+        def __init__(self):
+            super().__init__()
+
+        def forward(self, x):
+            return composite_gelu(x, approx)
+
+    input = torch.randn(32, 32)
```
ajakovljevicTT, line 64: The input should be parametrized

tests/torch/single_chip/ops/test_composite_ops.py
```diff
@@ -0,0 +1,147 @@
...
+
+    device = xm.xla_device()
+    model = model.to(device)
+    input = input.to(device)
+
+    output = model(input).to("cpu")
+
+    comparator = TorchComparator(
+        ComparisonConfig(
+            atol=AtolConfig(required_atol=0.02), pcc=PccConfig(required_pcc=0.99)
```
ajakovljevicTT, line 77: required_pcc is already 0.99 by default.

tests/torch/single_chip/ops/test_composite_ops.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+    )
+
+    comparator.compare(output, golden)
```
ajakovljevicTT, line 81: Have you tried using OpTester from the test infra for this?

python_package/tt_torch/composite_ops.py
```diff
@@ -0,0 +1,60 @@
...
+    input = builder.mark_outputs(input)
+
+    return input
+
+
+def install_composite_patches() -> None:
+    """
+    Replaces existing torch functions with composite ones.
+    """
+    F.gelu = composite_gelu
```
AleksKnezevic, line 60: I don't think we should monkeypatch torch.nn.functional. What if someone is running pytorch on GPU after loading our plugin?

AleksKnezevic (changes requested):

acolicTT:
> Can we add an example where we also patch the backward op? GELU has composite bwd that was added for jax [here](https://github.com/tenstorrent/tt-xla/blob/32ca4caf3489fcb72a0aca05cc85bacbf0f731e9/python_package/jax_plugin_tt/monkeypatch.py#L234).

@pglusacTT Gelu backward op handling is not implemented in MLIR, so this is a lie. I will add it in a separate PR and implement it for torch too.

python_package/tt_torch/composite_ops.py
```diff
@@ -0,0 +1,62 @@
...
+    return input
+
+
+"""
+Dictionary holding replacement composite functions for torch functions.
+"""
+replacements = {torch.nn.functional.gelu: composite_gelu}
+
+
+def handle_composite_ops(gm: torch.fx.GraphModule) -> None:
```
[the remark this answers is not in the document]
AleksKnezevic: Agreed

AleksKnezevic (approved):
Left a couple of questions, after addressing lgtm.

Issue #1485, opened 2025-09-26
Remove program cache workaround


odjuricicTT:
Once there is a proper way for a user to control device options (https://github.com/tenstorrent/tt-xla/issues/1480) the workaround for enabling program cache needs to be removed.

Issue #1480, opened 2025-09-26
Enable (re)configuring device options


pilkicTT:
There are certain options which can be set when opening the device, e.g. `enableProgramCache`. We should have a mechanism of controlling this.

Currently, we open the device when initializing `ClientInstance` (which is done on startup) and we don't have a way of modifying these settings at that point in time.

The same holds for modifying the settings at later points of the execution.

To resolve these issues we can:
- take a look if we could implement this behaviour within existing PJRT APIs
- possibly extend our binary to have APIs for global control of the opened devices

PR #1486, opened 2025-09-26
Add workaround to enable program cache

odjuricicTT:
### Problem description
There is not way for a user to enable program cache. Without program cache all model e2e performance measurements don't make sense as host overhead huge.

### What's changed
This change adds a temporary way to enable program cache. This is **NOT** intended to be widely used or exposed to users! It is a workaround to help unblock perf development and e2e perf measurements.

Progress on a proper solution for this problem is tracked here: https://github.com/tenstorrent/tt-xla/issues/1480
One that is completed, this workaround **WILL** be removed. Tracked here: https://github.com/tenstorrent/tt-xla/issues/1485

### Checklist

### DIFF
--- a/src/common/pjrt_implementation/client_instance.cc
+++ b/src/common/pjrt_implementation/client_instance.cc
@@ -410,7 +410,16 @@ ClientInstance::openMeshDevice(const std::vector<uint32_t> &mesh_shape) {
     tt::runtime::setFabricConfig(tt::runtime::FabricConfig::DISABLED);
   }
 
+  // TODO(odjuricicTT, #1485): This is a temporary way to enable program cache
+  // until we have a proper way for a user to pass device options. After that
+  // this should be removed. Issue for device options:
+  // https://github.com/tenstorrent/tt-xla/issues/1480
+  bool enableProgramCache =
+      std::getenv("TT_RUNTIME_ENABLE_PROGRAM_CACHE") != nullptr &&
+      std::string(std::getenv("TT_RUNTIME_ENABLE_PROGRAM_CACHE")) == "1";
+
   tt::runtime::MeshDeviceOptions options = tt::runtime::MeshDeviceOptions{
+      .enableProgramCache = enableProgramCache,
       .meshShape = mesh_shape,
   };
 


pilkicTT (approved):

src/common/pjrt_implementation/client_instance.cc
```diff
@@ -430,7 +430,16 @@ ClientInstance::openMeshDevice(const std::vector<uint32_t> &mesh_shape) {
     tt::runtime::setFabricConfig(tt::runtime::FabricConfig::DISABLED);
   }
 
+  // TODO(odjuricicTT, #1485): This is a temporary way to enable program cache
+  // until we have a proper way for a user to pass device options. After that
+  // this should be removed. Issue for device options:
+  // https://github.com/tenstorrent/tt-xla/issues/1480
+  bool enableProgramCache =
+      std::getenv("TT_RUNTIME_ENABLE_PROGRAM_CACHE") != nullptr &&
+      std::string(std::getenv("TT_RUNTIME_ENABLE_PROGRAM_CACHE")) == "1";
```
mrakitaTT, line 439: I would like to recommend a little bit cleaner way to achieve this, but if you are in a rush I don't mind merging this as-is and then following up with a cleanup.

So, when PJRT Client is created it accepts a `create_options` arg (can see [here](https://github.com/openxla/xla/blob/main/xla/pjrt/c/pjrt_c_api.h#L433C26-L433C40)). That arg is populated from ML frameworks when they spawn PJRT client, which is invoked by plugin registration, you can see it in JAX [here](https://github.com/jax-ml/jax/blob/31792df36d0679eef164bd3a680ab3ff1b893a00/jax/_src/xla_bridge.py#L558) and in torch_xla [here](https://github.com/pytorch/xla/blob/master/torch_xla/experimental/plugins.py#L53). We can populate the `create_options` map with `enableProgramCache` arg in our plugin initialization files (jax [here](https://github.com/tenstorrent/tt-xla/blob/main/python_package/jax_plugin_tt/__init__.py) and torch [here](https://github.com/tenstorrent/tt-xla/blob/main/python_package/torch_plugin_tt/__init__.py)). Since the plugin initialization is done automatically, you might wonder how then can user pass some option to plugin registration. Unfortunately, XLA team hasn't yet provided some good way to do that, which is why for example GPU plugin still uses env variables to populate the create options ([source](https://github.com/jax-ml/jax/blob/dd5371775ea35ff9755347c92bf1c80c4a055236/jaxlib/xla_client.py#L168)). Still, at least we will use ENV variables at one central place, instead of scattering them around our PJRT implementation code, making it easier to manage them.

FYI @pilkicTT @ajakovljevicTT

mrakitaTT (approved):

Issue #1519, opened 2025-09-30
Constant tensor lowered to vhlo as input


odjuricicTT:
`torch.ops.aten.scalar_tensor(0)` is lowered as:
```
%14 = "vhlo.constant_v1"() <{value = #vhlo.tensor_v1<dense<0.000000e+00> : tensor<1x192x28x28xbf16>>}> : () -> !vhlo.tensor_v1<1x192x28x28x!vhlo.bf16_v1> loc(#loc)
```

but `torch.ops.aten.scalar_tensor(6)` is lowered as as input to the main function:
```
%arg2: !vhlo.tensor_v1<!vhlo.bf16_v1> loc("xla__device_data")
```

This creates a whole host of problems for performance on mlir level since we don't know the value of this input tensor and cannot perform fusion patterns.

mrakitaTT:
FYI @ajakovljevicTT

odjuricicTT:
@ajakovljevicTT Why is this closed, do we track this somewhere else? If so can we please link to the issue that track this problem?

PR #1520, opened 2025-09-30
Add decomposition for clamp const inputs

odjuricicTT:
### Problem description
Because torch constants created with `torch.ops.aten.scalar_tensor(x)` (where x!=0 and x!=1) are lowered to vhlo as graph inputs we are not able to perform fusing in mlir. This causes performance problems. The issue for this is here: https://github.com/tenstorrent/tt-xla/issues/1519

### What's changed
Add a decomposition in torch passes to represent integer constants as additions for multiple ones. This gets optimized insto a single `vhlo.constant` later.

### Checklist


pilkicTT (approved):
wow :smile:

Issue #953, opened 2025-08-07
Cache device tensors for input buffers


mrakitaTT:
Currently we keep input buffers on host and call `to_layout` at the beginning of each execute to transfer them into layout which compiler decided (can decide to keep them on host or move to device in specific layout). To optimize performance we can cache tensors after `to_layout`, but we need to keep in mind that same input tensors can be used as input to multiple graphs where compiler can decide to use different layouts.

We have two proposals in mind:
- Performance over memory: To store a vector of runtime tensors created with each execute (if they are different) inside `BufferInstance`, and then in execute to go through this vector and check if there is already a tensor with required layout, if not create it and add it to the vector. During buffer destruction we deallocate all tensors.
- Memory over performance: To require executions on a single device to always be sequential and to overwrite the buffer tensor in each execute. I am not even sure if we are able to run multiple graphs at the same time on the same device, I know that runtime has the "program index" in its API, but I am not sure how it is used. But even if we are not able to do that, still we might have a case where user is iterating over batches and calling two graphs on the same input in each batch, then we will always be doing re-layouts.

Another optimization idea here is that we can mark buffer as "used" in the first execute, so in the later executes when we see that it was already used we can signal to const-eval that it is not dirty even if it requires re-layouting.

nvukobratTT:
@jameszianxuTT, is this one indeed in progress, or is it blocked on the Open/Close device issue?
- https://github.com/tenstorrent/tt-xla/issues/952 

For clarification, with the current approach for open/close device, did you have a time to sync with @pilkicTT and confirm if this issue is needed?

jameszianxuTT:
Hi @nvukobratTT, no, @pilkicTT resolved the issue for single and multidevice tensor persistence usecases.

I am doing some loosely related followup work in #1551, but it's a separate issue. I will close this issue.

PR #1523, opened 2025-09-30
[pjrt] input tensor persistence

pilkicTT:
This change implements initial version of input tensor persistence, i.e. leaving prepared tensors on the device for subsequent executions of the same program. The goal being to minimize host->device transfers between executions which is crucial for getting performance.

The `BufferInstance` now holds the host & prepared runtime tensors. The `m_host_runtime_tensor` is the old `m_runtime_tensor` (just renamed it).

Whenever we prepare input tensors for program execution, they get saved in the `BufferInstance` so that they can be reused on the subsequent execution. The condition for reuse is that the already prepared tensor is in the expected layout (from the flatbuffer binary) - this is needed for cases when the same buffer instance is used in multiple programs with different layouts.

The only complication is for sharded/replicated tensors; due to the difference between PJRT and our runtime - we have only one tensor object, while PJRT can have multiple buffer instances belonging to one tensor. For now, we simply hold the prepared runtime tensor in all buffer instances that represent it.

In future changes we can remove the host runtime tensors; I haven't done it now to simplify this change.

The deallocation logic is modified to just reset the tensor object, and let ref-counting do its magic.

NOTE: additionally the training test infra is fixed to run compiled graphs.

Issue #953


inc/common/pjrt_implementation/buffer_instance.h
```diff
@@ -89,8 +89,19 @@ class BufferInstance {
...
   // Returns the underlying runtime tensor created for this buffer.
-  const tt::runtime::Tensor &getRuntimeTensor() const {
-    return m_runtime_tensor;
+  const tt::runtime::Tensor &getHostRuntimeTensor() const {
+    return m_host_runtime_tensor;
+  }
+
+  // Returns the prepared runtime tensor created for this buffer on last
+  // execution.
+  const tt::runtime::Tensor &getPreparedTensor() const {
```
ajakovljevicTT, line 98: maybe getLastExecutionTensor (and member name changed accordingly)?

src/common/pjrt_implementation/buffer_instance.cc
```diff
@@ -67,22 +67,26 @@ BufferInstance::BufferInstance(PJRT_Buffer_Type data_type,
                                DeviceInstance *device, MemoryInstance *memory)
     : m_data_type(data_type), m_dimensions(dims, dims + num_dims),
       m_device(device), m_memory(memory),
-      m_runtime_tensor(nullptr, nullptr, tt::runtime::DeviceRuntime::TTNN),
+      m_host_runtime_tensor(nullptr, nullptr, tt::runtime::DeviceRuntime::TTNN),
       m_data_ready(false), m_data_ready_event(nullptr),
-      m_done_with_host_buffer_event(nullptr), m_data_deleted(false) {}
+      m_done_with_host_buffer_event(nullptr), m_data_deleted(false),
+      m_prepared_runtime_tensor(nullptr, nullptr,
```
ajakovljevicTT, line 73: can we use optional or not set this at all? This way it might be a bit more confusing in we just have a "dummy" tensor.

src/common/pjrt_implementation/buffer_instance.cc
```diff
@@ -137,8 +141,18 @@ void BufferInstance::deleteData() {
...
+    // reference count drops to zero.
+    m_host_runtime_tensor =
+        tt::runtime::Tensor(nullptr, nullptr, tt::runtime::DeviceRuntime::TTNN);
+  }
+
+  if (m_prepared_runtime_tensor.handle) {
+    // Just reset the tensor, deallocation happens automatically when
+    // reference count drops to zero.
+    m_prepared_runtime_tensor =
+        tt::runtime::Tensor(nullptr, nullptr, tt::runtime::DeviceRuntime::TTNN);
```
ajakovljevicTT, line 155: Again, some sort of invalidation (like std::nullopt) to me looks cleaner than resetting to dummy.

src/common/pjrt_implementation/loaded_executable_instance.cc
```diff
@@ -237,15 +237,45 @@ tt_pjrt_status LoadedExecutableInstance::getInputRuntimeTensors(
...
           BufferInstance::unwrap(argument_lists[device_index][arg_index]);
-      arg_tensors.push_back(buffer->getRuntimeTensor());
+      arg_buffers.push_back(buffer);
     }
 
+    // Assert that all buffer instances have the same prepared tensor.
+    // NOTE: In case of sharded tensor we have multiple buffer instances on the
+    // PJRT side, but on our side (tt-mlir runtime) we prepare a single
+    // multi-device tensor.
+    tt::runtime::Tensor prepared_tensor = arg_buffers[0]->getPreparedTensor();
```
ajakovljevicTT, line 253: Move this to a separate function in loaded_executable_instance.cc and .h

inc/common/pjrt_implementation/buffer_instance.h
```diff
@@ -179,8 +190,14 @@ class BufferInstance {
   // via `PJRT_Client_BufferFromHostBuffer_Args` and memory was not specified.
   MemoryInstance *m_memory;
 
-  // Underlying runtime tensor created for this buffer.
-  tt::runtime::Tensor m_runtime_tensor;
+  // Underlying host runtime tensor created for this buffer.
```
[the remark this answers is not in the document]
ajakovljevicTT: The comment is great, thanks! For me the important thing about this object is that is the shard that the Buffer Instance represents, not that it is on host (that is more of a consequence on how we set up our runtime), so I wanted to represent that in the naming. Pinging @mrakitaTT to see if he has any naming preferences, so you don't rename twice.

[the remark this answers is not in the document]
pilkicTT: also bear in mind that we'll likely restructure this further going forward, once we address the remaining issues with the multi-device tensors that we've talked about - removing the host tensor, having only shards in buffer instances, etc...

mrakitaTT (approved):
LGTM, waited a day to make sure I don't come up with some important edge case to cover.

src/common/pjrt_implementation/buffer_instance.cc
```diff
@@ -66,23 +66,23 @@ BufferInstance::BufferInstance(PJRT_Buffer_Type data_type,
...
                                const std::vector<std::uint32_t> &dimensions,
                                DeviceInstance *device, MemoryInstance *memory,
                                PJRT_Buffer_Type data_type)
     : m_data_type(data_type),
       m_dimensions(dimensions.begin(), dimensions.end()), m_device(device),
-      m_memory(memory), m_runtime_tensor(tensor), m_data_ready(false),
+      m_memory(memory), m_host_runtime_tensor(tensor), m_data_ready(false),
       m_data_ready_event(nullptr), m_done_with_host_buffer_event(nullptr),
-      m_data_deleted(false) {
+      m_data_deleted(false), m_prepared_runtime_tensor(std::nullopt) {
```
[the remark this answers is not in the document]
mrakitaTT: In PJRT world `BufferInstance` is expected to be "device buffer", and when the host buffer is needed the client should trigger the `PJRT_Buffer_ToHostBuffer` function, so we should be able to lazily do `untilizeToHost` in `BufferInstance::copyToHost` function which is called on `PJRT_Buffer_ToHostBuffer` API call. The only reason it wasn't done as such so far is because in case when output buffer is used as input buffer into another program, we needed its tensor to be the host tensor, that was the assumption of the previous logic which always called `toLayout` before execution, in order to keep things simple. In hindsight this wasn't actually a limitation since `toLayout` can work in both directions. Now with the optimization @pilkicTT added, we don't need the host tensor anymore and can pass the device buffer into `m_prepared_runtime_tensor` of the `BufferInstance`, and do the `untilizeToHost` separately in `BufferInstance::copyToHost`.

nvukobratTT (approved):

PR #1525, opened 2025-10-01
[pytest] mark xpass tests as failed

mstojkovicTT:
### Problem description
`@pytest.mark.xfail` hides expected failures, but when such test unexpectedly passes (XPASS), its still treated as success. This lets stale `xfails` accumulate.

### What's changed
Enabled `xfail_strict = true` so XPASS is now treated as a test failure.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,8 @@
 [pytest]
 
+# Fail on xpassed
+xfail_strict=true
+
 markers =
     push: marks tests as push, to be executed on PR pipeline
     nightly: marks tests as nightly, to be executed on nightly pipeline


mstojkovicTT:
cc: @nvukobratTT @vvukomanTT

mstojkovicTT:
> Do you know if it affects direct calls to `pytest.xfail()` ([docs](https://docs.pytest.org/en/stable/reference/reference.html#pytest.xfail)) which doesn't accept a strict option?

It doesnt interact with `pytest.xfail()` in any way (it will still mark it as XPASS and not fail).

Issue #1001, opened 2025-08-13
Pipe through KV cache tensors into tt-mlir


AleksKnezevic:

nvukobratTT:
@jameszianxuTT I assume you're pushing this effort through some Llama model. Can you clarify which one?

jameszianxuTT:
Hi @nvukobratTT, yes, that's correct. The first model I'm bringing up static cache on is the Transformers implementation of `meta-llama/Llama-3.2-3B`.  ([src](https://huggingface.co/meta-llama/Llama-3.2-3B))

I have previously brought it up through the torch-mlir path in tt-torch, this task is specifically the bringup through the torch-xla/tt-xla path.

jameszianxuTT:
> I don't have context around these, but I assume they are not related and required to have functional Llama 3.2 3B? These are more related to specific model sharding for multi-chip support?

Right. The first of these is required for static caching in general, the other two are required for tensor parallel static caching. They're not specific to llama, just background infrastructure items.

nvukobratTT:
Thanks! Potentially, the first point can be covered with an open/close device effort. I.e., simply persisting tensors on the device until it's closed. However, this is still WIP, so I can't comment with more details. Let's follow the details from the dependency PRs/Issues and see what the minimal requirements are to get Llama 3.2 3B running e2e with static KV cache enabled.

jameszianxuTT:
I have this working tensor parallel + single device on a test branch but pending two MLIR changes, without needing device persistence or tensor persistence. 



1. update_cache workaround patch for sharded tensors
2. [shlo scatterOp assertion bypass for static cache update](https://github.com/tenstorrent/tt-mlir/pull/5084) 

Some explanation about how this works without device persistence / tensor persistence:

In prefill, static cache tensors are returned to host by tt-xla in untilizeToHost and their underlying data is transferred to Output BufferInstances held by torch_xla. torch_xla then aliases the inplace-updated host data representing the static caches to Input BufferInstances in decode. 

Another way to explain the above is that if you compile the following test that performs inplace ops on a singular input tensor, we are actually able to get the right output because even though we have 3 `execute` calls with 3 different runtime tensors representing input tensor `x`, tt-xla transfers output data from device to host memory post-execution, and torch_xla orchestrates inplace updates of BufferInstance data to provide the appearance of tensor persistence without us actually having to manage that.

```
import pytest
import torch
import torch_xla.core.xla_model as xm
[...]

PR #1539, opened 2025-10-01
Bringup tensor parallel Llama 3.2 3B generative loop with static cache

jameszianxuTT:
### Ticket
#1001 

### Problem description
We don't have tensor parallel static cache in tt-xla CI.
### What's changed
- Add tensor parallel llama generative loop with static cache example
- Add tensor parallel llama decode and prefill single inference pytests for CI
- Add logic to swap between:
        torch_xla._XLAC._xla_sync_multi(list(output), self.devices, wait=False)
        torch_xla.sync()
        Depending on if output mutations are present. 

- The use of _xla_sync_multi causes incorrect tracing for non-functional graphs, i.e. those with side effects like buffer mutations. 

### Checklist
- [x] Merge tt-mlir change https://github.com/tenstorrent/tt-mlir/pull/5084 to fix sharded scatterOp 
- [x]  Merge tt-xla #1523 to avoid out of DRAM after >2 inference steps introduced by #1417

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -5,7 +5,7 @@
 option(USE_CUSTOM_TT_MLIR_VERSION "Flag to use TT_MLIR_VERSION set by the user" OFF)
 
 if (NOT DEFINED TT_MLIR_VERSION OR NOT USE_CUSTOM_TT_MLIR_VERSION)
-    set(TT_MLIR_VERSION "be97d17850a4b3a07dcf62b5172f6d6a4f3d64d7")
+    set(TT_MLIR_VERSION "274a5d02bf74aafe8b71247a1da05d75e24bbbde")
 endif()
 
 set(PROTOBUF_VERSION "v21.12") # same version as tt-metal uses


third_party/CMakeLists.txt
```diff
@@ -5,7 +5,12 @@
 option(USE_CUSTOM_TT_MLIR_VERSION "Flag to use TT_MLIR_VERSION set by the user" OFF)
 
 if (NOT DEFINED TT_MLIR_VERSION OR NOT USE_CUSTOM_TT_MLIR_VERSION)
-    set(TT_MLIR_VERSION "d5bf8aaf4aced9952b67f9ec5ec4d31fe16af56a")
+# base tt-mlir version 9d2d589165162a9117cb74cc5520fe13fefae2b6
+# revert revert of analyzemeshpass delta fe2f3609021951d24677e48adc49a9c49b01b295
+# tt-mlir with 3 patches 4e7b914c00cfa6045552bda0b71e936c48fd53fc
+# old - a96ee58c735f0a5567cb740fc3abe0ed85480061
+    set(TT_MLIR_VERSION "ef2fb2d552119987344d6095cff6baa068bcdd3d")
+
```
jameszianxuTT, line 13: Will be removed after mlir uplift of https://github.com/tenstorrent/tt-mlir/commit/3dddf576dc3a3d91665431eea54eada3bc5fcd5b

jameszianxuTT:
Passing result in CI: https://github.com/tenstorrent/tt-xla/actions/runs/18173616157/job/51734909001?pr=1539#step:15:31

jmcgrathTT (approved):

vvukomanTT (approved):

python_package/tt_torch/backend/backend.py
```diff
@@ -79,14 +82,30 @@ def __init__(self, module: torch.fx.GraphModule):
     def __call__(self, *args):
 
         output = self.module(*args)
-        # This tells torch-xla to cut the graph at only what is required to
-        # compute all tensors in the `output` list.
-        torch_xla._XLAC._xla_sync_multi(list(output), self.devices, wait=False)
+
+        gm_has_functional_output_kind: bool = True
+
+        for el in self.signature.output_specs:
```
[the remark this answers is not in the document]
jameszianxuTT: Based on discussion previously, I think we could try always using torch_xla.sync(). It seems like we are using sync_multi() as an "optimization" rather than any functional reasons.

> torch_xla.sync() creates a graph where unneeded tensors are returned.
> Occaisionally the graph it will compile with mark_step [torch_xla.sync] will have more outputs than intended [...] I think it might return all leaf nodes at that point if the output isnt explicitly the returned value of the graph

However, I think that might be better off as a separate PR to this one?

examples/pytorch/llama.py
```diff
@@ -0,0 +1,175 @@
...
+# --------------------------------
+def llama():
+
+    # Must be called at start of program.
+    setup_spmd()
+
+    # Set up config variables.
+    tokens_to_generate: int = 16
+    output_tokens: List[str] = []
+    model_hidden_layers: int = 28  # model config default
```
AleksKnezevic, line 64: please remove.

examples/pytorch/llama.py
```diff
@@ -0,0 +1,175 @@
...
+    batch_size: int = 1
+    max_cache_len: int = 32
```
AleksKnezevic, line 66: This should be bigger, I'd say at least 128.

examples/pytorch/llama.py
```diff
@@ -0,0 +1,175 @@
...
+    # Move model and inputs to device.
+    static_cache.key_cache = [k.to(device) for k in static_cache.key_cache]
+    static_cache.value_cache = [v.to(device) for v in static_cache.value_cache]
+    input_args["input_ids"] = input_args["input_ids"].to(device)
+    input_args["cache_position"] = input_args["cache_position"].to(device)
+
+    model = model.to(device)
+
+    # mark shardings on model and inputs.
+    xs.mark_sharding(input_args["input_ids"], mesh, (None, None))
```
AleksKnezevic, line 117: This is default behaviour. Are these needed?

examples/pytorch/llama.py
```diff
@@ -0,0 +1,175 @@
...
+        xs.mark_sharding(layer.mlp.up_proj.weight, mesh, ("model", None))
+        xs.mark_sharding(layer.mlp.gate_proj.weight, mesh, ("model", None))
+        xs.mark_sharding(layer.mlp.down_proj.weight, mesh, (None, "model"))
+
+        xs.mark_sharding(layer.self_attn.q_proj.weight, mesh, ("model", None))
+        xs.mark_sharding(layer.self_attn.k_proj.weight, mesh, ("model", None))
+        xs.mark_sharding(layer.self_attn.v_proj.weight, mesh, ("model", None))
+        xs.mark_sharding(layer.self_attn.o_proj.weight, mesh, (None, "model"))
+
+    model.compile(backend="tt")
```
AleksKnezevic, line 140: can you please use the `torch.compile` API?

tests/torch/multi_chip/n300/models/tensor_parallel/llama3/test_llama_step_n300.py
```diff
@@ -0,0 +1,189 @@
...
+def test_llama_step(run_mode):
+
+    # Must be called at start of program.
+    xr.set_device_type("TT")
+    enable_spmd()
+
+    # Set up config variables.
+    model_hidden_layers: int = 28
+    batch_size: int = 1
+    max_cache_len: int = 32
```
AleksKnezevic, line 56: Please increase.

AleksKnezevic (approved):
Few nits, otherwise looks great.

jameszianxuTT:
After rebasing it looks like we hit a new out of DRAM regression that I'm trying to track down. I'm still inclined to merge this even if the llama.py example doesn't work due to out of DRAM on 3rd token and investigate after merge with high priority to prevent any other regressions. 

the llama.py example is not currently tested in CI, it requires #1420 which is waiting on some other changes, so the out of DRAM issue won't block the PR gate

examples/pytorch/llama.py
jameszianxuTT: Technically no, but I wanted to be explicit that all model inputs and weights are being sharded by showing a mark_sharding call on all tensors. I can remove it if you want.

kmabeeTT (approved):

examples/pytorch/llama.py
AleksKnezevic: Doesn't seem needed, but also doesn't hurt, up to you.

tests/infra/utilities/xla_multichip_utils.py
```diff
@@ -0,0 +1,35 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
ajakovljevicTT, line 1: Can you rename this `torch_multichip_utils`? Keeping in line with `jax_multichip_utils` that exist.

tests/runner/test_utils.py
```diff
@@ -13,7 +13,7 @@
 from infra import ComparisonConfig, RunMode, TorchModelTester
 from tests.utils import BringupStatus, Category
 from third_party.tt_forge_models.config import Parallelism
-
+from infra.utilities.xla_multichip_utils import get_mesh
```
ajakovljevicTT, line 16: nit: move up in the imports, in line with `from infra import ComparisonConfig, RunMode, TorchModelTester`

.pre-commit-config.yaml
```diff
@@ -21,3 +21,8 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-added-large-files
+  - repo: https://github.com/pycqa/isort
```
jameszianxuTT, line 24: Hey @ajakovljevicTT I added this pre-commit hook to enforce python import sorting if that's a convention we want. I applied it to the imports in `tests/torch/multi_chip/n300/models/tensor_parallel/llama3/test_llama_step_n300.py`

tests/torch/multi_chip/n300/models/tensor_parallel/llama3/test_llama_step_n300.py
```diff
@@ -0,0 +1,189 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+from tests.infra.testers.single_chip.model.model_tester import RunMode
```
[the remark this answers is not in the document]
jameszianxuTT: What do you think of using the `https://github.com/pycqa/isort` pre-commit hook to enforce this? I think it would be better for you not to have to check for this manually and I don't see any other automated checks for python import sorting

examples/pytorch/llama.py
```diff
@@ -0,0 +1,171 @@
...
+    with torch.no_grad():
+        for step in range(tokens_to_generate):
+            output: CausalLMOutputWithPast = compiled_model(**input_args)
+            output_logits: torch.Tensor = output.logits.to("cpu")
+            output_text = tokenizer.decode(output_logits[:, -1].argmax(dim=-1))
+
+            print(output_text, end="")
+
+            # Update inputs for next iteration
+            next_token = output_logits[:, -1].argmax(dim=-1).unsqueeze(-1)
```
nvukobratTT, line 149: Why do we need to return each generated token back to the host, update inputs & positions, and then push back to the device? Isn't this handled by the `update_cache` op?

nvukobratTT (commented):
First batch, will review the second part later and/or tomorrow :))

examples/pytorch/llama.py
```diff
@@ -0,0 +1,171 @@
...
+            input_args["input_ids"] = next_token.to(device)
+
+            host_cache_pos = input_args["cache_position"].to("cpu")
+            host_cache_pos = torch.tensor([host_cache_pos[-1:] + 1])
+            input_args["cache_position"] = host_cache_pos.to(device)
+
+            # reapply shardings for static cache (i/o inplace mutated tensors since they lose sharding annotations).
```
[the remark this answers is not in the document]
jameszianxuTT: This is a torch_xla quirk specifically affecting tensors that are:
- sharded 
- mutated in place
- used as graph input and output
- reused in multiple executions

I extracted out a simple example with these conditions and was able to reproduce this behaviour, but haven't root caused as an intentional or unintentional behavior of torch_xla. Either way, I think addressing this involves a deeper debug and patching within torch_xla in a way that's beyond the scope of this PR.

examples/pytorch/llama.py
```diff
@@ -0,0 +1,171 @@
...
+
+def create_device_mesh() -> Mesh:
+    """
+    Create device mesh for tensor parallelism.
+
+    Returns:
+        Mesh object for SPMD operations
+    """
+    num_devices = xr.global_runtime_device_count()
+    mesh_shape = (1, num_devices)
```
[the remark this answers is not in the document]
jameszianxuTT: I have not, but I can do that.

examples/pytorch/llama.py
jameszianxuTT: No, update_cache just does an inplace update of the static cache tensor _given_ an update inputs and positions. It does not alter the update inputs (eg. K/V matrices) and positions themselves.

The movement of the cache position and next token entities to and from device is to prevent muddying the traced inference graph, which represents just a forward pass through the llama model. 

Cache position update and next token selection are features of the generation loop rather than the forward pass of the llama model. Ideally we don't need to implement a custom generation loop at all, but LazyTensor trace boundaries cause some issues when we try to use transformers `model.generate()` and I think that's being looked into.

examples/pytorch/llama.py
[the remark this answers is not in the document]
jameszianxuTT: Okay, this does not work on 1x1. It is fine on loudbox. 

That is a good catch. I think we might have some issues in general running with SPMD mode on for 1x1 meshes for specific compute graphs. This generative loop does work on n150, but breaks if adding SPMD / shard marking. 

I'll special case this demo to remove the SPMD stuff for a 1x1 mesh so it can perform as a general "llama generative" demo and make an issue about SPMD on 1x1 mesh.

The alternative to that would be to name this test "llama_tensor_parallel.py" and possibly make a "llama_single_device.py", but that would be worse in my opinion due to the excessive code duplication.

examples/pytorch/llama.py
```diff
@@ -0,0 +1,171 @@
...
+
+    # Generate inputs.
+    inputs = tokenizer.encode_plus(
+        input_prompt,
+        return_tensors="pt",
+        truncation=True,
+    )
+
+    # Instantiate static cache on host (device instantiation leads to trace of unfusable creation ops.)
+    static_cache: StaticCache = StaticCache(
```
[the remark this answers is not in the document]
jameszianxuTT: Also, even if we update the fusion pass, I think there may be more issues stemming from the static cache tensor originating from a constant. We don't want this to get constevalled or any other side effects to happen on the static cache. So we'd probably need to modify some argument types from constant -> input

Issue #1343, opened 2025-09-09
Dynamically pull test properties from tt-xla runtime for reporting, with optional fallback to static test properties (eg. skipped/xfailed tests)


jameszianxuTT:
See discussion here: https://github.com/tenstorrent/tt-xla/issues/1052#issuecomment-3267325379
For implementation details.

nvukobratTT:
@jameszianxuTT any luck with dynamic properties recording? :))

nvukobratTT:
@jameszianxuTT is currently focused on TP llama and tensor lifetime, so this issue will be prolonged. Therefore, we'll re-assign this to the next available engineer :))

jameszianxuTT:
Just a heads up @mstojkovicTT that we are already dynamically recording accuracy metrics (PCC/ATOL) captured at runtime for torch_xla tt-forge-models tests via test_models.py. 

See PR: https://github.com/tenstorrent/tt-xla/pull/1542

PR #1542, opened 2025-10-01
Return accuracy metrics from comparator.compare() to caller via ModelTester.test()

### LINKED ISSUE #1448: Pipe PCC results from runtime into Dynamic Torch ModelTester, and record to junitxml
It is helpful to extract actual PCC results from runtime into reporting in the junitxml. The actual accuracy information is not currently known at runtime and only guarded by an (optional) assert.

I have a test branch here with a hacked implementation to collect true accuracy data from runtime as a POC: https://github.com/tenstorrent/tt-xla/compare/main...jzx/report_pcc

The mechanism involves pushing up data through the test infra (eg. providing override for test_model => test_model_with_metrics) that bubbles up the actual accuracy results from the various comparators into the test script itself, at which point we can collect the return value and record_property it into the junitxml. Does this sound reasonable @mrakitaTT ?

It does work and I was able to make a spreadsheet showing the actual runtime PCC, threshold PCC, and assertion status of the 700+ models we have running through the forge model tester, but I plan to clean it up a bit.

Also partially addresses #1343. 

cc @kmabeeTT

jameszianxuTT:
### Ticket
Resolves part of #1343 by exposing a path to report real runtime accuracy metrics.
Resolves #1448 

### Problem description
There is no way to return computed PCC results from runtime to caller with existing test infra (comparator, modeltester).

### What's changed
- Split up comparator assertion into separate stages for 
(1) accuracy calculation
(2) result evaluation and 
(3) result assertion
- Whether or not the comparator.compare() function asserts directly is now controlled by ComparatorConfig.assert_on_failure boolean, which is default true for backwards compatibility
- `comparator.compare()` now directly returns the result of the comparison but remains stateless
- Expose manual assert trigger via `comparator._assert_on_results(ComparisonResult)`. The expected flow is for a caller to 
(1) execute a test and get the `ComparisonResult` from return value of `tester.test()`
(2) report them to CI with call to `record_property`
(3) call `_assert_on_results()` to exit
- Upload PCC/ATOL/thresholds/assertion enablement status to superset for forge models tests as a demonstration of the above flow

- Note:
There is currently no precedent for recording results for single tests that return multiple different results. Inference tests only return one ComparisonResult, but training tests return two - one for forward pass and one for backward pass. For training tests, both results are asserted on, but only the backwards accuracy result is published to the database.

### Checklist
- [x] New tests/infra_tests pass, checking that existing test comparator infra is not compromised for torch/jax tensors in a variety of failing and passing scenarios, checking for false pass and false fail. These tests do pass: https://github.com/tenstorrent/tt-xla/actions/runs/18410773126/job/52463267173#step:16:279

### DIFF
--- a/tests/runner/test_models.py
+++ b/tests/runner/test_models.py
@@ -20,7 +20,6 @@ from tests.runner.test_utils import (
 )
 from tests.utils import BringupStatus
 from third_party.tt_forge_models.config import Parallelism
-from tests.runner.test_config import PLACEHOLDER_MODELS
 
 # Setup test discovery using utility functions
 TEST_DIR = os.path.dirname(__file__)


jameszianxuTT:
Example result - note that PCC / atol is stuffed into tags. 
```xml
<?xml version="1.0" encoding="utf-8"?>
<testsuites name="pytest tests">
    <testsuite name="pytest" errors="0" failures="0" skipped="0" tests="1" time="45.721"
        timestamp="2025-10-01T20:21:39.626421+00:00"
        hostname="aus-wh-05-special-jameszianxu-for-reservation-44723">
        <testcase classname="tests.runner.test_models"
            name="test_all_models[mnist/pytorch-full-inference]" time="43.423">
            <properties>
                <property name="tags"
                    value="{'test_name': 'test_all_models', 'specific_test_case': 'test_all_models[mnist/pytorch-full-inference]', 'category': 'model_test', 'model_name': 'pytorch_mnist_base_cv_image_cls_custom', 'model_info': {'task': 'cv_image_cls', 'source': 'custom', 'framework': 'pytorch', 'model_arch': 'mnist', 'variant_name': 'base'}, 'run_mode': 'inference', 'bringup_status': 'PASSED', 'parallelism': 'single_device', 'pcc': 1.0, 'atol': None, 'pcc_threshold': None, 'atol_threshold': None, 'pcc_assertion_enabled': None, 'atol_assertion_enabled': False, 'comparison_passed': True, 'comparison_error_message': None}" />
                <property name="owner" value="tt-xla" />
                <property name="group" value="generality" />
            </properties>
        </testcase>
    </testsuite>
</testsuites>
```

tests/infra/comparators/comparator.py
```diff
@@ -19,28 +37,153 @@ class Comparator(ABC):
...
+        if self._comparison_config.assert_on_failure:
+            self.assert_on_results()
+
+        return self._comparison_result
+
+    def get_results(self) -> ComparisonResult:
+        """
+        Get comparison results.
+
+        Returns the ComparisonResult object containing
```
jameszianxuTT, line 98: oops, I will complete this comment shortly

tests/runner/test_utils.py
```diff
@@ -405,7 +408,7 @@ def record_model_test_properties(
     # database correctly in very short term for newly added TP models on n300-llmbox.
     # This will be replaced by something more robust in near future.
     arch = request.config.getoption("--arch")
-    if "llmbox" in arch:
+    if arch is not None and "llmbox" in arch:
```
jameszianxuTT, line 411: This is a side-fix, I noticed that we now fail to manually invoke these model tests without specifying an --arch option from CLI due to this branch, since it assumes `arch` has a value.

jameszianxuTT: nvm, @kmabeeTT spotted this this in #1559

AleksKnezevic (approved):

Issue #1383, opened 2025-09-15
Computation Client Must Be Initialized at Program Start


ndrakulicTT:
## Description

The TT PJRT computation client must be initialized at the very beginning of the program, before any PyTorch operations that might trigger XLA compilation or device interactions. Without this initialization, PyTorch can trigger actions that require TT PJRT before actual execution on the TT device, causing failures.

## Erorr

```
Traceback (most recent call last):
  File "/localdev/ndrakulic/tt-xla/main.py", line 30, in <module>
    output.sum().backward()
  File "/localdev/ndrakulic/tt-xla/venv/lib/python3.11/site-packages/torch/_tensor.py", line 648, in backward
    torch.autograd.backward(
  File "/localdev/ndrakulic/tt-xla/venv/lib/python3.11/site-packages/torch/autograd/__init__.py", line 353, in backward
    _engine_run_backward(
  File "/localdev/ndrakulic/tt-xla/venv/lib/python3.11/site-packages/torch/autograd/graph.py", line 824, in _engine_run_backward
    return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: 0 <= device.index() && device.index() < static_cast<c10::DeviceIndex>(device_ready_queues_.size()) INTERNAL ASSERT FAILED at "/pytorch/torch/csrc/autograd/engine.cpp":1523, please report a bug to PyTorch. 
```

## Required Fix

Add one of the following initialization calls at the start of your program:

```python
# Option 1: Direct initialization
torch_xla._XLAC._init_computation_client()

# Option 2: Implicit initialization via device access
xm.xla_device()
```

## Repro

```python
import os
import torch
import torch_xla
import torch_xla.runtime as xr
import torch_xla.core.xla_model as xm
import torch_xla._internal.pjrt as pjrt

xr.set_device_type("TT")

# Fails without this
torch_xla._XLAC._init_computation_client()
# or
# xm.xla_device()

inputs = torch.ones(10, 10)
weights = torch.ones(10, 10, requires_grad=True)

output = inputs @ weights
output.sum().backward()
cpu_grads = weights.grad.clone()

weights.grad.detach_()
weights.grad.zero_()

device = xm.xla_device()

[...]

PR #1339, opened 2025-09-09
Implement torch training infra

ndrakulicTT:
Fixes:
#1383
### Training mode support:

* Added a `training_mode` flag and context management using `torch.set_grad_enabled` in `TorchDeviceRunner`, allowing the runner to switch between inference and training modes.
* Updated `TorchModelTester` to configure models for training, set the runner to training mode, and skip compilation during training.
* Implemented a `_test_training` method in `TorchModelTester` to run training tests, compute gradients on both CPU and TT devices, and compare results.
* Added initialization for the TT PJRT computation client in `TorchDeviceConnector` see #1383
* Created a new training test configuration file and updated the main test config to merge inference and training configs.  
* Parameterized `run_mode` in `test_models.py` to run both inference and training tests.

### Checklist

- [ ] submit issue for compilation in training mode

PR #1550, opened 2025-10-02
Add torch_xla basic tests to onPR/onPush, return initialize_device_connectors, require tests to manually init xla computation client

jameszianxuTT:
### Ticket
#1556

### What's changed
- return `initialize_device_connectors` autouse fixture
  - remove  `torch_xla._XLAC._init_computation_client()` call from TorchDeviceConnector ctor, to avoid possible hangs when running pytorch and jax tests in the same process.
  - This means that tests will need to initialize the torch xla runtime manually, and most already do so by acquiring a device. Maybe this can be considered a reopening of #1383, since we don't have a good solution to that yet
- Add push marks to torchxla basic and torchxla custom ops tests
  - Due to --forked invocation ( #1554 ) add additional parallel splits to these test jobs so they don't take 1.5 hours to run
- xfail 26 torchxla basic ops tests failing due to PCC errors with unknown passing point #1555    
- In TorchDeviceRunner only move workload.compiled_executable .to device if .to is defined on the compiled_executable (not defined for customOps OpTester tests)

### Checklist

### DIFF
--- a/tests/torch/single_chip/test_torch_xla_basic.py
+++ b/tests/torch/single_chip/test_torch_xla_basic.py
@@ -77,29 +77,7 @@ def test_simple_mm_eager(bias):
     comparator.compare(output, golden)
 
 
-def test_relu6():
-    class Relu6(torch.nn.Module):
-        def forward(self, x):
-            return torch.nn.functional.relu6(x)
-
-    input_x = torch.randn(32, 32, dtype=torch.bfloat16)
-
-    model = Relu6()
-    golden = model(input_x)
-
-    device = xm.xla_device()
-    model = torch.compile(model.to(device), backend="tt")
-
-    output = model(input_x.to(device))
-
-    comparator = TorchComparator(
-        ComparisonConfig(
-            atol=AtolConfig(required_atol=0.02),
-        )
-    )
-    comparator.compare(output, golden)
-
-
+@pytest.mark.push
 def test_silu():
     class Silu(torch.nn.Module):
         def forward(self, x):
@@ -122,6 +100,7 @@ def test_silu():
     comparator.compare(output, golden)
 
 
+@pytest.mark.push
 def test_silu_with_dtype_promotion():
     class Silu(torch.nn.Module):
         def forward(self, x):
@@ -145,6 +124,7 @@ def test_silu_with_dtype_promotion():
     comparator.compare(output, golden)
 
 
+@pytest.mark.push
 def test_relu6():
     class Relu6(torch.nn.Module):
         def forward(self, x):
@@ -168,6 +148,7 @@ def test_relu6():
     comparator.compare(output, golden)
 
 
+@pytest.mark.push
 def test_mul():
     class Mul(torch.nn.Module):
         def forward(self, x, y):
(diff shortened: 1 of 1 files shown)


tests/infra/connectors/torch_device_connector.py
```diff
@@ -19,8 +19,6 @@ class TorchDeviceConnector(DeviceConnector):
     def __init__(self) -> None:
         super().__init__()
         xr.runtime.set_device_type("TT")
-        # Initialize TT PJRT computation client.
-        torch_xla._XLAC._init_computation_client()
```
jameszianxuTT, line 23: > This means that tests will need to initialize the torch xla runtime manually, and most already do so. Maybe this can be considered a reopening of https://github.com/tenstorrent/tt-xla/issues/1383, since we don't have a good solution to that yet

Per this comment - We probably need a better way to guarantee computation client initialization within test infra but putting that as a side effect of TorchDeviceConnector does not work since not all torch_xla tests use it. This caused a breaking change for torch_xla basic tests which was not noticed since they are not in CI

Solving this with an automatic test fixture that optionally initializes the computation client for torchxla tests doesn't seem like the best option. We'd still need to enforce that test groups running within the same process don't accidentally include both torchxla and jax tests. 

I think something like a custom fixture added over all torchxla tests that initializes the runtime if not initialized already might work, plus guards to check if jax is initialized in the same context [similar to how pytorch/xla does it](https://github.com/pytorch/xla/blob/1ab678799f2a9a448d202df0262a9cca9ea97bb1/torch_xla/_internal/jax_workarounds.py#L8-L12) might be a good option. We'd also need the reverse guard for jax (which I'm more unfamiliar with) to disallow initialization if torch_xla is detected to be initialized.

jameszianxuTT: I think the failure sequence you observed was specifically for running .backwards() as shown in reproducer of #1383. 

In that sequence:
1. set xr.runtime.set_device_type("tt")
2. run some operations including .backwards()
3. acquire device via xm.xla_device and implicitly cause runtime initialization
4. move weights/inputs to device
5. execute test
6. run .backwards() for ondevice graph and see autograd crash

I think the failure is related to autograd in (2) doing some weird stuff with XLA side effects, maybe causing gradients to get stuck on CPU or something so (6) fails. Based on findings in #1383 this seems to be a real bug, and unexpected that performing torch operations before acquiring a device has breaking side effects. i.e. doing steps in order (1-2-3 should be the same as doing them in order 1-3-2). It seems like it's only for training that this ordering difference matters.
 
The fix in #1339 was to enforce step order (1-3 ... ) before everything else by putting it in the TorchDeviceConnector initializer, but this seems only necessary for the particular training test and breaks other tests that don't construct a TorchDeviceConnector at start of test.

So, because this solution seems training-specific and breaks many other tests, I think it should be reverted as I have done in my PR. This means that for training graphs for now, test writer must remember to initialize device before running torch operations on CPU due to the bug described in https://github.com/pytorch/xla/issues/4174. 

Maybe in the future we can implement some guarded framework auto-initialization feature to each test, but for sake of getting torch_xla basic tests into CI sooner rather than later I suggest a simple reversion for now.

jmcgrathTT (approved):

nsumrakTT (approved):

jameszianxuTT:
Merging this now.
cc @jazpurTT as it adds conv2d tests which may complicate your uplift.

Issue #1538, opened 2025-10-01
LLM specific tests


AleksKnezevic:
There are many ops that we'll be [fusing](https://github.com/tenstorrent/tt-mlir/issues/4214) specifically for LLM perf.

Please add unit tests for these for some relevant models (llama 3b, qwen3, bge-m3, gpt-oss, stable diffusion to start). 

Please add the following tests for each of the models:

- attention (prefill/decode)
  - concat heads
  - split heads
  - SDPA
- rotary embedding

I have an attention prefill test in `aknezevic/llm_ops` as an example running [LlamaAttention](https://github.com/huggingface/transformers/blob/94bbf8e19911a3aed089d47094c836a18717c9c0/src/transformers/models/llama/modeling_llama.py#L224). I extracted the shapes running the full model and setting a breakpoint in the attention layer, i.e. in `venv/lib/python3.11/site-packages/transformers/models/llama/modeling_llama.py:241`

for concat heads, you'll want to run:
```python
attn_output.reshape(*input_shape, -1).contiguous()
```
for create heads:
```python
        query_states = self.q_proj(hidden_states).view(hidden_shape).transpose(1, 2)
        key_states = self.k_proj(hidden_states).view(hidden_shape).transpose(1, 2)
        value_states = self.v_proj(hidden_states).view(hidden_shape).transpose(1, 2)
```
for SDPA:
```python
        attention_interface: Callable = eager_attention_forward
        if self.config._attn_implementation != "eager":
            attention_interface = ALL_ATTENTION_FUNCTIONS[self.config._attn_implementation]

        attn_output, attn_weights = attention_interface(
            self,
            query_states,
            key_states,
            value_states,
            attention_mask,
            dropout=0.0 if not self.training else self.attention_dropout,
            scaling=self.scaling,
            **kwargs,
        )
```
and for rotary embedding it will be similar to the attention test, but you want to target [LlamaRotaryEmbedding](https://github.com/huggingface/transformers/blob/94bbf8e19911a3aed089d47094c836a18717c9c0/src/transformers/models/llama/modeling_llama.py#L95)

AleksKnezevic:
@ssaliceTT can you please add rotary embeddings tests. 
@gengelageTT can you please add attention tests.

PR #1557, opened 2025-10-02
Add llama attention unit tests

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/1538

### Problem description
We want unit tests for attention layers of models that will benefit from the ops being [fused](https://github.com/tenstorrent/tt-mlir/issues/4214) specifically for LLM perf. These should test attention prefill/decode, concat heads, split heads (create heads) and SDPA.

### What's changed
Added attention prefill/decode, concat heads, create heads and SDPA unit tests for LlamaAttention. Runs for all variants of the llama model in tt-forge-models `llama/causal_lm`. Marked with `@pytest.mark.push` so they run in CI.

### Checklist


AleksKnezevic (dismissed):

jameszianxuTT (changes requested):
It looks like this is too many tests to put into OnPush, which now runs for 2.5 hours on n300.

I suggest conditionally only running the 1024 seqlen type in onPush, and the other seqlen in nightly. Or not parameterizing on seqlen at all, as I don't think that matters for the purpose of testing the functionality of these unit tests. 

It would also resolve the out of DRAM issues for the larger 2048/4096 seqlen tests (except for llama 70b, which I doubt can run at all and should be generally xfailed?).

btw: pytest supports [per-parameterization marks](https://docs.pytest.org/en/stable/example/parametrize.html#set-marks-or-test-id-for-individual-parametrized-test), just wrap each parameterization in a pytest.param object and set its `mark` field. 

Also, you can [xfail during runtime without a mark](https://docs.pytest.org/en/stable/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail) and not at collection time, which might help the problem of resolving variants at collection time.

AleksKnezevic (approved):
Small nit, please remove commented out `#atol = ...`

Issue #1496, opened 2025-09-29
Backend Mismatch


devisettymahidhar608:
I've observed some inconsistent behavior in the **Torch framework** during model testing.
For **model testing**:

- When using the [CPU](https://github.com/tenstorrent/tt-xla/blob/main/tests/infra/testers/single_chip/model/torch_model_tester.py#L97), the backend is set to Inductor.

- When using the [TT](https://github.com/tenstorrent/tt-xla/blob/main/tests/infra/testers/single_chip/model/torch_model_tester.py#L105) device, the backend is set to tt.

For **operator tests**:

- Both the [CPU](https://github.com/tenstorrent/tt-xla/blob/main/tests/infra/testers/single_chip/op/op_tester.py#L86) and the [TT ](https://github.com/tenstorrent/tt-xla/blob/main/tests/infra/testers/single_chip/op/op_tester.py#L62)device use tt as the backend.

Could you please clarify which one is correct? Should I raise a PR to fix it?

Thank You.

AleksKnezevic:
@mrakitaTT and I discussed this a while ago. torch models do not need a CPU compile step, so no need to use inductor at all. When executing on CPU we should just run the model with inputs/weights on CPU, and use backend=TT for on device operation. Please do submit a PR.

kmabeeTT:
Don't think so, closing it.

PR #1562, opened 2025-10-03
Configured the backend as 'tt' for running on the CPU.

devisettymahidhar608:
### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist

### DIFF
--- a/tests/infra/testers/single_chip/model/torch_model_tester.py
+++ b/tests/infra/testers/single_chip/model/torch_model_tester.py
@@ -121,7 +121,7 @@ class TorchModelTester(ModelTester):
 
         Compiles for inductor backend by default.
         """
-        self._compile_for_backend(workload, backend="inductor")
+        self._compile_for_backend(workload, backend="tt")
 
     # @override
     def _compile_for_tt_device(self, workload: Workload) -> None:


AleksKnezevic (changes requested):
This is the CPU codepath (for generating golden outputs that we want to compare device outputs against). Your change would compile and run this on our device, so we'd be comparing device outputs to device outputs. 

This part does not need change. I believe the issue found was in op tester where the wrong backend was set.

devisettymahidhar608:
Hi @AleksKnezevic , 
As discussed in the [issue ticket](https://github.com/tenstorrent/tt-xla/issues/1496), the backend for CPU should be set to **"tt"**, not **"inductor"**. The op_tester was already using **"tt"** as the backend, but there was a mismatch in model_tester. This PR fixes that inconsistency.

AleksKnezevic:
Perhaps I was a little unclear in the ticket, what I said was:
> torch models do not need a CPU compile step, so no need to use inductor at all. When executing on CPU we should just run the model with inputs/weights on CPU, and use backend=TT for on device operation. Please do submit a PR.
To re-iterate: 
for CPU, we do not need a compile step. we can just run the model as is, without compiling. 
for device, we need to compile with backend="tt"
Two changes are possible here, either, 1) we change the op tester to use backed="inductor" for CPU tests, or, 2) we refactor the tester to remove compilation for CPU tests. 

Please modify this PR to do 1, change the op tester backend to use "inductor". There are a few things that need to be addressed in the refactor, so we can postpone that.

PR #1591, opened 2025-10-06
Add combined test preset for running on-pr manually

brataTT:
### Ticket
None (slack thread)

### Problem description
When testing metal uplift, running equivalent of on-pr tests using manual test workflow (Run Test) required either
- manually copying multiple preset contents, or
- separate CI runs, which duplicated the build process and led to repeated 1-hour build steps (longer if image build is needed and they are run in parallel).

### What's changed
- Create basic-and-models-push.json combining basic-test.json and model-test-push.json
  - This can be selected to test uplift more easily

### Checklist

### DIFF
--- a/.github/workflows/test-matrix-presets/basic-and-models-push.json
+++ b/.github/workflows/test-matrix-presets/basic-and-models-push.json
@@ -1,0 +1,13 @@
+[
+  { "runs-on": "wormhole_b0",        "name": "run_jax",               "dir": "./tests/jax/single_chip",                 "test-mark": "push and not large", "parallel-groups": 2 },
+  { "runs-on": "wormhole_b0",        "name": "run_torch",             "dir": "./tests/torch/single_chip",               "test-mark": "push", "parallel-groups": 3  },
+  { "runs-on": "n150",               "name": "run_large_jax_models",  "dir": "./tests/jax/single_chip",                 "test-mark": "push and large", "shared-runners": "true" },
+  { "runs-on": "p150",               "name": "run_jax",               "dir": "./tests/jax/single_chip",                 "test-mark": "push and not large", "parallel-groups": 2 },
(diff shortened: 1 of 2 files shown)


.github/workflows/test-matrix-presets/basic-and-models-push.json
```diff
@@ -0,0 +1,13 @@
+[
```
nsumrakTT, line 1: nitpick: I would prefer if you add line with # comment and describe what this test suite is all about, just for other people to have  information why it exists and its intended usage.

nsumrakTT (approved):

.github/workflows/test-matrix-presets/custom-basic-and-models-push.json
```diff
@@ -0,0 +1,13 @@
+[
```
brataTT: added a description, and a small function to filter out comments, but I'll leave preferred comment syntax for devex to plan out

jameszianxuTT: One unfortunate scenario might be that someone adds some huge model tests and then doubles the parallel-groups for the impacted job to account for that. 

Uplifter won't know that happened until their run takes a very long time.

Also any path renaming / mark renaming may cause this to break, and 

I suggest an alternative API than a hardcoded test group - maybe 
`def read_preset_test_entries(file_path)` could be modified to handle a semicolon delimited filepath array?


Then check for this custom option in the `file_path` and do a 
```
for file_path in file_paths.split(;)
    with open(file_path, "r") as f:
        matrix = json.load(f)
```

then you would change the workflow_dispatch option to `model-test-push.json;basic-test.json`

brataTT: but yes, agree that this is a problem for keeping parity between different presets

in current approach, anyone changing one preset needs to be aware and check if other presets need similar change

vmilosevic (approved):

ajakovljevicTT (approved):

PR #1597, opened 2025-10-07
Fixing the nightly

ajakovljevicTT:
Solving a couple of issues regarding the current failing CI nightly:
- When lots of tests are added in one go, the estimation for the timeout is way lower than it should be, leading to timeout, so added a check which check ran tests against `.test_durations` file and sets the timeout to 4h if there is. a mismatch (meaning new tests were added).
- We still have out of disk memory error on large models, possibly due to one model crashing, have to do a more detailed check, skipped for now.
- One model was failing on bad pcc, so changed the config accordingly.
- One of the xfailing torch model tests fail on RAM OOM, so skipping it

### DIFF
--- a/tests/jax/single_chip/models/vit/vit_large_patch16_384/test_vit_large_patch16_384.py
+++ b/tests/jax/single_chip/models/vit/vit_large_patch16_384/test_vit_large_patch16_384.py
@@ -5,10 +5,7 @@
 
 import pytest
 from infra import RunMode
-from utils import (
-    BringupStatus,
-    Category,
-)
+from utils import BringupStatus, Category, failed_runtime
 
 from ..tester import ViTTester
 from third_party.tt_forge_models.config import Parallelism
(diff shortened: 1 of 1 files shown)


.github/workflows/call-test.yml
```diff
@@ -274,14 +274,69 @@ jobs:
...
 
-        ESTIMATED_DURATION=$(echo "$COLLECTED_OUTPUT" | grep "estimated duration:")
-        ESTIMATED_DURATION=$(echo "$ESTIMATED_DURATION" | sed 's/.*estimated duration: \([0-9.]*\)s.*/\1/' | awk '{print int($1)}')
-        # multiply the estimate with 3 until estimates are more precise
-        ESTIMATED_DURATION=$((ESTIMATED_DURATION * 3 / 60))
-
-        MIN_TEST_DURATION=30
-        if [ $ESTIMATED_DURATION -lt $MIN_TEST_DURATION ]; then
-          ESTIMATED_DURATION=$MIN_TEST_DURATION
+        # Get list of test IDs in a simpler format using pytest -q
```
vmilosevic, line 277: Looks good, but let me try doing a simpler version which uses BASE_PYTEST_CMD

vmilosevic (approved):

PR #1614, opened 2025-10-08
Add isort pre-commit plugin

jameszianxuTT:
isort enforces python import sorting so this doesn't need to be checked manually on PRs.

We have hundreds of files that don't abide by this convention and it seems that adding a pre-commit hook triggers all files to be swept, so fixes need to be made to all files by pre-commit run --all-files.


AleksKnezevic (approved):

jameszianxuTT:
@nsumrakTT It seems to be a frequent nit comment when PRs are reviewed in this repo. I assume it's for aesthetic reasons. 

I just feel that if it's going to come up as a comment during review we should decide whether to enforce it properly (eg. via auto-checks) or not enforce it at all. 

cc @ajakovljevicTT

PR #1690, opened 2025-10-14
Add data parallel support for torch testing and initial set of 75 passing models

gengelageTT:
### Ticket
N/A

### Problem description
We want to add support to runner/test_models.py and the torch_model_tester infra so that single chip models can be tested in data parallel (with batched inputs). This was supported in tt-torch. 

### What's changed
- @kmabeeTT added `Parallelism.DATA_PARALLEL` as an option in runner/test_config.py and added ~75 passing tests to CI 
- added changes in runner/test_utils.py to batch inputs, create a shard spec function and create a mesh when `self.parallelism == Parallelism.DATA_PARALLEL`
- added support in torch_device_runner.py for shard spec functions that take in `args` and `kwargs`(data parallel) instead of `model` (tensor parallel)

### Checklist

### DIFF
--- a/tests/runner/test_utils.py
+++ b/tests/runner/test_utils.py
@@ -283,16 +283,24 @@ class DynamicTorchModelTester(TorchModelTester):
             inputs = self.loader.load_inputs()
 
         if self.parallelism == Parallelism.DATA_PARALLEL:
+
+            def batch_tensor(tensor, num_devices):
+                if isinstance(tensor, torch.Tensor):
+                    if tensor.dim() == 0:
+                        return tensor.repeat(num_devices)
+                    else:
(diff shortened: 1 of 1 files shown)


tests/runner/test_config/test_config_inference_data_parallel.py
```diff
@@ -0,0 +1,244 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
AleksKnezevic: Would it make sense to merge this with test_config_inference? Asking for opinion, not necessarily to change it

kmabeeTT (approved):
Awesome!

Issue #1265, opened 2025-09-02
Only reset torch seed and dynamo state if the test is a PyTorch test.


LPanosTT:
Currently, in `tests/conftest.py` we reset the torch seed and dynamo state between every pytest. For jax tests this is useless and wastes time. We should find a way to only perform such tasks if the next test is a PyTorch test.

kmabeeTT:
A previous version of that change in conftest.py looked at path of test being run (which I didn't love) and so I had to locally extend it to this to catch test_models.py which runs tt-torch tests.  Anyways, just mentioning this in case file path is used again in the future (I hope not).

```
    # Apply torch workaround to torch directory tests or test_models.py which runs torch tests
    dir_str = str(request.fspath.dirname)
    file_str = str(request.fspath)
    is_torch_test = ("torch" in dir_str) or ("tests/runner/test_models.py" in file_str)
```

PR #1704, opened 2025-10-15
Uplift tt-mlir and attempt hacked up exposure of internal global PJRTClient to test infra via ctypes

jameszianxuTT:
### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist

### DIFF
--- a/.github/workflows/test-matrix-presets/basic-test-nightly.json
+++ b/.github/workflows/test-matrix-presets/basic-test-nightly.json
@@ -1,6 +1,6 @@
 [
   { "runs-on": "wormhole_b0",        "name": "run_jax",               "dir": "./tests/jax/single_chip",                 "test-mark": "nightly and not large", "parallel-groups": 2 },
-  { "runs-on": "wormhole_b0",        "name": "run_torch",             "dir": "./tests/torch/single_chip",               "test-mark": "nightly" },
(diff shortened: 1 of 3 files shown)


jameszianxuTT:
@sgligorijevicTT This seems to resolve the failure mode where a local n300 multichip test hangs on repeat invocation without `--forked`. 

The reason that hang happened appears to be that the ClientInstance dtor we relied on for MeshDevice cleanup intentionally is not called by xla, which means we dangle an open MeshDevice by running literally anything. At the next time we try to acquire a dangled device (i.e. on repeat `pytest` invocation) we hang because it's UB to dangle devices like that and runtime has no device autoclose mechanism. 

I am not sure how to properly fix the fact that xla doesn't give us a teardown method. 

This hacky implementation exposes the PJRT client as a global entity and binds a C API function to release the mesh devices attached to that Client. The API function is exposed and "callable" from user code through test infra using `ctypes`/interrogating the `pjrt_plugin_tt.so` which seemed to be the least intrusive change, I didn't want to set up pybind infra just for an experiment. 

I then added an autouse fixture to release mesh devices after a test completes (yield -> release mesh devices), and to do so only for torch tests as suggested by Kyle in a previous issue #1265

PR #1719, opened 2025-10-17
Program Cache enabled by default, disable with optional TT_RUNTIME_ENABLE_PROGRAM_CACHE=0

kmabeeTT:
### Ticket
None

### Problem description
- It was requested that we strongly consider enabling program cache by default.  I tested the change to do that against tt-xla nightly CI on branch (link below) and everything continues to pass. 
- Maybe program-cache issues will be exposed down the road, but the idea here is to globally enable it so we can catch issues sooner rather than later.

### What's changed
- Make program cache enabled by default, flip usage of temp env-var to be TT_RUNTIME_ENABLE_PROGRAM_CACHE=0 to disable it (was previously TT_RUNTIME_ENABLE_PROGRAM_CACHE=1 to enable), considered renaming env-var but this seemed less confusing since it matches the API flag name.
- Update few tests/examples that would manually enable program cache with env-var, no longer needed.

### Checklist
- [x] Ran nightly Oct 16 (passing) with this change: https://github.com/tenstorrent/tt-xla/actions/runs/18571768016
- [x] Run nightly Oct 28 (matches main) with this change: https://github.com/tenstorrent/tt-xla/actions/runs/18896258322

### DIFF
--- a/examples/vllm/Qwen3-Embedding-4B/service.sh
+++ b/examples/vllm/Qwen3-Embedding-4B/service.sh
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-TT_RUNTIME_ENABLE_PROGRAM_CACHE=1 LOGGER_LEVEL=DEBUG vllm serve Qwen/Qwen3-Embedding-4B \
+LOGGER_LEVEL=DEBUG vllm serve Qwen/Qwen3-Embedding-4B \
     --max-model-len 16384 \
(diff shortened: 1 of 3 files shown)


odjuricicTT:
Nice! Definitely makes sense to have it on by default.

kmabeeTT:
Some debate offline about the usefulness of this change right now, and ownership, long term rollout. Going to close this for now.

AleksKnezevic (approved):

jnie-TT (approved):

PR #1740, opened 2025-10-21
Fix single test dispatch

vvukomanTT:
### Ticket
/

### Problem description
Workflow `Run Test Single` fails when ran on CIv1 runners because github's `fromJson()` method doesn't convert booleans in string representation to booleans in boolean representation.

### What's changed
- Strictly check for possible values of the `shared-runners` field.
- Convert "true" and "false" strings to booleans explicitly.
- Added a test workflow to validate current and future behaviour.

This will make the workflow more deterministic.

### Checklist


.github/workflows/pr-test-generate-test-matrix.yml
```diff
@@ -0,0 +1,279 @@
+name: Test Matrix Generator
```
nsumrakTT: Not in workflows, should this be there?
Should it have name with naming conventions?

nsumrakTT (approved):

PR #1754, opened 2025-10-22
Putting jax models in tt_forge_models test infra

ajakovljevicTT:
Changing `test_models.py`, to now run both `jax` and `pytorch` models. To enable this, I did some infra changes:

- Separated `DynamicTorchModelTester`, and added `DynamicJaxModelTester` and `DynamicJaxMultiChipModelTester`.
- All of the dynamic tester now have either a `TorchDynamicLoader` or `JaxDynamicLoader`, which is instatiated with a loader gotten from `tt_forge_models` 
- To support our custom tensor parallel jax models, their implementation has been moved to tt_forge_models, and must be merged there before this PR merges (PR here https://github.com/tenstorrent/tt-forge-models/pull/206)
- Deleted all the related inference tests in jax, left training tests to be moved to this infra later (cc @vladimirjovanovicTT )
- Data parallel test also to be added later

### DIFF
--- a/.github/workflows/test-matrix-presets/model-test-xfail-weekly.json
+++ b/.github/workflows/test-matrix-presets/model-test-xfail-weekly.json
@@ -1,4 +1,4 @@
 [
-  { "runs-on": "n150", "name": "run_forge_models_torch", "dir": "./tests/runner/test_models.py", "test-mark": "weekly and n150 and (known_failure_xfail or not_supported_skip)", "parallel-groups": 1 },
-  { "runs-on": "p150", "name": "run_forge_models_torch", "dir": "./tests/runner/test_models.py", "test-mark": "weekly and p150 and (known_failure_xfail or not_supported_skip)", "parallel-groups": 1 }
+  { "runs-on": "n150", "name": "run_forge_models_torch", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "weekly and n150 and (known_failure_xfail or not_supported_skip)", "parallel-groups": 1 },
+  { "runs-on": "p150", "name": "run_forge_models_torch", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "weekly and p150 and (known_failure_xfail or not_supported_skip)", "parallel-groups": 1 }
 ]


.github/workflows/test-matrix-presets/model-test-experimental.json
```diff
@@ -1,4 +1,6 @@
 [
-  { "runs-on": "n150", "name": "run_forge_models_torch", "dir": "./tests/runner/test_models.py", "test-mark": "unspecified and inference and single_device", "parallel-groups": 3 },
-  { "runs-on": "p150", "name": "run_forge_models_torch", "dir": "./tests/runner/test_models.py", "test-mark": "unspecified and inference and single_device", "parallel-groups": 3 }
+  { "runs-on": "n150", "name": "run_forge_models_torch", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "unspecified and inference and single_device", "parallel-groups": 3 },
```
vmilosevic: @vvukomanTT can we have "dir" be a list of paths to take the tests from? 
Pythest should support running a tests from a list of paths

tests/jax/single_chip/models/vit/vit_large_patch32_224_in21k/test_vit_large_patch32_224_in21k.py
```diff
@@ -35,27 +28,13 @@ def training_tester() -> ViTTester:
...
 
-@pytest.mark.model_test
-@pytest.mark.record_test_properties(
-    category=Category.MODEL_TEST,
-    model_info=MODEL_INFO,
-    parallelism=Parallelism.SINGLE_DEVICE,
-    run_mode=RunMode.INFERENCE,
-    bringup_status=BringupStatus.FAILED_RUNTIME,
-)
 @pytest.mark.xfail(
```
mmilosevicTT: Please remove inference stuff completely. This mark will be moved to training, which is not what we want. 

On another note, it might be beneficial to first add infra and then have single clean-up PR for these tests when we add training part as well.

.github/workflows/call-test.yml
```diff
@@ -260,11 +260,15 @@ jobs:
     # Ensure test_config files for tt-forge-models test are valid, especially on PR's for uplifting tt-forge-models
     # but still run tests even in case of failure here, will still treat overall job as failed.
     - name: Validate Forge Models Tests test config file
-      if: ${{ (matrix.build.name == 'run_forge_models_torch' || matrix.build.name == 'run_forge_models_torch_multichip') && strategy.job-index == 0 }}
+      if: ${{ (matrix.build.name == 'run_forge_models_torch' || matrix.build.name == 'run_forge_models_jax' || matrix.build.name == 'run_forge_models_torch_multichip') && strategy.job-index == 0 }}
```
vmilosevic: It would be better if we don't have this control logic based on build name, even if we have to add new field to config.

tests/jax/single_chip/models/vit/vit_large_patch32_224_in21k/test_vit_large_patch32_224_in21k.py
ajakovljevicTT: @mmilosevicTT I put both the infra and the test deletions in the PR so we can watch for infra regressions (which do happen), and not have double recording of tests

.github/workflows/call-test.yml
kmabeeTT: Yeah, would involve download+install of wheel which seemed like waste to run 10-15 second job that spins up pytest to validate the config file which is why I (for efficiency) bundled it in this way to existing job.

To be fair this condition being modified to contain build name exists a few times in call-test.yml though, so just moving out validate job wouldn't solve that bigger problem.  It could be simplifed in couple ways:

1. Some new flag maybe though haven't throught this through what this means
2. replace 3 condition or check with 1 condition checking if build name starts with `run_forge_models` which would catch `run_forge_models_torch` , (recently/temporarily-added to workaround hangs) `run_forge_models_multichip` and this new `run_forge_models_jax`

tests/infra/testers/single_chip/model/jax_model_tester.py
```diff
@@ -144,9 +145,19 @@ def _get_forward_method_kwargs(self) -> Mapping[str, Any]:
         if isinstance(self._model, FlaxPreTrainedModel):
             kwargs = {
                 "params": self._input_parameters,
-                "train": False if self._run_mode == RunMode.INFERENCE else True,
                 **self._input_activations,
             }
+
+            # Only add 'deterministic' if the model accepts it
+            try:
+                sig = inspect.signature(self._model.__call__)
```
mmilosevicTT: train parameter needs to be added to kwargs as well as static_args for the previous infra to work for training.

.github/workflows/test-matrix-presets/model-test-experimental.json
vmilosevic: This is beyond the scope of this PR, but Ideally I would like us to have test suite definition that doesn't deal with jobs, just saying which test we want to run on which hardware.
Then tests are automatically distributed to a number of jobs, dynamically based on test duration, for example into 30 minute jobs. 
That way we can simplify test config, optimal hardware usage and lowest wait time.

.github/workflows/call-test.yml
kmabeeTT: James' point is valid... this current change isn't sufficient to look at runs-on.  I prefer to go back to my suggestion in prev message using just 2 build names in json files and conditional checks like we have today, it basically makes all of the changes here in this discussion moot since you're just renaming two build names in their declaration and usage as follows, which simplfies things:

```
run_forge_models_torch                   => run_forge_models
run_forge_models_torch_multichip => run_forge_models_multichip
```

.github/workflows/call-test.yml
```diff
@@ -274,11 +276,11 @@ jobs:
     # Ensure test_config files for tt-forge-models test are valid, especially on PR's for uplifting tt-forge-models
     # but still run tests even in case of failure here, will still treat overall job as failed.
     - name: Validate Forge Models Tests test config file
-      if: ${{ (matrix.build.name == 'run_forge_models_torch' || matrix.build.name == 'run_forge_models_torch_multichip') && strategy.job-index == 0 }}
+      if: ${{ (matrix.build.name == 'run_forge_models' || matrix.build.name == 'run_forge_models_multichip') && strategy.job-index == 0 }}
       shell: bash
       run: |
         source venv/activate
-        python -m pytest -vv tests/runner/test_models.py::test_all_models --validate-test-config
+          python -m pytest -vv tests/runner/test_models.py::test_all_models --validate-test-config
```
kmabeeTT: I cleaned up this code in this file to make the changes as minimal as possible. Left a whitespace here by accident, feel free to remove when you have more substantiate change to re-push, don't want to re-trigger CI for just this whitespace fix.

ajakovljevicTT:
> Good stuff here, haven't been able to finish looking at it all - since lots of changes here. Would you consider splitting this into 2 PR's @ajakovljevicTT ?
> 
> 1. Infra changes to support JAX, CI updates we talked about, move/rename test_config files (the important stuff, but less files)
> 2. Follow up right after for the changes to populate JAX test_config files and delete hand written files (looots of changes)
> 
> If you saved a copy of branch, and dropped all the test changes and jax test_config files, might accomplish step 1 above.
> 
> Benefits:
> 
> 1. Allows you to not chase conflicts in config files everyday (they change often) and get important step in to start
> 2. Merge sooner :)
> 3. Less risk, arguably easier to review, etc.

@kmabeeTT The issue here is that the current changes to infra would break some of the manually written jax tests, mostly the multichip ones, and that would then need change. Meaning that I would have to change those tests, just to delete them very soon, which points to the changes being tightly coupled. Is it possible to disregard the test_config files for jax and deletion of the hand written files, as these are mostly trivial, and focus in the review on the changes to the infra (the first point in your list)?

.github/workflows/test-matrix-presets/basic-test.json
```diff
@@ -2,10 +2,8 @@
   { "runs-on": "wormhole_b0",        "name": "run_jax",               "dir": "./tests/jax/single_chip",                 "test-mark": "push and not large", "parallel-groups": 2 },
   { "runs-on": "n150",               "name": "run_torch",             "dir": "./tests/torch/single_chip",               "test-mark": "push and not llmbox", "parallel-groups": 3, "shared-runners": "true" },
   { "runs-on": "wormhole_b0",        "name": "run_meta_infra_tests",  "dir": "./tests/infra_tests/single_chip",         "test-mark": "push" },
-  { "runs-on": "n150",               "name": "run_large_jax_models",  "dir": "./tests/jax/single_chip",                 "test-mark": "push and large", "shared-runners": "true" },
```
kmabeeTT: We had a bunch of tests run here and now we don't, consider replacing this with something else to not lose coverage in onPR/onPush like was done in custom-basic-and-models-push.json?

.github/workflows/test-matrix-presets/model-test-experimental.json
```diff
@@ -1,4 +1,6 @@
 [
-  { "runs-on": "n150", "name": "run_forge_models_torch", "dir": "./tests/runner/test_models.py", "test-mark": "unspecified and inference and single_device", "parallel-groups": 3 },
-  { "runs-on": "p150", "name": "run_forge_models_torch", "dir": "./tests/runner/test_models.py", "test-mark": "unspecified and inference and single_device", "parallel-groups": 3 }
+  { "runs-on": "n150", "name": "run_forge_models", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "unspecified and inference and single_device", "parallel-groups": 3 },
+  { "runs-on": "p150", "name": "run_forge_models", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "unspecified and inference and single_device", "parallel-groups": 3 },
+  { "runs-on": "n150", "name": "run_forge_models", "dir": "./tests/runner/test_models.py::test_all_models_jax", "test-mark": "unspecified and inference and single_device", "parallel-groups": 3 },
```
kmabeeTT: with only 6 unspecified tests in experimental for jax compared to 54 (was ~100 recently) for torch, I think you can get away for parallel-groups:1 for now to keep things simpler (one place to look) and increase later as needed?

.github/workflows/test-matrix-presets/model-test-passing.json
```diff
@@ -1,7 +1,11 @@
...
-  { "runs-on": "n300", "name": "run_forge_models_torch_multichip", "dir": "./tests/runner/test_models.py", "test-mark": "data_parallel and n300 and expected_passing", "parallel-groups": 3 }
-
+  { "runs-on": "n150", "name": "run_forge_models", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "n150 and expected_passing", "parallel-groups": 10 },
+  { "runs-on": "p150", "name": "run_forge_models", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "p150 and expected_passing", "parallel-groups": 10 },
+  { "runs-on": "n300-llmbox", "name": "run_forge_models_multichip", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "tensor_parallel and n300-llmbox and expected_passing", "parallel-groups": 4 },
+  { "runs-on": "n300", "name": "run_forge_models_multichip", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "data_parallel and expected_passing", "parallel-groups": 3 },
+  { "runs-on": "n150", "name": "run_forge_models", "dir": "./tests/runner/test_models.py::test_all_models_jax", "test-mark": "n150 and expected_passing and not large", "parallel-groups": 10 },
```
kmabeeTT: Only ~60 jax tests today compared to ~400 torch, so probably should reduce from 10 jobs to 3 each for p150,150 single chip which is already pretty lightly loaded groups.

tests/runner/test_models.py
```diff
@@ -76,6 +83,30 @@ def test_all_models(
     request,
     capteesys,
 ):
+    # Fix venv isolation issue: ensure venv packages take precedence over system packages
```
kmabeeTT: Is this new venv/path chunk of code needed so that the new optional per-model system-requirements.txt installation works?   Can it go in a funtion (moved to tests/runner/test_utils.py maybe) so logic can be re-used between tests here?

tests/runner/test_models.py
```diff
@@ -4,27 +4,34 @@
...
 import pytest
-from infra import RunMode
+from infra import DynamicJaxMultiChipModelTester, RunMode
+from infra.testers.single_chip.model import (
+    DynamicJaxModelTester,
+    DynamicLoader,
+    DynamicTorchModelTester,
+    JaxDynamicLoader,
+    TorchDynamicLoader,
+)
```
kmabeeTT: It was nice that these classes were previously in same folder tests/runner/ which is the only place they are used so it was obvious to people (and recent CI optimization to allow PR changes to "only tests/runner/ folder" to only run forge-models push tests).  Them being in a `single_chip` folder (despite working for singlechip, multichip tests here) is also tiny bit misleading.  On the other hand, the classes are now closes to the other existing test infra code, so mixed feelings here.  Anyways, just a thought...

tests/runner/conftest.py
```diff
@@ -115,9 +128,32 @@ def pytest_sessionfinish(session, exitstatus):
...
+        "pytorch" in nodeid or "test_all_models_torch" in nodeid
+        for nodeid in _collected_nodeids
+    )
+    has_jax_tests = any(
+        "jax" in nodeid and "test_all_models_jax" in nodeid
+        for nodeid in _collected_nodeids
+    )
+
+    # Only validate configs for the framework(s) that were actually collected
+    if has_torch_tests and has_jax_tests:
```
kmabeeTT: Just for less code to go wrong, prefer for this to be removed and we just always validate all configs always. Like before, there wasn't any use case to only validate single_device configs, or only validate tensor_parallel configs, or only inference or training - I don't forsee useful case where we only want to ensure jax configs are good, or only ensure torch configs are good... we basically want them all to be good, all the time.  CI has single command to validate all, and I usually just find myself copy-pasting the same command from CI. Could be replaced with:

```
combined_test_config = torch_test_config | jax_test_config
```

kmabeeTT (approved):
Sorry for delay, finished looking at stuff (kind of skimmed the jax infra since not seen jax tests before), leaving approval, just some minor things/ideas and one bug to fix.  

If you want to avoid rebase loop of hell (I have some changes planned for test_config files for some promoted tests), consider putting this in separate PR just to get it merged easier (comment out reference to non-existent jax test_config files)

2025-10-21 (7534c2ca) by ajakovljevicTT (Author ajakovljevic@tenstorrent.com) : Changes to yaml

tests/runner/test_models.py
ajakovljevicTT: Agreed, will think about it a bit more.

ajakovljevicTT: Moved everything to `tests/runner/`

tests/runner/conftest.py
ajakovljevicTT: Edit: Added specific checking for configs for both torch and jax in the github job

kmabeeTT:
Looks awesome, thx for adjusting with changes and rebasing.  Only small problem is 2 new CI json files added here that need update to specify `test_all_models_torch` testname like was done for existing .json files.

```
.github/workflows/test-matrix-presets/model-test-passing-weekly.json
.github/workflows/test-matrix-presets/model-test-xfail-weekly.json
```

Making this change will re-trigger onPR and delay this PR from being merged by 2 hours, and we have some other changes that are dependent on this, so I propose to make this "zero risk" change, and forge merge this PR so long as nobody judges me for it (It's already passing and approved, it's silly to delay it by 2 hours to make a typo fix).  

Or we can merge as-is and fix it in follow up PR (which actually won't require onPR to rerun since just changing a weekly .json file, recent optimization released last night helps here).

PR #1766, opened 2025-10-22
Dump flatbuffers and preserve older versions of files when dumping

### LINKED ISSUE #1761: Change flatbuffer dump
Dumping flatbuffer using the serialization approach (https://github.com/tenstorrent/tt-xla/pull/947) fails for some models that generate multiple files in the persistent cache dir. 

It would be very useful to have it done the same way as it is done for IR modules in https://github.com/tenstorrent/tt-xla/pull/1612.

sgligorijevicTT:
### Ticket
closes #1761

### Problem description
Perf team consistently ran into problems with the serialization based approach for saving artifacts, and we still need to save flatbuffers

### What's changed
Now setting `export_path` also dumps the flatbuffer.
I also added logic to avoid overwriting existing files by appending a timestamp, as multiple graphs executing seems to be a frequent problem.

### Checklist


src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -1064,6 +1083,13 @@ ModuleBuilder::buildModuleForTTNNRuntime(
   tt::runtime::Binary flatbuffer(nullptr);
   tt_pjrt_status status = createFlatbufferBinary(mlir_module, input_shardings,
                                                  output_shardings, flatbuffer);
+
+  if (compile_options.export_path.has_value()) {
```
mrakitaTT: This should be done after the status is checked, otherwise flatbuffer might be invalid

src/common/pjrt_implementation/module_builder/module_builder.cc
```diff
@@ -76,6 +77,23 @@ namespace tt::pjrt::module_builder {
 
 const std::string c_mlir_format_name = "mlir";
 
+// Helper function to move an existing file to a new path with a timestamp
+// suffix, to avoid overwriting.
+static void preserveOlderFile(const std::filesystem::path &path) {
```
mrakitaTT: Instead of doing this, why not just always append the timestamp to the file path?

sgligorijevicTT:
@ajakovljevicTT PTAL

sgligorijevicTT:
@vkovacevicTT was stuck in review, got it unblocked now and I'll merge this now. FYI, as per review comments, now all IRs and flatbuffers will be serialized to files with a unix timestamp in the file name, hopefully this doesn't cause you any difficulties in opening the right file.

PR #1802, opened 2025-10-24
Add BufferInstance session-unique global ID 

jameszianxuTT:
### Ticket
None

### Problem description
It's hard to tell if two BufferInstances are the same, which can help for memory leak and tensor persistence debug investigations. BufferInstance pointers are highly unstable as they are deallocated unpredictably by the framework and reused at subsequent allocations.

### What's changed
Add BufferInstance UID system to attach a session-unique, atomically incrementing uint64_t ID to each bufferInstance. This implementation mimics tt-mlir runtime's [global tensor ID implementation](https://github.com/tenstorrent/tt-mlir/blob/00d3097a44e7007ba8194d7bb146ebc40a3012d0/runtime/lib/common/types.cpp#L127-L130) with a static atomic counter incremented with atomic fetch_add.

Access via bufferInstance member function `getUID()` returning uint64_t

### Checklist

### DIFF
--- a/pjrt_implementation/src/api/buffer_instance.cc
+++ b/pjrt_implementation/src/api/buffer_instance.cc
@@ -32,9 +32,6 @@
 
 namespace tt::pjrt {
 
-// Initialize the static UID counter
-std::atomic<uint64_t> BufferInstance::s_next_uid{0};
-
 std::unique_ptr<BufferInstance> BufferInstance::createInputBufferInstance(
     PJRT_Buffer_Type data_type, const std::int64_t *dims, size_t num_dims,
     DeviceInstance *device, MemoryInstance *memory) {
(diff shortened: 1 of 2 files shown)


pjrt_implementation/src/api/buffer_instance.cc
```diff
@@ -32,6 +32,9 @@
 
 namespace tt::pjrt {
 
+// Initialize the static UID counter
+std::atomic<uint64_t> BufferInstance::s_next_uid{0};
```
acolicTT: It's maybe easier to create static function in buffer_istance.h that returns next global id, and in cpp file just init this one as static and return next id. Something like:
```
static uint64_t NextUID
{
    static std::atomic<uint64_t> uid{0};
    return uid.fetch_add(1, std::memory_order_relaxed);
}
```
This way, we don't have to duplicate fetch_add logic for each call.

PR #1930, opened 2025-10-31
Fixing xfails on nightly CI

ajakovljevicTT:
Some models are failing on nightly CI either because of the issues with loading yolox, or because they were not marked as large. Fixing now.


nvukobratTT (approved):
cc @chandrasekaranpradeep @pdeviTT. I'll assign you two to check it out (since we've already solved this issue on the FFE side).

PR #1936, opened 2025-10-31
Cleanup compile config and add performance user docs

odjuricicTT:
### Problem description
We are missing instructions for users to increase performance on models that they are compiling.

### What's changed
* Added user docs for optimizer and other perf features
* Replace most perf related compile flags with `optimization_level`
* Uplift mlir to compatible commit

### Checklist


docs/src/performance.md
```diff
@@ -0,0 +1,193 @@
...
+#### Level 0 (Default)
+- Most optimization passes disabled
+- **Use for:** Iterating fast, safest option
+- **Compilation time:** Fastest
+- **Runtime performance:** Slowest
+
+#### Level 1 (Recommended)
+- Basic optimizations enabled
+  - Enables optimizer passes for basic op tuning
+  - Conv2D weights preprocessing and fusion patterns
```
nvukobratTT: Think about how to structure these with a bit more detail. Ideally, we would have it as is, as just a point to the documentation section with more details (e.g., a sub-page under the performance or features section). 

Not a hard requirement here, but we should follow up sooner

docs/src/performance.md
```diff
@@ -0,0 +1,193 @@
...
+- **Use for:** General model compilation, good balance
+- **Compilation time:** Moderate
+- **Runtime performance:** Good
+
+#### Level 2
+- Advanced optimizations enabled, all level 1 plus:
+  - Memory layout analysis for sharded layouts
```
nvukobratTT: Similar to above, a bit more details on what and why. E.g., layout analysis with a goal of doing XY in favor of better performance.

docs/src/performance.md
```diff
@@ -0,0 +1,193 @@
...
+  - Maximize usage of fast SRAM memory instead of DRAM
```
nvukobratTT: Using DRAM interleave in previous optimization options is a noticeable point, and we should mention it at least in level 2.

docs/src/performance.md
```diff
@@ -0,0 +1,193 @@
...
+The third iteration is for good luck :)
+
+All of the above is a one time fixed cost and all subsequent iterations of the model will be orders of magnitude faster.
+
+---
+
+## 3. Data Formats
+
+TT Hardware supports multiple lower precision data formats ([docs](https://docs.tenstorrent.com/tt-metal/latest/ttnn/ttnn/tensor.html#data-type)):
+* bfloat16
```
nvukobratTT: Metal docs have more DF outlined compared to ours. What about matching those, and/or removing bullet points here?

nvukobratTT: Also, we should emphasize that we're supporting these on the kernel level. However, from the frontend level user can use any data format they have available in the framework. However, the compiler will fallback to the closest available if requested one isn't directly supported, or we'll do accumulation in some cases (e.g., matmul).

examples/pytorch/compiler_options.py
```diff
@@ -26,12 +26,8 @@ def mnist_with_compiler_options():
 
     device = xm.xla_device()
 
-    # Define compiler options
     options = {
```
nvukobratTT: 🫶

examples/pytorch/mnist_performant.py
ajakovljevicTT: I would change the naming to mnist_optimization_lvl2 or similar

docs/src/performance.md
```diff
@@ -0,0 +1,193 @@
...
+* Model compilation and optimization
+* Op kernel compilation
+* Transferring of model weights to device
+* Const-eval of model weight and constants
+* Caching of op kernels on device
+
+The second iteration is needed for:
+* Capturing runtime trace to reduce op dispatch overhead (Section 4)
+
+The third iteration is for good luck :)
```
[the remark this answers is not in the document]
AleksKnezevic: lol

docs/src/performance.md
```diff
@@ -0,0 +1,193 @@
...
+
+All of the above is a one time fixed cost and all subsequent iterations of the model will be orders of magnitude faster.
```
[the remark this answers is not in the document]
AleksKnezevic: AFAIK, no, part of this has to do with the fact that harvesting options differ between chips, so kernels are not the same.

PR #1947, opened 2025-10-31
Sharded attention tests for Llama 3, Qwen 3, Qwen 2.5, Falcon, Gemma, Mistral

ssaliceTT:
### Ticket
N/A

### Problem description
Graph tests were missing sharded attention coverage for key production models (Llama 3, Qwen 3, Qwen 2.5, Falcon, Gemma, Mistral). Without these tests, we couldn't verify that attention layers correctly distribute computation across multiple devices.

### What's changed
- Added shared attention_prefill, attention_decode and spda tests for the 6 model families in `test_attention.py`
- Falcon-specific changes: Uses fused query_key_value projection instead of separate Q/K/V projections so used a custom wrapper function used to handle this. Also, SDPA test directly calls torch.nn.functional.scaled_dot_product_attention for Falcon test only.

### Checklist

### DIFF
--- a/tests/torch/single_chip/graphs/test_attention.py
+++ b/tests/torch/single_chip/graphs/test_attention.py
@@ -49,10 +49,10 @@ from third_party.tt_forge_models.mistral.pytorch.loader import (
 from third_party.tt_forge_models.mistral.pytorch.loader import (
     ModelVariant as MistralModelVariant,
 )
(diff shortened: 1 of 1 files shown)


AleksKnezevic:
Also, there are 43 files changed in the submodule update, is it the right version?

PR #1967, opened 2025-11-03
Initial support for nnx in test infra

vzeljkovicTT:
### Problem description
Our test infrastructure does not support Flax NNX model types

### What's changed
Added support for Flax NNX models in the test infrastructure


tests/infra/testers/single_chip/model/jax_model_tester.py
```diff
@@ -103,16 +107,23 @@ def _initialize_workload(self) -> None:
         args = self._get_forward_method_args()
         kwargs = self._get_forward_method_kwargs()
         forward_static_args = self._get_static_argnames()
-        forward_method_name = self._get_forward_method_name()
 
         assert (
             len(args) > 0 or len(kwargs) > 0
         ), f"Forward method args or kwargs or both must be provided"
-        assert hasattr(
```
ajakovljevicTT: Lets change this assert to still check for forward_method_name if the model is not nnx

tests/infra/testers/single_chip/model/jax_model_tester.py
```diff
@@ -103,16 +107,23 @@ def _initialize_workload(self) -> None:
         args = self._get_forward_method_args()
         kwargs = self._get_forward_method_kwargs()
         forward_static_args = self._get_static_argnames()
-        forward_method_name = self._get_forward_method_name()
```
ajakovljevicTT: Could this just be __call__() for nnx?

tests/infra/testers/single_chip/model/jax_model_tester.py
```diff
@@ -162,6 +173,8 @@ def _get_forward_method_kwargs(self) -> Mapping[str, Any]:
                     )
             except:
                 pass
+        elif isinstance(self._model, nnx.Module):
```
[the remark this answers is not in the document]
mmilosevicTT: By default we use this only for exampleModel, where we do not need any parameter, as we have `.train()` option. So I would say we only need to pass the model to train mode.

PR #1970, opened 2025-11-04
Add custom test for Surya OCR model

meenakshiramanathan1:
### Ticket
Fixes [#243 ](https://github.com/tenstorrent/tt-forge-models/issues/243)

### Problem description and changes

-  Model fails with below error from torch dynamo

```
            try:
                value.flags.writeable = True
            except ValueError:
                # One can not easily make nditer elements writable,
                # but warning is not the end of the world
>               assert isinstance(value.base, np.nditer)
E               AssertionError: 
E               
E               from user code:
E                  File "tt-xla/venv/lib/python3.11/site-packages/surya/detection/__init__.py", line 59, in torch_dynamo_resume_in_prepare_image_at_59
E                   img = np.asarray(img, dtype=np.uint8)

venv/lib/python3.11/site-packages/torch/_dynamo/variables/builder.py:1887: AssertionError
```

- The failure came from using np.asarray on a PIL Image, which can return a non-writeable view backed by PIL’s memory. TorchDynamo then tries to toggle writeable and hits an internal assertion. Switching to np.array(..., copy=True) forces a fresh, contiguous, writeable NumPy buffer (no odd base, flags.writeable=True). This change has been brought in by monkey-patching this change. 

- Custom test has been created to accomodate the above mentioned changes and model fails with `E       torch._dynamo.exc.TorchRuntimeError: Dynamo failed to run FX node with fake tensors: call_function <Wrapped method <original mul>>(*(FakeTensor(..., size=(1200, 1200, 3), dtype=torch.float64), 0.00392156862745098), **{}): got AttributeError("'ndarray' object has no attribute 'mul'")
`

- This failure is added as xfail in the inference test and will be handled in a future PR.

### Logs
[surya_ocr_text_xla.log](https://github.com/user-attachments/files/24100096/surya_ocr_text_xla.log)

### DIFF
--- a/tests/torch/models/surya_ocr/model_utils.py
+++ b/tests/torch/models/surya_ocr/model_utils.py
@@ -1,0 +1,20 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import numpy as np
+import torch
+from PIL import Image
+
+
+def _prepare_image(self, img):
+    new_size = (self.processor.size["width"], self.processor.size["height"])
+
+    img.thumbnail(new_size, Image.Resampling.LANCZOS)
+    img = img.resize(new_size, Image.Resampling.LANCZOS)
+    # Original line
+    # img = np.asarray(img, dtype=np.uint8)
+    # New line (patch): replace original with explicit copy
+    img = np.array(img, dtype=np.uint8, copy=True)
+    img = self.processor(img)["pixel_values"][0]
+    img = torch.from_numpy(img)
+    return img
--- a/.test_durations
+++ b/.test_durations
@@ -834,6 +834,7 @@
     "tests/torch/models/resnet/test_resnet.py::test_torch_resnet_inference[optimization_level_0-float32]": 22.216,
     "tests/torch/models/resnet/test_resnet.py::test_torch_resnet_inference_trace": 24.383,
     "tests/torch/models/stable_diffusion_3_5/test_stable_diffusion_3_5_medium.py::test_torch_stable_diffusion_3_5_medium_inference": 0.002,
+    "tests/torch/models/surya_ocr/test_surya_ocr_text.py::test_torch_surya_ocr_text_inference": 1500,
     "tests/runner/test_models.py::test_all_models_torch[xception/pytorch-xception41-single_device-training]": 203.077,
     "tests/runner/test_models.py::test_all_models_torch[wide_resnet/pytorch-wide_resnet101_2.timm-single_device-training]": 146.852,
     "tests/runner/test_models.py::test_all_models_torch[regnet/pytorch-regnet_y_3_2gf-single_device-training]": 151.887,
--- a/tests/runner/test_config/torch/test_config_placeholders.yaml
+++ b/tests/runner/test_config/torch/test_config_placeholders.yaml
@@ -65,8 +65,6 @@ PLACEHOLDER_MODELS:
     bringup_status: NOT_STARTED
   Command A Reasoning:
     bringup_status: NOT_STARTED
-  SURYA OCR:
-    bringup_status: NOT_STARTED
   SAM_large_sa1b:
     bringup_status: NOT_STARTED
   Motif-Technologies/Motif-Image-6B-Preview:
--- a/tests/torch/models/surya_ocr/test_surya_ocr_text.py
+++ b/tests/torch/models/surya_ocr/test_surya_ocr_text.py
@@ -1,0 +1,72 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import inspect
+
+import pytest
+from infra import RunMode
+from utils import BringupStatus, Category
+
+import third_party.tt_forge_models.suryaocr.pytorch.loader as surya_loader
+from tests.runner.requirements import RequirementsManager
+from third_party.tt_forge_models.config import Parallelism
+from third_party.tt_forge_models.suryaocr.pytorch.loader import (
+    ModelLoader,
+    ModelVariant,
+)
+
+from .tester import SuryaOCRTester
+
+VARIANT = ModelVariant.OCR_TEXT
+MODEL_INFO = ModelLoader._get_model_info(VARIANT)
+
+
+# ----- Fixtures -----
+
+
+@pytest.fixture
+def inference_tester() -> SuryaOCRTester:
+    return SuryaOCRTester(ModelVariant.OCR_TEXT)
+
+
+@pytest.fixture
+def training_tester() -> SuryaOCRTester:
+    return SuryaOCRTester(ModelVariant.OCR_TEXT, run_mode=RunMode.TRAINING)
+
+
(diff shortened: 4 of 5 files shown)


saiarthiraguram (approved):

meenakshiramanathan1:
This PR will be made ready fore review once this tt_forge_models [commit](https://github.com/tenstorrent/tt-forge-models/pull/240) gets tracked in xla.

tests/torch/single_chip/models/surya_ocr/model_utils.py
```diff
@@ -0,0 +1,98 @@
...
+        self.key_cache.append(new_layer_key_cache)
+        self.value_cache.append(new_layer_value_cache)
+
+    self.attention_mask = torch.zeros(
+        (self.max_batch_size, self.max_cache_len), device=device, dtype=torch.long
+    )
+    self.text_token_counts = [
+        torch.zeros(self.max_batch_size, dtype=torch.long, device=device)
+        for _ in range(self.num_layers)
+    ]
```
vkovinicTT: Why do we need this?

meenakshiramanathan1: This line is part of [code](https://github.com/datalab-to/surya/blob/5497449bfab790aa43e56693f302c50b9c352d5c/surya/foundation/cache.py#L54) from surya package, only [this](https://github.com/tenstorrent/tt-xla/blob/b3f11ed7d0923786ebab872c5cd5706d1f241ebf/tests/torch/single_chip/models/surya_ocr/model_utils.py#L70) part has been patched.

tests/torch/single_chip/models/surya_ocr/tester.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(
+        self,
+        variant: ModelVariant,
+        comparison_config: ComparisonConfig = ComparisonConfig(),
+        run_mode: RunMode = RunMode.INFERENCE,
+    ) -> None:
+        self._model_loader = ModelLoader(variant)
+        super().__init__(comparison_config, run_mode)
+
+    # @override
```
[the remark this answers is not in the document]
meenakshiramanathan1: It is just an optional comment present in [other](https://github.com/tenstorrent/tt-xla/blob/a45335dfde8cbe091f4ae86b3720d7648824f46a/tests/torch/single_chip/models/resnet/tester.py#L29) custom test tester files as well.

tests/torch/single_chip/models/surya_ocr/model_utils.py
```diff
@@ -0,0 +1,101 @@
...
+    # Cache init is taken from huggingface StaticCache - https://github.com/huggingface/transformers/blob/67ddc82fbc7e52c6f42a395b4a6d278c55b77a39/src/transformers/cache_utils.py#L1125
+    self.key_cache: list[torch.Tensor] = []
+    self.value_cache: list[torch.Tensor] = []
+    cache_shape = (
+        self.max_batch_size,
+        self.num_key_value_heads,
+        self.max_cache_len,
+        self.head_dim,
+    )
+    device = torch.device(device) if device is not None else None
```
vkovinicTT: In the reference [implementation](https://github.com/huggingface/transformers/blob/67ddc82fbc7e52c6f42a395b4a6d278c55b77a39/src/transformers/cache_utils.py#L1155) they have:
```python
        for idx in range(config.num_hidden_layers):
            if layer_device_map is not None:
                layer_device = layer_device_map[idx]
```
but we here always assign the same device, what if we are running model on multiple devices? Why not stick to the original implementation?

tests/torch/single_chip/models/surya_ocr/model_utils.py
```diff
@@ -0,0 +1,101 @@
...
+    for _ in range(config.num_hidden_layers):
+        new_layer_key_cache = torch.zeros(cache_shape, dtype=self._dtype, device=device)
+        new_layer_value_cache = torch.zeros(
+            cache_shape, dtype=self._dtype, device=device
+        )
+        # New line (patch): guard static address marking to avoid Dynamo forbidden-callable cases
+        if not torch._dynamo.is_compiling():
+            torch._dynamo.mark_static_address(new_layer_key_cache)
+            torch._dynamo.mark_static_address(new_layer_value_cache)
```
vkovinicTT: How does this affects our compile flow, since the original [implementation](https://github.com/huggingface/transformers/blob/67ddc82fbc7e52c6f42a395b4a6d278c55b77a39/src/transformers/cache_utils.py#L1161) has this comment?
```python
# Note: `mark_static_address` is used to tag the cache as a fixed data pointer,
# preventing compiled graph breaks when updating the cache.
```
I think (this is just my hypothesis) that since we don't mark static address in compile flow, we will trigger graph breaks, needing further compiling and in that process we might hit the limit of recompiles (as can be seen in the log?). Can we verify whether it is okay not to mark static address if we compile model?

vkovinicTT:
> @vkovinicTT Removed the static cache related patch applied because in latest main, model is failing with a different issue which has been added as xfail. This issue will be solved in a new PR and if the static cache related issue is reproduced again, it will be handled in next PR keeping into consideration the comments left in this PR related to static cache changes. Latest Log: [surya_ocr_text_xla.log](https://github.com/user-attachments/files/24099996/surya_ocr_text_xla.log)

Can we try to fix this new issue, since what I can see from the logs, it doesn't look like it is something that hard to fix?

meenakshiramanathan1:
@ajakovljevicTT Can you please review this PR?

tests/torch/single_chip/models/surya_ocr/model_utils.py
```diff
@@ -0,0 +1,48 @@
...
+
+    if torch.cuda.is_available():
+        return "cuda"
+
+    if torch.backends.mps.is_available():
+        return "mps"
+
+    # Only consider XLA when explicitly enabled; avoid auto-detect to prevent
+    # importing torch_xla in environments/tests that expect CPU behavior.
+    enable_xla = os.getenv("SURYA_ENABLE_XLA", "").lower() in ("1", "true", "yes", "y")
```
ajakovljevicTT: can you explain in more detail why do we do this?

meenakshiramanathan1: This is being done so that TORCH_DEVICE_MODEL won’t import torch_xla unless SURYA_ENABLE_XLA is explicitly set. Without this guard, merely detecting the device could import torch_xla early, initialize PJRT with the wrong/default backend, and eventually would fail with `RuntimeError: Can't change device type after XLA runtime is initialized`

tests/torch/models/surya_ocr/model_utils.py
```diff
@@ -0,0 +1,20 @@
...
+from PIL import Image
+
+
+def _prepare_image(self, img):
+    new_size = (self.processor.size["width"], self.processor.size["height"])
+
+    img.thumbnail(new_size, Image.Resampling.LANCZOS)
+    img = img.resize(new_size, Image.Resampling.LANCZOS)
+    # Original line
+    # img = np.asarray(img, dtype=np.uint8)
```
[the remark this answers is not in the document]
ajakovljevicTT: Noted, makes sense, thanks!

Issue #1929, opened 2025-10-31
Yolo models have issues with loading yolox==0.3.0 from tt-forge models


ajakovljevicTT:
Currently yolo models have issues with loading yolox==0.3.0 from tt-forge-models requirements.txt file, needs further investigation.

chandrasekaranpradeep:
@ajakovljevicTT  Can you please share the link of yolox issues in nightly and also the tests cases which are failing with the issue?

chandrasekaranpradeep:
@ajakovljevicTT / @nvukobratTT, I’ve identified a YOLOX issue in the tt-xla nightly — [link](https://github.com/tenstorrent/tt-xla/actions/runs/18958627822/job/54141082605#step:16:822).
Could you please confirm if this is the issue that needs to be addressed?

PR #1974, opened 2025-11-04
Fix yolox nightly issue

chandrasekaranpradeep:
Fixes https://github.com/tenstorrent/tt-xla/issues/1929

### **Summary**
In the nightly pipeline, YOLOX model tests were failing with the following error:  
[AssertionError: torch is required for pre-compiling ops, please install it first.](https://github.com/tenstorrent/tt-xla/actions/runs/18958627822/job/54141082605#step:16:822)

This error occurred even though `torch` was already installed.  
The issue appeared only when installing YOLOX through a subprocess, because the `pip` installation inside the subprocess used an **isolated build environment** by default.  
In this isolated environment, preinstalled dependencies like `torch` were not accessible — causing YOLOX’s setup script to fail when trying to import `torch` for pre-compiling ops.

### **Root Cause**
- `pip` uses *build isolation* by default, creating a temporary environment without access to already installed `site-packages`.  
- YOLOX’s `setup.py` attempts to import `torch` during build time.  
- Since the isolated environment doesn’t have `torch`, the build failed, leading to yolox test failure.

### **Fix**
To resolve the YOLOX installation issue, build isolation was **disabled** during installation.

**Changes made:**
- Renamed the requirements file from `requirements.nodeps.txt` → `requirements.nodeps.nobuildisolation.txt` for the YOLOX model in the `tt_forge_models` repository.  
- Added an optional installation step in the `RequirementsManager` to install the `requirements.nodeps.nobuildisolation.txt` file **without dependencies** and **without build isolation**.

Tested and verified these fix in the nightly pipeline for yolox - https://github.com/tenstorrent/tt-xla/actions/runs/19040102718/job/54376712258

### DIFF
--- a/tests/runner/requirements.py
+++ b/tests/runner/requirements.py
@@ -108,6 +108,26 @@ class RequirementsManager:
             _dbg(f"[Requirements] __enter__: installing (no-deps) -r {nodeps_path}")
             self._pip(("install", "--no-input", "--no-deps", "-r", nodeps_path))
 
+        # Optional: a sibling file 'requirements.nodeps.nobuildisolation.txt' for packages to install without dependencies and isolated environment
+        nodeps_no_build_isolation_path = os.path.join(
+            os.path.dirname(self.requirements_path),
+            "requirements.nodeps.nobuildisolation.txt",
+        )
+        if os.path.isfile(nodeps_no_build_isolation_path):
+            _dbg(
+                f"[Requirements] __enter__: installing (no-deps and no-build-isolation) -r {nodeps_no_build_isolation_path}"
+            )
+            self._pip(
+                (
+                    "install",
+                    "--no-input",
+                    "--no-deps",
+                    "--no-build-isolation",
+                    "-r",
+                    nodeps_no_build_isolation_path,
+                )
+            )
+
         _dbg("[Requirements] __enter__: running pip freeze (after)")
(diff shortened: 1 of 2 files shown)


tests/runner/requirements.py
```diff
@@ -108,6 +108,26 @@ def __enter__(self) -> "RequirementsManager":
...
+        )
+        if os.path.isfile(nodeps_no_build_isolation_path):
+            _dbg(
+                f"[Requirements] __enter__: installing (no-deps and no-build-isolation) -r {nodeps_no_build_isolation_path}"
+            )
+            self._pip(
+                (
+                    "install",
+                    "--no-input",
+                    "--no-deps",
```
ajakovljevicTT: Do we also run with no deps?

ajakovljevicTT:
Can you also set the models in question to xfail (with the appropriate reasons) instead of skip? And do you have a xfail run that passed (if not, please run only that via the `Run Tests` pipeline)

saiarthiraguram (approved):

chandrasekaranpradeep:
> Can you also set the models in question to xfail (with the appropriate reasons) instead of skip? And do you have a xfail run that passed (if not, please run only that via the `Run Tests` pipeline)

Updated test config for yolox

ajakovljevicTT (approved):

PR #1999, opened 2025-11-05
[pjrt] ensure client destruction on process exit

### LINKED ISSUE #1824: torch_xla ClientInstance destructor doesn't get called leading to leaked devices and mesh
This means that tt-xla tests going through the torch_xla path in a multidevice environment will softlock the environment after the test process exits. Any future tests run on that machine will hang, until tt-smi reset is issued.

pilkicTT:
`torch_xla` doesn't call `PJRT_ClientDestroy` properly. This means that we are not closing the devices properly.

Recently, this started causing hangs on `n300` boards on subsequent execution of tests.

This PR introduces a global singleton object which will ensure that we are properly destroying the client instance on process shutdown. The singleton serves as a fallback mechanism if the framework doesn't call `PJRT_ClientDestroy` - like in the case of `torch_xla`.

Additionally, optimizer sub-meshes are now closed after each compilation; this previously was needed to avoid hangs, but now it causes them. Leaving the mechanism of persisting optimizer submesh in the code base, so that we can play with it if needed. Obviously, we need to dig deep into these issues to fix them properly.

NOTE: this does not solve the case when the process terminates abruptly, e.g. in case of `SIGSEGV` (segmentation fault). For this, ideally we would want a fix on `tt-metal` side.

Closes #1824

### DIFF
--- a/pjrt_implementation/inc/api/client_instance.h
+++ b/pjrt_implementation/inc/api/client_instance.h
@@ -47,14 +47,14 @@ class ModuleBuilder;
 class GlobalClientInstanceSingleton {
 public:
   static ClientInstance *getClientInstance();
-  static PJRT_Error *init_client();
-  static void destroy_client();
+  static PJRT_Error *initClient();
+  static void destroyClient();
 
 private:
   GlobalClientInstanceSingleton(std::unique_ptr<ClientInstance> client_instance)
       : m_client_instance(std::move(client_instance)) {}
 
-  bool is_initialized() const { return m_client_instance != nullptr; }
+  bool isInitialized() const { return m_client_instance != nullptr; }
 
   static GlobalClientInstanceSingleton &getInstance();
   std::unique_ptr<ClientInstance> m_client_instance;
--- a/pjrt_implementation/src/api/client_instance.cc
+++ b/pjrt_implementation/src/api/client_instance.cc
@@ -37,7 +37,7 @@
 
 namespace tt::pjrt {
 
-PJRT_Error *GlobalClientInstanceSingleton::init_client() {
+PJRT_Error *GlobalClientInstanceSingleton::initClient() {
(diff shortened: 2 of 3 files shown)


jameszianxuTT:
cc @kmabeeTT

pjrt_implementation/inc/api/client_instance.h
```diff
@@ -34,6 +34,32 @@ namespace module_builder {
...
+// `PJRT_Client_Destroy` API properly and `tt-metal` currently cannot recover
+// (on n300 boards) if we do not properly close all previously opened devices -
+// which is done on client destruction.
+//
+// NOTE: This serves only as a fallback option if `PJRT_Client_Destroy` is not
+// called by the framework.
+class GlobalClientInstanceSingleton {
+public:
+  static ClientInstance *getClientInstance();
+  static PJRT_Error *init_client();
```
sshonTT: I think we need to use camel case? (`initClient`)

pjrt_implementation/src/api/client_instance.cc
```diff
@@ -446,6 +477,13 @@ ClientInstance::openMeshDevice(const std::vector<uint32_t> &mesh_shape) {
   return tt::runtime::openMeshDevice(options);
 }
 
+void ClientInstance::closeOptimizerSubmesh() {
```
sshonTT: Can we also call this from destructor of `ClientInstance`?

sshonTT (approved):

jameszianxuTT (approved):

kmabeeTT:
Thank you! Works good for me, am able to run back to back pytest cmds on n300-llmbox now without hangs (and needing to call tt-smi -r between tests).  I still get DRAM leak between tests in same pytest cmd, but that's a separate issue to figure out.

aleksandarcolic22414 (approved):

Issue #1052, opened 2025-08-20
Improve tt-xla superset integration


mrakitaTT:
Currently tt-xla has a superset integration similar to other frontends, but it records properties statically for each pytest. This requires us manually updating the properties per test. We should explore how this works in tt-torch and see what improvements we can make.

There are also some features missing currently:
- we don't record the `parallelism` mode
- we don't use `ModelInfo` from tt-forge-models and log these fields manually
- we are missing data for large models (added with special marker in [this CI](https://github.com/tenstorrent/tt-xla/pull/852))

nvukobratTT:
We shared our feedback for Superset on Slack. 

@mrakitaTT @sdjukicTT, would you like to summarize the next step for this issue, or do you think we're fine with closing it and continuing further improvements through other issues?

nvukobratTT:
> Thanks for the context! For dynamic recording, let's create a new issue and link it to this one aa s parent:
> 
> * [Feature parity #991](https://github.com/tenstorrent/tt-xla/issues/991)
> 
> [@mrakitaTT](https://github.com/mrakitaTT) [@sdjukicTT](https://github.com/sdjukicTT), can you confirm if other parts are in place?

I'm closing this issue. 

In sum
- `parallelism` is now recorded
- model auto-discovery is using `ModelInfo` for property recording
- large models are recently enabled: https://github.com/tenstorrent/tt-xla/pull/852\

The additional part is related to dynamic property recording; however, this is tracked through a separate issue:
- https://github.com/tenstorrent/tt-xla/issues/1343

Feel free to re-open if something wasn't covered for PyTorch parity.

PR #2012, opened 2025-11-06
Add dynamic tracking of BringupStatus for test infra

mstojkovicTT:
### Ticket
#1343

### Problem description
explained in #1343 and #1052

### What's changed
This PR is solving the part where we want to dynamicly check for bringup status of models, without reading from `config.yaml`-s. I also change a bit how some functions work to be able to add this change.

How it works:
We create file `_bringup_stage.txt` at the start of test using `pytest.fixture` and deleting it at the end of the single test using the same fixture. We write to that file using `LOG_BRINGUP_STAGE` API implemented for logging the current bringup stage (we override the file every time we hit different bringup stage). Later on, we just read the last checkpoint we reached and classify it as one of the `BringupStatus` enums.

Left the logic for us to be able to write `reasons` and `bringup_status` for models that segfault the CI, so we need to mark them as `ModelTestStatus.NOT_SUPPORTED_SKIP`.

To enable this feature, use `ENABLE_BRINGUP_STAGE_LOGGING=1` env var inside nightly CI.

### DIFF
--- a/tests/runner/test_models.py
+++ b/tests/runner/test_models.py
@@ -19,6 +19,7 @@ from tests.runner.test_utils import (
     ModelTestStatus,
     fix_venv_isolation,
     record_model_test_properties,
+    update_test_metadata_for_exception,
 )
 from tests.runner.testers import (
     DynamicJaxModelTester,
@@ -84,6 +85,7 @@ def test_all_models_torch(
     record_property,
     test_metadata,
     request,
+    capteesys,
 ):
     # Fix venv isolation issue: ensure venv packages take precedence over system packages
     fix_venv_isolation()
@@ -127,9 +129,9 @@ def test_all_models_torch(
                 Comparator._assert_on_results(comparison_result)
 
         except Exception as e:
-            # TODO: remove this once we have a better way to set the reason dynamically.
-            # and handle it in record_model_test_properties.
-            setattr(test_metadata, "runtime_reason", str(e))
+            err = capteesys.readouterr().err
+            # Record runtime failure info so it can be reflected in report properties
+            update_test_metadata_for_exception(test_metadata, e, stderr=err)
             raise
         finally:
             # If there are multiple comparison results, only record the first one because the
@@ -205,6 +207,7 @@ def test_all_models_jax(
     record_property,
     test_metadata,
     request,
+    capteesys,
 ):
     # Fix venv isolation issue: ensure venv packages take precedence over system packages
     fix_venv_isolation()
@@ -255,9 +258,9 @@ def test_all_models_jax(
                 Comparator._assert_on_results(comparison_result)
 
         except Exception as e:
-            # TODO: remove this once we have a better way to set the reason dynamically.
-            # and handle it in record_model_test_properties.
-            setattr(test_metadata, "runtime_reason", str(e))
+            err = capteesys.readouterr().err
+            # Record runtime failure info so it can be reflected in report properties
+            update_test_metadata_for_exception(test_metadata, e, stderr=err)
             raise
         finally:
             # If there are multiple comparison results, only record the first one because the
--- a/tests/runner/test_utils.py
+++ b/tests/runner/test_utils.py
@@ -195,6 +195,26 @@ def parse_last_bringup_stage() -> BringupStatus | None:
     return stage_to_status.get(stage_name)
 
 
+def update_test_metadata_for_exception(
+    test_metadata, exc: Exception, stderr: str
+) -> None:
+    """
+    Inspect exception message and set `failing_reason` and `runtime_reason` on `test_metadata`.
+    """
+    try:
+        message = str(exc)
+    except Exception:
+        message = repr(exc)
+
+    # Find failing reason by raised exception
+    failing_reason = FailingReasonsFinder.find_reason_by_exception(exc)
+
+    # TODO: remove this once we have a better way to set the reason dynamically.
+    # and handle it in record_model_test_properties.
+    setattr(test_metadata, "runtime_reason", message)
+    setattr(test_metadata, "failing_reason", failing_reason)
+
+
 # This is needed for combination of pytest-forked and using ruamel.yaml
 # ruamel returns ScalarFloat/ScalarString types (subclasses of float/str).
 # pytest-forked uses Python's marshal, which rejects non-builtin subclasses inside the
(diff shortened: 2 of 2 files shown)


tests/runner/test_utils.py
```diff
@@ -22,6 +23,12 @@
 from tests.utils import BringupStatus, Category
 from third_party.tt_forge_models.config import Parallelism
 
+# Path to bringup stage file at repo root
+_REPO_ROOT = os.path.dirname(
+    os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+)
+BRINGUP_STAGE_FILE = os.path.join(_REPO_ROOT, "._bringup_stage.txt")
+
```
sdjukicTT: Do we need to join with _REPO_ROOT here?

pjrt_implementation/src/api/module_builder/module_builder.cc
```diff
@@ -227,6 +227,7 @@ ModuleBuilder::buildModule(
     const std::unordered_map<std::string, std::string> &compile_options_map,
     ClientInstance *client_instance) {
   DLOG_F(LOG_DEBUG, "ModuleBuilder::buildModule");
+  LOG_BRINGUP_STAGE("FE_COMPILATION_START");
```
sgligorijevicTT: Arguably this is already `FE_COMPILATION_END` ... As soon as you get to the plugin, the frontend has already done it's job in my book.
I guess there's a bit of room to argue if "frontend SHLO" passes are part of frontend or MLIR, but realistically they're closer to being part of mlir compilations(I mean they are already in MLIR). FE compilation IMO mostly encompasses tracing, which is not our job.

mstojkovicTT: I think they are in ownership of FE devs so we want to record them as FE fail. If you think that MLIR people should look at these error, we can change this.

tests/runner/test_utils.py
```diff
@@ -159,43 +162,34 @@ def _normalize_markers(self, markers_value):
...
-# and this would be updated to use actual bringup_status achieved by a model.
-def update_test_metadata_for_exception(
-    test_metadata, exc: Exception, stderr: str
-) -> None:
+def parse_last_bringup_stage() -> BringupStatus | None:
     """
-    Inspect exception message, stderr and set `runtime_bringup_status` and `runtime_reason` on `test_metadata`.
+    Read the current stage from file and map to BringupStatus.
+
+    This function reads the structured logging marker written to ._bringup_stage.txt (at repo root)
```
jameszianxuTT: Nit: Maybe add this to .gitignore? No guarantees the os.remove() will trigger in case of a segfault

pjrt_implementation/src/api/flatbuffer_loaded_executable_instance.cc
```diff
@@ -410,6 +410,7 @@ void FlatbufferLoadedExecutableInstance::releaseResources() {
 tt_pjrt_status FlatbufferLoadedExecutableInstance::execute(
     PJRT_LoadedExecutable_Execute_Args *args) {
   DLOG_F(LOG_DEBUG, "FlatbufferLoadedExecutableInstance::Execute");
+  LOG_BRINGUP_STAGE("RUNTIME_EXECUTION_START");
```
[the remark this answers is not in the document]
mstojkovicTT: > I would like to capture a history of bringup statuses in _bringup_stage.txt as it would be an interesting artifact (as it allows us to observe graph breaks), but if that adds too much complexity then it's not necessary.

Do you want us to just append to file instead of overriding it, and leave it for offline debugging? Or would you like to record it somewhere (in XML for example)?

Keep in mind that file is deleted after the test is finished, so you would have to put some breakpoints ether way to get the file.

tests/runner/test_models.py
```diff
@@ -129,9 +127,9 @@ def test_all_models_torch(
                 Comparator._assert_on_results(comparison_result)
 
         except Exception as e:
-            err = capteesys.readouterr().err
-            # Record runtime failure info so it can be reflected in report properties
-            update_test_metadata_for_exception(test_metadata, e, stderr=err)
+            # TODO: remove this once we have a better way to set the reason dynamically.
+            # and handle it in record_model_test_properties.
+            setattr(test_metadata, "runtime_reason", str(e))
```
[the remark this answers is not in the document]
mstojkovicTT: `capteesys` was only used inside the runtime logic for getting the bringup status. Since we no longer use that logic, and for runtime reason we just used exception as string, i removed it.

tests/runner/conftest.py
```diff
@@ -34,6 +36,21 @@ def pytest_addoption(parser):
     )
 
 
+def pytest_runtest_setup(item):
```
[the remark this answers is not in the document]
mstojkovicTT: You are right, it would be better to have fixture for this and not use hooks. Will change that!

tests/runner/test_utils.py
mstojkovicTT: Added to .gitignore :))

pjrt_implementation/src/api/module_builder/module_builder.cc
nvukobratTT: Yup, this is a fine line depending on how we're interpreting these stages, and both suggestions are valid from different standpoints. 

For this use case, these will be used to get a rough overview of where we are with the current models and how we're going to prioritize each of them with different teams to tackle them. Therefore, having SHLO of FE semantically and historically makes sense, mostly as FE folks were the ones pushing support for this. 

P.S. Even if you check our current ownership chart, you'll see that the start of MLIR ownership is marked with TTIR, which indicates that we're expecting that SHLO will be part of the FE ownership going forward.

pjrt_implementation/src/api/flatbuffer_loaded_executable_instance.cc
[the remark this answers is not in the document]
jameszianxuTT: Thanks for the explanation. I don't want to add too much complex logic for no reason to manage this file's persistence.

I think one useful and pretty much "free" datapoint we can get is by stuffing a a history of per-test bringup_reason into `tags->bringup_reason_history`. That might give us some useful debug information in the future. However, I do not want to add too much complexity here without an immediate need, so I think this can always be revisited/amended pretty easily in a future PR.

sdjukicTT (approved):

PR #2085, opened 2025-11-10
enable bringup_stage_logging

mstojkovicTT:
### What's changed
Enable `BringupStage` logging in CI.

Edit: also removed bringup status from configs.

### DIFF
--- a/tests/runner/test_config/torch/test_config_inference_data_parallel.yaml
+++ b/tests/runner/test_config/torch/test_config_inference_data_parallel.yaml
@@ -68,7 +68,6 @@ test_config:
     supported_archs: [n300]
     assert_pcc: false
     status: EXPECTED_PASSING
-    bringup_status: INCORRECT_RESULT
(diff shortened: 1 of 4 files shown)


.github/workflows/call-test.yml
```diff
@@ -316,6 +316,7 @@ jobs:
         HF_HOME: /mnt/dockercache/huggingface
         GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
         HF_TOKEN: ${{ secrets.HF_TOKEN }}
+        ENABLE_BRINGUP_STAGE_LOGGING: 1
```
nvukobratTT: Should we remove the bringup status for all nightly models now that this feature is enabled?

nvukobratTT (approved):

.github/workflows/call-test.yml
mstojkovicTT: Yes, but we will need to keep models marked as `NOT_SUPPORTED_SKIP` bringup status until we find a better way to store info for them

Issue #1966, opened 2025-11-03
Making onPr more reliable


ajakovljevicTT:
Currently, our onPR run has several issues, making debugging and test stability worse, and wasting more engineering time than is needed:
- onPR runs all ops parametrized, which rarely fail or catch bugs, but waste compute resources
- onPR does not have a representable subset od torch/jax models, and we catch most of the regression in nightly
- On nightly fails a lot due to this, and since it takes up to 6h, we cannot run on uplift or on every PR

Some feature requests that could solve it:
- get a subset of torch and jax models in in onPR
- triage the current tests in on PR
- change the onPR to run the tests added in that specific PR, even if they are added as nightly

Tagging @vmilosevic and @nsumrakTT if they have more ides.

nvukobratTT:
cc @kmabeeTT

nvukobratTT:
Hey guys, any action points on this one? 

Seems like we can kick off these two right away:
- get a subset of torch and jax models in onPR
- triage the current tests in on PR

@ajakovljevicTT, would you like to push on these two points?

> change the onPR to run the tests added in that specific PR, even if they are added as nightly

This one probably requires a bit more discussion, probably something that @kmabeeTT @nsumrakTT  can push/own? Does this make sense? Other suggestions?

PR #2089, opened 2025-11-10
Add some models from torch runner to onpush

sgligorijevicTT:
### Ticket
Addresses part of #1966 , specifically getting much more converage on onPr at a (hopefully) reasonable CI time cost

### Problem description
Our onPR coverage is not that extensive and that allows mistakes to slip trough

### What's changed
I have added a push market to a small subset of models to torch runner config. We already have a decent-ish converge in Jax config so I didn't touch that part(*yet*)
Rationale for each of the models added:
- CLIP - tests quickgelu, pretty important model(it's one part of Stable Diffusion), and it doesn't work rn to boot, we should fix sooner rather than later
-  Mamba - Non transformer LM, representative of most SSMs
- YOLOv3 - relatively typical CNN, tho YOLO loss and postprocessing can get gnarly, very useful to test the backward for this
- EfficientNet - relatively small CNN, mostly representative of a typical CNN
- Hrnet - tests strided convolutions
- Bert - very typical transformer encoder, tests gelu too
- Opt 350M - One of the rare postnorm transformers, this can cause funky numerics in backward.
- Beit - at one point we ran into a weird gather due to positional embeddings in Beit, so I want to test that
- llama 3.2 1b - Almost all LLMs these days are llama likes, gives us a lot of bang for the buck to test
- ViT - RED, tests patch embeddings, prototype for all non CNN vision models
- Resnet 50 - RED, very typical CNN. Tests maxpool and avgpool
- Vovnet - the oneshot agreggation has potential to catch if we do something silly re lifetime management in graphs
- Gemma 2 2B - Tests prepostnorm/"sandwitch norm" and GeGLU. Ideally this would have been Gemma 2 9B as that one also has tanh softcapping.

### Checklist

### DIFF
--- a/tests/runner/test_config/torch/test_config_inference_single_device.yaml
+++ b/tests/runner/test_config/torch/test_config_inference_single_device.yaml
@@ -44,6 +44,7 @@ test_config:
   clip/pytorch-base_patch16-single_device-full-inference:
     status: KNOWN_FAILURE_XFAIL
     reason: "RuntimeError('Check failed: handle->HasValue(): Trying to access XLA data for tensor with ID 3802 while an async operation is in flight: UNKNOWN_SCALAR[]') - https://github.com/tenstorrent/tt-xla/issues/1306"
+    markers: ["extended"]
 
   clip/pytorch-base_patch32-single_device-full-inference:
     status: KNOWN_FAILURE_XFAIL # Newly exposed in Sept 6 due to tt-mlir uplift.
@@ -87,6 +88,7 @@ test_config:
     arch_overrides:
       n150:
         required_pcc: 0.98 # Is lately 0.992 but close to threshold.
+    markers: ["extended"]
 
   openpose/v2/pytorch-single_device-full-inference:
     assert_pcc: false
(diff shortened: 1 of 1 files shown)


ajakovljevicTT (approved):

AleksKnezevic:
@sgligorijevicTT, how much does this increase on push CI by?

sgligorijevicTT:
Looks like ~40minutes extra, which is a lot admittedly.
Three longest running tests are:
```
385.84s call     tests/runner/test_models.py::test_all_models_torch[hrnet/pytorch-hrnet_w18-single_device-full-inference]
340.53s call     tests/runner/test_models.py::test_all_models_torch[mamba/pytorch-mamba-790m-hf-single_device-full-inference]
289.53s call     tests/runner/test_models.py::test_all_models_torch[gemma/pytorch-google/gemma-2-2b-it-single_device-full-inference]
```
Hrnet can go, as I see now that we have mobiletnetv1 on onpr which also has a strided conv somewhere, and then maybe one of the other two can also go. Mamba takes longer but tests new arch, meanwhile gemma is closer to a llama but takes less time...

AleksKnezevic:
@nvukobratTT we can have graph tests that test individual unique subgraph components if we cannot achieve similar coverage from ops. There is already a subsection of such tests in `torch/single_device/graph_tests`. If we want to test unique attention mechanisms, embeddings, etc, I'm ok with adding tests there. 

Our on PR is already too long, My pushback to adding all these models is that I don't see how/when they'll be replaced by subtests in the future. It's much easier to add things in CI than it is to remove them, and I don't want to get into the same state we had with buda CI where we had hundreds of models spinning without providing any meaningful testing that we couldn't achieve from subgraphs / ops.

Issue #1940, opened 2025-10-31
Memory leaking on device (metal OOM issue)


gengelageTT:
### Describe the issue
Running multiple tests without closing the device in between can cause tests to fail with out of DRAM when they should be passing.
```
Out of Memory: Not enough space to allocate 117440512 B DRAM buffer across 12 banks, where each bank needs to store 9805824 B, but bank size is only 1073741792 B
```
This causes issues in CI where multiple tests recorded as failing when they should be passing. [example here](https://github.com/tenstorrent/tt-xla/actions/runs/18973924069/job/54190515895#step:16:388) <- this test passes locally 

Running tests with consteval disabled seems to resolve this issue, but we want to keep consteval enabled.

### How to repro
Running the following command (on an n300) will cause the first 2 tests to fail with out of DRAM and the third test to fail unexpectedly with the same error. The third test passes when run on its own.
```
pytest -svv tests/runner/test_models.py::test_all_models[llama/causal_lm/pytorch-llama_3_8b_instruct-data_parallel-full-inference] tests/runner/test_models.py::test_all_models[vit/pytorch-vit_l_16-data_parallel-full-inference] tests/runner/test_models.py::test_all_models[llama/causal_lm/pytorch-llama_3_2_3b_instruct-data_parallel-full-inference]
```
If you run the same command with consteval disabled [here](https://github.com/tenstorrent/tt-xla/blob/main/pjrt_implementation/src/api/compile_options.cc#L45), only the first test fails.

[Note]: Some data parallel tests are currently failing since program cache is enabled by default. This issue can be resolved by applying [this change ](https://github.com/tenstorrent/tt-xla/commit/0cabdba6b4d8e2450c53e646f13f9be290139f2c)

jameszianxuTT:
From testing, I do not see that LoadedExecutableInstances are ever destructed during execution of multiple tests within the same process. Not sure what the correct lifetime of these entities should be, or if this is side effect of something in the framework holding onto executables (eg. some sort of compilation cache, which I'm pretty sure exists since we can skip compilation when executing a test in duplicate in the same proc.)

If we don't have a route through PJRT to destruct a LoadedExecutableInstance/clear LEI cache, then we will need some bypass path to either clear consteval cache or clean up LoadedExecutableInstances, which would tie into this solution: https://github.com/tenstorrent/tt-xla/pull/1952

[...]

PR #2108, opened 2025-11-12
[bugfix] Closing device on tt::runtime::submit failure

acolicTT:
It might happen that tt::runtime::submit fails (sometimes with exceptions from metal). Since those exceptions are never caught, we are leaving mesh device open and in potentially bad state. With this change, we will always close mesh device on tt::runtime::submit failure.

One concrete problem occurs when buffer allocations fails in metal, which leaves allocator in a bad state. Next time allocator is called (in the next test run within the same process, for example), it will fail with the same allocation error even if the memory is available.

In the future, we might want to wrap all calls to runtime with introduced function wrapper (std::invoke) and close mesh device on failure.

Issue: https://github.com/tenstorrent/tt-xla/issues/1940

### DIFF
--- a/pjrt_implementation/inc/utils/utils.h
+++ b/pjrt_implementation/inc/utils/utils.h
@@ -10,6 +10,7 @@
 #include <optional>
 #include <sstream>
 #include <type_traits>
+#include <variant>
 #include <vector>
 
 namespace tt::pjrt::utils {
@@ -28,7 +29,7 @@ template <typename T> std::string to_string(const std::vector<T> vec) {
 // Invokes function with arguments inside try/catch block.
 // Return type is std::optional with derived type from calling function, so user
 // can check return value for nullopt (or with implicit bool operator).
-// If function return type is void, std::optional will hold bool.
+// If function return type is void, std::optional will hold std::monostate.
 //
 // Example usage:
 // std::vector<int> my_fn(int first, int second);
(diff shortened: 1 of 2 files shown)


pjrt_implementation/inc/utils/utils.h
```diff
@@ -21,6 +25,44 @@ template <typename T> std::string to_string(const std::vector<T> vec) {
...
+// if (!r)
+//  ... failed - handle error ....
+//
+// or keep going if successful:
+// std::vector<int>& output = *r;
+// ...
+//
+template <class Fn, class... Args,
+          class ReturnType = std::invoke_result_t<Fn, Args...>>
+std::optional<
```
sgligorijevicTT: minor philosophical nit, feel free to ignore:
It somewhat bothers me that in the case T==void, this return optional<bool> rather than bool, where the Egaged(false) path is never used. As I understand it, this is to workaround the fact that void is not a regular type. Logically what you want here is an optional<monostate>, which is isomorphic to bool.
TLDR:
Replace `optional<T!=void?T:bool>` with `T!=void?optional<T>:bool`. Admittedly this does involve uglifying the code by copy pasting the if constexpr to the return after catch.
OR, replace bool with monostate

sgligorijevicTT (approved):
Discussed offline, there's a concern about how this interacts with prepared inputs(Runtime Stitching) and, after #1657 is merged, with results not yet pulled to host. But as I understand, we already have the same problem with close-then-open mesh procedure we do when executing graphs with different world sizes.

nvukobratTT:
@jameszianxuTT thanks for the repro! This seems orthogonal to this PR. Can we open a separate issue for potential DRAM leaks with the details above, and merge this one that solves the current use case and implements a sort of graceful shutdown :))

PR #2113, opened 2025-11-12
Fix NOT_STARTED and bring back xfail models

mstojkovicTT:
### Problem description
We lost `NOT_STARTED` models from superset dashboards due to change where we didnt run whole `tests/runner/test_models.py`.

Also there was a mistake at #1979 where torch xfail test were deleted from json, added them as well here.

### What's changed
Added `placeholder` and `n150 and (known_failure_xfail or not_supported_skip) and not large` models to `model-test-xfail`.


kmabeeTT (approved):
Looks good now!  Tried out a test and see NOT_STATED bringup status written out now to junit xml:

```
    <testcase classname="tests.runner.test_models" name="test_placeholder_models[Boltz2]" time="0.326">
      <properties>
        <property name="error_message" value="Not yet started or WIP"/>
        <property name="tags" value="{'test_name': 'test_placeholder_models', 'specific_test_case': 'test_placeholder_models[Boltz2]', 'category': 'model_test', 'model_name': 'boltz2', 'model_info': {}, 'run_mode': 'inference', 'bringup_status': 'NOT_STARTED', 'failing_reason': {'name': None, 'description': None, 'component': None}, 'parallelism': 'single_device', 'arch': None}"/>
        <property name="owner" value="tt-xla"/>
        <property name="group" value="red"/>
      </properties>
```

One downside of recent refactor/cahnges with dynamic bringup_status is that this change is even neeeded, but also test_config_placeholders.yml cannot set other hardcoded bringup statuses other than `NOT_STARTED` currently like they could before. We don't do it often, but we did it once before for some models that weren't yet merged and running in nightly, but that weren't really `NOT_STARTED`, so we'd set something else.  That's okay for now, it's not popular and eventually this placeholder list/test should die..

vmilosevic (approved):

nvukobratTT (approved):
Looks good! Thanks!

PR #2116, opened 2025-11-12
Expose L1_SMALL_SIZE

### LINKED ISSUE #2112: Expose L1_SMALL size
We ran into multiple problems with tt-mlir L1_SMALL_SIZE set in [tt-mlir/include/Constants.h.](https://github.com/tenstorrent/tt-mlir/blob/main/include/Constants.h) where we needed to increase L1_SMALL size in order for some models to work.

We should expose L1_SMALL_SIZE so it can be set per model, like it is in tt-metal.

vkovacevicTT:
### Ticket
Closes #2112 
Closes https://github.com/tenstorrent/tt-mlir/issues/5768

### Problem description
We should have L1_SMALL_SIZE exposed so we can set it per model.

### What's changed
Added `TT_RUNTIME_L1_SMALL_SIZE` env var that sets `l1SmallSize` in `MeshDeviceOptions`.

### DIFF
--- a/pjrt_implementation/src/api/client_instance.cc
+++ b/pjrt_implementation/src/api/client_instance.cc
@@ -466,9 +466,19 @@ ClientInstance::openMeshDevice(const std::vector<uint32_t> &mesh_shape) {
     traceRegionSize = std::stoull(std::getenv("TT_RUNTIME_TRACE_REGION_SIZE"));
   }
 
+  // TODO(vkovacevic, #1480): This is a temporary way to set L1 small size
+  // until we have a proper way for a user to pass device options. After that
+  // this should be removed. Issue for device options:
+  // https://github.com/tenstorrent/tt-xla/issues/1480
+  std::optional<size_t> l1SmallSize = std::nullopt;
+  if (std::getenv("TT_RUNTIME_L1_SMALL_SIZE") != nullptr) {
+    l1SmallSize = std::stoull(std::getenv("TT_RUNTIME_L1_SMALL_SIZE"));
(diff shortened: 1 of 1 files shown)


pjrt_implementation/src/api/client_instance.cc
```diff
@@ -466,9 +466,19 @@ ClientInstance::openMeshDevice(const std::vector<uint32_t> &mesh_shape) {
     traceRegionSize = std::stoull(std::getenv("TT_RUNTIME_TRACE_REGION_SIZE"));
   }
 
+  // TODO(vkovacevic, #1480): This is a temporary way to set L1 small size
```
nvukobratTT: > We should have L1_SMALL_SIZE exposed so we can set it per model.

Can we share more reasoning on why the frontend should care about this configuration? I.e., why is MLIR/Metal not handling it? Please provide more details.

vkovacevicTT: This is a runtime thing set in tt-mlir as a constant.
Some new models we add (e.g. Vanilla UNet) require more L1_SMALL then anticipated, thus setting this constant to max(L1_SMALL_SIZE_REQUIREMENT) for all models is unpredictable (and we might be wasting L1).

I agree that from the user perspective this is just another confusing parameter, but currently we would like to be flexible in setting this per model, like it is in tt-metal.

cc @odjuricicTT

vkovacevicTT:
Closing this since we don't want to go with the env variable and are looking at setting this as a compile option.

PR #2124, opened 2025-11-12
Add llama rms norm unit test

gengelageTT:
### Ticket
N/A

### Problem description
We want to add a unit test for [LlamaRMSNorm](https://github.com/huggingface/transformers/blob/5689dd6b8ed0257c6c43c0937d0d224d05774ecc/src/transformers/models/llama/modeling_llama.py#L62).

### What's changed
Added `test_llama_rms_norm` unit test to new `tests/torch/single_chip/graphs/test_rms_norm.py` following same structure as attention unit tests.

### Checklist

### DIFF
--- a/tests/torch/single_chip/graphs/test_rms_norm.py
+++ b/tests/torch/single_chip/graphs/test_rms_norm.py
@@ -1,25 +1,12 @@
 # SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
 #
 # SPDX-License-Identifier: Apache-2.0
-
-from typing import Callable
-
-import numpy as np
 import pytest
 import torch
 import torch_xla
-import torch_xla.core.xla_model as xm
 import torch_xla.runtime as xr
 from infra import Framework, run_graph_test
 from infra.comparators.comparison_config import ComparisonConfig, PccConfig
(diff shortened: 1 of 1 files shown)


tests/torch/single_chip/graphs/test_rms_norm.py
```diff
@@ -0,0 +1,92 @@
...
+)
+
+MODEL_LOADER_MAP = {
+    "llama": LlamaModelLoader,
+}
+
+AVAILABLE_VARIANT_MAP = {
+    "llama": [
+        "llama_3_8b",
+        "llama_3_8b_instruct",
```
ndrakulicTT: `_instruct` models should have the same architecture as the base versions so maybe we don't need it for the graph tests?

ndrakulicTT (changes requested):
Sorry for blocking, but what is executions time on this tests as they will be part of the onPR tests?

gengelageTT:
@ndrakulicTT  These tests are marked as nightly and shouldn't be part of onPR tests. They ran in 5m40s [here](https://github.com/tenstorrent/tt-xla/actions/runs/19307334336/job/55221960253)

I'll remove the `_instruct` variants and look into if it's possible to avoid loading the models.

Issue #1969, opened 2025-11-03
torch.fmax, torch.fmin, torch.nextafter do nothing on device + ND segfault/PCC drop when combined with program cache and async tensor return to host


jameszianxuTT:
When bringing up #1657, a rebase past #1719 to enable program cache by default reveals an interesting nondeterministic combined failure.

When running one of the fmax, fmin and nextafter tests, we hit nondeterministic PCC drops (to the same PCC when it drops), and rarer nondeterministic segfaults.  

`pytest -svv tests/torch/single_chip/ops/test_eltwise_binary_ops.py::test_fmin`
`pytest -svv tests/torch/single_chip/ops/test_eltwise_binary_ops.py::test_fmax`
`pytest -svv tests/torch/single_chip/ops/test_eltwise_binary_ops.py::test_nextafter`

The nondeterminism can be observed in a loop, where installing `pytest-loop` and running pytests with `loop=100` we can see between 1-20 failures, or an immediate segfault on the first execution.

This does not appear to be a runtime stitching failure with #1657, since there is only a single compute graph generated for each of these optests and the failure (PCC drop/segfault) may occur on the first loop. 

Instead, it seems to be some revealed bug related to _changing the timing_ for when tensors are returned from device to host, from being immediate post-device-execution, to being around 800ms afterwards (which is the point - to return results lazily).

Moreover, these 3 optests, which on their face are distinct, compile to the same, pointless compute graph.

jameszianxuTT:
Another datapoint with MLIR debug build +  `TTMLIR_RUNTIME_LOGGER_LEVEL=Debug TT_RUNTIME_ENABLE_PROGRAM_CACHE=1 LOGGER_LEVEL=DEBUG pytest -svv tests/torch/single_chip/ops/test_eltwise_binary_ops.py::test_fmin`

<img width="750" height="167" alt="Image" src="https://github.com/user-attachments/assets/42c033ff-502d-4119-a23d-176d171410bb" />

In this "program" nothing at all executes.

PR #2125, opened 2025-11-12
Lock m_copy_to_host_thread with mutex to avoid copyToHost race condition

jameszianxuTT:
### Problem description
There is an arcane, double race condition described in #1969, exposed by combined changes in #1657 + #1719 and occurring nondeterministically on 3/100s of optests, + no model tests.

This race condition causes nondeterministic segfaults and tensor output data clobbering due these conditions:

1. Metal program cache is not thread safe (https://github.com/tenstorrent/tt-metal/issues/32113) and produces all sorts of race conditions if you trigger multiple concurrent ttnn op dispatches which are entirely untested in ttnn
2. Changes in #1657 introduce an op dispatch for untilize to host for bf16 tensors specifically (due to a tt-mlir runtime optimization), and this dispatch can have multiple instances of itself occurring concurrently in copyToHost due to:
3. An second race condition in our PJRT bufferInstance copyToHost that allows multiple copyToHost threads to be spawned simultaneously when triggered in close proximity Also, there is no protection at all for different BufferInstance instances calling copyToHost concurrently, which newly becomes a problem in #1657 which adds a metal dispatch toHost call inside the async thread. For example, if the framework ran a program with multiple outputs and tried to retrieve them all at the same time in concurrent threads, each grabbing a single bufferInstance's contents.

### What's changed
Lock bufferInstance copy to host thread with a global mutex so two simultaneously initiated copyToHost threads cannot trigger concurrently, even between multiple BufferInstances. This still allows any individual copyToHost thread to act asynchronously, and better encodes the intent to avoid triggering multiple copyToHost threads simultaneously.

It also device to host buffer transfers. However, as it seems that metal does not support concurrent d2h transfers, we are constrained to his anyways.

### Checklist

### DIFF
--- a/pjrt_implementation/src/api/buffer_instance.cc
+++ b/pjrt_implementation/src/api/buffer_instance.cc
@@ -135,10 +135,11 @@ void BufferInstance::deleteData() {
   }
 
   // Wait if there is a copy to host in progress.
-  std::lock_guard<std::mutex> copy_lock(s_copy_to_host_thread_mutex);
+  std::unique_lock<std::mutex> copy_lock(s_copy_to_host_thread_mutex);
   if (m_copy_to_host_thread) {
     m_copy_to_host_thread->join();
   }
+  copy_lock.unlock();
 
   // Just reset the tensors, deallocation happens automatically when
   // reference count drops to zero.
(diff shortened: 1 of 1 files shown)


jameszianxuTT:
Proper qualification of this PR will involve proving that it averts the issue in #1969 when coupled with the triggering change #1657

jameszianxuTT:
A couple patches later and this does work. 

It is worth considering if serializing the (to be) lazy output return to host between all BufferInstances is acceptable, as this is a compromise we have to make to implement #1657. I think we will be paying this cost (serially) for all output runtime tensors in the original case, which would forcibly happen at every execution. 

Now, our copyToHost is slower, but in sum we won't be copying as many buffers to host (eg. static caches), so it does seem worth it.

Issue #2128, opened 2025-11-13
Investigate missing failing reasons for INCORRECT_RESULTS


meenakshiramanathan1:
Most of the current INCORRECT_RESULTS are being reported without a failing reason. Recently introduced regex patterns (via a recent PR) don’t seem to match many of the current PCC/Accuracy failures. We need to investigate whether the existing regex checks in tests/infra/utilities/failing_reasons/checks_xla.py need correction or if new ones should be added to properly capture these failures.

PR #2130, opened 2025-11-13
Fix Unclassified and N/A Failures

meenakshiramanathan1:
### Ticket

- https://github.com/tenstorrent/tt-xla/issues/2128

### Problem description

- Most of the current INCORRECT_RESULTS are being reported without a failing reason.

### What's changed

- Majority of the unclassified errors were due to Out of Memory error and the current OOM rule only matches when the last traceback line is from infra/runners/torch_device_runner.py, it misses cases where the OOM appears only in runtime logs and not on the exception’s last line or message. Now the current change broadens matching to also look for the OOM phrase anywhere in the captured long traceback

- New Failing reasons has been added for the below cases which were originally captured as unclassified as failing reason
   a. RuntimeError: "deformable_im2col" not implemented for 'BFloat16'
   b. RuntimeError: 'nms_kernel' not implemented for 'BFloat16'
   c. TypeError: AutoShape.forward() takes from 2 to 5 positional arguments but 7 were given
   d. TorchRuntimeError: Dynamo failed to run FX node with fake tensors
   e. RuntimeError: The tensor has a non-zero number of elements, but its data is not allocated yet
   f. Phi3VForCausalLM.forward got unexpected keyword 'max_new_tokens'
   g. AssertionError at torch_model_tester.py:70: assert isinstance(self._model, torch.nn.Module)

### DIFF
--- a/tests/infra/utilities/failing_reasons/checks_xla.py
+++ b/tests/infra/utilities/failing_reasons/checks_xla.py
@@ -505,7 +505,134 @@ class FailingReasons(Enum):
                     ),
                 ],
                 error_log=[
-                    M.last_line(M.contains("infra/runners/torch_device_runner.py:")),
+                    M.any(
+                        # Accept when comparator/runner is the last line
+                        M.last_line(
+                            M.contains("infra/runners/torch_device_runner.py:")
+                        ),
+                        # Also accept when the OOM message appears in the captured longrepr anywhere
+                        M.regex(
+                            "Out of Memory: Not enough space to allocate .* B (?:L1|L1_SMALL|DRAM) buffer across .* banks"
+                        ),
+                    ),
+                ],
+            ),
+        ],
+    )
+
+    TORCHVISION_DEFORM_IM2COL_BFLOAT16_NOT_IMPLEMENTED = FailingReason(
+        description="deformable_im2col not implemented for bfloat16",
+        checks=[
+            # RuntimeError: "deformable_im2col" not implemented for 'BFloat16'
+            ExceptionCheck(
+                class_name="RuntimeError",
+                error_log=[
+                    M.contains("deformable_im2col"),
+                    M.contains("not implemented"),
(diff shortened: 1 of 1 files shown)


saiarthiraguram (approved):
LGTM!

tests/infra/utilities/failing_reasons/checks_xla.py
```diff
@@ -505,7 +505,172 @@ def find_by_description(cls, desc: str) -> Optional["FailingReasons"]:
...
+            ),
+        ],
+    )
+
+    TORCHVISION_DEFORM_IM2COL_BFLOAT16_NOT_IMPLEMENTED = FailingReason(
+        description="deformable_im2col not implemented for bfloat16",
+        checks=[
+            # RuntimeError: "deformable_im2col" not implemented for 'BFloat16'
+            ExceptionCheck(
+                class_name="RuntimeError",
```
vbrkicTT: Component information is missing. It's optional but may be helpful in superset.

tests/infra/utilities/failing_reasons/checks_xla.py
```diff
@@ -505,7 +505,172 @@ def find_by_description(cls, desc: str) -> Optional["FailingReasons"]:
...
+            ExceptionCheck(
+                class_name="TypeError",
+                message=[
+                    M.contains("Phi3VForCausalLM.forward"),
+                    M.contains("unexpected keyword argument"),
+                    M.contains("max_new_tokens"),
+                ],
+                error_log=[
+                    M.any(
+                        M.contains("Phi3VForCausalLM.forward"),
```
vbrkicTT: Message (exception message) and error log usually contains some same information. Usually error_log includes all information from message so there is no need to check for same message in both places.

Example `Phi3VForCausalLM.forward` is part of exception message. If we check it in message section via M.contains usually it's redundant with same check in error_log

PR #2180, opened 2025-11-14
Refactoring jax traning tests to go through automatic test discovery infra

ajakovljevicTT:
Refactoring jax training tests, which were written manually, to align with our automatic test discovery infra.
Couple of things done in this PR:
- Added training support in test_all_models_jax in test_models.py
- Added configs for jax training (for now, they are all set to skip, except the one passing test, do not put pressure on the nightly CI)
- Changed the loaders in tt-forge-models (PR here: https://github.com/tenstorrent/tt-forge-models/pull/271), to align with these changes. Note that tt-forge-models PR needs to go into meain first for this to merge.


tests/runner/testers/jax/dynamic_jax_model_tester.py
```diff
@@ -7,9 +7,9 @@
 import inspect
 
 import jax
-import jax.numpy as jnp
 from infra.comparators import ComparisonConfig
 from infra.testers.single_chip.model import JaxModelTester, RunMode
+from transformers import FlaxPreTrainedModel
```
ndrakulicTT: future question: do we have any plan how we will handle transformer v5 where jax is deprecated, My idea was that we can get around that limitation by using `RequirementsManager` but that will probably not work if we include it in global scope

tests/runner/test_utils.py
```diff
@@ -378,4 +386,7 @@ def record_model_test_properties(
     if test_metadata.status == ModelTestStatus.NOT_SUPPORTED_SKIP:
         pytest.skip(reason)
     elif test_metadata.status == ModelTestStatus.KNOWN_FAILURE_XFAIL:
+        # Ensure we have a reason for xfail (fixes crash when reason is None)
```
ndrakulicTT: This should be fixed with: #2149

PR #2193, opened 2025-11-14
[vLLM plugin] Rename all tests to single chip tests

### LINKED ISSUE #2192: [vLLM plugin] Update testing to support single/multi chip tests
All the existing tests are added for single chip, though there is no way to distinguish between single chip and multi chip tests. Tests need to be divided between single chip and multi chip; so that they can be executed on appropriate devices.

mmanzoorTT:
### Ticket
closes #2192 

### Problem description
Existing tests assume a single chip, but there’s no clear separation between single-chip and multi-chip tests.

### What's changed
- Created a dedicated directory for single-chip tests.
- Updated CI JSON files to reference the new paths for single-chip tests.

### Checklist


AleksKnezevic (changes requested):
I would rather not have the tests in separate places. There's no need that a test that we have for N150 shouldn't also work on llmbox in data-parallel. Let's keep all the tests in one place and use `@pytest.mark.single_device` and `@pytest.mark.llmbox`

PR #2215, opened 2025-11-17
Model `paged_fill_cache` and `paged_scaled_dot_product_attention_decode`. Alter vLLM plugin to utilize paged attention.

LPanosTT:
**THIS PR IS DEPENDENT ON UPLIFTING TT-MLIR BEYOND:** `36e172c46b99e9fc63e64bbd79836dcce0a344ac`

### Problem description
Our vLLM plugin uses static cache and static attention. We wish to use paged attention

### What's changed
- Model `tt.paged_fill_cache`
- Model `tt.paged_scaled_dot_product_attention_decode`
- Add unit tests for both.
- Altered vLLM plugins attention module, and model runner to operate using paged attention/KV cache
- Renamed variables and purged unused ones to more accurately represent what they are for TT's use case
- Move page table and cache position to device in model runner instead of within the attention module
- Generative models no longer require an attention mask as the combination of `cache_position` and `is_causal = True` allows the op to attend correctly.
- Increased context length for TinyLlama demo, alter client to keep track of conversation history for a better "chat" experience.
### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/platform.py
+++ b/integrations/vllm_plugin/vllm_tt/platform.py
@@ -127,11 +127,6 @@ class TTPlatform(Platform):
         from vllm.config import CompilationLevel, CUDAGraphMode
 
         cache_config = vllm_config.cache_config
-        if cache_config.gpu_memory_utilization > 0.2:
-            logger.warning(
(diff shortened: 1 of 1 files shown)


LPanosTT:
I would have to keep how the KV cache is generated and some other things in place as well. vLLM is built for using paged attention and so this will be the version we will wish to push forward with. I do not think it is worth adding this complexity. @AleksKnezevic what do you think?

Issue #1535, opened 2025-10-01
Add infra to assert whether fusion patterns work


AleksKnezevic:
Add infra to FE tests that will run certain ops (rotary embedding, attention, concat/split heads) and assert that fusion patterns in tt-mlir worked. 

This is important as it will allow us to catch if different fusion patterns are needed after model changes (i.e. transformers updates).

PR #2230, opened 2025-11-18
Bringup LLVM FileCheck infra to assert that fusion patterns still work

jameszianxuTT:
### Ticket
#1535 

### Problem description

We have no integration tests testing the entire FE -> Compiler path for integrity of the compiled IR. There are cases where the compiled binary produces semantically correct, but suboptimal results.

This regression case can happen if we update a dependency causing the traced graph to be different (eg. transformers uplift). Then, we might see the traced SHLO being different, and subsequent fusion passes failing to execute.

### How to add and check new filechecks 
- For the test(s) you want to add filechecks for, add `filechecks` list value to the config yaml under the test name and specify which filecheck "golden" files to test against. Filecheck goldens live in `tests/filecheck/[name]_[ttir/ttnn].mlir`, for example `tests/filecheck/concatenate_heads.ttnn.mlir`. 
  - For reference of what this looks like, check the config change to `vit/pytorch-base-single_device-full-inference` in this PR
-  If needed, add more filecheck goldens as `tests/filecheck/[name]_[ttir/ttnn].mlir`

### What's changed
- Add tests/filecheck folder to hold common-pattern filechecks. 
- Add infra in test-models.py to search for a list of filechecks inside each test's yaml config
- Add filecheck_utils to execute a filecheck given a source IR directory path string and patterns filename string list [WIP] 
- Create demo for vit/pytorch_base to look for concat heads + split qkv and heads filechecks
  - Done by adding `filechecks` list in test config yaml, pointing to files in tests/filechecks 
- Modify `--serialize` API in test_models and base testers to use a common test name sanitizer in test utils
- Fix test name to not create subdirs during serialization by adding / to the character blacklist.
- Add filecheck infra testcases 
- Improve test config linter to account for new optional filechecks field

### Note

This feature requires a FileCheck binary available in the PATH. This is available if the tt-mlir toolchain has been built, and is currently found in `/opt/ttmlir-toolchain/bin/FileCheck` in IRD. These tests are for internal use only, and I do not expect them to be distributed to customers, so I have not attempted to package the FileCheck binary with tt-xla. 

### Checklist

### DIFF
--- a/tests/infra/utilities/filecheck_utils.py
+++ b/tests/infra/utilities/filecheck_utils.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
 #
 # SPDX-License-Identifier: Apache-2.0
 
@@ -8,6 +8,8 @@ FileCheck utilities for verifying IR patterns in tests.
 
 import subprocess
 from pathlib import Path
+
+from tests.infra.utilities import sanitize_test_name
 
 FILECHECK_PATH = Path("tests/filecheck")
 
@@ -37,11 +39,12 @@ def run_filecheck(
     irs_dir = Path(irs_filepath)
 
     if not irs_dir.exists():
-        raise FileNotFoundError(f"IR directory does not exist: {irs_filepath}")
+        raise FileNotFoundError(
+            f"IR directory does not exist: {irs_filepath}. This is probably because you didn't run the test with pytest --serialize option."
+        )
 
     # Build a mapping of IR types to IR files for the current test
     # We need to match files that correspond to this specific test_node_id
-    from tests.infra.utilities.utils import sanitize_test_name
 
     sanitized_test_id = sanitize_test_name(test_node_name)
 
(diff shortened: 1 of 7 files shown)


jameszianxuTT:
@AleksKnezevic I went with FileCheck for this implementation because that's what we're already using in tt-mlir to match patterns of arbitrary complexity. 

There are a few core problems to address and FileCheck seems like the most elegant solution here:
1. How do we describe the fusion patterns we want to match? (i.e. format of test goldens - I chose MLIR)
2. How do we describe which fusion patterns to match against? (I chose YAML list against common filecheck folder, thanks @kmabeeTT for suggestion)
3. Given an IR and a fusion pattern, how do we determine if the fusion pattern is in the IR? (I chose filecheck, based on decision in (1))

It would be relatively easy to rip out the [core _run_filecheck](https://github.com/tenstorrent/tt-xla/pull/2230/files#diff-5bda927b47396a94901b0a722aaabfb2686d4c307ab06c782dae8cb3c7a45aa0R136-R141) function and replace it with some string matcher, but it becomes much harder to write test goldens for complex patterns.

Alternatively, it is possible to use the tt-mlir dialect python nanobinds to walk through the IR and load it into python datastructures. You can see an [example of that here](https://github.com/tenstorrent/tt-xla/pull/2230/commits/e65d8eed80b3e1636f0a08d9ac7454fd02eb1369#diff-a4f1237da65ee36d22f20619fb2610c22e8b039bfcc5ec257b93d542a52407e6R6), and it works pretty well. However, we would need to come up with some serialized test golden format and write a parser for it (basically, invent our own way to serialize the graph of a test pattern, put that pattern into tests/filechecks/...). It seems like we'd go around existing MLIR tooling in that case to reinvent the wheel for no benefit. 

I am open to other suggestions if using FileCheck seems inappropriate. My main concern w/ filecheck is distribution, but I do not think we will be shipping these tests to end users.

tests/infra/utilities/filecheck_utils.py
```diff
@@ -0,0 +1,171 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
```
jameszianxuTT: TBD - fix - 2025
Also, this is run already in on pr -> basic tests -> meta infra tests.

Issue #2153, opened 2025-11-13
model class loader


AleksKnezevic:
Currently our graph unit tests (attention, mlp, norm) etc load the entire model, with pre-trained weights and then only run a small part of it. This take more time in CI. 

An improvement can be made to expose the model config from tt-forge-models and then just instantiate an i.e. attention module with that config.

Issue #2151, opened 2025-11-13
Remove instruct variants in torch graph tests


gengelageTT:
`_instruct` variants have the same architecture as their base variant so we don't need them tested in unit tests

[see context here](https://github.com/tenstorrent/tt-xla/pull/2124#discussion_r2524077418)

PR #2264, opened 2025-11-20
Torch graph test fixes and improvements

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/2153 https://github.com/tenstorrent/tt-xla/issues/2151

### Problem description
There are 4 issues with the torch graph tests currently:

1. `is_llmbox()` and `@pytest.mark.llmbox` is currently blocking certain tests (attention, mlp) from running on single chip (both locally and in CI).
2. Issue 2151: some instruct variants are tested unnecessarily
3. Issue 2153: we are loading the entire model for each graph test which is time consuming. We can load just the desired module instead.
4. [EDIT] @ssaliceTT : Multi-chip Gemma/Llama/Qwen2.5 attention tests were failing because of mesh configs. Tiny models with awkward head counts (e.g. Gemma 1.1 2B with 1 KV head, Qwen2.5 with 2 KV heads, TinyLlama with 4) couldn’t shard cleanly, so we needed adaptive mesh selection to account for it.

### What's changed
1. Added parametrization `is_llmbox` for `@pytest.mark.llmbox` (True) and `@pytest.mark.single_chip`(False) so that tests can be run on one device or in tensor parallel. I assumed that we want to be running both versions in CI, but if that’s not the case this can be changed.
2. Removed instruct variants when the equivalent base variant is already tested.
3. Now calling `loader.load_config(variant=variant)` instead of `loader.load_model(variant=variant)` and then instantiating the module (ex. LlamaAttention, LlamaMLP, etc.) with the config. This shows the following improvements in test durations locally:
    - tests/torch/single_chip/graphs/test_rms_norm.py : `6 passed in 278.47s (0:04:38)` → `12.77s`
    - tests/torch/single_chip/graphs/test_rotary_emb.py: `22 passed in 327.68s (0:05:27)` → `32.29s`
4. [EDIT] @ssaliceTT : 
- Made the Llama and Gemma tests pick 1×8 only when both attention+KV head counts divide 8, otherwise fall back to 2×4 (and skip if neither works).
- Gemma attention tests leaves KV replicated when there’s just one head.
- Qwen2.5 tests now skip the 0.5B/1.5B/3B variants as they have only 2 KV heads which doesnt shard cleanly on 1x8 or 2x4 mesh.

### Checklist

### DIFF
--- a/tests/torch/single_chip/graphs/test_attention.py
+++ b/tests/torch/single_chip/graphs/test_attention.py
@@ -1539,6 +1539,16 @@ def test_gemma_attention_prefill(seq_len, variant, variant_config, is_llmbox):
                 shard_specs[attention.o_proj.weight] = (None, "model")
                 return shard_specs
 
+        elif num_heads % 8 == 0 and num_key_value_heads == 1:
+            # for small variants like gemma 1.1 2b with 8 attn heads and 1 kv head
+            mesh_shape = (1, num_devices)
+
+            def get_shard_spec(attention, args, kwargs):
+                shard_specs = {}
+                shard_specs[attention.q_proj.weight] = ("model", None)
+                shard_specs[attention.o_proj.weight] = (None, "model")
+                return shard_specs
+
         else:
             batch_size = 2
             mesh_shape = (2, num_devices // 2)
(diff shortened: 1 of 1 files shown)


gengelageTT:
All torch graph tests are passing in CI with these changes here: https://github.com/tenstorrent/tt-xla/actions/runs/19552238101

PR #2284, opened 2025-11-24
[vLLM plugin] Add test for Qwen3-Embedding-0.6B model

### LINKED ISSUE #2283: [vLLM plugin] Add test for qwen3-embedding-0.6B model

mmanzoorTT:
### Ticket
closes #2283 

### What's changed
Add test for Qwen3-Embedding-0.6B model

### Checklist


AleksKnezevic:
@kmabeeTT, these aren't model weights, they're golden tensors. We can have them in S3, but they're only ~20kB, so not strictly needed.

kmabeeTT:
Ah thanks yeah I see that now that the 6 files are all small.  Talked to Asif offline too and there was interest here, so gave some details how to do it, and requested his username to have access, and in the meantime I copied the 6 .pt files to this folder `test_files/vllm/pooling/baseline/` inside s3 bucket, which can be accessed like `file = get_file("test_files/vllm/pooling/baseline/qwen3_embedding_0.6B_baseline.pt")` after setting `IRD_LF_CACHE` locally (CI does it too) to local mirror/cache of the s3 bucket (path for this is kept private, shared offline).

AleksKnezevic:
yeah, they can def be in s3, just shouldn't be blocking.

PR #2303, opened 2025-11-24
Add automated performance bisect tools for tt-xla and tt-mlir

rpavlovicTT:
This commit introduces three new scripts for automating performance regression investigation through git bisect:

1. **bisect_perf.sh** - Standalone bisect script for tt-xla commits
   - Tests individual tt-xla commits for performance regressions
   - Supports custom benchmark commands, thresholds, and metric patterns
   - Handles build failures and benchmark crashes gracefully (exit 125)
   - Can be used directly with: git bisect run ./scripts/bisect_perf.sh

2. **bisect_ttmlir_perf.sh** - Standalone bisect script for tt-mlir commits
   - Tests individual tt-mlir commits within the tt-mlir submodule
   - Dynamically modifies CMakeLists.txt to test specific tt-mlir versions
   - Supports incremental builds for faster iteration
   - Logs all tests to /tmp/bisect_ttmlir_*.log for later analysis

3. **bisect_perf_auto.sh** - Master orchestrator (RECOMMENDED)
   - Fully automated two-phase bisect: tt-xla first, then tt-mlir if needed
   - Automatically detects tt-mlir uplift commits
   - Drills down into tt-mlir submodule to find exact regression commit
   - Comprehensive logging and final summary report
   - Single command operation: ./scripts/bisect_perf_auto.sh -g <good> -b <bad>

4. **cmake_fix.patch** - CMakeLists.txt patch for incremental builds
   - Removes --target tt-alchemist to enable faster incremental builds
   - Applied automatically by bisect scripts

All scripts support:
- Custom benchmark commands (-c, --command)
- Custom performance thresholds (-t, --threshold)
- Custom metric extraction patterns (-p, --pattern)
- Comprehensive help documentation (-h, --help)
- No hardcoded paths - works from any tt-xla location

Default configuration:
- Benchmark: resnet model with batch size 8, bfloat16, 128 loops
- Threshold: 680 samples/second
- Pattern: "Sample per second:\s*\K[0-9.]+"

Example usage:
  ./scripts/bisect_perf_auto.sh -g 051ebb20 -b HEAD

Successfully tested and used to identify tt-mlir commit a868fa2a9 ([TTIRFusing] Improve ScaledSumToMeanPattern) as the root cause of a performance regression in resnet (from ~700 to ~574 samples/sec).

### DIFF
--- a/scripts/bisect_perf.sh
+++ b/scripts/bisect_perf.sh
@@ -11,6 +11,7 @@
 #   -c, --command COMMAND       Benchmark command to run (required if no defaults)
 #   -t, --threshold THRESHOLD   Performance threshold value (required if no defaults)
 #   -p, --pattern PATTERN       Grep pattern to extract metric (default: "Sample per second:\s*\K[0-9.]+")
+#   -r, --revert COMMIT         Revert this tt-mlir commit if present in history (useful for isolating additional regressions)
 #   -h, --help                  Show this help message
 #
 # EXAMPLES:
@@ -31,6 +32,9 @@
 #     -t 500 \
 #     -p "Sample per second:\s*\K[0-9.]+"
 #
+#   # Revert a known bad tt-mlir commit to isolate additional regressions
+#   git bisect run ./scripts/bisect_perf.sh -r 70efb12f5f75c7b1642542e7c3ce330c472ea038
+#
 # EXIT CODES:
 #   0   - Good commit (performance >= threshold)
 #   1   - Bad commit (performance < threshold)
@@ -45,6 +49,7 @@ DEFAULT_PATTERN="Sample per second:\s*\K[0-9.]+"
 BENCHMARK_COMMAND=""
 PERF_THRESHOLD=""
 METRIC_PATTERN=""
+REVERT_COMMIT=""
 
 show_help() {
     sed -n '2,/^$/p' "$0" | grep '^#' | sed 's/^# \?//'
@@ -63,6 +68,10 @@ while [[ $# -gt 0 ]]; do
             ;;
         -p|--pattern)
             METRIC_PATTERN="$2"
+            shift 2
+            ;;
+        -r|--revert)
+            REVERT_COMMIT="$2"
             shift 2
             ;;
         -h|--help)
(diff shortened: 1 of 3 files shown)


scripts/cmake_fix.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/third_party/CMakeLists.txt b/third_party/CMakeLists.txt
+index 24d8f161..d7c72fac 100644
+--- a/third_party/CMakeLists.txt
++++ b/third_party/CMakeLists.txt
+@@ -85,7 +85,7 @@ else()
+         BINARY_DIR ${TTMLIR_BUILD_DIR}
+ 
+         BUILD_COMMAND env ${WITH_METAL_RUNTIME_ROOT_SET} ${CMAKE_COMMAND} --build <BINARY_DIR>
+-        COMMAND ${CMAKE_COMMAND} --build <BINARY_DIR> --target tt-alchemist
```
odjuricicTT: @svuckovicTT Can this patch just be part of the changes?

svuckovicTT: Synced offline, it's okay, will remove soon as there will be no need for it.

AleksKnezevic:
@jameszianxuTT and @brataTT, FYI since you already had some bisect scripts

rpavlovicTT:
> Great stuff. Learned some things from it.
> 
> Wasn't aware we could use `git bisect view` that way. In uplift scripts, we use it to get a status/summary of long running bisects instead. Separating logs by commits instead of commands is also a good idea. I'll check if I can steal these ideas to improve metal uplift scripts too.
> 
> Concerns
> 
> * I don't see any clean-build step.
>   That'll make running this really fast 👍 .
>   But if there are unreliable results at any point, it may be useful to add that step.
>   With that said, I have no specific examples of when this can happen or if this is a valid concern.
> * Patches: If mlir introduces breaking changes that required changes in tt-xla uplift PR, there will be a larger number of untestable changes.
>   In metal uplift, we solve this by keeping separate downstream fix patches per breaking upstream commit.
>   Then, we apply those patches starting from the breaking change onwards.
>   The unrolling @jameszianxuTT mentioned simplifies this.

About your concerns -- 
1. I intentionally let it build incrementally just to keep progress quicker. Ideally, we should fall back to clean build if build fails. 
2. That is a great pointer. This seems to me as very difficult problem. As I understand it, we could have a tt-mlir uplift commit that contains some tt-xla code changes, and we don't know which tt-mlir commits contained in the uplift work with these changes and which don't. 
I'm curious about `separate downstream fix patches per breaking upstream commit.` -- how do you create these patches?

Anyway as I said, this is a demo version I made quickly with Claude just to help us out.

scripts/bisect_perf.sh
```diff
@@ -0,0 +1,151 @@
...
+fi
+
+# Update submodules to match this commit's version
+echo "Updating submodules..."
+git submodule update --init --recursive
+
+# Build with CMake
+echo "Building project..."
+cmake -G Ninja -B build -DCMAKE_BUILD_TYPE=Release
+if ! cmake --build build; then
```
odjuricicTT: If torch-xla version has changes in https://github.com/tenstorrent/tt-xla/blob/main/python_package/requirements.txt we should probably install the new version.

PR #2306, opened 2025-11-24
Adding experimental_bfp8_weights compile option to compiler config

### LINKED ISSUE #2305: Add experimental_bfp8_weights compile option to compiler config
After [PR](https://github.com/tenstorrent/tt-mlir/issues/5882) is consumed in tt-xla, we should add experimental_bfp8_weights flag to compiler config and lower it to compile options in pjrt.

dgolubovicTT:
### Ticket
Closes #2305 

### Problem description
Experimental weights on bfp8 support has been added to tt-mlir in [PR](https://github.com/tenstorrent/tt-mlir/issues/5882).
After this is consumed in tt-xla, we should add the flag experimental_bfp8_weights to compiler config

### What's changed
- Adding experimental_bfp8_weights compile option to compiler config
- Adding matmul and linear single op tests with experimental_bfp8_weights flag on and off.

### Checklist
Added single op tests for torch matmul and linear operators with second operand as parameter, and `experimental_bfp8_weights` true.


pjrt_implementation/inc/api/compile_options.h
```diff
@@ -42,6 +42,15 @@ struct CompileOptions {
   // and user should provide and get tensors of common dtype.
   bool enable_bfp8_conversion = false;
 
+  // Enables experimental BFP8 weight conversion in MLIR.
+  bool experimental_bfp8_weights = false;
```
AleksKnezevic: Can we have a better name than experimental? Open to whatever, but perhaps `enable_amp`,`enable_bfp8_mm_weights`...

dgolubovicTT: I am not hardly opinionated on the name, especially since we have enable_bfp8_conversion and it is also "experimental" in a sence that it is not on by default...

nvukobratTT: I don't have hard preferences for the naming; we can go with `enable_weight_bfp8_conversion ` as outlined. In the future, the flag that will encapsulate this feature should be named closely to "automatic mixed precision", without the experimental prefix. All sub-features around it (e.g., matmul weights on bfp8_b, conv2d weights on bfp8_b, etc.) that will be nice to have for experimenting and debugging purposes should go under "experimental_" to explicitly outline that these are not features meant to be used commonly (if ever). 

In sum, we can do the rename here, but let's think about the rest of the sub-features we'll add and how overall mixed-precision will be controlled (e.g., different optimization levels, similar to the optimizer case).

nvukobratTT (approved):

ajakovljevicTT (approved):

ajakovljevicTT (approved):

pjrt_implementation/inc/api/compile_options.h
odjuricicTT: Since Darko is OOO, i renamed to flag to `experimental_enable_weight_bfp8_conversion`. That should keep it consistent with the other one, explain what it is doing better and have the experimental warning.

If there are no objections i'll merge this tomorrow morning.

PR #2307, opened 2025-11-24
Add Xfail Preset to Weekly Workflow

mmilosevicTT:
### Ticket
N/A

### Problem description
Currently, weekly only runs passing tests. Having this in mind, training tests will not be picked up by the current desing.

### What's changed
Adding xfail preset to the weekly job so that we actually follow the nightly structure of running tests.

### Checklist

### DIFF
--- a/.github/workflows/schedule-weekly.yml
+++ b/.github/workflows/schedule-weekly.yml
@@ -27,8 +27,6 @@ jobs:
     uses: ./.github/workflows/call-test.yml
     secrets: inherit
     needs: [ build-image, build-ttxla ]
-    # This ensures the job runs regardless of success or failure of `weekly_tests`:
-    if: success() || failure()
     with:
       test_suite: model-test-passing-weekly.json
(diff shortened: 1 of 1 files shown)


.github/workflows/schedule-weekly.yml
```diff
@@ -35,10 +35,23 @@ jobs:
       artifact_release_run_id: ${{ needs.build-ttxla.outputs.artifacts_run_id }}
       wheel_release_artifact_name: ${{ needs.build-ttxla.outputs.wheel_artifact_name }}
 
+  test_forge_models_xfail:
+    uses: ./.github/workflows/call-test.yml
+    secrets: inherit
+    needs: [ build-image, build-ttxla ]
+    # This ensures the job runs regardless of success or failure of `weekly_tests`:
+    if: success() || failure()
```
vvukomanTT: There is no "weekly_tests" job/step, so these lines serve no puropse.
I know that you didn't introduce this concept here in the first place, but please remove the same two lines from `test_forge_models_passing` above.

PR #2320, opened 2025-11-25
[PJRT] Initial changes for PJRT unit/mock tests

acicovicTT:
### Ticket

N/A

### Problem description

PJRT API needs a framework for executing tests in isolation, in addition to existing e2e integration workflows (Python tests). The idea is to use this framework to create unit and mock tests which will put PJRT code in the focus of testing. This is the initial PR which introduces the necessary changes for building and running tests, and some example unit tests for PJRT utils. The problem of mocking PJRT dependencies (tt-mlir) will be solved in a future PR (this will also make it easier to review this one).

### What's changed

Read full description of new changes in the new `tests/pjrt/README.md`.

### Best way to review

0. New README.md
1. Top-level CMakeLists.txt, then new leaf CMakeLists.txt
2. Tests

### What is **not** included in this PR

- Meaningful unit / mock tests, apart from example unit tests that validate PJRT helpers, and can be used as an example of how to add new tests.
- Changes that integrate new tests with CI. This will be done in a different PR, after more tests have been added.

### Checklist

### DIFF
--- a/tests/pjrt/test_utils.cc
+++ b/tests/pjrt/test_utils.cc
@@ -39,9 +39,7 @@ protected:
     std::filesystem::remove(TMP_LOG_FILE_NAME);
   }
 
-  void TearDown() override {
-    std::filesystem::remove(TMP_LOG_FILE_NAME);
-  }
+  void TearDown() override { std::filesystem::remove(TMP_LOG_FILE_NAME); }
 };
 
 const char *PJRTUtilLoggingTests::TMP_LOG_FILE_NAME = "tmp_test.log";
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -14,6 +14,9 @@ third_party/tt-mlir
 .cache/
 compile_commands.json
 
+# CTest
+Testing/
```
mrakitaTT: Maybe we should move this out of the root (if the path can be customized), for example into "build" or "tests/pjrt" folders, so we don't have "tests" and "Testing" folders in the root.

tests/pjrt/test_utils.cc
```diff
@@ -0,0 +1,123 @@
...
+// This file incorporates work covered by the following copyright and permission
+// notice:
+// SPDX-FileCopyrightText: Copyright 2023 The IREE Authors
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// https://llvm.org/LICENSE.txt
+
+// GTest includes
+#include <gtest/gtest.h>
+
+// PJRT implementation includes
```
mrakitaTT: Nit: in tt-xla we follow Google's order of includes: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

sdjukicTT (approved):

.gitignore
acicovicTT: This is already the case, it was just that once I accidentally configured root for build and I saw that it created this Testing/ folder. So, this folder will not be created in repo root, but I just added it just in case someone in the future makes the same mistake as me.

tests/pjrt/test_utils.cc
```diff
@@ -0,0 +1,123 @@
...
+// C++ standard library includes
+#include <cstdlib>
+#include <filesystem>
+#include <vector>
+
+namespace tt::pjrt::tests {
+
+// ---------- Fixture classes --------------------------------------------------
+
+class PJRTUtilLoggingTests : public ::testing::Test {
```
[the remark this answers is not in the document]
acicovicTT: Sure, will do that. It will be used only in this file though

PR #2324, opened 2025-11-25
Fix deadlock in concurrent copyToHost, add testcases

jameszianxuTT:
### Ticket
Fixes a deadlock spotted by @pilkicTT in #2125 

### Problem description
There is an obvious deadlock in the implementation of #2125 after changing to unique_lock, and the exact mechanism is discussed in the comments of #2125. 

### What's changed
- Use separate locks for:
  - (Instance lock) Preventing multiple framework threads from concurrently joining on the same copyToHost thread for the **same** BufferInstance, which is C++ UB
  - (Static lock) Preventing multiple framework threads from concurrently running runtime toHost() between **different** BufferInstances, due to metal's lack of op-dispatch thread safety. This is locked from within the async copyToHost thread
- Reset BufferInstance m_copy_to_host_thread after joining. This is a separate bug, where a later framework thread could join on an already-destructed thread
- Add testcases for (1) concurrent copyToHost calls on the same bufferInstance and (2) concurrent copyToHost calls on multiple bufferInstances

### Checklist

### DIFF
--- a/pjrt_implementation/src/api/buffer_instance.cc
+++ b/pjrt_implementation/src/api/buffer_instance.cc
@@ -151,6 +151,7 @@ void BufferInstance::deleteData() {
   std::unique_lock<std::mutex> copy_lock(m_copy_to_host_thread_mutex);
   if (m_copy_to_host_thread) {
     m_copy_to_host_thread->join();
+    m_copy_to_host_thread.reset();
   }
   copy_lock.unlock();
 


jameszianxuTT:
I have checked that the new testcases actually do cause the framework to initiate concurrent BufferInstance copyToHost calls. These testcases would have caught the deadlock you observed @pilkicTT and fail in main right now.

I am not sure if there is a legitimate usecase to trigger concurrent copyToHost on a single BufferInstance, but it does not seem like the framework prevents it, so this is a more graceful handling than a deadlock. This deadlock is not encountered in any CI tests though.

pjrt_implementation/src/api/buffer_instance.cc
```diff
@@ -148,7 +148,7 @@ void BufferInstance::deleteData() {
   }
 
   // Wait if there is a copy to host in progress.
-  std::unique_lock<std::mutex> copy_lock(s_copy_to_host_thread_mutex);
+  std::unique_lock<std::mutex> copy_lock(m_copy_to_host_thread_mutex);
   if (m_copy_to_host_thread) {
     m_copy_to_host_thread->join();
```
pilkicTT: you can reset the thread pointer here as well

pjrt_implementation/src/api/buffer_instance.cc
```diff
@@ -148,7 +148,7 @@ void BufferInstance::deleteData() {
   }
 
   // Wait if there is a copy to host in progress.
-  std::unique_lock<std::mutex> copy_lock(s_copy_to_host_thread_mutex);
```
pilkicTT: why not lock_guard and keep it for the whole duration of the function?

pilkicTT (approved):
thanks!

Issue #2309, opened 2025-11-25
gemma model PCC fails and PCC higher than 1.0


kmabeeTT:
Nothing new, but I am using 2f603024 from Nov 24.

Some variants show PCC higher than 1.0 Example (when adding comparator.py debug print)

```
tests/runner/test_models.py::test_all_models_torch[gemma/pytorch-google/gemma-1.1-2b-it-data_parallel-full-inference]
PCC result: 1.0200762748718262, threshold: 0.99
```

Some variants show low PCC (this is run on blackhole p150) since it is a RED model variant. It does not fit on n150, but maybe if number of layers is reduced slightly it could fit for debug purposes.

```
pytest -svv tests/runner/test_models.py::test_all_models_torch[gemma/pytorch-google/gemma-1.1-7b-it-data_parallel-full-inference] --arch p150

FAILED tests/runner/test_models.py::test_all_models_torch[gemma/pytorch-google/gemma-1.1-7b-it-data_parallel-full-inference]
AssertionError: Comparison result 0 failed: PCC comparison failed. Calculated: pcc=0.9412137269973755. Required: pcc=0.99.
```

PR #2333, opened 2025-11-26
Perform accuracy comparisons in Float64

sgligorijevicTT:
### Ticket
Related to #2309

### Problem description
Some models were reporting PCC>1.0 on the dashboard.

### What's changed
All accuracy calculations are now done in float64 to minimize numerical stability issues.
Allclose checking is now done per pair of tensors rather than globally, to avoid cases where in a big collection some pairs fail causing pcc to be ran for all pairs and resulting in nans.

### Checklist

### DIFF
--- a/tests/runner/test_config/torch/test_config_inference_single_device.yaml
+++ b/tests/runner/test_config/torch/test_config_inference_single_device.yaml
@@ -79,6 +79,7 @@ test_config:
     status: EXPECTED_PASSING
 
   xglm/pytorch-xglm-1.7B-single_device-full-inference:
+    required_pcc: 0.98 # PCC was >0.99 while it was calculated in f32. ATOL is outrageously bad so a drop is not unexpected.
     status: EXPECTED_PASSING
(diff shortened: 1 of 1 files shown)


sgligorijevicTT:
Heads up, I am piggybacking another change onto this PR. @mmilosevicTT ran into a bug when comparing gradients, namely we checked if all tensors in the pytree were allclose as a requirement for skipping pcc, and that's a problem as it can happen that some tensors fail allclose but others don't and cause some pccs to be nan. I pushed the allclose check into the per tensor lambda to rectify this.

sgligorijevicTT:
Heads up, running weekly(abeit on an unrebased version from a good while ago) on this uncovered 1 new regression below the threshold, xglm. I skipped it in this PR alongside codegemma that was known from before.
If anything extra falls, Andrej and I will deal with it. Either way, I think we would rather have a more accurate picture of accuracy than fudging the math to make more models 'pass'.

PR #2359, opened 2025-11-27
[PJRT] Modular way to build tests, and new unit tests

acicovicTT:
This PR introduces a modular way of organizing tests, and adds test coverage for some of the core components.

### CMake Build System Changes

The build system has been improved to support modular, per-target test organization:

- Each library target (TTPJRTApi, TTPJRTUtils) now declares its `UNIT_TESTS_DIR` property.
  - Does not impose any kind of format for test directory names or file names.
- Enables test sources to live alongside implementation code.
- Enables one test target to collect and build all the other test targets.

### Unit Tests

- Old unit tests have been refactored to address some comments from the previous PR.
- New unit tests have been added for some of the core components (`ErrorInstance`, `EventInstance`, ...).

### Checklist

### DIFF
--- a/pjrt_implementation/src/api/unit_tests/inc/unit_test_utils.h
+++ b/pjrt_implementation/src/api/unit_tests/inc/unit_test_utils.h
@@ -41,26 +41,26 @@ class PJRTComponentUnitTests : public ::testing::Test {
 protected:
   // Runs before every test.
   void SetUp() override {
-    device_ = DeviceInstance::createInstance(
+    m_device = DeviceInstance::createInstance(
         TT_DEVICE_HOST,
         /*is_addressable=*/true, LOCAL_HARDWARE_ID_UNDEFINED, TT_ARCH_WH);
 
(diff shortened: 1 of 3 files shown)


pjrt_implementation/src/api/unit_tests/inc/unit_test_utils.h
```diff
@@ -0,0 +1,68 @@
...
+    device_as_vector_.push_back(device_.get());
+
+    // DEVNOTE: Intentionally not set to device during setup.
+    default_memory_ =
+        MemoryInstance::createInstance(device_as_vector_, DEFAULT_MEMORY_ID,
+                                       /*is_host_memory=*/true);
+  }
+
+  // Device that is unique per-test.
+  std::unique_ptr<DeviceInstance> device_;
```
ajakovljevicTT: SHould we follow naming conventions such as in other parts of the repo? So this would be `m_device`, etc.

Issue #1051, opened 2025-08-20
☔ Perf benchmarking requirements


odjuricicTT:
Currently generality models tests and performance benchmarking model tests are completely separate. This is not ideal since there is a lot of common infra that can be shared. Parts of perf benchmarks should be integrated into the generality testing infra and potentially enabled for all models.

Here is a high level summary of perf benchmarking requirements:
1. Measure e2e perf
A few different metrics for different models architectures e.g. fps, tps, ttft etc...
Example:
    ```python
      # Warm up
      for _ in range(3):
        compiled_model(inputs[i])
      
      # Measure
      start = time.time()
      for i in range(loop_count):
          output = compiled_model(inputs[i])
      end = time.time()
    ```
2. Measure device perf
Requires build with tracy enabled (different from release build used for e2e perf).
In an ideal world this can look like:
    ```python
      # Warm up
      for _ in range(3):
        compiled_model(inputs[i])
      
      tracy.start()
      output = compiled_model(inputs[i])
      tracy.stop()
    ```
4. Accuracy / eval score on real dataset
5. Measure cpu perf
6. Be able to save all measurements to superset
7. Save multiple files to artifacts
    * ttir
    * ttnn ir
    * flatbuffer
    * device perf report

vcanicTT:
There is [Pydantic Model](https://github.com/tenstorrent/tt-github-actions/blob/main/.github/actions/collect_data/src/pydantic_models.py), for performance we class BenchmarkRun and class BenchmarkMeasurement. These classes contains all data we send to the database as a JSON.

[Here](https://github.com/tenstorrent/tt-forge/blob/main/benchmark/tt-forge-fe/resnet_hf.py#L217) is an example, how we form JSON with results.

kmabeeTT:
Thx for details! I like the idea of unifying infra, and being able to generate benchmarks from within tt-xla itself especially if it will reduce code duplication.  That being said, am not familiar at all with the perf benchmarking aside from what's described here, or really what this will look like when integrated into tt-xla model testing infra.  At the moment the test_models.py stuff I added is very light weight and simple, so slight worry is it getting complicated.

Can you share some more details on what the main code/infra that is shared today that we would be eliminating?

odjuricicTT:
[...]

PR #2378, opened 2025-11-29
e2e perf benchmarking

ssaliceTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/1051

### Problem description
Generality models tests and performance benchmarking model tests are completely separate right now. This is not ideal since there is common infra that can be shared. Parts of perf benchmarks should be integrated into the tt-xla generality testing infra.

#### Model tester:
- Added perf measurement loop in model_tester.py to collect e2e timing (warmup + timed iterations)
- Perf measurements run by default; can be disabled with --disable-perf-measurement.
#### Benchmark output:
- Created helper functions `create_benchmark_result` and `create_measurement` to format perf measurements into the pydantic models data format required by the data collection pipeline.
- The two helper functions are intentionally kept as simple as possible to start and can be expanded as needed.
- Benchmark results print to console
- JSON report files generated when --perf-report-dir=<path> is provided (no files created locally unless explicitly requested)
#### CI:
- Added --perf-report-dir pytest argument, passed in call-test.yml
- Added --perf-id pytest argument, passed in call-test.yml
- Added artifact upload step for benchmark reports (perf-reports-*)

### Samples
See below for sample runs and collect-data actions done through these changes and the superset dashboard created with it:
perf-test run: https://github.com/tenstorrent/tt-xla/actions/runs/20186676550
collect-data run: https://github.com/tenstorrent/tt-xla/actions/runs/20186907215/job/57958566636
Superset dashboard: https://superset.tenstorrent.com/superset/dashboard/397/?native_filters_key=waW5_L6Gqe4

### Checklist

### DIFF
--- a/.github/workflows/call-test.yml
+++ b/.github/workflows/call-test.yml
@@ -329,7 +329,7 @@ jobs:
         fi
         python -m pytest ${{ env.VERBOSITY }} ${{ env.BASE_PYTEST_CMD }} \
                --junitxml=${{ steps.strings.outputs.test_report_path }} \
-               --perf-report-dir=${{ steps.strings.outputs.per_report_dir }} \
+               --perf-report-dir=${{ steps.strings.outputs.perf_report_dir }} \
                --perf-id=${{ steps.fetch-job-id.outputs.job_id }} \
                2>&1 | tee pytest.log
 


jameszianxuTT (commented):
Overall, looks good, but I have a few questions before I approve. 

Also, let's rebase to see if the multihost llama fails again. It's passing in onPush so I really have no idea how it's failing here. Maybe you picked up a changed dependency or something, but I don't think it's due to anything you did in this PR.

Nice work!

ssaliceTT:
I'm having troubling figuring out what to do for enabling/disabling perf benchmarking (if needed)

For enabling/disabling perf collection, a few options:
- Pytest marker (@pytest.mark.perf) to tag to tests as James suggested above.
- CLI flag (--collect-perf) – enable perf collection for any test in a run. 
- New workflow for just perf-tests where we only run some of them (probably not the best as we'd be making jobs for testing perf which defeats the purpose of putting the loop in the model_tester.)

Any thoughts on what is needed? @AleksKnezevic @jameszianxuTT

jameszianxuTT:
@ssaliceTT I think we'll end up removing the [perf benchmark workflow in tt-forge](https://github.com/tenstorrent/tt-forge/blob/main/.github/workflows/perf-benchmark.yml) and port it to tt-xla. 

I think what this will eventually look like is an additional testgroup or even a different test config in nightly tests JSON for running perf, that collects all tests marked with @perf (we might need to change test_models.py test configs to add those markers) and executes them.

For the purposes of this change though, I suggest you make a new testgroup to run in basic-tests.json as a PoC, and measure perf on a quick handful of models as a test of the perf infra. It seems like you can add `markers: ["push", "weekly", ...]`  to some basic test in tests/runner/test_config/torch/test_config_inference_single_device.yaml and then filter by that mark to select which tests to run in that job.

Finally, to actually trigger perf run you probably do want a CLI flag which is read inside test_models.py to switch between the normal testrunner and the perf runner. One hacky way to set the flag would be in `.github/workflows/call-test.yml` to read the `MARKS` in  `Set pytest command` and if `perf` is found, to forward the --run-with-perf flag (or whatever you want to call it)

Ideally though, we should have some infra for passing through arbitrary pytest args in the JSON configs, I'll make an issue for devex for that.

Also cc @odjuricicTT, this change brings in the infra to run perf tests inside tt-xla and upload them to superset.

jameszianxuTT:
WRT the fatal error in the multihost tests, I can bring up that API next. It would suggest that this change somehow introduced a graph break or something similar due to the multiple invocations of each pytest.

As a temporary solution, please modify https://github.com/tenstorrent/tt-xla/blob/8bd4a56d392ca507ebd27a235c05049374176097/tests/torch/multi_host/llmbox/test_multihost.py#L54 to add your disable perf CLI option so it doesn't run. I will patch the distributed runtime and enable perf on multihost once I can get that working, but I don't want to block this PR.

tests/runner/conftest.py
```diff
@@ -34,6 +34,12 @@ def pytest_addoption(parser):
         default=False,
         help="Fail if test_config files and collected test IDs are out of sync",
     )
+    parser.addoption(
+        "--disable-perf",
```
AleksKnezevic: similar here, `disable-perf-measurement`

jameszianxuTT (approved):
Would prefer to avoid creating an env variable for the sake of passing a pytest config option to test infra + suggested a fix for the multihost test failure.

Other than that, looks good!

.github/workflows/call-test.yml
```diff
@@ -323,6 +323,7 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
         HF_TOKEN: ${{ secrets.HF_TOKEN }}
         ENABLE_BRINGUP_STAGE_LOGGING: 1
+        JOB_ID: ${{ steps.fetch-job-id.outputs.job_id }}
```
nsumrakTT: From my point of view it would be much better to pass this as pytest parametar along with directory and name.
So, instead of having --disable-perf, you can have for example: 
`--enable-perf=test_reports_benchmarks --perf-id= ${{ steps.fetch-job-id.outputs.job_id }} `
The reason is local usage where you would have files generated by default in preset dir that is not in .gitignore which creates dev frustration. 
Another thing that you lack is documentation. In a few weeks, someone will ask around what is this JOB_ID or why perf files has trailing _0000... Not to mention our clients that could run and ask questions on emitted files...

ssaliceTT: Agreed that makes sense. I made the following changes based on feedback:
- Removed JOB_ID env var so it's not used anywhere anymore.
- Added --perf-report-path=<path>. If provided, generates the benchmark JSON file at the specified path. If not provided, no files are generated.
- --disable-perf-measurement disables the perf measurement loop in the tester entirely, so no measurements are collected

How it works now:
- Perf measurement runs by default (unless --disable-perf-measurement is passed)
- `create_benchmark_results` prints perf measurements to console when measurements are collected
- However, report files are only written when --perf-report-path is specified. Files are generated in the format report_perf_*.json. This would be used in CI.

I'm hoping this should cover all the concerns.

.github/workflows/call-test.yml
```diff
@@ -359,6 +361,29 @@ jobs:
...
+      if: success() || failure()
+      with:
+        report_paths: ${{ steps.strings.outputs.test_report_path }}
+        check_name: TT-XLA Tests
+        comment: true
+        updateComment: true
+        detailed_summary: true
+        group_suite: true
+        token: ${{ github.token }}
+
```
nsumrakTT: Please remove this. Test reports are now generated per group not per job.

.github/workflows/call-test.yml
```diff
@@ -129,6 +129,7 @@ jobs:
         echo "work-dir=$(pwd)" >> "$GITHUB_OUTPUT"
         echo "build-output-dir=$(pwd)/build" >> "$GITHUB_OUTPUT"
         echo "test_report_path=report_$JOB_ID.xml" >> "$GITHUB_OUTPUT"
+        echo "perf_report_path=report_perf_$JOB_ID.json" >> "$GITHUB_OUTPUT"
```
nsumrakTT: I don't understand this. perf_report_path is not path but file name.
In test_utils.py you open file with "w" which means it will overwrite existing file. If you run multiple tests, would it create multiple benchmark reports, or it will create single file with all reports.
Or it will generate benchmark json only the last test?

Issue #2383, opened 2025-12-01
Codegen erroring out


svuckovicTT:
Error msg:
```
2025-12-01 12:04:00.175 (   1.055s) [        F6C0F000]loaded_executable_insta:276      1| LoadedExecutableInstance::PJRT_LoadedExecutable_GetExecutable
2025-12-01 12:04:00.175 (   1.055s) [        F6C0F000]loaded_executable_insta:295      1| LoadedExecutableInstance::PJRT_LoadedExecutable_AddressableDevices
2025-12-01 12:04:00.175 (   1.055s) [        F6C0F000] executable_instance.cc:108      1| ExecutableInstance::PJRT_Executable_OptimizedProgram
2025-12-01 12:04:00.175 (   1.055s) [        F6C0F000] executable_instance.cc:108      1| ExecutableInstance::PJRT_Executable_OptimizedProgram
2025-12-01 12:04:00.177 (   1.056s) [        F6C0F000] executable_instance.cc:218      1| ExecutableImage::PJRT_Executable_OutputMemoryKinds
2025-12-01 12:04:00.177 (   1.056s) [        F6C0F000] executable_instance.cc:82       1| ExecutableInstance::PJRT_Executable_NumReplicas
2025-12-01 12:04:00.177 (   1.056s) [        F6C0F000] executable_instance.cc:166      1| ExecutableInstance::PJRT_Executable_Fingerprint
2025-12-01 12:04:00.177 (   1.057s) [        F6C0F000] executable_instance.cc:218      1| ExecutableImage::PJRT_Executable_OutputMemoryKinds
2025-12-01 12:04:00.177 (   1.057s) [        F6C0F000] executable_instance.cc:108      1| ExecutableInstance::PJRT_Executable_OptimizedProgram
2025-12-01 12:04:00.177 (   1.057s) [        F6C0F000] executable_instance.cc:108      1| ExecutableInstance::PJRT_Executable_OptimizedProgram
2025-12-01 12:04:00.179 (   1.058s) [        F6C0F000] executable_instance.cc:108      1| ExecutableInstance::PJRT_Executable_OptimizedProgram
2025-12-01 12:04:00.179 (   1.058s) [        F6C0F000] executable_instance.cc:108      1| ExecutableInstance::PJRT_Executable_OptimizedProgram
[...]

PR #2389, opened 2025-12-01
set tensor retain on graph outputs-aliased-to-inputs for SOLoadedExecutableInstance

jameszianxuTT:
### Ticket
Partial resolution to #2383

### Problem description
SOLoadedExecutableInstance::prepareInputTensor was missing the output-alias-to-input tensor retention setting leading to error:
> Prepared input tensor should have retain=true or it may be deallocated during execution.

Under specific runtime-stitching scenarios. This failure mode was originally encountered during bringup of FlatbufferLoadedExecutableInstance and patched for that class, but not for the SO* class, and the above error log was added as an additional safety check.

### What's changed
Port patch prepareInputTensor from FlatbufferLoadedExecutableInstance to SOLoadedExecutableInstance. This ensures that whichever codepath we source input tensors from - either from host or from outputs of previous execution - the tensor is retained and not accidentally deallocated during execution.

### Checklist

### DIFF
--- a/pjrt_implementation/src/api/so_loaded_executable_instance.cc
+++ b/pjrt_implementation/src/api/so_loaded_executable_instance.cc
@@ -205,8 +205,21 @@ SOLoadedExecutableInstance::prepareInputTensor(
   // As a hack, we'll always use whatever we have prepared.
   if (prepared_tensor.has_value()) {
     DLOG_F(LOG_DEBUG,
-           "Reusing already prepared input tensor for argument index %zu",
-           arg_index);
+           "Reusing already prepared input tensor for argument index %zu with "
+           "shape %s and buffer UID %zu",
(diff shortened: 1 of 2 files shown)


pjrt_implementation/src/api/so_loaded_executable_instance.cc
```diff
@@ -205,8 +205,21 @@ SOLoadedExecutableInstance::prepareInputTensor(
   // As a hack, we'll always use whatever we have prepared.
   if (prepared_tensor.has_value()) {
```
acicovicTT: Just an idea, should we extract the code in this `if` and in `FlatbufferLoadedExecutableInstance::prepareInputTensor ... if(has_expected_layout) { ... }` to a private function in the parent class?

Issue #1419, opened 2025-09-18
Automatic bisecting for nightly models


nvukobratTT:
## Summary
Similar to what we had on tt-forge-fe, we should have on tt-xla regarding nightly job bisecting. Additionally, I would expand the request to see how we can do a further granular bisecting of related MLIR and Metal modules.

E.g., if some model regresses on the frontend, we can pinpoint which metal commit caused the regression.

nsumrakTT:
We talked about tt-metal->tt-mlir bisecting but we don't have capacity to perform such operation as tt-metal has 30-100 commits per day.

tt-xla bisect is on the way, we had to make changes and stabilize workflows in order to introduce nightly bisect tool.

nsumrakTT:
Currently algorithm bisects tt-xla commits between nightlies and, when uplift is detected, it tries to bisect tt-mlir commits between uplifts. Further automatic tt-metal uplift bisect proves to be very complicated as tt-metal API changes prevents builds. However, @rpavlovicTT has built manual bisect script with claude support that might help:
https://github.com/tenstorrent/tt-xla/pull/2371  

Additionally, PCC drop analyze is added to the Fail Inspector which shows PCC drops between nightlies.

PR #2403, opened 2025-12-02
Nightly fail inspector

nsumrakTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/1419

### Problem description
Bisect commits for failed nightly tests looking for the first bad result.

### What's changed
Workflow is set to run automatically after failed nightly. In this mode it will send results to the slack.

However, it can be also run with manual dispatch. In such case one should provide **nightly's** run id. There is also an option to provide test filter by test name (eg. if single or specific test(s) need to be bisected) but this test has to fail in selected nightly and not to fail in previous nightly. There is also an option to run bisect multiple number of times useful for bisect of flaky tests. 

### Checklist

### DIFF
--- a/.github/workflows/workflow-run-fail-inspect.yml
+++ b/.github/workflows/workflow-run-fail-inspect.yml
@@ -182,7 +182,7 @@ jobs:
           fi
 
       - uses: slackapi/slack-github-action@v1.26.0
-        if: ${{ steps.inspect-failures.outputs.send_msg }}
+        if: ${{ steps.inspect-failures.outputs.send_msg && github.event_name == 'workflow_run' }}
         with:
           payload: ${{ steps.inspect-failures.outputs.send_msg }}
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow-run-fail-inspect.yml
```diff
@@ -157,8 +182,315 @@ jobs:
...
+            echo "Bad commit: $bad_commit"
+            echo "Tests:"
+            echo "$tests"
+
+            # Check if commit file already exists
+            if [[ -f ".bisect-commits/$bad_commit" ]]; then
+              echo "$tests" >> ".bisect-commits/$bad_commit"
+            else
+              echo "$tests" > ".bisect-commits/$bad_commit"
+            fi
```
vvukomanTT: Couldn't we just use 
```
echo "$tests" >> ".bisect-commits/$bad_commit"
```
instead of the entire if block here?

">>" will create a file if it doesn't exist.

Issue #2407, opened 2025-12-02
Swin einsum error


vkovacevicTT:
Swin fails on [latest perf benchmark nightly](https://github.com/tenstorrent/tt-forge/actions/runs/19847809624/job/56868529851) with error:
```
Traceback (most recent call last):
  File "/localdev/vkovacevic/tt-xla/../tt-forge/benchmark/benchmark.py", line 291, in <module>
    main()
  File "/localdev/vkovacevic/tt-xla/../tt-forge/benchmark/benchmark.py", line 280, in main
    results = run_benchmark(config)
              ^^^^^^^^^^^^^^^^^^^^^
  File "/localdev/vkovacevic/tt-xla/../tt-forge/benchmark/benchmark.py", line 111, in run_benchmark
    results = module.benchmark(config)
              ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/localdev/vkovacevic/tt-forge/benchmark/tt-xla/swin.py", line 273, in benchmark
    return test_swin_torch_xla(
           ^^^^^^^^^^^^^^^^^^^^
  File "/localdev/vkovacevic/tt-forge/benchmark/tt-xla/swin.py", line 152, in test_swin_torch_xla
    golden_output = framework_model(golden_input)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/torch/nn/modules/module.py", line 1751, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/torch/nn/modules/module.py", line 1762, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/torchvision/models/swin_transformer.py", line 608, in forward
    x = self.features(x)
        ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/torch/nn/modules/module.py", line 1751, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/torch/nn/modules/module.py", line 1762, in _call_impl
[...]

PR #2412, opened 2025-12-02
Fix einsum override for linear functions

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/2407

### Problem description
Swin fails on [latest perf benchmark nightly](https://github.com/tenstorrent/tt-forge/actions/runs/19847809624/job/56868529851) with error:
```
res = torch.einsum("...mk,...kn->...mn", args[0], args[1])
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/torch/functional.py", line 422, in einsum
    return _VF.einsum(equation, operands)  # type: ignore[attr-defined]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: einsum(): subscript k has size 384 for operand 1 which does not broadcast with previously seen size 96
```

We see this issue because [linear functions transpose the weight](https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.linear.html) before computing a regular matmul, so for `func.__name__ == "linear"` the einsum equation should be `"...mk,...nk->...mn"` ([n, k]) instead of `"...mk,...kn->...mn"` ([k, n]).

### What's changed
- Changed the einsum equation in `TorchFunctionOverride` to be `"...mk,...nk->...mn"` when `func.__name__ == "linear"`.
- Added a linear_torch_override unit test to replicate the conditions that caused a failure in swin and ensure the correct einsum equation is used for linear functions.

### Checklist

### DIFF
--- a/tests/torch/single_chip/ops/test_linear.py
+++ b/tests/torch/single_chip/ops/test_linear.py
@@ -4,9 +4,12 @@
 
 import pytest
 import torch
+import torch.nn.functional as F
 from infra import Framework, run_op_test_with_random_inputs
+from infra.comparators.torch_comparator import TorchComparator
 from utils import Category
 
+from tests.infra.comparators.comparison_config import ComparisonConfig
 from tests.infra.testers.compiler_config import CompilerConfig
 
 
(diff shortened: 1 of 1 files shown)


tests/torch/single_chip/ops/test_linear.py
```diff
@@ -47,3 +47,36 @@ def test_linear(
...
+    input_tensor = torch.randn(1, 128, 128, in_features, dtype=dtype)
+    weight = torch.randn(out_features, in_features, dtype=dtype)
+    bias = torch.randn(out_features, dtype=dtype)
+
+    # Golden output
+    golden_output = torch.einsum("...mk,...nk->...mn", input_tensor, weight)
+    golden_output = golden_output + bias
+
+    # Compute actual output with override active (eager mode, not compiled)
+    with TorchFunctionOverride():
```
AleksKnezevic: This should be applied by default? Is it needed?

tests/torch/single_chip/ops/test_linear.py
```diff
@@ -47,3 +47,36 @@ def test_linear(
...
+    dtype = torch.bfloat16
+    in_features = 96
+    out_features = 384
+
+    input_tensor = torch.randn(1, 128, 128, in_features, dtype=dtype)
+    weight = torch.randn(out_features, in_features, dtype=dtype)
+    bias = torch.randn(out_features, dtype=dtype)
+
+    # Golden output
+    golden_output = torch.einsum("...mk,...nk->...mn", input_tensor, weight)
```
AleksKnezevic: Shouldn't the golden be the actual linear op, then you can compare that the einsum decomp is correct? Not sure if you can do that if `TorchFunctionOverride` is on by default though...

AleksKnezevic (approved):
Couple nits, overall looks great, thanks!

PR #2420, opened 2025-12-03
Improve Pytorch locations for codegen

svuckovicTT:
### Ticket
Part of https://github.com/tenstorrent/tt-mlir/issues/4933

### Problem description
Location data on pytorch ops aren't very useful for codegen purposes.

### What's changed
Updated location data to include full stack of module paths, properly formatted; each op location has info on the line of the op, its function, and the name of the function. Debug mode is also included.

### Checklist

### DIFF
--- a/python_package/tt_torch/backend/metadata_propagation.py
+++ b/python_package/tt_torch/backend/metadata_propagation.py
@@ -60,21 +60,16 @@ class EmitLoc:
     func_name: str
     op_line_num: int
     op_name: str
-
-    # DEBUG
-    is_debug: bool = False
     op_index: int = -1
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


python_package/tt_torch/backend/metadata_propagation.py
```diff
@@ -242,15 +378,18 @@ def _set_metadata(
         in the HLO IR. Since all outputs come from the same operation node, setting metadata
         on one output is sufficient to label the entire operation.
         """
+        pass
```
vkovinicTT: Should this pass be here? If we are not using this method, we can just remove it (but I think it's needed since without it you won't set locations in the `torch-xla`).

Issue #1670, opened 2025-10-13
Incremental tt-mlir build doesn't work


rpavlovicTT:
### Describe the bug

After make a change in third_party/tt-mlir/src/tt-mlir code and invoking `cmake --build build` it fails with 
```
      copying build/lib/tt_alchemist/templates/cpp/standalone/ttnn-install/lib/libbenchmark_main.so -> build/bdist.linux-x86_64/wheel/./tt_alchemist/templates/cpp/standalone/ttnn-install/lib
      copying build/lib/tt_alchemist/templates/cpp/standalone/ttnn-install/lib/libdevice.so -> build/bdist.linux-x86_64/wheel/./tt_alchemist/templates/cpp/standalone/ttnn-install/lib
      copying build/lib/tt_alchemist/templates/cpp/standalone/ttnn-install/bin/capnpc-capnp -> build/bdist.linux-x86_64/wheel/./tt_alchemist/templates/cpp/standalone/ttnn-install/bin
      copying build/lib/tt_alchemist/templates/cpp/standalone/ttnn-install/bin/flatc -> build/bdist.linux-x86_64/wheel/./tt_alchemist/templates/cpp/standalone/ttnn-install/bin
      copying build/lib/tt_alchemist/templates/cpp/standalone/ttnn-install/bin/capnp -> build/bdist.linux-x86_64/wheel/./tt_alchemist/templates/cpp/standalone/ttnn-install/bin
      copying build/lib/tt_alchemist/templates/cpp/standalone/ttnn-install/bin/capnpc-c++ -> build/bdist.linux-x86_64/wheel/./tt_alchemist/templates/cpp/standalone/ttnn-install/bin
      copying build/lib/tt_alchemist/templates/python/local/run -> build/bdist.linux-x86_64/wheel/./tt_alchemist/templates/python/local
      copying build/lib/tt_alchemist/templates/python/local/utils.py -> build/bdist.linux-x86_64/wheel/./tt_alchemist/templates/python/local
      copying build/lib/tt_alchemist/templates/python/local/__init__.py -> build/bdist.linux-x86_64/wheel/./tt_alchemist/templates/python/local
      copying build/lib/tt_alchemist/templates/python/local/my_get_device.py -> build/bdist.linux-x86_64/wheel/./tt_alchemist/templates/python/local
      copying build/lib/tt_alchemist/templates/python/standalone/todo -> build/bdist.linux-x86_64/wheel/./tt_alchemist/templates/python/standalone
[...]

PR #2421, opened 2025-12-03
Update the way tt-alchemist is built

acicovicTT:
### Ticket

#1670 (partially addressed)

### What's changed?

Updates how the tt-alchemist lib binary path is constructed and removes a custom step of building tt-alchemist (no longer needed) when including tt-mlir source tree during cmake config.

### How is the change validated?

1. Re-ran cmake config step to apply the changes.
2. `python examples/pytorch/codegen/resnet.py`
3. `cd resnet50_codegen && ./run` (note: in order for `run` to succeed, a small fix on tt-mlir side is needed as well)


svuckovicTT (approved):
LGTM!

PR #2436, opened 2025-12-04
Fixing issues with data parallel yolo in CI

ajakovljevicTT:
The yolov6l data_parallel test fails with `RuntimeError: PyTorch is not linked with support for jax devices` when ran in CI, but passes when run in isolation.
 
 When data_parallel tests run, they call `xs.mark_sharding()` during TT device execution his function internally calls `maybe_get_torchax()`, which imports `torchax`. On import, torchax registers "jax" as a PyTorch accelerator via `torch.utils.rename_privateuse1_backend('jax') `.
 
 This causes `torch.accelerator.is_available() `to return True for all subsequent tests in the pytest session.

When yolov6l runs on CPU using torch.compile(backend='inductor'), the model contains operations that create FX subgraphs with no tensor inputs, which call `torch.accelerator.current_device_index()`. But since Since "jax" was registered as an accelerator but PyTorch isn't actually linked with jax device support, current_device_index() fails.

I've added a context manager _mask_jax_accelerator() that temporarily makes torch.accelerator.is_available() return False when the jax accelerator is registered and I applied this during  `_run_on_cpu()`.

This is just a workaround, as I think this is a bug in torch that we should raise as an issue. Pinging @AleksKnezevic and @jameszianxuTT to see if they have any thoughts on this.

### DIFF
--- a/tests/infra/testers/single_chip/model/torch_model_tester.py
+++ b/tests/infra/testers/single_chip/model/torch_model_tester.py
@@ -4,6 +4,7 @@
 
 import collections
 import os
+from contextlib import contextmanager
 from typing import Any, Dict, Mapping, Sequence, Set, Tuple
 
 import torch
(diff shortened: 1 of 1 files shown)


jameszianxuTT (approved):
Thanks for looking into this and making a fix! I don't see a better way to implement this.

PR #2437, opened 2025-12-04
Adding first iteration of SDXL pipeline to  tt-xla

ppadjinTT:
## TL;DR

This example file can run inference of Stable Diffusion XL through the compiler, using a custom SDXL Pipeline class.
It performs text2image generation

## More details

Currently, we defined our own SDXLPipeline class that has almost the same API as StableDiffusionXLPipeline from Huggingface diffusers package. We lower only the UNet component to device at the moment (other components should be lowered soon, but UNet is by far the biggest bottleneck for performance).

Current benchmarks - Wormhole singlechip (no optimizer):
Warm single image generation duration (num_inference_steps=50): 115.3s
Of which UNet inference steps: 110.5s

## Future improvements
1. Lowering other model components (text_encoder, text_encoder_2, vae.decoder)
2. Low hanging fruit operation fusions:
- SDPA - SDPA fusing pattern has recently been added to MLIR, but some adapting is required so it can catch the `torch.nn.functional.scaled_dot_product_attention` operation
- Fusing attention preparing to nlp_create_qkv_heads/split_heads ops

### DIFF
--- a/examples/pytorch/sdxl-pipeline.py
+++ b/examples/pytorch/sdxl-pipeline.py
@@ -244,7 +244,9 @@ class SDXLPipeline:
                 )  # (2B, 4, H, W) if do_cfg else (B, 4, H, W)
                 model_input = self.scheduler.scale_model_input(model_input, timestep)
                 model_input = tt_cast(model_input)
-                timestep = tt_cast(timestep.unsqueeze(0))
+                timestep = tt_cast(
+                    timestep.unsqueeze(0)
+                )  # unsqueeze is necessary due to https://github.com/tenstorrent/tt-xla/issues/2408
                 encoder_hidden_states = tt_cast(encoder_hidden_states)
                 pooled_text_embeds = tt_cast(pooled_text_embeds)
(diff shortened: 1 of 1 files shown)


ppadjinTT:
> Awesome!!! How long does it take to generate an image?

620s after compilation :( It is much slower than TTNN SDXL inference (they get around 10-15s if I understood correctly), but there are some obvious improvements we can make. I'm hoping we'll get few X faster inference in the next 2-3 weeks

UPDATE: I just rebased to new tt-xla main and ran the benchmark again and we got 420s inference :D . Not sure what caused the speedup, I'll investigate

AleksKnezevic:
Even more awesome! The unet in SD has attention blocks, so some of the fusion patterns would have really helped. Are you running with optimizer enabled? The resnets block will probably really benefit from that. You can also try enabling tracing, see [example](https://github.com/tenstorrent/tt-xla/blob/main/examples/pytorch/mnist_performant.py). Might get a boost too.

Issue #2291, opened 2025-11-24
Multidevice models failing on nightly CI


ajakovljevicTT:
### Describe the bug

A set of nightly models is failing with bad PCC on the nightly runs:
```
tests/runner/test_models.py::test_all_models_torch[falcon/pytorch-tiiuae/Falcon3-10B-Base-tensor_parallel-full-inference]
tests/runner/test_models.py::test_all_models_torch[mistral/pytorch-7b_instruct_v03-tensor_parallel-full-inference]
tests/runner/test_models.py::test_all_models_torch[mistral/pytorch-magistral_small_2506-tensor_parallel-full-inference]
tests/runner/test_models.py::test_all_models_torch[mistral/pytorch-mistral_nemo_instruct_2407-tensor_parallel-full-inference]
tests/runner/test_models.py::test_all_models_torch[mistral/pytorch-devstral_small_2505-tensor_parallel-full-inference]
tests/runner/test_models.py::test_all_models_torch[mistral/pytorch-mistral_small_24b_instruct_2501-tensor_parallel-full-inference]
```

hkwonTT:
One model is still failing with a low PCC reason even though I reverted my change. [LINK](https://github.com/tenstorrent/tt-xla/actions/runs/19913366502/job/57086920437)

```
FAILED tests/runner/test_models.py::test_all_models_torch[falcon/pytorch-tiiuae/Falcon3-10B-Base-tensor_parallel-full-inference] - AssertionError: Comparison result 0 failed: PCC comparison failed. Calculated: pcc=0.9896957874298096. Required: pcc=0.99.
```

I found the commit that causes this problem using bisect: [1c186b4](https://github.com/tenstorrent/tt-mlir/commit/1c186b475470693966f7675d7b1a4b0cb900bb5b)

Can we close this issue and open another one to track this new low-PCC problem?

PR #2451, opened 2025-12-04
Add 3x TP models to model-test-extended.json to qualify against tt-mlir uplifts

kmabeeTT:
### Ticket
#2291 

### Problem description
- Recent regression in TP models not caught during tt-mlir uplift is ~fixed, we want to protect it from happening next time as talked about in 2291

### What's changed
- Add 3 TP models from mistral, gemma, qwen3-embedding to model-test-extended.json group, runtime is not much longer than single test.

### Checklist
- [x] Tested on branch here: https://github.com/tenstorrent/tt-xla/actions/runs/19946424360 adds 10 minute job with 3 tests in 2-4 minute range.

### DIFF
--- a/.github/workflows/test-matrix-presets/model-test-extended.json
+++ b/.github/workflows/test-matrix-presets/model-test-extended.json
@@ -1,4 +1,4 @@
 [
   { "runs-on": "n150", "name": "extended_models", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "n150 and extended and expected_passing", "parallel-groups": 2 },
-  { "runs-on": "n300-llmbox", "name": "run_forge_models_multichip", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "n300-llmbox and extended and expected_passing", "parallel-groups": 1 }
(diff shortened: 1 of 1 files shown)


nsumrakTT (approved):

.github/workflows/test-matrix-presets/model-test-extended.json
```diff
@@ -1,3 +1,4 @@
 [
-  { "runs-on": "n150", "name": "extended_models", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "extended and expected_passing", "parallel-groups": 2 }
+  { "runs-on": "n150", "name": "extended_models", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "n150 and extended and expected_passing", "parallel-groups": 2 },
+  { "runs-on": "n300-llmbox", "name": "run_forge_models_multichip", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "n300-llmbox and extended and expected_passing", "parallel-groups": 1 }
```
vvukomanTT: We can omit `"parallel-groups": 1` if the test suite isn't split into multiple groups.

Issue #2396, opened 2025-12-02
models failing in nightly with  TypeError and torch._dynamo.exc.TorchRuntimeError


ctr-pmuruganTT:
Few models failed in [nightly](https://github.com/tenstorrent/tt-xla/actions/runs/19842210049/attempts/1) 

```
tests/runner/test_models.py::test_all_models_torch[phi3/causal_lm/pytorch-microsoft/Phi-3-mini-4k-instruct-single_device-full-inference]

tests/runner/test_models.py::test_all_models_torch[phi3/causal_lm/pytorch-microsoft/Phi-3-mini-128k-instruct-single_device-full-inference]

tests/runner/test_models.py::test_all_models_torch[arnold/pytorch-vizdoom_2017_track2_rnn-single_device-full-inference]

tests/runner/test_models.py::test_all_models_torch[arnold/pytorch-vizdoom_2017_track1_rnn-single_device-full-inference]

tests/runner/test_models.py::test_all_models_torch[arnold/pytorch-deathmatch_shotgun_rnn-single_device-full-inference]

tests/runner/test_models.py::test_all_models_torch[transfuser/pytorch-single_device-full-inference] 
```

The above models failed with the error as follows

```
TypeError: arange() received an invalid combination of arguments - got (int, int, device=torch.Size, dtype=torch.dtype), but expected one of:
E            * (Number end, *, Tensor out = None, torch.dtype dtype = None, torch.layout layout = None, torch.device device = None, bool pin_memory = False, bool requires_grad = False)
E            * (Number start, Number end, *, torch.dtype dtype = None, torch.layout layout = None, torch.device device = None, bool pin_memory = False, bool requires_grad = False)
E            * (Number start, Number end, Number step = 1, *, Tensor out = None, torch.dtype dtype = None, torch.layout layout = None, torch.device device = None, bool pin_memory = False, bool requires_grad = False)
```

`tests/runner/test_models.py::test_all_models_torch[transfuser/pytorch-single_device-full-inference] ` model failed with 

```
 torch._dynamo.exc.TorchRuntimeError: Dynamo failed to run FX node with fake tensors: call_function <Wrapped method <original div>>(*(FakeTensor(..., size=()), 2), **{}): got AttributeError("'ndarray' object has no attribute 'div'")
E       
E       from user code:
E          File "/__w/tt-xla/tt-xla/third_party/tt_forge_models/transfuser/pytorch/src/model.py", line 389, in forward
[...]

PR #2452, opened 2025-12-04
Fix torch_function_override nightly errors

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/2396

### Problem description
The use of TorchFunctionOverride is causing the following errors in nightly tests.
**Tensor attribute access error:**
```
TypeError: arange() received an invalid combination of arguments - got (int, int, device=torch.Size, dtype=torch.dtype), but expected one of:
E            * (Number end, *, Tensor out = None, torch.dtype dtype = None, torch.layout layout = None, torch.device device = None, bool pin_memory = False, bool requires_grad = False)
E            * (Number start, Number end, *, torch.dtype dtype = None, torch.layout layout = None, torch.device device = None, bool pin_memory = False, bool requires_grad = False)
E            * (Number start, Number end, Number step = 1, *, Tensor out = None, torch.dtype dtype = None, torch.layout layout = None, torch.device device = None, bool pin_memory = False, bool requires_grad = False)
```
This caused by calling `torch.compiler.is_compiling()` for every function that passes through `__torch__function__`. It causes tensor attributes accesses to always return the value of the first attribute accessed. 
 
 [EDIT]: transfuser is failing with a new torch override related error after rebasing, so it will be fixed in a seperate pr.

### What's changed
The **tensor attribute error** is resolved when we only call `torch.compiler.is_compiling()` if the function is linear or matmul:
```
if (
     func.__name__ == "matmul" or func.__name__ == "linear"
) and not torch.compiler.is_compiling():
```
Added a unit test `test_tensor_attributes_torch_override` that is now passing with the above change.

### Checklist


gengelageTT:
Tests that were failing in nightly are passing with these changes here: https://github.com/tenstorrent/tt-xla/actions/runs/19944651753

gengelageTT:
Returning `NotImplemented` for torch functions that expect numpy inputs [caused failures](https://github.com/tenstorrent/tt-xla/actions/runs/19947020299/job/57199479878?pr=2452#step:17:374) so it now checks that `func` isn't a known numpy compatible function before returning `NotImplemented`. [Running nightly on this branch](https://github.com/tenstorrent/tt-xla/actions/runs/19971194084) to check that no numpy compatible functions were missed.

python_package/tt_torch/torch_overrides.py
```diff
@@ -1,15 +1,56 @@
 # SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
 #
 # SPDX-License-Identifier: Apache-2.0
+import numpy as np
 import torch
 from torch.overrides import TorchFunctionMode
 
+# Functions that accept numpy arrays and convert them
+NUMPY_COMPATIBLE_FUNCS = {
```
jameszianxuTT: QQ - where does this list come from? Is there some more authoritative source than inlining a string array into an override?

PR #2479, opened 2025-12-06
[vLLM plugin] Create attention mask for batched inputs

### LINKED ISSUE #2478: [vLLM plugin] Add support for batched attention mask
We are using attention mask in every scenario for pooling models with encoder-only attention layers. Update attention mask generation to create mask for batched inputs. 
Refactor/parameterize the test for batched inputs for encoder-only and decoder-only model.

mmanzoorTT:
### Ticket
closes #2478 

### Problem description
We are using attention mask in every scenario for encoder-only pooling models.

### What's changed
- Update attention mask generation to create mask for batched inputs.
- Refactor/parameterize the test for batched inputs for encoder-only and decoder-only models.

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/pooling_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/pooling_runner.py
@@ -132,8 +132,6 @@ MIN_NUM_SEQS = 1
 def generate_attn_mask(
     context_lens: torch.Tensor,
     num_query_tokens: int,
-    num_query_heads: int,
-    max_model_len: int,
     batch_size: int,
     is_decoder_only: bool,
     dtype: torch.dtype,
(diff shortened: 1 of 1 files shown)


integrations/vllm_plugin/vllm_tt/pooling_runner.py
```diff
@@ -134,58 +134,73 @@ def generate_attn_mask(
...
 
-    Note:
-        Decoder-only attention layers require causal attention (triangular) masks.
-        Encoder-only attention layers require bidirectional self-attention (square) masks.
+    - Decoder-only models use causal (triangular) attention, but only when the
+      input contains multiple concatenated sequences (i.e., sequences that must
+      not attend to each other). A single contiguous sequence requires no causal
+      mask.
+
+    - Encoder-only models use bidirectional self-attention (square masks). These
```
AleksKnezevic: We need to measure the performance overhead for this. Is it possible to use a batch*seq-len mask if we have a batch of padded (non-concatenated) sequences?

seq-len*seq-len masks will also greatly reduce the max seq-len we support due to their memory requirements.

mmanzoorTT: We are already using `input[-1]` as seq-len; not max-seq-len. For batch-1; seq-len is `sum (all input sequences) + padding` and for batch>1; seq-len is `max(all input sequences) + padding`.

Issue #1002, opened 2025-08-13
SHLO op-by-op mode


AleksKnezevic:
Enable stableHLO op-by-op mode using infra brought up by @kmitrovicTT

mrakitaTT:
For the reference: https://github.com/tenstorrent/tt-xla/pull/599

AleksKnezevic:
It's not too slow, the largest model we have runs for ~40 minutes I think. @jameszianxuTT or @kmabeeTT might have a more correct answer. Also, this is op-by-op from pytorch, which should be a little slower than op-by-op from SHLO. 

There is unique op filtering per model (based on input shapes and attributes), so you only run each op once within a model, but not globally. We have all of our WIP models running in CI nightly and all of the already brought up ones weekly. I would like to maintain a similar pipeline as the number of models scales. 

Reporting is in superset for a general overview, and there is also a nice spreadsheet that's generated with additional information (see [models_op_per_op_2025-08-15_main_87144d5a.xlsx](https://github.com/tenstorrent/tt-torch/actions/runs/16982739887/artifacts/3772458440) as an example).

There's more we can do with this info. I have a not fully flushed out plan on building op chains and doing sweeps like that for increased coverage on multiple platforms with limited CI.

kmabeeTT:
Yeah that is pretty much correct, quick look at nightly (pretty much all torch op by op flow) and I see many complete in minutes, some bigger models are 40 minutes to 1 hour ([link](https://github.com/tenstorrent/tt-torch/actions/runs/16982739887/job/48146444861)).  Speed is not terrible, but not as fast as I'd like, could probably be improved.  Having that XLS generated with results was very useful for debug, triaging, driving closure, I would also like it to be maintained.

nvukobratTT:
@ddilbazTT, any new updates on the state of this effort? :))

nvukobratTT:
We confirmed offline that this effort will be handed off, as @ddilbazTT is currently focused on other priorities.

AleksKnezevic:
@sdjukicTT any updates?

nvukobratTT:
One of the reasons I'm more in favor of better integration with tt-xla is that there are known issues/limitations that should be addressed to achieve tighter integration with our frontend (which contains all model tests). Solving the current limitations will enable us to achieve both goals:
1. Have the ability to analyze models locally
2. Easily enable analysis and record details on CI through existing model testing logic

[...]

PR #2494, opened 2025-12-08
Op by op mode test

sdjukicTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/1002
Solution from this PR has drawbacks explained in [issue comment](https://github.com/tenstorrent/tt-xla/issues/1002#issuecomment-3627583883); it is temporary until a solution that is appropriate for both CI and local testing is implemented. 

### Problem description
We want to have a test that tests models in op-by-op mode, which should:
- run the model to get the StableHLO IR;
- wrap each op in its own module;
- compile and run each op module;
- create a report with results.

### What's changed
- added `--dump-irs` option to `test_all_models_torch/jax` methods -> used instead of `--serialize` because it dumps IR modules even if compilation didn't finish successfully (we use StableHLO IR for op_by_op for now and existence of it is enough)
- removed `full` from configs/parametrization/test_id creation -> op_by_op is just another mode of testing, it doesn't need to have its own config
For example, if we collect tests with 
`pytest --collect-only -q tests/runner/test_models.py -m "expected_passing and n150"`
we will get op_by_op tests for models that are expected to pass in regular (full) mode.
- extracted shared logic for JAX and Torch tests to reduce code duplication
- added test_all_models_op_by_op method that can be called to test a model in op by op mode

For this test to work:
1. project needs to be built with `-DTTMLIR_ENABLE_BINDINGS_PYTHON=ON`
2. system descriptor needs to be saved using `ttrt query --save-artifacts`
3. for the report to be created, test should be called with `--json-report --json-report-file=report.json`

### Notes
For now, only single chip models were considered.
If this test is ran on N300 machine, it might leave it in inconsistent state, needing a reset.

### Checklist

### DIFF
--- a/tests/runner/test_models.py
+++ b/tests/runner/test_models.py
@@ -392,7 +392,7 @@ def test_all_models_op_by_op(
         f"[{model_test_id}-{parallelism.value}-{run_mode.value}]"
     )
 
-    pytest_cmd = [sys.executable, "-m", "pytest", pytest_node_id, "-v", "--dump-irs"]
+    pytest_cmd = [sys.executable, "-m", "pytest", pytest_node_id, "-sv", "--dump-irs"]
 
     subprocess_result = subprocess.run(
         pytest_cmd,


tests/runner/utils/dynamic_loader.py
```diff
@@ -444,6 +453,8 @@ class JaxDynamicLoader(DynamicLoader):
     particularly for discovering JAX model loader files.
     """
 
+    framework = Framework.JAX
```
ajakovljevicTT: can we put this in `__init__()`? Same with torch above

tests/runner/test_models.py
```diff
@@ -227,124 +356,131 @@ def test_all_models_torch(
...
+    variant, ModelLoader = test_entry.variant_info
+    model_info = ModelLoader.get_model_info(variant=variant)
 
-                # Check if filecheck patterns are specified
-                pattern_files = (
-                    test_metadata.filechecks
-                    if hasattr(test_metadata, "filechecks")
-                    else None
-                )
+    nodeid = (
```
ajakovljevicTT: Can we be a little more descriptive here? Both with naming of the variable and have some sort of a comment above

ajakovljevicTT (approved):

sdjukicTT:
> > For now, only single chip models were considered.
> > If this test is ran on N300 machine, it might leave it in inconsistent state, needing a reset.
> 
> Can you clarify on this?

Multichip models have different IR and weren't taken into account in this PR, they require extra support.

Completely unrelated: when this op-by-op test for single chip models is ran on N300, a reset afterwards is sometimes needed. I didn't research this issue much because this solution is temporary and I prioritized developing the long term script solution.

PR #2500, opened 2025-12-08
Set tensor retain on output buffer instance constructor

jameszianxuTT:
### Ticket
Partial resolution to #2383, as alternative to worse solution in #2389 

### Problem description
Previously in #1657, output BufferInstances had tensor retain set lazily at execution if output is reused as input. This is not necessary, and introduces a duplicate complexity in the loadedExecutableInstance classes that should really belong in BufferInstance implementation.

### What's changed
In the case where a bufferInstance is constructed with an live device tensor, set retain=True inside the constructor rather than lazily doing so at execution time.

### Checklist


sgligorijevicTT (approved):
Much cleaner! Thanks for the fix

PR #2534, opened 2025-12-10
[vLLM plugin] Convert invalid min_context_len flag to valid length

### LINKED ISSUE #2533: [vLLM plugin] minimum input length assertion bug
vLLM determines the padded input length using the range [min_context_len, max_num_batched_tokens] (inclusive). In contrast, tt-metal imposes a stricter requirement: the input length must be a multiple of 32. The vLLM plugin did not enforce this divisibility constraint and only checked that min_context_len was greater than zero. This causes vLLM plugin to crash if `min_context_len<32`.

mmanzoorTT:
### Ticket
closes #2533 

### Problem description
vLLM plugin does not use the same constraints as required by tt-metal for input length.

### Constraints:
- tt-metal requires input length to be multiple of 32.
- vLLM plugin uses power of 2 for padding sizes to keep number of precompile graphs manageable.

### What's changed
- Ensure min_context_len flag satisfies the constraint.
- Round up min_context_len if necessary, and log a warning when adjustments are made.
- Remove the assertion and improve logging for generated paddings. 
- Simplify/clean the padding logic to match pooling runner constraints.

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/pooling_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/pooling_runner.py
@@ -1962,11 +1962,38 @@ def _get_padded_num_reqs_with_upper_limit(x: int, upper_limit: int) -> int:
     return min(res, upper_limit)
 
 
+def _adjust_min_token(min_token_size: int) -> int:
+    """
+    Ensure min_token_size is a power of two and >= 32 (divisible by 32).
+
+    If min_token_size already meets the constraint, return it unchanged.
+    Otherwise, round it up to the next power of two (minimum 32).
+    """
+    # Check if min_token_size satisfies the constraints.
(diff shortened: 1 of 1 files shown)


ljovanovicTT (approved):

kmabeeTT (approved):
This is first VLLM code I've ever looked at, so more questions than answers, but seems okay and matches ticket/comments (plus a bit more: logging cleanup, paddings simplification, might be worth mentioning as bulletpoint in description).  

Were there some specific models that this code exposed an issue in?  And ticket mentioned "This causes vLLM plugin to crash if min_context_len<32.", but does this PR solve that crash somehow, or just assert earlier?

mmanzoorTT:
Thanks @kmabeeTT I agree that rounding up `min_context_len` will improve user-experience. I have updated the PR. Now `min_context_len` is rounded up (if required) and a warning message is also logged.

Issue #1494, opened 2025-09-29
gemma-2-27b fails to fit on t3k (n300-llmbox) via Tensor Parallel


kmabeeTT:
The test is gaining Tensor Parallel support for T3K via https://github.com/tenstorrent/tt-xla/pull/1429 but somewhat surprisingly does not fit on n300-llmbox (t3k):

```
pytest -svv tests/runner/test_models.py::test_all_models[gemma/pytorch-google/gemma-2-27b-it-full-inference] --arch n300-llmbox |& tee gemma2_27b.log


Out of Memory: Not enough space to allocate 2359296000 B DRAM buffer across 12 banks, where each bank needs to store 196614144 B, but bank size is only 1073741792 B
```

Meanwhile the 9B version fits and runs (`tests/runner/test_models.py::test_all_models[gemma/pytorch-google/gemma-2-9b-it-full-inference]`)

Issue #2308, opened 2025-11-24
Live on-device BufferInstances will be deallocated if the MeshDevice is reshaped before their last usage


jameszianxuTT:
Currently, we cache an `m_prepared_runtime_tensor` inside buffer instance, representing live laid out, on-device runtime tensors. This is either a singledevice or multidevice tensor, but maps to a singular tt::runtime::Tensor instance.

In the PJRT runtime, it is assumed that these m_prepared_runtime_tensors are **always valid and allocated** as long as the BufferInstance itself exists. 

There are a few things that can deallocate runtime tensors:
1. ttnn deallocator pass - Which doesn't trigger on bufferInstance cached tensors because bufferInstances represent inputs or outputs. We set the retain bit on all input bufferInstances, which means that the ttnn deallocator pass or any calls to runtime::deallocateTensor will be a no-op. Outputs are not eagerly deallocated anymore after #1657, so they will always be alive - if they are resubmitted as inputs to a different graph, they go through the same path to set the retain bit, and will not be auto-deallocated.
2. shared_ptr refcount - We are holding a reference to the live tensor in the BufferInstance, so refcount won't drop and trigger deallocation
3. (implicit) MeshDevice closing deallocates all live tensors.  (eg. if we reshape a mesh, we close and reopen it).

The 3rd case will happen unintentionally if:
1. there's a graph break causing the trace of a replicated graph that could run on 1x1 -> auto-reshaped to 1x8 by #1560. This will cause a reshape of the underlying meshdevice of the original mesh shape (eg. as requested by the model loader mesh_shape), if that original mesh shape is nonlinear, eg. 2x4 as in falcon / qwen. 
2. If models running on different mesh shapes run back to back in the same process, and for some reason share python objects pointing to the same allocated tensor as in #1020 

This will lead to segfault if we try to reuse a tensor that was allocated for execution in one graph -> deallocated during mesh reshape between graph executions -> reused as input in a future subgraph execution.

Some possible solutions:
1. Use a proper meshDevice reshape. It is mentioned in a comment that this causes hangs, but should be investigated further. #1436. Also, I am not sure if reshaping mesh device will prevent deallocation of live tensors.
2. Track liveness of BufferInstances from some singleton weak_ptr map, and transfer extant bufferInstances back to host prior to a reshape. This will be expensive and is pretty hacky. 
[...]

PR #2535, opened 2025-12-10
Enable multi-chip op testing and add Gemma TP op tests

gengelageTT:
### Ticket
N/A

### Problem description
A strategy to help handle large weights for Gemma-2-27b is to [add the following shard specs](https://github.com/tenstorrent/tt-xla/issues/1494#issuecomment-3575898877):
```
        shard_specs[model.model.embed_tokens.weight] = ("batch", "model")
        shard_specs[model.lm_head.weight] = ("batch", "model")
```
[Edit]: or `(None, "batch")` for mesh_shape `(2, 4)`

 We want to add unit tests for sharding these ops (`nn.Embedding`,`nn.Linear`) for this model. The op tester infra doesn't support passing in `mesh` or `shard_spec_fn` yet.
 
### What's changed
- Added support for `mesh` and `shard_spec_fn` as optional inputs to `run_op_test()` (same as done in `run_graph_test()`) and updated `test_spmd_sharding` to use the op_tester.

- Added tensor-parallel op tests for Gemma-2-27b `lm_head` and `embed_tokens` [Edit]: and parametrized by `mesh_shape` and `shard spec` to test both possible shard specs.

### Checklist

### DIFF
--- a/tests/torch/ops/test_embedding.py
+++ b/tests/torch/ops/test_embedding.py
@@ -34,8 +34,15 @@ class Embedding(torch.nn.Module):
 
 @pytest.mark.nightly
 @pytest.mark.llmbox
+@pytest.mark.parametrize(
+    "mesh_shape,shard_spec",
+    [
+        ((1, 8), ("batch", "model")),
+        ((2, 4), (None, "batch")),
+    ],
+)
 @pytest.mark.record_test_properties(category=Category.OP_TEST)
-def test_gemma2_27b_embed_tokens():
+def test_gemma2_27b_embed_tokens(mesh_shape, shard_spec):
     loader = GemmaModelLoader(variant=GemmaModelVariant.GEMMA_2_27B_IT)
     config = loader.load_config()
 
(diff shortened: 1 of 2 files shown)


kmabeeTT:
Nice tests and changes! Those shard specs you linked to from my comment were an initial guess (then there has been a lot of offline discussion about them and debug with @hshahTT, trying alternatives but no hard conclusions), I think it makes sense to go with this for now though, unless Het has alternative that is worth using.

hshahTT:
Oh that happens because it's a (1, 8) mesh with "batch" = 1 and "model" = 8. So `(None, "batch")` is the same as `(None, None)` (since the batch axis only has 1 device).

Could you change this to a (2, 4) mesh with "batch" = 2 and "model" = 4 instead?

gengelageTT:
Tests passing with new parametrization by mesh_shape here: https://github.com/tenstorrent/tt-xla/actions/runs/20142144630

kmabeeTT:
Unfortunately mesh_shape paramaterization on test (mixing mesh shapes in same CI job) will not work after imminent transformers uplift [PR](https://github.com/tenstorrent/tt-xla/pull/2545) according to Jona, which brings an assert on mixing not being allowed, and explicit change to separate models into separate CI jobs to not mix 1x8 and 2x4 mesh_shape models.   

To avoid breakage once his change lands, we'd have to do the same here (split tests into separate CI jobs), which is annoying... or temporarily stick to the same mesh_shape across the board for all tests in same job while this is an issue (pending resolution of https://github.com/tenstorrent/tt-xla/issues/2308, not been looked into too deep yet).

PR #2553, opened 2025-12-12
Extract characteristic summary message per failing reason

### LINKED ISSUE #2225: Extract characteristic summary message per failing reason
Usually exception mapped to failing reason has a characteristic message that helps identifying additional details about exception.
That characteristic message is sometimes part of error message thats represented at start of exception message, but sometimes is in the middle of error log.
Failing reasons should support extracting that message via additional regexes.

Example of failing reason definition:
```
summary=[
    M.regex("Out of Memory: Not enough space to allocate .* B L1 buffer across .* banks, where each bank needs to store .* B, but bank size is only .* B"
]
```

Example pytest tags:
```
'failing_reason': {'name': 'OUT_OF_MEMORY', 'description': 'Out of memory', 'component': 'Metal', summary: 'Out of Memory: Not enough space to allocate 44235264 B L1 buffer across 48 banks, where each bank needs to store 921568 B, but bank size is only 1331648 B'}
```

vobojevicTT:
Resolves #2225

### DIFF
--- a/tests/infra/utilities/failing_reasons/checks_xla.py
+++ b/tests/infra/utilities/failing_reasons/checks_xla.py
@@ -7,7 +7,7 @@
 from enum import Enum
 from typing import Optional
 
-from .utils import ExceptionCheck, FailingReason, M
+from .utils import ME, ExceptionCheck, FailingReason, M
 
 
 class ComponentChecker(Enum):
@@ -558,6 +558,11 @@ class FailingReasons(Enum):
                         ),
                     ),
                 ],
+                summary=[
+                    ME.regex(
+                        "(TT_THROW: Statically allocated circular buffers on core range)"
+                    ),
+                ],
             ),
             # Circular buffers exceed L1 size
             ExceptionCheck(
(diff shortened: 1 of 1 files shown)


tests/infra/utilities/failing_reasons/utils.py
```diff
@@ -73,9 +81,26 @@ def last_line(checker: MessageCheckerType) -> str:
         )
 
 
+class MessageExtractor:
+    @staticmethod
+    def regex(pattern: str) -> str:
+        """Extract a substring from the message using the given regex pattern."""
+
+        def extractor(ex_message: str) -> Optional[str]:
```
vbrkicTT: Maybe better to use verb `extract` for the method name

tests/runner/test_models.py
```diff
@@ -153,8 +154,9 @@ def test_all_models_torch(
                 validate_filecheck_results(filecheck_results)
 
         except Exception as e:
-            out = capteesys.readouterr().out
-            err = capteesys.readouterr().err
+            # out = capteesys.readouterr().out
+            # err = capteesys.readouterr().err
+            out, err = capfd.readouterr()
```
vbrkicTT: This will not be needed after #2662

vbrkicTT (commented):
It's better to wait until #2662 is merged.

tests/infra/utilities/failing_reasons/checks_xla.py
```diff
@@ -558,37 +558,73 @@ def find_by_description(cls, desc: str) -> Optional["FailingReasons"]:
...
             ExceptionCheck(
                 stderr=[
                     M.contains("circular buffers"),
                     M.contains("beyond max L1 size"),
                 ],
+                summary=[
+                    ME.regex(
+                        "(TT_THROW: Statically allocated circular buffers on core range)"
+                    ),
+                ],
```
vbrkicTT: We should later cleanup redundant checks. For now it's fine to make robust this way.

vbrkicTT (approved):
Aside from small cleanups everything looks good to me.

vbrkicTT (approved):

PR #2558, opened 2025-12-12
Roll back tt_forge_models to avoid transformers changes 

kmabeeTT:
(Speculative PR in case it is needed... will adjust description more later today if this is needed/not-needed).

### Ticket
None

### Problem description
- tt-forge-models changes across several models that depend on transformers uplift in tt-xla (pending, not yet released) made it into tt-xla via automatic uplift job
- There is a chance transformers uplift might not land today (perf-benchmarks are pending/running) and if it doesn't we want to not break nightly models, so pre-emptively get this PR ready to roll it back in case that happens
- If transformers uplift does land then this PR can be closed, not needed.

### What's changed
- Roll back to tt-forge-model change before transformers changes

### Checklist


kmabeeTT:
Marking closed, transformers uplift landed, this is not needed.

PR #2572, opened 2025-12-15
Add debug build workflow with ctest and code coverage

acicovicTT:
## Issues

- https://github.com/tenstorrent/tt-xla/issues/2429
- https://github.com/tenstorrent/tt-xla/issues/2430

## Summary

- Implements a `debug` build type within the existing `call-build.yml` workflow, building with debug info and code coverage instrumentation
- Runs C++ unit tests after the build
- Uploads gtest XML report and test logs as artifacts (`CMakeLists.txt` is updated so that gtest emits this report), with the expected artifact name format compliant with superset
- Generates and uploads code coverage to Codecov
- Integrate the new workflow into `pr-main.yml` and `push-main.yml` to run in parallel with other build jobs
- **Removes** the existing `codecov` build type that was added so integration tests could generate code coverage info
- **Removes** the steps from `call-test.yml` that were uploading code coverage
- **Removes** other references to codecov in integration test configs and other GitHub workflows

### DIFF
--- a/.github/workflows/call-build-debug.yml
+++ b/.github/workflows/call-build-debug.yml
@@ -33,13 +33,13 @@ jobs:
             submodules: recursive
             repository: 'tenstorrent/tt-xla'
 
-      - name: Fetch job id
+      - name: Fetch job ID
         id: fetch-job-id
         uses: tenstorrent/tt-github-actions/.github/actions/job_id@main
         with:
           job_name: "Build tt-xla (debug)"
 
-      - name: Override tt-mlir SHA mlir_override is set
+      - name: Override tt-mlir SHA if mlir_override is set
         if: ${{ inputs.mlir_override }}
         shell: bash
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/call-build-debug.yml
```diff
@@ -0,0 +1,205 @@
...
+        if: success() || failure()
+        uses: actions/upload-artifact@v4
+        with:
+            name: ${{ needs.check-existing-artifact.outputs.build_artifact_name }}
+            path: ${{ steps.strings.outputs.build-output-dir }}/artifact.tar
+
+      - name: Remove tar archive
+        if: success() || failure()
+        shell: bash
+        run: rm -f ${{ steps.strings.outputs.build-output-dir }}/artifact.tar
```
nsumrakTT: I don't think we need build directory anymore (we needed it for codecov wheel tests because of gcda files)

nsumrakTT (approved):
just a few nits

Issue #2463, opened 2025-12-05
Sharded embeddings


AleksKnezevic:
There's an issue when we try to shard embeddings weights on hidden dim. It should be pretty simple, embeddings sharded on hidden-dim, all-gather results, hidden states become replicated and we continue execution as expected. Currently, we get this:

```
#loc1 = loc("p0.2")
#loc2 = loc("p1.7")
#loc3 = loc("p2.16")
#loc4 = loc("p3.21")
#loc5 = loc("p4.29")
#loc6 = loc("p5.34")
#loc7 = loc("p6.70")
#loc8 = loc("p7.91")
#loc9 = loc("p8.98")
#loc10 = loc("p9.101")
#loc11 = loc("p10.118")
#loc12 = loc("p11.138")
#loc13 = loc("p12.228")
#loc14 = loc("p13.237")
#loc27 = loc("reduce.50")
#loc99 = loc("dot.194")
#loc107 = loc("reduce.208")
module @SyncTensorsGraph.255 attributes {mhlo.cross_program_prefetches = [], mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
  sdy.mesh @mesh = <["_axis_0"=2, "_axis_1"=4]> loc(#loc)
[...]

PR #2573, opened 2025-12-15
Add sharding constraint API for intermediate tensors

sshonTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/2463

### Problem description
Unlike JAX, there is no way to express sharding constraints on the intermediate tensor, so it makes less-efficient graph with ccl ops

### What's changed
Introduce user-facing sharding API to apply sharding constraints on intermediate tensor outputs when compiling with torch.compile
In tt-torch python module,
 - Add sharding.py with sharding_constraint_hook() function
~~- Registers forward hook internally~~
~~- Returns RemovableHandle for optional hook removal~~
 - Returns the hook that user can use wherever they want (to use forward, pre_forward, backward,...)
 - Includes input validation and error handling

Usage:
```
  from tt_torch import sharding_constraint_hook 
  hook = sharding_constraint_hook(model.embed_tokens, mesh, ("batch", None, None))
  model.embed_tokens.register_forward_hook(hook)
```

Expected result:
  stablehlo.custom_call with `@tt.sharding_constraint` op is generated immediately after the target op. for example, 
`%16 = stablehlo.custom_call @tt.sharding_constraint(%15) {mhlo.frontend_attributes = {xla.sdy.sharding = "#sdy.sharding_per_value<[<@mesh, [{\22_axis_0\22}, {}, {}]>]>"}} `
and it will be converted to `sdy.constraint_op` in the stablehlo pipeline in tt-mlir, and it will be converted to corresponding ccl op at the end of the pipeline.

### Checklist

### DIFF
--- a/tests/filecheck/sharding_constraints.ttir.mlir
+++ b/tests/filecheck/sharding_constraints.ttir.mlir
@@ -1,0 +1,4 @@
+//CHECK: ttir.gather
+//CHECK-SAME: -> tensor<32x128
+//CHECK: ttir.all_gather
+//CHECK-SAME: -> tensor<1x32x512
--- a/tests/torch/test_torch_xla_basic.py
+++ b/tests/torch/test_torch_xla_basic.py
@@ -2,6 +2,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 import os
+import shutil
+from pathlib import Path
 
 import numpy as np
 import pytest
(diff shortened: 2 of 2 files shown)


python_package/tt_torch/sharding.py
```diff
@@ -0,0 +1,98 @@
...
+    if not hasattr(mesh, "axis_names"):
+        raise ValueError("mesh must have 'axis_names' attribute")
+
+    # Convert to sdy.sharding string at hook creation time
+    sdy_sharding = _partition_spec_to_sdy_sharding(mesh, partition_spec)
+
+    def hook(mod, input, output):
+        return torch.ops.tt.sharding_constraint(output, sdy_sharding)
+
+    return module.register_forward_hook(hook)
```
AleksKnezevic: What do you think of simply returning the hook and having the user call `register_forward_hook` from their code? ie, something like:

```python
hook = sharding_constraint_hook(mesh, ("batch, None))
model.embedding.weight.register_forward_hook(hook)
```

This way users can also register `pre_forward` or `backward` hooks.

sshonTT:
> Two general questions:
> 
> 1. Do these overrides work for both compiled and eagerly executed modules?
> 2. Do we suspect and perf impact? That is, does this introduce graph breaks?

@AleksKnezevic 
1. I haven't checked it, but it should work as it follows other custom_op semantic.
2. I haven't seen graph break from this

Issue #2586, opened 2025-12-16
Jax mnist multichip fails with Error code 13


ndrakulicTT:
### Describe the bug

Action from tt-blacksmith: https://github.com/tenstorrent/tt-blacksmith/actions/runs/20222195874?pr=318

### Steps to reproduce the issue

For the repro:
1. activate xla env
2. clone and enter tt-blacksmith
4. `pip install wandb && pip install -e .`
5. `python https://github.com/tenstorrent/tt-blacksmith/blob/main/blacksmith/experiments/torch/mnist/tensor_parallel/test_mnist_training.py`
I am also uploading the small repro that can be run outside of tt-blacksmith

[repro.py](https://github.com/user-attachments/files/24187967/repro.py)

PR #2600, opened 2025-12-16
Fix buffer lifetime management for mesh reshape scenarios

ajakovljevicTT:
This PR adds buffer tracking infrastructure to manage buffer lifetimes across mesh reshape operations. When we reshape meshes between graph runs, buffers from previous executions could hold stale device tensor references after mesh reshape, causing assertion failures in prepareInputTensor where buffers in the same sharded group had inconsistent prepared_tensor states. The fix introduces registerBuffer/unregisterBuffer methods to ClientInstance that track all buffers created with a client reference. A new materializeAllBuffersToHost() method is called during closeMeshDevice() to copy device tensors to host for buffers that don't have host tensors, and clear all prepared_tensor references to ensure consistent state. All buffer creation sites now pass the client instance so buffers are properly tracked.

Partially fixes https://github.com/tenstorrent/tt-xla/issues/2308
Partially fixes https://github.com/tenstorrent/tt-xla/issues/2586

### DIFF
--- a/pjrt_implementation/src/api/unit_tests/buffer_instance_tests.cc
+++ b/pjrt_implementation/src/api/unit_tests/buffer_instance_tests.cc
@@ -13,6 +13,7 @@
 
 // PJRT implementation headers
 #include "api/buffer_instance.h"
+#include "api/client_instance.h"
 
 namespace tt::pjrt::tests {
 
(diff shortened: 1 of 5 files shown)


pjrt_implementation/src/api/buffer_instance.cc
```diff
@@ -100,9 +107,18 @@ BufferInstance::BufferInstance(
   if (m_prepared_runtime_tensor.has_value()) {
     tt::runtime::setTensorRetain(*m_prepared_runtime_tensor, /*retain=*/true);
   }
+
+  if (m_client) {
```
sgligorijevicTT: This seems like an invariant, assert seems more appropriate

sgligorijevicTT (approved):
For a hack it't a not *that* bad, but let's hope it gets ripped out ASAP. Approving.

jameszianxuTT:
Perhaps you could add my repro to the test_tensor_persistence testgroup?

https://github.com/tenstorrent/tt-xla/issues/2308#issuecomment-3582509822

Didn't want to add it before since it would leave the device in a bad state, but it could help catch regressions

also cc @kmabeeTT, this fix should make it so that after transformers uplift, we can run falcon 2x4 and llama 1x8 in the same testgroup

pjrt_implementation/src/api/unit_tests/buffer_instance_tests.cc
```diff
@@ -28,16 +29,22 @@ class BufferInstanceUnitTests : public PJRTComponentUnitTests {
...
-  std::unique_ptr<BufferInstance> createInputBuffer() {
+  std::unique_ptr<BufferInstance> createInputBuffer(ClientInstance *client) {
     return BufferInstance::createInputBufferInstance(
         DEFAULT_DATA_TYPE, DEFAULT_DIMS, DEFAULT_NUM_DIMS, m_device.get(),
-        m_default_memory.get());
+        m_default_memory.get(), client);
   }
 
+  // Mock client for buffer tracking (uninitialized, no hardware access).
+  std::unique_ptr<ClientInstance> m_mock_client;
```
acicovicTT: Maybe it would be beneficial for other tests to have this in the base class `PJRTComponentUnitTests`

PR #2618, opened 2025-12-17
PCC Promotions (enable/raise) based on Dec 16 nightly

kmabeeTT:
### Ticket
None

### Problem description
- There are 16 cases where PCC is higher than threshold and assert_pcc:false can be removed (especially bunch of red language models recently worked on)
- There are 16 cases where required_pcc threshold can be raised (or removed if 0.99) - all generality models.

### What's changed
- Make updates for these cases to test_config files based on dec 16 nightly run-id [20287272543](https://github.com/tenstorrent/tt-xla/actions/runs/20287272543)
- They were made using promote_from_junit_guidance.py script which I hope to put up PR for soon after more testing. Will include some details of it's usage for reference in this PR since there are several interesting cases.
- Some whitespace fixes introduced from ruamel.yaml round trip yaml editing in a couple places

### Checklist
- [x] Tested on branch: https://github.com/tenstorrent/tt-xla/actions/runs/20316330562

### DIFF
--- a/tests/runner/test_config/torch/test_config_inference_single_device.yaml
+++ b/tests/runner/test_config/torch/test_config_inference_single_device.yaml
@@ -511,8 +511,8 @@ test_config:
   vit/pytorch-base-single_device-full-inference:
     status: EXPECTED_PASSING
     filechecks:
-    - concatenate_heads.ttnn.mlir
-    - split_query_key_value_and_split_heads.ttnn.mlir
+      - concatenate_heads.ttnn.mlir
+      - split_query_key_value_and_split_heads.ttnn.mlir
     markers: ["push"]
 
   vit/pytorch-large-single_device-full-inference:


tests/runner/test_config/torch/test_config_inference_single_device.yaml
```diff
@@ -515,8 +511,8 @@ test_config:
   vit/pytorch-base-single_device-full-inference:
     status: EXPECTED_PASSING
     filechecks:
-      - concatenate_heads.ttnn.mlir
-      - split_query_key_value_and_split_heads.ttnn.mlir
+    - concatenate_heads.ttnn.mlir
+    - split_query_key_value_and_split_heads.ttnn.mlir
```
jameszianxuTT: Why remove the indent? Is this valid YAML?

PR #2638, opened 2025-12-19
Add test case and infra for supporting priority to sharding

sshonTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/2463#issuecomment-3665964705

### Problem description
When sharding conflict is anticipating, there is no way for user to choose which dimension should be propagated, and let the Shardy make decision.

### What's changed
It introduces user defined sharding priority when it expect to be conflicted.
It already supports from torch-xla and tt-mlir.

### Checklist


sshonTT:
Leave it to draft until torch-xla and tt-mlir change is merged.

AleksKnezevic (approved):
As long as we can upstream changes, lgtm

PR #2730, opened 2026-01-02
Show deprecated logger level warning only if new env variable is not set

mmanzoorTT:
### Problem description
If both old and new logging env variables are set then logger is disabled due to old/deprecated env variable.

### What's changed
Enable logging if new env variable is set and ignore the deprecated env variable.


mmanzoorTT:
For background... I had set old `LOGGER_LEVEL` env variable and then set new `TTXLA_LOGGER_LEVEL` env variable without unsetting the old one. I had to investigate why logging is not working.

Issue #2081, opened 2025-11-10
[Torch] Make experimental PyTorch backend to be default one


nvukobratTT:
## Summary
This change: https://github.com/tenstorrent/tt-xla/pull/2029 introduced the "experimental" PyTorch backend that is focused on being more aligned with the original "openxla" introduced by torch-xla (more details are explained in the PR description).

## Overview
The main issue is that our custom backend ("tt") is completely detached from the "openxla" one that is already supported by PyTorch. One of the issues that has arisen recently was runtime overhead due to the way we're compiling our models through the "tt" backend. 

There are multiple steps to cover and to experiment with, but for a starting point, we should check and iterate over implementations of the few existing backends:
- `openxla` - `pytorch/pytorch/torch/_dynamo/backends/torchxla.py`
- `xla` - `pytorch/xla/torch_xla/distributed/xla_backend.py`

Useful resources for more details about custom backends:
- `pytorch/pytorch/docs/source/torch.compiler_custom_backends.md`

---

_Note: This issue contains a problem overview; more specific tasks/features will be tracked through the sub-issues._

PR #2734, opened 2026-01-05
Torch experimental backend: Constant handling

### LINKED ISSUE #2820: [Experimental Compile Torch] Model OOM
Gemma 2 9B tensor parallel model hits OOM errors exclusively with experimental compile

sgligorijevicTT:
### Ticket
part of broader efforts toward #2081 
Also piggybacking handling of CPU constants, closes #2820 

### Problem description
Lifted constants weren't being handled on the experimental compile path. This resulted in errors akin to `KeyError: "L['self'].model.lifted_tensor_0"`
Edit: Another issue is that with handling of lifted constants, they could have been on the CPU, which works but forces graph breaks

### What's changed
Added a case to handle constants.
Edit: Also force move everything not on xla to xla. For now we only observed it to be necessary for constants, for all non constants we allow the move but warn if it happens.

### Checklist

### DIFF
--- a/python_package/tt_torch/backend/backend.py
+++ b/python_package/tt_torch/backend/backend.py
@@ -136,7 +136,8 @@ class XLAExecutor:
             if arg.device.type != "xla":
                 if spec.kind != InputKind.CONSTANT_TENSOR:
                     print(
-                        f"Found an argument on non-XLA device which was not a lifted constant: {spec.target}.\n You might not have intended to pass a non XLA argument to TT compile. Force moving the argument to XLA."
+                        f"Found an argument on non-XLA device which was not a lifted constant: {spec.target}.\n"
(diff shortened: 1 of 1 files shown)


python_package/tt_torch/backend/backend.py
```diff
@@ -112,11 +112,15 @@ def _build_params_and_consts(self, ep: ExportedProgram) -> Tuple[torch.Tensor]:
 
         # Export keeps a state dict for lifted params/buffers.
         state = ep.state_dict
+        constants = ep.constants
 
         # Map from placeholder name -> tensor.
         total_args = tuple()
         encountered_user_input = False
         for spec in sig.input_specs:
+            if spec.kind == InputKind.CONSTANT_TENSOR:
```
nvukobratTT: How about adding support for other input types as well?

vkovinicTT (approved):

PR #2742, opened 2026-01-06
Add test_llms_decode_torch which discovers models with load_inputs_decode() and uses it for Decode-Only LLMs

kmabeeTT:
### Ticket
None

### Problem description
- We want to add decode tests for all of the LLMs we have running, but specifically for the multi chip LLMs

### What's changed
- Add RunPhase enum for future expansion with LLM_DECODE, LLM_PREFILL to start and report on it in junit xml, and use it to select func in model loader. Easy to expand to prefill-only tests in future.  Include DEFAULT enum values for existing tests which is slightly cleaner than None reported in junit xml.
- Separate test test_llms_decode_torch rather than bundling into existing test_all_models_torch is cleaner, avoids hacks, impacting existing test names. Just run on inference, single_device and tensor_parallel for now.
- Update TorchComparator equal, pcc, atol, allclose comparisons to ignore StaticCache objects as workaround for #2743 
     - Exposed by enabling use_cache=True in Decode-Only LLM model tests, this avoids "TypeError: equal(): argument 'input' (position 1) must be Tensor, not StaticCache"
     - In the future (after Transformers uplift settles which impacts things, we wish to de-risk) we will re-evluate checking StaticCache internals in these functions.
- This test relies on tt_forge_models changes to add load_inputs_decode() to several LLMs uplifted to tt-xla on Jan7.
- Not including these tests in CI to start. Would require new test_config files to track the status of these tests and which are expected passing for each arch. Will do in follow up PR.
- Fix unrelated op-by-op flow typo nodeid -> pytest_node_id in error message

### Checklist
- [x] Tested locally handful of models that are discovered with tt-forge-models uplift, passing

### DIFF
--- a/tests/infra/comparators/torch_comparator.py
+++ b/tests/infra/comparators/torch_comparator.py
@@ -14,6 +14,24 @@ from .comparison_config import AllcloseConfig, AtolConfig, ComparisonConfig, Pcc
 class TorchComparator(Comparator):
     """Comparator for Torch tensors/pytrees."""
 
+    # LLM decode tests may include transformers StaticCache objects in outputs (e.g., past_key_values).
+    # These are not torch.Tensors, so we detect them and treat matching StaticCache leaves as equal.
+    # TODO https://github.com/tenstorrent/tt-xla/issues/2743: Enable checking for allclose, pcc, atol, equal.
+    @staticmethod
+    def _is_static_cache(x: object) -> bool:
+        # Avoid importing transformers at module import time.
+        try:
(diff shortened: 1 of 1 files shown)


kmabeeTT:
From offline discussion with @jameszianxuTT was suggested to de-risk Transformers uplift by keeping StaticCache checking disabled here (change is included to do this) until dust settles after uplift.   I'll see if I can get checking working/enabled in meantime though and if it passes like we think it might.

AleksKnezevic (approved):

PR #2751, opened 2026-01-08
Exclude large RED language models that don't fit on single chip, run/report on n300-llmbox

kmabeeTT:
### Ticket
None

### Problem description
- Lots of clutter (SKIPPED tests reported as failed runtime) with tests known not to fit on single chip that need to be run on n300-llmbox or larger via tensor parallel. Would be great to not have to sift through and look at these results.

### What's changed
 - Go through results from latest nightly run-id 20766433803 and use supported_archs or EXCLUDE_MODEL in test_config yml as needed to target archs that they fit on
 - Many were already running p150 + n300-llmbox, but many were too big for n150 and p150 and need n300-llmbox or galaxy
 - Keep 7B/8B param models as SKIP for n150 since bfp8 weights will be tried soon and should allow them to fit.
- Enable PCC check for qwen_3/causal_lm/pytorch-8b-single_device-inference but drop threshold to 0.985. It's reporting 0.99001 lately, too close to 0.990 threshold

### Checklist
- [x] Visual check, test_config validation pass still runs.  Will include source list of models targeted used for this exercise in PR comment.

### DIFF
--- a/tests/runner/test_config/torch/test_config_inference_single_device.yaml
+++ b/tests/runner/test_config/torch/test_config_inference_single_device.yaml
@@ -1604,8 +1604,12 @@ test_config:
     bringup_status: FAILED_RUNTIME
 
   qwen_3/embedding/pytorch-embedding_8b-single_device-inference:
-    supported_archs: ["p150"] # Runs as tensor_parallel for wormhole.
     status: EXPECTED_PASSING
+    arch_overrides:
+      n150:
+        status: NOT_SUPPORTED_SKIP
+        reason: "Too large for single chip"
+        bringup_status: FAILED_RUNTIME
 
   gpt_neo/sequence_classification/pytorch-gpt_neo_125M-single_device-inference:
     status: EXPECTED_PASSING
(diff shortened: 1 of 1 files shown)


kmabeeTT:
> Thanks for this @kmabeeTT! Just wanted to check are we sure that all of these models cannot be ran even theoretically on one device? If they can, maybe it makes sense to leave them as skipped for now?

Np. A few (7 from above) are 7B/8B param models that could theoretically fit on n150 with bfp8 weights, but that is separate thing to look into which hasn't been started yet (getting all 7B/8B param models properly working with bfp8 weights and running on n150 like they do already on p150 with fp16 weights). On the fence about keeping those as SKIP for n150 or not.

ajakovljevicTT:
> Thx @ajakovljevicTT you're right, it's fair point - discussed with Aleks today about BFP8 weights topic, and it could be evaluated now/soon across all models, so I'll revert the changes to the 7B/8B models here that are good candidateas for BFP8 weights allowing them to fit on n150.

Great, thanks for following up on this! I'll approve this PR not to block you on merging when you revert the changes to the 7B/8B models

PR #2783, opened 2026-01-09
Add multi-input support for MultiChip JAX test infra

vzeljkovicTT:
This PR standardizes input handling across the test infrastructure for all NNX models (including EasyDel source models) by:

  - Wrapping all model kwarg inputs in dictionaries to standardize multi-input handling
  - Converting PartitionSpec entries to tuples for multichip models to enable proper input partitioning

Additionally, in this PR it is included adding recursive mesh updating for EasyDel multimodal models (e.g., CLIP) that have nested configs (text_config, vision_config)

### DIFF
--- a/tests/infra/workloads/jax_workload.py
+++ b/tests/infra/workloads/jax_workload.py
@@ -74,7 +74,7 @@ class JaxMultichipWorkload(Workload):
 
     def set_nnx_model_mesh_recursive(self, device_mesh):
         """Set mesh on EasyDL model config and recursively on sub-configs.
-        
+
         EasyDL models can have nested config objects (text_config, vision_config)
         that also need mesh assignment for proper distributed execution.
         """


tests/infra/runners/jax_device_runner.py
```diff
@@ -117,11 +117,14 @@ def _put_multichip_workload_on_device(
 
         kwargs_on_device = {}
         for key, arg in multichip_workload.kwargs.items():
-            device_arg = self._put_sharded_arg_on_multichip_device(
-                arg,
-                multichip_workload.device_mesh,
-                multichip_workload.in_specs[spec_index],
-            )
+            if key not in multichip_workload.static_argnames:
```
mrakitaTT: Add comment to explain why we are doing this only for non-static arguments, also it might make sense to write this code as `if key in multichip_workload.static_argnames ... else ...`.

vzeljkovicTT: I kept the if key not in static_argnames pattern for consistency with the single chip implementation (line ~86). Both use the same logic structure. Added a comment to explain why we handle non-static arguments differently. :)

PR #2786, opened 2026-01-09
[FX fusing] Add torch.FX fusing pass

mstojkovicTT:
### Ticket
fixes: #2788

### Problem description
Described in issue.

### Summary
- Add fusion pass module that detects patterns in `PyTorch FX graphs` and replaces them with single high-level operations
- Implement `RMSNorm` fusion pattern that matches `LlamaRMSNorm` module
- Provide `FusionProvider` base class with auto-registration via `__init_subclass__`

### What's changed
  - `tt_torch/fusion_passes/utils.py` - Core utilities: `FusionPattern` dataclass, `apply_fusion_pattern()`
  - `tt_torch/fusion_passes/providers.py` - `FusionProvider` + provider implementations (`RMSNormFusionProvider`)
  - `tt_torch/fusion_passes/__init__.py` - Public exports
  - `tt_torch/backend/passes.py` - Entry point `run_fusion_passes()`
  - `tests/torch/ops/test_fusion_ops.py` - E2E and Graph tests using transformers `LlamaRMSNorm`

### Adding new patterns
``` python
 class MyPatternFusionProvider(FusionProvider):
      """Add to providers.py - auto-registers via __init_subclass__."""

      @property
      def name(self) -> str:
          return "my_pattern"

      @staticmethod
      def pattern(x: Tensor, y: Tensor, dtype) -> Tensor:
          """Use method calls (.add, .mul) and parameters for wildcards."""
          return x.add(y).mul(2).to(dtype)  # dtype matches any dtype

      @staticmethod
      def replacement(x: Tensor, y: Tensor, dtype) -> Tensor:
          return torch.some_fused_op(x, y)
```

### DIFF
--- a/python_package/tt_torch/fusion_passes/providers.py
+++ b/python_package/tt_torch/fusion_passes/providers.py
@@ -27,7 +27,6 @@ class FusionProvider(ABC):
     2. Implement the `name` property
     3. Implement the `pattern` static method (the pattern to match)
     4. Implement the `replacement` static method (the replacement function)
-    5. Optionally override `get_patterns()` to customize pattern generation
     """
 
     _registered_providers: List[Type["FusionProvider"]] = []
@@ -60,7 +59,7 @@ class FusionProvider(ABC):
         pass
 
     def get_patterns(self) -> List[FusionPattern]:
-        """Return list of FusionPatterns. Override to customize (e.g. add match filters)."""
+        """Return list of FusionPatterns."""
         return [FusionPattern(self.name, self.pattern, self.replacement)]
 
 
--- a/python_package/tt_torch/fusion_passes/utils.py
+++ b/python_package/tt_torch/fusion_passes/utils.py
@@ -2,11 +2,11 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-from dataclasses import dataclass, field
-from typing import Callable, List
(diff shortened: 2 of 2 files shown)


tests/torch/ops/test_fusion_ops.py
```diff
@@ -0,0 +1,57 @@
...
+
+    This is the exact pattern from HuggingFace transformers LlamaRMSNorm.
+    """
+
+    def __init__(self, hidden_size, eps=1e-6):
+        super().__init__()
+        self.weight = nn.Parameter(torch.ones(hidden_size))
+        self.variance_epsilon = eps
+
+    def forward(self, hidden_states):
```
AleksKnezevic: Why not load the rms norm from huggingface? That way, we can catch if the pattern changes and our re-writer needs to as well. Also, are you actually asserting that your pattern is hit?

mstojkovicTT: Yeah, importing from `huggingface` makes much more sense for reasons you have said, I am just so used to writing tests with my class that i missed that... For the testing, I was thinking about how to test that as well, i pushed initial version of how that might look like, but will take some more time to think about the best way that this can be done, if you have some ideas / examples of how this should be done in a more clean way, it will be more then welcome :))

PR #2827, opened 2026-01-14
On PR test trigger changes and labels changes

vvukomanTT:
### Ticket
#804 

### Problem description
1. Performance Benchmark tests are getting triggered on tt-forge-models -> tt-xla uplifts, we don't want that anymore
2. Performance Benchmark tests failing in the On PR workflow wasn't failing the workflow, so fails are silent

### What's changed
1. Performance Benchmark tests are now getting triggered only on tt-mlir -> tt-xla uplifts
2. Performance Benchmark tests failing in the On PR workflow fail the entire workflow
3. We are running only a subset of Performance Benchmark tests in the tt-mlir -> tt-xla uplift PRs

Additional bug I found:
4. Subset of model tests that are ran on uplifts (test_models_on_uplift) could silently fail, fixed it

### Checklist

### DIFF
--- a/.github/workflows/manual-nightly-release.yml
+++ b/.github/workflows/manual-nightly-release.yml
@@ -41,7 +41,7 @@ jobs:
 
   build-release-docker:
     uses: ./.github/workflows/call-build-release-docker.yml
-    needs: [ build-ttxla ]
+    needs: [ publish-release-wheel, build-ttxla ]
     secrets: inherit
     with:
       wheel_release_artifact_run_id: ${{ needs.build-ttxla.outputs.artifacts_run_id }}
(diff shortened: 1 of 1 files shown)


nsumrakTT (approved):

.github/workflows/schedule-uplift.yml
```diff
@@ -83,7 +83,7 @@ jobs:
           commit-message: "Uplift ${{ env.TT_MLIR_SUBMODULE_PATH }} to ${{ env.LATEST_TT_MLIR_VERSION }} ${{ env.TODAY }}"
           title: "Uplift ${{ env.TT_MLIR_SUBMODULE_PATH }} to ${{ env.LATEST_TT_MLIR_VERSION }} ${{ env.TODAY }}"
           body: "This PR uplifts the ${{ env.TT_MLIR_SUBMODULE_PATH }} to the ${{ env.LATEST_TT_MLIR_VERSION }}"
-          labels: uplift
+          labels: uplift-mlir
```
nsumrakTT: @ajakovljevicTT please check if this is ok for uplift process.

acicovicTT:
Can we hold off on the merging of this PR, or remove the perf benchmarks as a requirement in a PR on its own? We are yet to discuss whether we want to block PRs on benchmarks or not

vkovacevicTT:
@ajakovljevicTT @vvukomanTT 
> I'm not sure we should be failing our onPRs with benchmark tests.

Agreed. Perf benchmark should not block uplifts since perf tests are still failing very often.

> If these perf tests won't fail the uplift PR itself, I wouldn't even run them at all.
Nigthly perf tests in tt-forge will catch the regressions from tt-mlir -> tt-xla uplifts and it will be easy to pinpoint the commit from a one day timespan.

This also makes sense.

In any case, we should consider switching to Shared runners for perf regression checks as mentioned [here](https://github.com/tenstorrent/tt-forge/issues/804#issuecomment-3750588052).

.github/workflows/schedule-uplift.yml
kmabeeTT: (makes sense to me, differentiates between mlir uplift and tt-forge-models uplift as described)

vmilosevic (approved):

Issue #2857, opened 2026-01-16
Reports don't get generated when process dies


pglusacTT:
Currently, when running model tests and the process dies, reports do not get generated and put into superset.

PR #2858, opened 2026-01-16
Fix backward in backend

### LINKED ISSUE #1391: Enable using torch.compile for training workloads.
With compile for inference we get:
```
E       RuntimeError: Trying to backward through prims.convert_element_type.default but no autograd formula was registered. Please use register_autograd to add one.
```

pglusacTT:
### Ticket
Closes #1391

### Problem description
Backward pass was not working with `torch.compile`.

### What's changed
- Register backwards for `mark_argument_attributes` and  `sharding_constraint`.
- Remove `aten._to_copy` decomposition that breaks backward.

### DIFF
--- a/tests/torch/training/test_basic.py
+++ b/tests/torch/training/test_basic.py
@@ -2,8 +2,13 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 import pytest
+import torch
 import torch_xla
+import torch_xla.core.xla_model as xm
+import torch_xla.distributed.spmd as xs
 import torch_xla.runtime as xr
+from infra.utilities.torch_multichip_utils import enable_spmd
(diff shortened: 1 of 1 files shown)


pglusacTT:
For testing this functionality we should determine how many models should be tested in eager vs compile mode.
I've ran the weekly workflow but there is an issue which prevents us from seeing how many models work in compile mode, which will influence this decision (if there is parity maybe we should fully switch to compile).

Once #2857 is fixed, I'll rerun it and add tests.

@vladimirjovanovicTT what do you think?

mstojkovicTT (approved):
Looks good, thanks for the effort to make compile work with training!

ndrakulicTT (approved):
Lets just add simple training on n300 to check if everything works for multichip other then that LGTM

Issue #2846, opened 2026-01-16
"Static CBs grow ... is beyond max L1 size" in new n300-llmbox TP test_models.py::test_llms_torch decode-only tests Jan 15


kmabeeTT:
A new fail needs bisect/triage/debug.

Kind of annoying since these are new tests I'm trying to add via https://github.com/tenstorrent/tt-xla/pull/2809 (I suspect this will show up in existing nightly in tt-xla tonight too), but this error showed up in branch CI run here: https://github.com/tenstorrent/tt-xla/actions/runs/21047142550/job/60526585716 based on commit 063472cab , previously passed few hours earlier in branch CI run here: https://github.com/tenstorrent/tt-xla/actions/runs/21041261101 based on commit be941780b.  Just these commits in window between passing base and failing commit.

```
2026-01-15 (063472cab) by GitHub (Author mstojkovic@tenstorrent.com) : [FX fusing] Add torch.FX fusing pass (#2786)
2026-01-15 (b5b524b9b) by GitHub (Author ppadjin@tenstorrent.com) : Adding option to put vae and clip on device (#2832)
2026-01-15 (efaac38c7) by GitHub (Author 160603639+odjuricicTT@users.noreply.github.com) : Add optimization_level to TTConfig for vLLM compile (#2663)
2026-01-15 (e3d95418b) by GitHub (Author ajakovljevic@tenstorrent.com) : Uplift third_party/tt-mlir to cb33dc71f08c218deee35bdf69443b17bab848b7 2026-01-15 (#2758)
```

Error and affected tests:

```
TT_THROW: Statically allocated circular buffers on core range [(x=0,y=0) - (x=0,y=0)] grow to 1511680 B which is beyond max L1 size of 1499136 B (assert.hpp:104)

FAILED tests/runner/test_models.py::test_llms_torch[qwen_3/causal_lm/pytorch-14b-llm_decode-tensor_parallel-inference] - RuntimeError: Bad StatusOr access: INTERNAL: Error code: 13
FAILED tests/runner/test_models.py::test_llms_torch[qwen_3/causal_lm/pytorch-32b-llm_decode-tensor_parallel-inference] - RuntimeError: Bad StatusOr access: INTERNAL: Error code: 13
[...]

PR #2865, opened 2026-01-16
[FX Fusing] Turn off by default

mstojkovicTT:
### Ticket
#2846

### Problem description
- Many TP n300-llmbox red models hitting "TT_THROW: Statically allocated circular buffers on core range [(x=0,y=0) - (x=0,y=0)] grow to 1511680 B which is beyond max L1 size of 1499136 B (assert.hpp:104)" in latest nightly, bisected to #2786 changes pretty quickly, needs debug.

### What's changed
Turn off `fx fusing` by default until stable for multichip

### DIFF
--- a/python_package/tt_torch/backend/backend.py
+++ b/python_package/tt_torch/backend/backend.py
@@ -35,7 +35,7 @@ def torch_pass_pipeline(
 
     # Run fusion passes to detect and fuse multi-op patterns
     # This runs before composite_ops to allow fused patterns to be wrapped as composites
-    enable_fusion_passes = options is None or options.get(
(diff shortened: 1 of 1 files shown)


AleksKnezevic (approved):

kmabeeTT:
Been testing this, unfortunately it doesn't help to solve the L1 OOM failure.   Can we revert change instead?

kmabeeTT:
@nvukobratTT - push tests are improved recently, but regrettably we do not have n300-llmbox TP tests in push/pr, so it is not caught, mainly due to lack of machines (n300-llmbox in other jobs are pending quite a lot lately)... I agree though, would like to add, so I will go ahead and add one job.

Issue #2892, opened 2026-01-19
Python aborts when generating pytest report


ndrakulicTT:
It is possible for the `tt-mlir` runtime to crash during Torch autograd execution (a scenario that was expected). When this occurs, if the last Python function in the traceback (specifically `TorchFunctionOverride::__torch_function__`) contains Torch XLA tensors as arguments, pytest attempts to generate a string representation (`repr`) of these inputs for the failure report. This action triggers a call to `toHost`, which subsequently causes the process to segfault.

Issue #2814, opened 2026-01-13
xr.clear_computation_cache() error on test teardown if test fails in mlir runtime


ndrakulicTT:
### Describe the bug

In [nightly](https://github.com/tenstorrent/tt-xla/actions/runs/20939830171/job/60199528172) and [training weekly](https://github.com/tenstorrent/tt-xla/actions/runs/20875338451), there were errors for xfailed models, caused by `xr.clear_computation_cache()` on test teardown. It only appears if the the test has metal error in mlir runtime

 
### Steps to reproduce the issue

In latest main run for example: `pytest -svv tests/runner/test_models.py::test_all_models_torch[dla/pytorch-dla60x_c-single_device-training]`
or for small repro run:
```py
import torch_xla
import torch
import torch_xla.runtime as xr
import torch_xla.core.xla_model as xm
import time
xr.set_device_type("TT")

class Model(torch.nn.Module):
    def forward(self, x, y):
        return x @ y

device = torch_xla.device()
model = Model().to(device)
model.compile(backend="tt")

def test_oom(N, M):
    device = torch_xla.device()
    a = torch.randn(N, M).to(device)
    b = torch.randn(M, N).to(device)
    
    c = model(a, b)
    # torch_xla.sync()

print("Test 1")
test_oom(2**13, 4) # This should pass

try:
    print("Test 2")
    test_oom(2**16, 4) # This should fail
except Exception as e:
    print(e)

# Sleep for 10 seconds to allow the program before to be executed on the device
time.sleep(10)

# Running anything after catching exception will fail
print("Test 3")
test_oom(2**13, 4) # This should pass, 
# xr.clear_computation_cache()
```

Error:
```
2026-01-13 10:05:41.992 | critical |          Always | TT_FATAL: Out of Memory: Not enough space to allocate 17179869184 B DRAM buffer across 12 banks, where each bank needs to store 1431658496 B, but bank size is only 1073741792 B (assert.hpp:103)
2026-01-13 10:05:42.025 (   1.001s) [        ECFF9640]                utils.h:62     ERR| Exception:
{TT_FATAL @ /localdev/ndrakulic/tt-xla/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/tt_metal/impl/allocator/bank_manager.cpp:429: address.has_value()
info:
[...]

PR #2896, opened 2026-01-19
Skip timeout tests and apply torch_xla.sync workarounds

ndrakulicTT:
### Ticket

Skip tests affected by tt-mlir#6563
Workarounds for  #2892 and #2814

### What's changed

- Remove group of training weekly large tests as there is none currently
- revert `wait=False` in training tests and try catch sync as workaround for #2814
- add tb=short as workaround for #2892
- skip tests that take too long on eio

### Checklist

* [x] New/Existing tests provide coverage for changes


tests/conftest.py
```diff
@@ -488,6 +489,12 @@ def clear_torchxla_computation_cache():
     This helps avoid consteval-associated DRAM leaks as described in https://github.com/tenstorrent/tt-xla/issues/1940
     """
     yield
+    try:
+        torch_xla.sync(wait=True)
+    except Exception as e:
```
ajakovljevicTT: Can we also print out what the error was?

Issue #2961, opened 2026-01-23
Need for unified logger system in python


mstojkovicTT:
### Is your feature request related to a problem? Please describe

Currently, we have logger system for cpp part of the repo (pjrt plugin) using `TTXLA_LOGGER_LEVEL` env var, but we dont have anything for `python_package` or python code as a whole. 

While looking trough the repo, I also found that `vllm` has its own [logger](https://github.com/tenstorrent/tt-xla/blob/ef48658b03c4f3c42cea9f6dc38118c53b1d8b68/integrations/vllm_plugin/vllm_tt/logger.py#L10). 

I think not having proper unified (or in this case any) logging system is a terrible user experience, and we should make efforts to change this.

PR #2968, opened 2026-01-26
Add logging to python part of the repo

mstojkovicTT:
### Ticket
fixes: #2961

### Problem description
Described in issue.

### What's changed
  - Add centralized logging configuration module (`ttxla_tools/logging.py`) for consistent logging across all `tt-xla` Python packages 

Edit: upload example of `TTXLA_LOGGER_LEVEL=DEBUG` for `phi2` model                                                                                        
[debug.log](https://github.com/user-attachments/files/24940651/debug.log)

### DIFF
--- a/tests/infra/testers/single_chip/model/jax_model_tester.py
+++ b/tests/infra/testers/single_chip/model/jax_model_tester.py
@@ -21,7 +21,6 @@ from infra.utilities import (
     random_tensor,
 )
 from infra.workloads import Workload
-from loguru import logger
 from transformers.modeling_flax_utils import FlaxPreTrainedModel
 
 from tests.infra.testers.compiler_config import CompilerConfig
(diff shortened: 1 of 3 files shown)


mstojkovicTT:
> Is there any background on the choice of the standard logging lib vs loguru or other logging packages?

@jameszianxuTT There is not much of a difference, besides the colored prints out of the box. Main reason i was for logging was because of the similarity and not having another dependency for the repo, but taking into consideration that we already have it in env, and that we use it in cpp, I think loguru is better choice now. 

Also, i saw that we have custom `VERBOSE` level in cpp, I wont change that in this PR, but because we diverge there (python has `TRACE` that is actually logically same as what we use `VERBOSE` for in cpp), maybe we should change the cpp `VERBOSE` to `TRACE` simply because i think devs are more familiar with `TRACE` for logging.

ppadjinTT (approved):

jameszianxuTT (approved):

jameszianxuTT:
>Also, i saw that we have custom VERBOSE level in cpp, I wont change that in this PR, but because we diverge there (python has TRACE that is actually logically same as what we use VERBOSE for in cpp), maybe we should change the cpp VERBOSE to TRACE simply because i think devs are more familiar with TRACE for logging.

I added that originally, I have no problem changing from VERBOSE to TRACE; it was an arbitrary choice

PR #2993, opened 2026-01-27
Implemented pjrt tensor

acolicTT:
Since PJRT does not operate on tensors but on BufferInstances (where each BufferInstance represent single tensor shard), we will use our own tensor abstraction.

Each shard will hold reference to pjrt tensor that is part of.

Tensor can be constructed from shards or from an existing tensor.

Whenever tensor is constructed, it is moved into tensor pool and whenever it is destructed, it is moved out.

This abstraction gives us control over host and device tensors and hides complexity behind simple APIs.

### DIFF
--- a/pjrt_implementation/inc/api/buffer_instance.h
+++ b/pjrt_implementation/inc/api/buffer_instance.h
@@ -155,8 +155,8 @@ public:
     m_pjrt_tensor.reset(std::move(pjrt_tensor), this);
   }
 
-  PjrtTensorRef &pjrtTensor() { return m_pjrt_tensor; };
-  const PjrtTensorRef &pjrtTensor() const { return m_pjrt_tensor; };
+  PjrtTensorRef &getPjrtTensor() { return m_pjrt_tensor; };
+  const PjrtTensorRef &getPjrtTensor() const { return m_pjrt_tensor; };
 
   const tt::runtime::Tensor &runtimeTensor() const {
     return m_pjrt_tensor->runtime_tensor();
--- a/pjrt_implementation/inc/api/tensor.h
+++ b/pjrt_implementation/inc/api/tensor.h
@@ -40,7 +40,8 @@ class BufferInstance;
 // This abstraction gives us control over host and device tensors and hides
 // complexity behind simple APIs.
 class PjrtTensor {
-  // Prevents direct construction. Use PjrtTensor::init instead.
+  // Prevents direct construction. Use static functions below for
+  // initialization.
   struct Private {
     explicit Private() = default;
   };
@@ -111,7 +112,7 @@ private:
 
 private:
   // Tensor unique identifier. For now, used for debug only.
-  const uint64_t m_uid{next_uid()};
+  const uint64_t m_uid;
 
   // Tensor shards. Each shard hold pjrt tensor reference to this pjrt tensor.
   std::vector<BufferInstance *> m_shards;
@@ -135,7 +136,7 @@ private:
 // void test_function(const BufferInstance* buf) {
 //     // This will be allowed if we store std::shared_ptr<PjrtTensor> in
 //     // BufferInstance.
-//     uf->pjrtTensor()->move_to_host();
+//     buf->pjrtTensor()->move_to_host();
 // }
 //
 // With this simple wrapper, we have tied 'const'ness of BufferInstance and
--- a/pjrt_implementation/src/api/buffer_instance.cc
+++ b/pjrt_implementation/src/api/buffer_instance.cc
@@ -233,7 +233,7 @@ void BufferInstance::copyFromHost(
 
 void BufferInstance::copyFromBuffer(BufferInstance *src_buffer) {
   DLOG_F(LOG_DEBUG, "BufferInstance::copyFromBuffer");
-  assert(src_buffer->pjrtTensor() && "Source buffer has no data.");
+  assert(src_buffer->getPjrtTensor() && "Source buffer has no data.");
 
   ::tt::target::DataType runtime_data_type =
       tt::pjrt::data_type_utils::convertPJRTToRuntimeDataType(
@@ -249,7 +249,7 @@ void BufferInstance::copyFromBuffer(BufferInstance *src_buffer) {
   tt::runtime::Tensor runtime_tensor = tt::runtime::createOwnedHostTensor(
       /* data= */ nullptr, shape, strides, element_size, runtime_data_type);
 
-  src_buffer->pjrtTensor()->move_to_host();
+  src_buffer->getPjrtTensor()->move_to_host();
   tt::runtime::memcpy(runtime_tensor, src_buffer->runtimeTensor());
 
   PjrtTensor::from_runtime_tensor({this}, std::move(runtime_tensor));
--- a/pjrt_implementation/src/api/tensor.cc
+++ b/pjrt_implementation/src/api/tensor.cc
@@ -64,7 +64,8 @@ PjrtTensor::from_runtime_tensor(std::vector<BufferInstance *> shards,
 
 PjrtTensor::PjrtTensor(Private, std::vector<BufferInstance *> shards,
                        tt::runtime::Tensor rt_tensor)
-    : m_shards{std::move(shards)}, m_runtime_tensor{std::move(rt_tensor)} {
+    : m_uid{next_uid()}, m_shards{std::move(shards)},
+      m_runtime_tensor{std::move(rt_tensor)} {
 
   TensorPool::insert(this);
 }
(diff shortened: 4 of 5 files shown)


pjrt_implementation/src/api/tensor.cc
```diff
@@ -0,0 +1,304 @@
...
+  m_tensors.erase(tensor);
+}
+
+void PjrtTensorPool::clear() {
+
+  const std::scoped_lock lock{m_mtx};
+  m_tensors.clear();
+}
+
+// Moves all tenors to host.
```
acicovicTT: nit: tensors

pjrt_implementation/inc/api/tensor.h
```diff
@@ -0,0 +1,244 @@
...
+    return m_tensor.operator->();
+  }
+  PjrtTensor *operator->() noexcept { return m_tensor.operator->(); }
+
+  const PjrtTensor &operator*() const noexcept { return *m_tensor; };
+  PjrtTensor &operator*() noexcept { return *m_tensor; };
+
+private:
+  std::shared_ptr<PjrtTensor> m_tensor;
+  BufferInstance *m_shard{nullptr};
```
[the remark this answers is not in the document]
acicovicTT: > Shard is part of one tensor, right?

When I initially was reviewing this change I thought that may not be true since I assumed that the purpose of this class was to link each shard with each tensor, but that is not the case, so you can disregard my initial question.

When you say automatic memory management, do you mean automatic destruction of `PjrtTensor` when last `PjrtTensorRef` is destroyed?

pjrt_implementation/inc/api/tensor.h
```diff
@@ -0,0 +1,242 @@
...
+       const tt::runtime::Device &device,
+       const std::optional<const tt::runtime::Layout> &layout,
+       const std::vector<std::uint32_t> &mesh_shape,
+       const std::unordered_map<std::string, std::string> &strategy);
+
+  // Initializes tensor from an existing device tensor.
+  static PjrtTensor &init(std::vector<BufferInstance *> shards,
+                          tt::runtime::Tensor device_tensor);
+
+public: // Constructors needs to be public for std::shared_ptr.
```
ajakovljevicTT: Could we use a concept similar to `make_unique_enabler` that we already do at some points?

pjrt_implementation/inc/api/tensor.h
```diff
@@ -0,0 +1,242 @@
...
+    return std::find(m_shards.begin(), m_shards.end(), shard) != m_shards.end();
+  }
+
+  void remove_shard(const BufferInstance *shard) noexcept;
+
+  void move_to_host() noexcept;
+
+private:
+  static PjrtTensor &
+  init_from_existing(const std::vector<BufferInstance *> &shards,
```
ajakovljevicTT: nit: `initialize`

pjrt_implementation/inc/api/tensor.h
```diff
@@ -0,0 +1,242 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
```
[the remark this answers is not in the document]
acolicTT: High, high level nit: I think that it would be more readable if we would delete both src, inc, src/api, inc/api folders and have all of our 10 files on a single place. We are following google approaches, which might sometimes (but I am not convinced) make sense for a billion lines c++ projects. Here, it is just a waist of time and mental effort.

Look at the git repo: https://github.com/git/git
If they made it, I think that we can make it too.

We are not shipping libs, there is no point in API and implementation separation, we are calling 20 functions using pjrt which does everything for us and we should stop complicating everything.

pjrt_implementation/inc/api/tensor.h
acolicTT: Looks like we can remove most of these make_unique_enablers, because we just giving memory to PJRT, we are not deleting it, so I am not sure why are we making everything as a unique_ptr. For example, every event instance is released after it is created. Every BufferInstance too. 
What do you think about deleting those, and the rest that should be handled by us, we can move to this implementation?

pjrt_implementation/inc/api/buffer_instance.h
```diff
@@ -175,25 +151,40 @@ class BufferInstance {
...
+
+  const tt::runtime::Tensor &runtimeTensor() const {
+    assert(m_pjrt_tensor && "Accessing non-existing PJRT tensor");
+    return m_pjrt_tensor->runtime_tensor();
+  }
+
+  tt::runtime::Tensor &runtimeTensor() {
+    assert(m_pjrt_tensor && "Accessing non-existing PJRT tensor");
+    return m_pjrt_tensor->runtime_tensor();
+  }
```
pilkicTT: these assertions should probably go to the `operator->()` of the `PjrtTensorRef`?

pjrt_implementation/inc/api/tensor.h
```diff
@@ -0,0 +1,228 @@
...
+class PjrtTensorPool {
+public:
+  void insert(PjrtTensor *tensor);
+  void erase(PjrtTensor *tensor);
+  void clear();
+  void move_tensors_to_host();
+  bool contains(PjrtTensor *tensor);
+
+private:
+  std::mutex m_mtx;
```
pilkicTT: if you want const apis, you can use `mutable` here

pilkicTT (approved):

pjrt_implementation/src/api/buffer_instance.cc
```diff
@@ -365,109 +301,41 @@ std::vector<std::uint32_t> BufferInstance::calculateStrides(
...
+
+  // TODO(acolic): Copying in a separate thread is left to match previous
+  // behavior. Check whether it is needed: it does not make much sence to
+  // create new thread for each shard, because tensors are moved once from
+  // device when onHost is called on a first shard; on the other hand, there is
+  // no sence to create new thread for memcpy because framework would wait on a
+  // copy anyway, right? Also, creating std::thread for memcpy might be overhead
+  // with performance loss. We must measure.
+  // Also, std::thread (as all objects from std::) has a value semantic, so it
+  // does not make any sence to create std::thread as a unique_ptr.
```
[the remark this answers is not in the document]
acolicTT: Makes sence :D

pjrt_implementation/src/api/flatbuffer_loaded_executable_instance.cc
```diff
@@ -56,110 +57,32 @@ FlatbufferLoadedExecutableInstance::prepareInputTensor(
...
-
-  return tt::runtime::toLayout(input_tensor, runtime_device, layout,
-                               tt::runtime::getTensorRetain(input_tensor));
+  return tensor.runtime_tensor();
 }
 
 void FlatbufferLoadedExecutableInstance::fillPJRTOutputLists(
-    const std::vector<tt::runtime::Tensor> &output_tensors, size_t num_devices,
+    const std::vector<tt::runtime::Tensor> &output_tensors,
+    const tt::runtime::Device &device, size_t num_devices,
```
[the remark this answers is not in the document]
acolicTT: Yes, I forgot to delete it.

pjrt_implementation/inc/api/tensor.h
```diff
@@ -0,0 +1,242 @@
...
+    return has_layout(m_runtime_tensor, layout);
+  }
+
+  void relay(const tt::runtime::Device &device,
+             const tt::runtime::Layout &layout) {
+    relay(m_runtime_tensor, device, layout);
+  }
+
+private: // members
+  const uint64_t m_uid{nextUID()};
```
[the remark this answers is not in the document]
mrakitaTT: I personally prefer when I can see in one place how each class member is initialized, instead of going back and forth between class constructor and header. I also think it can avoid bugs with double initialization or when specific initialization ordering is necessary. So I would recommend avoiding default initialization in the header and moving it into constructor.

pjrt_implementation/inc/api/tensor.h
```diff
@@ -0,0 +1,205 @@
...
+// which ensures that non-const methods from pjrt tensor (which might modify
+// this object) are only called on non-const this object.
+// This is needed because if we are calling non-const pjrt tensor method, it
+// could modify us, which we don't want.
+//
+// Example:
+// void test_function(const BufferInstance* buf) {
+//     // This will be allowed if we store std::shared_ptr<PjrtTensor> in
+//     // BufferInstance.
+//     uf->pjrtTensor()->move_to_host();
```
mrakitaTT: buf*

pjrt_implementation/src/api/buffer_instance.cc
```diff
@@ -365,109 +301,41 @@ std::vector<std::uint32_t> BufferInstance::calculateStrides(
...
-  }
+  auto rt_data_type =
+      tt::pjrt::data_type_utils::convertPJRTToRuntimeDataType(m_data_type);
+
+  // TODO(acolic): Copying in a separate thread is left to match previous
+  // behavior. Check whether it is needed: it does not make much sense to
+  // create new thread for each shard, because tensors are moved once from
+  // device when onHost is called on a first shard; on the other hand, there is
+  // no sense to create new thread for memcpy because framework would wait on a
+  // copy anyway, right? Also, creating std::thread for memcpy might be overhead
```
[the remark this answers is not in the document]
jameszianxuTT: - torch-xla framework may trigger multiple concurrent copyToHost, for example it copies all outputs to host at once in a parallelFor. We must serialize this since untilize to host is not thread safe from metal side and induce a bad data race. I added some unit tests to cover this behaviour already
- You are right that it doesn't make sense to create new threads for each shard being copied to host, this was a known limitation at the time since we had no way of copying a single shard to host

jameszianxuTT (approved):
One nit, can be ignored for now. Thanks for making this change, this is a nice abstraction

Issue #2995, opened 2026-01-27
Experimental compile required a hack for BEiT, also regressed PCC


sgligorijevicTT:
Without any hacks, BEiT models fail with experimental compile with a cryptic error suggesting an unsqeeze at index 1 was attempted on a scalar(where valid indices for unsqueeze are 0 and -1).
This was rootcaused to `_to_functional_tensor(scalarvalue)` object making it into the fx graph, where the `_to_copy` node which otherwise also does resizing fails to resize it, effectively eating up one dimension of the input causing the unsqueeze to fail.

This was worked around by decomposing `_to_copy` to a form that explicitly expands the operand when needed.

After this, device mismatch type errors were encountered(seems to be related to forming indices for gather like operations inside a relative position bias block within attention, code mentions that it's weird and there just to make compiles work), those were worked around by added an extra pass in a specific location to move all creation ops to XLA, adding yet another layer of forcing everything to XLA and nothing to CPU.

Ultimately this results in beit base regressing PCC from >0.99 to <0.3. Large seems unaffected.

Quick and dirty investigations hint that the source of the faulty op sequence is a decomposition of `select_scatter`, avoiding decomposing it on the aten level should allow us to at least remove the need for the first hack from this issue. To be investigated at one point.

Issue #2994, opened 2026-01-27
Experimental compile causes YOLO v6 data parallel tests to fail


sgligorijevicTT:
Previously passing data parallel tests for YOLOv6 (all 4 variants) now fail. After turning on debug logs, I can see the image is compiled for a 1 device mesh while executing on 2 device mesh, which faults.

PR #2996, opened 2026-01-27
Experimental compile on by default

### LINKED ISSUE #2081: [Torch] Make experimental PyTorch backend to be default one

### LINKED ISSUE #2087: Graph tests focused on our backend changes
Once we autline the diff https://github.com/tenstorrent/tt-xla/issues/2086, we should focus on enabling graph testing specific to our backend (i.e., graph tracing, in-place updates, bwd generation, etc.). E.g., for different passes we're introducing, and different ones we're missing, we should create a set of tests that will:
- Over time, confirm functionality
- Outline missing steps we should introduce (xfails)
- Etc.

sgligorijevicTT:
Heads up, despite the small diff, this is a big change.

### Ticket
closes #2081 closes #2087

### Problem description
Our current XLAExecutor performed fx->XLA op by op dispatch every execution which introduces high python overhead. This needs to be fixed.

### What's changed
Turns on experimental compile on by default
Adds needed hacks to make BEiT pass, this still results in BEiT base regressing on PCC quite severely https://github.com/tenstorrent/tt-xla/issues/2995
Marks YOLOv6 data parallel tests as failing https://github.com/tenstorrent/tt-xla/issues/2994
Adds tests to ensure experimental compile indeed avoids overhead, and that legacy compile indeed runs into the overhead
Adds tests to ensure mutation is handled correctly(this is not actually specific to our compile, but it was mentioned as one thing that might possibly warrant testing therefore it's hitching a ride on this PR)
Adds an xfailing test to confirm we indeed don't use AOTAutograd and do break if a backward over compile is attempted. In agreement with training team, adding AOTAutograd is the next step.

Heads up, I also observed vgg models failing on *some* CI runs, but I couldn't reproduce it locally so not failing anything here

Also, reviewers, read the issues linked in the "what's changed" to understand some of the changes

### Checklist

### DIFF
--- a/python_package/tt_torch/codegen.py
+++ b/python_package/tt_torch/codegen.py
@@ -34,7 +34,10 @@ def codegen_py(
     }
     torch_xla.set_custom_compile_options(real_compile_options)
     device = xm.xla_device()
-    model.compile(backend="tt")
+    # Using legacy compile is a temporary hack to 1) Make MetaDataProp work 2) Decrease odds of codegenning for graphs that never need to get executed.
+    # New compile serves to primarily reduce execution overhead, which we don't really care about if executing only once for codegen.
+    # TODO(sgligorijevic): Clean this up
+    model.compile(backend="tt", options={"tt_legacy_compile": True})
     model = model.to(device)
     args = [arg.to(device) for arg in args if isinstance(arg, torch.Tensor)]
     kwargs = {k: v.to(device) for k, v in kwargs.items() if isinstance(v, torch.Tensor)}
@@ -58,7 +61,10 @@ def codegen_cpp(
     }
     torch_xla.set_custom_compile_options(real_compile_options)
     device = xm.xla_device()
-    model.compile(backend="tt")
+    # Using legacy compile is a temporary hack to 1) Make MetaDataProp work 2) Decrease odds of codegenning for graphs that never need to get executed.
+    # New compile serves to primarily reduce execution overhead, which we don't really care about if executing only once for codegen.
+    # TODO(sgligorijevic): Clean this up
+    model.compile(backend="tt", options={"tt_legacy_compile": True})
     model = model.to(device)
     args = [arg.to(device) for arg in args if isinstance(arg, torch.Tensor)]
     kwargs = {k: v.to(device) for k, v in kwargs.items() if isinstance(v, torch.Tensor)}


python_package/tt_torch/backend/passes.py
```diff
@@ -228,3 +228,22 @@ def bypass_dtype_promotion_and_redundant_cast(gm, example_inputs):
         gm = bypass_dtype_promotion_and_redundant_cast(gm, example_inputs)
 
     return gm
+
+
+def replace_cpu_device_with_xla(gm):
```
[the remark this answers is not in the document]
ndrakulicTT: Can this model be compiler for example for cuda? If not I would suggest that we just disable torch.compile for that model, so we dont introduce workarounds for models that are not meant to be compiled

python_package/tt_torch/backend/backend.py
```diff
@@ -213,7 +218,14 @@ def xla_backend(gm, example_inputs, options=None):
     module, graph_signature, node_info = torch_pass_pipeline(
         gm, example_inputs, options
     )
-    experimental_compile_enabled = (
-        options.get("tt_experimental_compile", False) if options else False
-    )
-    return XLAExecutor(module, graph_signature, node_info, experimental_compile_enabled)
+    legacy_compile_enabled = False
+    if options:
```
mstojkovicTT: nit: maybe we dont need two nested if statements, and instead can just do:

``` python
if (val := options.get("tt_experimental_compile")) is not None:
    print('Warning: Experimental compile is now the default...')
    legacy_compile_enabled = not bool(val)
```

mstojkovicTT: Also, please use `from ttxla_tools.logging import logger` instead of `print` here

PR #3041, opened 2026-01-29
Use uv for faster package installation and remove preinstalled Python packages

### LINKED ISSUE #3139: Use uv package manager in tt-xla
Part of: https://github.com/tenstorrent/tt-xla/issues/3138

Add support for the "uv" package manager in tt-xla.

vmilosevic:
### Ticket
Closes #3139

### Problem description
Use uv for faster package installation and remove preinstalled Python deps from Docker base image.

### What's changed
- Install uv in Dockerfile.base instead of preinstalling Python packages
- explicitly use CPU-only PyTorch packages (torch==2.9.0+cpu, torchvision==0.24.0+cpu) and the PyTorch CPU index URL
- Update activate script to auto-detect and use uv when available (fallback to pip)
- Add UV_INDEX_STRATEGY=unsafe-best-match to resolve PyTorch packages across multiple indexes
- Remove SYSTEM_DIST_PACKAGES from PYTHONPATH since packages are now installed in venv
- Update CI workflows to use uv pip commands
- Filter --extra-index-url from setup.py requirements to avoid pip warnings
- Add --extra-index-url to requirements files and pin CPU-specific torch variants
- Pin librosa dependencies to prevent version conflicts
- Remove requirements.txt from Docker tag hash calculation

### Checklist


sgligorijevicTT:
> I think it would be good to test out the ird image first so it doesn't impact dev workflows. In my (limited) experience with uv, I think we'd need to use `uv run python ...` and `uv run pytest ...` instead of the normal venv workflow. I will have a look

From my, also limited, experience, using activate and python as usual works fine. One of the promises of uv is that `uv pip` should be a drop in replacement for pip.

ajakovljevicTT (approved):

PR #3042, opened 2026-01-29
[tools] package `tracy` and `ttnn`

### LINKED ISSUE #996: [Tooling] Tracy integration
## Summary
Clean Tracy integration for workload profiling. 

Depend on this PR on tt-metal [wip/review]:
- https://github.com/tenstorrent/tt-metal/pull/23550

pilkicTT:
This PR enables packaging of `ttnn` and `tracy` into our wheel, enabling users to use the tools as they would in `tt-metal` repo.

The main problem around packaging these modules is that they expect stuff installed at certain places, which differ from our installation.

To resolve these issues, we create a wrapper modules (for both packages) which sole purpose is to set up the paths properly before delegating all execution to the original modules. Since the wrapping mechanism is the same for both packages, this is extracted as a separate module in `pjrt_plugin_tt`.

Both packages are wrapped so that in the source tree we will have:
```
python_package/
├── tracy
│   ├── __init__.py
│   ├── __main__.py
│   └── _original -> <path to tracy in the install dir>
├── ttnn
│   ├── __init__.py
│   ├── __pycache__
│   └── _original -> <path to ttnn in the install dir>
```

The `_original` are symlinked directories, which are created during build.

The wrapping mechanism involves some python magic to do everything so that the original python modules behave exactly the same as if they were not wrapped. Also, I've tested if the python LSPs can handle this, and it looks like we should be good. Finally, even though i would like these wrappers to not exist, it seems that they shouldn't cause big maintenance problems.

Once we figure out and resolve our packaging, probably by making `tt-metal` one of our python requirements, this logic can go away.

The `setup.py` is modified so it picks up these new packages during `build_py`. Requirements are also extended by the `ttnn`/`tracy` requirements - we could also compute their requirements dynamically, but I wanted to go simple for now.

Also, as part of this change, we are building with profiler enabled by-default (same as `tt-metal`). This can be changed if we observe some performance issues, or we add `tracy` instrumentation on our end which causes perf issues.

A separate job for testing tools (outside of our venv) is added. The intention behind this is to have a pipeline for testing wheel installations on clean environment - so that we can make sure that everything we package works.

Closes #996

### DIFF
--- a/python_package/requirements.txt
+++ b/python_package/requirements.txt
@@ -7,9 +7,9 @@ loguru==0.7.3
 torch-xla@https://pypi.eng.aws.tenstorrent.com/torch-xla/torch_xla-2.9.0%2Bgit061c1e7-cp311-cp311-linux_x86_64.whl
 
 # TTNN + tracy requirements
-networkx>=3.1
-graphviz>=0.20.3
-seaborn>=0.13.2
-pyyaml>=5.4
-click>=8.1.7
-pandas>=2.0.3
+networkx==3.6.1
+graphviz==0.21
+seaborn==0.13.2
+pyyaml==6.0.1
+click==8.3.1
+pandas==3.0.0


pilkicTT:
>With this approach we have duplicate _ttnn.so lib, as whl (zip) doesn't support symlinks (instead it duplicates files). Maybe we can try RPATH and something like this inside wrapper:

@nsumrakTT 
i was thinking to leave that as a follow up to trim down the install dir, since the `_ttnn.so` should only be used by `ttnn` python package (so it doesn't make much sense to be in the lib dir)... but anyhow, it's not too much of additional complexity, so i've implemented something similar to your suggestion... now we have only two copies of `_ttnn.so` :smile:

nsumrakTT (approved):
The _orginal symlinks are copied inside wheel becuase whl (zip format) doesn't support symlinks. However, this is much smaller than _ttnn.so and we can address this later. I wonder if we need ttnn dir inside tt-metal in the first place?

PR #3161, opened 2026-02-04
Add cache positions to attention tests

sgligorijevicTT:
### Ticket
Addresses 1/2 of #3165 

### Problem description
Attention tests failed on nightly after experimental compile was added. I rootcaused this to a constant getting embedded into the graph instead of an input argument coming in, causing a conversion pattern to fail in MLIR. This is because cache positions weren't being passed into the test and instead getting defaulted.

### What's changed
This change both papers over the lack of generality of a conversion pattern in MLIR **and makes the test more representative of how models would be used in the wild**

Conversion will be fixed too, at a later date.

### Checklist

### DIFF
--- a/tests/torch/graphs/test_attention.py
+++ b/tests/torch/graphs/test_attention.py
@@ -304,7 +304,7 @@ def test_llama_attention_decode(variant, variant_config, arch):
     )
     past_key_states = static_cache
 
-    cache_positions = torch.randint(0, max_cache_len, (batch_size,), dtype=torch.long)
+    cache_positions = torch.randint(0, max_cache_len, (seq_len,), dtype=torch.long)
 
     run_graph_test(
         attention,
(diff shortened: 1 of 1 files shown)


AleksKnezevic (approved):
Thanks for fixing this! We currently have a couple of prefill graphs going through push, do you think we should switch one of them to decode?

tests/torch/graphs/test_attention.py
```diff
@@ -304,9 +304,17 @@ def get_shard_spec(attention, args, kwargs):
     )
     past_key_states = static_cache
 
+    cache_positions = torch.randint(0, max_cache_len, (batch_size,), dtype=torch.long)
```
AleksKnezevic: Pretty sure this should be seq_len instead of batch_size. Take a look at how llama [initializes](https://github.com/huggingface/transformers/blob/ace7c370145a4a4861383672daa7095a0c8e43fc/src/transformers/models/llama/modeling_llama.py#L400C1-L403C14) if not provided

Issue #3019, opened 2026-01-28
Prefill batch size per device coverage


vladimirjovanovicTT:
Title: Prefill batch size per device coverage

Description
- Initial target: Llama 3.2 1B; reuse pattern for larger models.
- Add tests with batch_per_device = 1 and >1 (e.g., 4 if memory allows).
- Use distinct texts per sample; input lengths should roughly match the active seq_len tier.
- Combine with selected seq_len tiers (at least the smallest tier).
- Correctness: use PCC for now; TODO to adopt per-token KL later.

Tasks
- Parameterize batch_per_device; generate distinct prompts per sample aligned to seq_len.
- Add memory checks; skip/xfail high-bpd on low-memory setups.
- Report batch_per_device, shapes, and effective input lengths in test logs.

Acceptance criteria
- Coverage includes bpd=1 and bpd>1; runs are stable and produce PCC metrics.

vladimirjovanovicTT:
Update: Integrate with ModelTestConfig and tests/runner/test_models.py

- Where: tests/runner/test_models.py
- Implementation details:
  - Add pytest parametrization for batch_per_device (1 and >1).
  - Generate distinct prompts per sample; input lengths should roughly match the active seq_len tier.
  - Thread batch_per_device into ModelTestConfig.
- ModelTestConfig:
  - Add field: batch_per_device.
  - Validate: respects device/memory constraints; default to 1.
- Prefill path:
  - Use ModelTestConfig.batch_per_device when building batched inputs; report PCC.

Issue #3018, opened 2026-01-28
Prefill sequence length coverage


vladimirjovanovicTT:
Title: Prefill sequence length coverage

Description
- Initial target: Llama 3.2 1B; reuse pattern for larger models.
- Support a configurable sequence length grid from 128 up to the model’s max sequence length.
- For Llama 3.2 1B, start with a small subset (e.g., 128, 512, 4k) and expand.
- Generate inputs whose tokenized length is close to the selected seq_len.
- Use per-batch padding to the longest sequence in the batch.
- Correctness: use PCC for KV cache matrices and logits for now.
- TODO: add per-token KL on softmaxed logits once supported.

Tasks
- Implement seq_len grid generator (min=128, max from model config; flexible step strategy).
- Implement input generator producing tokenized lengths close to target seq_len.
- Implement batch padding-to-longest behavior.
- Compute PCC; emit metrics in test logs.
- Document chosen lengths and observed memory characteristics for reference.

Acceptance criteria
- Infra runs from 128 up to max model seq_len; initial subset executes for Llama 3.2 1B.
- PCC metrics reported; thresholds remain TODO to baseline.

PR #3164, opened 2026-02-04
Add infra support for testing larger seq_len in single chip prefill tests

umalesTT:
### Ticket
Closes https://github.com/tenstorrent/tt-xla/issues/3019, https://github.com/tenstorrent/tt-xla/issues/3018

### Problem description
Up to now, we tested prefill with only ```seq_len = 128```, while larger models / tasks require significantly larger ```seq_len```. 

### What's changed
In this PR we setup infra to support such cases for single chip and give example for Llama 3.2 1B testing. 
Also, ```batch_size``` testing is expanded to include ```batch_size > 1``` and such option is added for both prefill and decode tests - although only prefill uses this for now. 
These tests are executed on [realistic inputs](https://github.com/tenstorrent/tt-forge-models/pull/453) with realistic lengths compared to ```seq_len```.

More details [here](https://docs.google.com/document/d/1nmFd002Ycv8wadpyOs1ZMMkV6zV6v0pWCl-LZsap1J8/edit?tab=t.0).

PS. Some miscellaneous fixes throughout infra are also applied within this PR

### Checklist

### DIFF
--- a/tests/runner/test_models.py
+++ b/tests/runner/test_models.py
@@ -433,9 +433,18 @@ def test_llms_torch(
     test_entry, run_phase = test_entry_and_phase
 
     if run_phase == RunPhase.LLM_DECODE:
+        # Decode tests don't parametrize on sequence length (default is seq_len = 1).
+        if sequence_length is not None:
+            pytest.skip("Decode tests do not support sequence_length parameterization")
+        # Decode tests for now run only batch_size = 1.
+        if batch_size != 1:
+            pytest.skip("Decode tests currently only support batch_size=1")
         request.node.add_marker(pytest.mark.llm_decode)
 
     if run_phase == RunPhase.LLM_PREFILL:
+        # Sequence length should be specified for prefill tests.
+        if sequence_length is None:
+            pytest.skip("Sequence length must be specified for prefill tests")
         request.node.add_marker(pytest.mark.llm_prefill)
 
     test_metadata.batch_size = batch_size
--- a/tests/runner/test_config/torch_llm/test_config_inference_single_device.yaml
+++ b/tests/runner/test_config/torch_llm/test_config_inference_single_device.yaml
@@ -180,7 +180,7 @@ test_config:
   # Llama 3.2 1B Prefill tests with different seq_len and batch_size
   llama/causal_lm/pytorch-llama_3_2_1b-llm_prefill-seq_128-batch_1-single_device-inference:
     status: EXPECTED_PASSING
-    markers: [push]
+    markers: [nightly]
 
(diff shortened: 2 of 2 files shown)


tests/runner/test_models.py
```diff
@@ -374,6 +375,14 @@ def test_all_models_jax(
 # LLM Specific tests for decode and prefill phases. Separate test to avoid impacting
 # original test names in test_all_models_torch and no need for collection-time deselection logic.
 
+# Sequence lengths for prefill testing (None = decode phase uses default)
+PREFILL_SEQ_LENS = [128, 1024, 2048, 4096, 8192]
+LLM_SEQ_LENS = [None] + PREFILL_SEQ_LENS
```
ndrakulicTT: @kmabeeTT is there plan that decode also supports multiple different seq lens/batch sizes? If yes then maybe we can leave todo here as `None` is not intuitive in my opinion.

tests/runner/test_models.py
```diff
@@ -374,6 +375,14 @@ def test_all_models_jax(
...
+
+# Batch sizes for prefill testing 
+PREFILL_BATCH_SIZES = [1, 2]
+LLM_BATCH_SIZES = PREFILL_BATCH_SIZES  # Same batch sizes for both decode and prefill
+
```
ndrakulicTT: Maybe just my preference but it is more visible that this is just the list inside the `pytest.parametrize`

tests/runner/test_models.py
```diff
@@ -426,12 +455,26 @@ def test_llms_torch(
     """PyTorch LLM model test (decode/prefill phases) - delegates to shared implementation."""
     test_entry, run_phase = test_entry_and_phase
 
-    # Add phase-specific marker for filtering
+    # Skip invalid combinations: decode needs sequence_length=None (seq1), batch_size=1 for now
+    # (batch_size>1 supported by loader, but YAML entries not added yet - remove skip when ready)
```
ndrakulicTT: Do we need this, as in CI we only run tests with expected passing/xfail/skip. If the test is not in yaml configs it will not have any of these marks.

tests/runner/test_models.py
kmabeeTT: It has not really been considered to be honest.  Needs some thought. 

In the meantime.... follow up to this comment 

> "in CI we only run tests with expected passing/xfail/skip. If the test is not in yaml configs it will not have any of these marks."

 .. consider that in general, experimental nightly will (for some tests) run all discovered tests not mentioned in test_config files, so adding many many mroe combos that will be run in experimental nightly may not be desired either.

PR #3174, opened 2026-02-05
Add qb2-blackhole support to Nightly CI support + 4x TP tests in onPush/onPR

kmabeeTT:
### Ticket
None

### Problem description
- We have our first QuietBox 2 machine in CI this week and need to start qualifying our models in preparation for product launch
- Just a single machine to start without s3-bucket mirror, more machines coming later this month in cloud, so for now we need to limit the usage of this machine by not rolling out QB2 tests in the same frequency as other archs.
- We'd also like to have TP tests in onPush/onPR, was missing until now, use the otherwise-idle during the daytime CI machine for this

### What's changed
 - Add qb2 specific nightly CI jobs via dedicated model-test-passing-qb2.json, model-test-xfail-qb2.json with usage only in new QB2 Nightly on main branch (exclude from official nightly for now).
 - Add qb2-blackhole to ALLOWED_ARCHES and default_archs in conftest.py
 - Add qb2-blackhole to supported_archs in ~420 test config entries
 - Add ~50 explicit qb2-blackhole arch_overrides modelled after p150 to start, refine/improve later.
 - Add qb2-blackhole arch_overrides for 2x lower PCC (0.98) and 8x s3-bucket-missing fails
 - Tag 4x TP tests for QB2 onPush/onPR and upate model-test-push.json (<10 min runtime)

### Checklist
- [x] Initial run of nightly models as stable commit yesterday passing: https://github.com/tenstorrent/tt-xla/actions/runs/21660160320 - will run again closer to when this is released with the actual changes in this PR
- [x] test run of new models on QB2 added to push today, passing: https://github.com/tenstorrent/tt-xla/actions/runs/21697421408

### DIFF
--- a/.github/workflows/schedule-nightly-qb2.yml
+++ b/.github/workflows/schedule-nightly-qb2.yml
@@ -1,4 +1,4 @@
-name: On nightly
+name: On nightly QB2
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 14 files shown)


kmabeeTT:
Thanks @vmilosevic , I pushed change here to split these tests to their own dedicated QB2 nightly job after offline discussion (TLDR; folks want these stable/expected passing tests reported on in superset and experimental nightly is for not-stable/not-expected-passing models and isn't reported on), take another look?

kmabeeTT:
outdated, superceded by other work.

PR #3224, opened 2026-02-09
Fixed bug with moving pjrt tensors to host

acolicTT:
When moving pjrt tensor to host, it is assumed that tt::runtime::toHost will return separate runtime tensor for each shard, which is not true for sharded tensors created with strategy identity.

When sharded tensors are prepared as input tensors with strategy identity, runtime tensor from first shard will be used as a common runtime tensor in newly created pjrt tensor, and runtime tensors from other shards will be discarded.

Problem occurs when onHost is called, because onHost will retrieve single runtime tensor which needs to be shared among all shards.

Since I did not handle that case properly, all kind of nonsenses happened in tests that were moving tensors to host and reusing those tensors in next iterations.
One example is:
tt-blacksmith/blacksmith/experiments/jax/mnist/multi_chip/data_parallel/test_pure_jax_mnist.py

This commit handles case when sharded tensor created from strategy identity is moved back to host.

### DIFF
--- a/pjrt_implementation/src/api/tensor.cc
+++ b/pjrt_implementation/src/api/tensor.cc
@@ -97,9 +97,13 @@ void PjrtTensor::ensure_layout(const tt::runtime::Device &device,
 // If sharded tensor is already on the host, toHost won't do anything (each
 // shard will have it's own pjrt tensor).
 // If Sharded tensor is moved from the device, we need to create new pjrt tensor
-// for each shard (since they shared that same runtime tensor).
+// for each shard (since they shared the same runtime tensor on device). For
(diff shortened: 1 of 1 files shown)


ajakovljevicTT (approved):

jameszianxuTT:
Please add a testcase to cover this failure condition explicitly if possible. I added a space for these tests in [test_tensor_persistence](https://github.com/tenstorrent/tt-xla/blob/f323dc8fb51ef62462c285837b78224f9b1a3704/tests/torch/graphs/test_tensor_persistence.py#L4) and am a bit surprised none of those tests caught this failure

edit - though, I guess those tests run on singledevice, so that may be the issue

PR #3229, opened 2026-02-09
Use dict-based approach for node info propagation

svuckovicTT:
### Ticket
N/A

### Problem description
The metadata propagation system uses a counter-based approach that assumes a 1:1 mapping between FX `call_function` nodes and dispatched aten operations. This breaks when a single FX node decomposes into multiple aten ops at dispatch time (e.g., `torch.matmul` decomposes into `expand → view → bmm → view`). When this happens, the counter drifts and subsequent nodes get assigned incorrect metadata, leading to wrong source locations in the HLO IR.

### What's changed
Replaced the counter-based approach with an FX Interpreter + ContextVar design:

`extract_nodes_info()` now returns a `dict[str, str]` (node name → metadata) instead of a positional `list[str]`.
New `MetadataInterpreter` subclass of `torch.fx.Interpreter` executes the graph node-by-node, setting a `ContextVar` with the current FX node's metadata before each node runs. This way all aten ops dispatched from that node — regardless of how many — see the same metadata.
`MetadataDispatchMode` is simplified: it no longer maintains a counter or stores node info. It just reads the `ContextVar` and attaches metadata to the result tensor.
This ensures correct metadata attribution for any FX node, independent of how many aten operations it decomposes into at runtime.

### Checklist

### DIFF
--- a/python_package/tt_torch/backend/metadata_propagation.py
+++ b/python_package/tt_torch/backend/metadata_propagation.py
@@ -411,10 +411,9 @@ class MetadataInterpreter(Interpreter):
             if metadata:
                 # Set context before executing - all dispatched ops will see this
                 token = _current_node_metadata.set(metadata)
-                try:
-                    return super().run_node(node)
-                finally:
-                    _current_node_metadata.reset(token)
+                result = super().run_node(node)
+                _current_node_metadata.reset(token)
+                return result
 
         return super().run_node(node)
 
@@ -443,8 +442,8 @@ class MetadataDispatchMode(TorchDispatchMode):
        torch_xla._XLAC._set_xla_custom_op_name_prefix(). torch-xla traces back from the
        tensor to find which operation produced it, then labels that operation node in HLO.
 
-    This approach correctly handles FX nodes that decompose into multiple aten operations,
-    unlike a counter-based approach which assumes 1:1 mapping between FX nodes and dispatches.
+    This is non-invasive: it doesn't modify the FX graph or operation semantics, only adds
+    debugging metadata to the generated XLA HLO IR for better observability.
     """
 
     def __torch_dispatch__(self, func, types, args=(), kwargs=None):


python_package/tt_torch/backend/metadata_propagation.py
```diff
@@ -361,10 +379,44 @@ def extract_nodes_info(graph_module: torch.fx.GraphModule) -> list[str]:
...
+            # Get metadata for this node
+            metadata = self.node_info.get(node.name)
+
+            if metadata:
+                # Set context before executing - all dispatched ops will see this
+                token = _current_node_metadata.set(metadata)
+                try:
+                    return super().run_node(node)
+                finally:
+                    _current_node_metadata.reset(token)
```
vkovinicTT: Why `try` with `finally`? Why not just:
```python
result = super().run_node(node)
_current_node_metadata.reset(token)
return result
```

python_package/tt_torch/backend/backend.py
```diff
@@ -182,10 +186,13 @@ def __call__(self, *args):
...
-            # MetadataDispatchMode intercepts tensor operations via TorchDispatchMode and
-            # attaches FX metadata (module hierarchy, file, line) to XLA tensors.
-            with MetadataDispatchMode(self.node_info):
-                output = self.module(*args)
+            # Use MetadataInterpreter + MetadataDispatchMode to correctly track metadata
+            # even when FX nodes decompose into multiple aten operations at dispatch time.
+            # MetadataInterpreter sets a context variable for each FX node, and
+            # MetadataDispatchMode reads it to attach the correct metadata to each dispatch.
+            with MetadataDispatchMode():
+                interp = MetadataInterpreter(self.module, self.node_info)
```
vkovinicTT: Currently `MetadataDispatchMode` only handles setting metadata to XLA tensors while `MetadataInterpreter` only handles extraction. Since we're transitioning to `MetadataInterpreter`, can we integrate both setting and extracting metadata logic into it and remove `MetadataDispatchMode` entirely?

svuckovicTT: Not that I'm aware of - using both is what makes this possible.

The interpreter goes thru the FX graph and sets a context variable for each op. After each op is "interpreted", python dispatcher runs for that op, before the interpreted runs again for the next op.

Pytorch dispatcher routes these ops to backend implementations - this is where we get multiple aten ops for a single FX graph op. So for all the ops being dispatched, we read the context var's value that was previously set by the FX graph interpreter - we do this in `__torch_dispatch__`.

So this is the whole idea - mapping dispatched aten ops to the FX graph ops.

python_package/tt_torch/backend/metadata_propagation.py
svuckovicTT: Overly safe, will remove.

Issue #2970, opened 2026-01-26
Update model naming


nvukobratTT:
Currently, model architecture and variant name are more in the form of encoding rather than the actual name. We should update them to represent human-readable names used for model presentation purposes. Here is the rough query that contains part of the models for naming reference:
- https://superset.tenstorrent.com/superset/dashboard/443/?native_filters_key=OjxQA9ItdqE

E.g. for change
```
whisper => Whisper
large_v3 => Large v3
```

PR #3235, opened 2026-02-09
Update model naming to human readable format

ctr-lelanchelianTT:
### Ticket
closes https://github.com/tenstorrent/tt-xla/issues/2970

### Problem description
Currently, model architecture and variant name are more in the form of encoding rather than the actual name. We should update them to represent human-readable names used for model presentation purposes

### What's changed
Fixed variant names to human readable formats

### Checklist

I have ran the pipeline making this pr point out to the corresponding changes in tt-forge models https://github.com/tenstorrent/tt-forge-models/pull/472 and the pipeline is passed


kmabeeTT:
You'll want to do search and replace in .test_durations too which contains all the old test names

Issue #3251, opened 2026-02-10
gpt-oss-20b batch_size=32 hang during decoder


ddilbazTT:
While testing the benchmark script with tt-forge for gpt-oss-20b, I encountered a hang when batch_size=32 and input prompt = "Here is an exaustive list of the best practices for writing clean code:" 

However, I was able to overcome this error by
- reducing the batch_size = 16 OR 
- changing the input prompt to "Explain quantum mechanics."

Currently, there isn't a definitive reason why this hang is happening and it needs further investigation. When I enabled runtime debug messages, this is the last message being printed out:
```
[32m            RuntimeTTNN[0m | [1m[38;5;240m   DEBUG[0m | Executing operation: %7812 = "ttnn.slice_static"(%7811) <{begins = [0 : i32, 0 : i32, 0 : i32, 0 : i32], ends = [32 : i32, 1 : i32, 14 : i32, 32 : i32], step = [1 : i32, 1 : i32, 1 : i32, 1 : i32]}> : (tensor<32x1x14x64xbf16, #ttnn.ttnn_layout<(d0, d1, d2, d3) -> (d0 * 32 + d1 * 32 + d2, d3), <1x1>, memref<32x2x!ttcore.tile<32x32, bf16>, #ttnn.buffer_type<dram>>, <interleaved>>>) -> tensor<32x1x14x32xbf16, #ttnn.ttnn_layout<(d0, d1, d2, d3) -> (d0 * 32 + d1 * 32 + d2, d3), <1x1>, memref<32x1x!ttcore.tile<32x32, bf16>, #ttnn.buffer_type<dram>>, <interleaved>>> loc("slice.1176")

```

However, I consulted with @sshonTT who advised I look into the last CCL executed, which is:
```
[...]

PR #3284, opened 2026-02-12
Add gpt-oss-20b to llm benchmark with batch size = 16 

ddilbazTT:
### Ticket
None

### Problem description
Add gpt-oss-20b to benchmark

### What's changed
Several notes about gpt-oss-20b:
- Does not work with optimizer enabled due to the following errors:
    - https://github.com/tenstorrent/tt-mlir/issues/6840 [FIXED]
    - https://github.com/tenstorrent/tt-mlir/issues/6949 [OPEN]
- Encountered a runtime hang that could be overcome by either 
    - changing the input prompt to "Explain quantum mechanics." 
    - reducing the batch size
    - I went about by changing the batch size, and left a comment which includes reference to https://github.com/tenstorrent/tt-xla/issues/3251

### Checklist
- [X] https://github.com/tenstorrent/tt-xla/actions/runs/21948333500/job/63392825380

### DIFF
--- a/tests/benchmark/test_llms.py
+++ b/tests/benchmark/test_llms.py
@@ -67,6 +67,7 @@ def test_llm(
         experimental_enable_permute_matmul_fusion: Enable permute matmul fusion optimization
         read_logits_fn: Function to extract logits from model output
         required_pcc: Required PCC threshold
+        num_layers: Number of layers to override
     """
     model_loader = create_model_loader(
         ModelLoaderModule, num_layers=num_layers, variant=variant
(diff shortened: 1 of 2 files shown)


tests/benchmark/test_llms.py
```diff
@@ -185,8 +187,6 @@ def test_llm_tp(
         output_file=output_file,
         mesh_config_fn=mesh_config_fn,
         shard_spec_fn=shard_spec_fn,
-        batch_size=32,
```
ddilbazTT: No need to specify batch_size and input_sequence_length as test_llm will use the default values (same as the ones used here) unless changed. Needed to edit batch_size since I want to change that for gpt_oss_20b.

PR #3299, opened 2026-02-13
[pjrt] add tracy zones

### LINKED ISSUE #3295: PJRT Host Tracy Zones
To have a complete picture when profiling with `tracy`, add zones to our PJRT implementation.

pilkicTT:
Tracy zones are added when compiled with `TTXLA_TRACY_ZONES=ON`.

All PJRT APIs are covered for now (except for error instance related stuff), we can iterate on this if we notice that we are getting too much noise; at this moment that doesn't seem to be the case.

Additionally, some expensive threads/methods are instrumented as well.

Cmake changes are to enable including needed `tracy` files. We will see if it makes sense to install those in `tt-mlir` installation, so that we can pick them up from there, but for now we take them from the `tt-metal` path.

The way we set up include dirs is refactored, so that each lib defines its own include dirs and propagates them via `PUBLIC` if needed.

Closes #3295

### DIFF
--- a/pjrt_implementation/src/api/CMakeLists.txt
+++ b/pjrt_implementation/src/api/CMakeLists.txt
@@ -38,9 +38,19 @@ add_library(TTPJRTApi
     "tensor_pool.cc"
 )
 
-target_include_directories(TTPJRTApi PUBLIC
+target_include_directories(TTPJRTApi
+PRIVATE
+    # Includes used in .cpp files.
+
+    # Used for tracy zones in PJRT.
+    ${PROJECT_SOURCE_DIR}/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/tt_metal/third_party/tracy/public
+PUBLIC
+    # Includes used in public headers.
+    ${PROJECT_SOURCE_DIR}/pjrt_implementation/inc
     ${TTMLIR_TOOLCHAIN_DIR}/src/shardy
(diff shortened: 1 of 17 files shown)


pjrt_implementation/src/api/CMakeLists.txt
```diff
@@ -54,6 +64,14 @@ PUBLIC
     TTPJRTUtils
 )
 
+# Enable Tracy profiling if requested
+if(TTXLA_TRACY_ZONES)
```
nvukobratTT: How about putting this behind the debug flag? 

I'm not opposed to introducing an additional flag, but it'll be good if we can manage fewer to control what we're building and when. E.g., I would assume that most will use Tracy for debugging models, so having it enabled during debug build by default seems reasonable, as you already have overhead introduced.

pjrt_implementation/src/api/buffer_instance.cc
```diff
@@ -16,6 +16,9 @@
 #include <stdexcept>
 #include <thread>
 
+// tracy includes
+#include "tracy/Tracy.hpp"
```
nvukobratTT: One general note (not required for this PR). Let's consider adding a dedicated documentation section for Tracy. Now that we have it integrated and enabled runtime host zones, we should outline how to set it up, run, and use (+ ref to metal docs if they exist for more details).

PR #3306, opened 2026-02-13
Improve benchmarks PCC computation

vkovacevicTT:
### Problem description
In Performance benchmark PCC for LLMs is calculated based on the first decode token. 
We should check it for every decode token.

### What's changed
LLM PCC for all decode tokens is checked by doing:
- Generate all decode logits during benchmark run
- Generate CPU golden logits with teacher-forcing using benchmark decode tokens as ground truth
- Calculate PCC for every decode output, assert on first failure

### Checklist
- [x] [Single Vision and Encoder test](https://github.com/tenstorrent/tt-xla/actions/runs/23061879362)
- [ ] [n150 LLMs](https://github.com/tenstorrent/tt-xla/actions/runs/23594173334)
- [ ] [n300 LLMs](https://github.com/tenstorrent/tt-xla/actions/runs/23594191213)
- [ ] [galaxy LLMs](https://github.com/tenstorrent/tt-xla/actions/runs/23609387522)

### DIFF
--- a/tests/benchmark/benchmarks/llm_benchmark.py
+++ b/tests/benchmark/benchmarks/llm_benchmark.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Built-in modules
+import copy
 import os
 import socket
 import sys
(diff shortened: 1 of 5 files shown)


tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -524,10 +527,30 @@ def benchmark_llm_torch_xla(
...
+    min_pcc = min(pcc_values)
+    avg_pcc = sum(pcc_values) / len(pcc_values)
+
+    print(f"PCC Statistics (over {num_pcc_tokens} tokens):")
+    print(f"  MAX PCC: {max_pcc:.6f}")
+    print(f"  MIN PCC: {min_pcc:.6f}")
+    print(f"  AVG PCC: {avg_pcc:.6f}")
+
+    # Use MAX PCC as final PCC value
+    pcc_value = min_pcc
```
bmalesevicTT: Shouldn't the comment and the PR desc say "use min pcc"?

vladimirjovanovicTT (changes requested):
We need slight improvements to this approach to reflect causal nature of LLM decode process:
- I propose to check PCC for output tensors sequentially - we check PCC for each decode step, and assert on the first output that fails. When one token differs between CPU and TT, we can expect the following tokens to diverge, so PCC information for following tokens is of limited use.
- We should enable teacher forcing behavior as described in https://github.com/tenstorrent/tt-forge/pull/859 (regenerating golden outputs). 

Mid-term plan (IMO not needed for this PR), as discussed in https://docs.google.com/document/d/1nmFd002Ycv8wadpyOs1ZMMkV6zV6v0pWCl-LZsap1J8/edit?tab=t.0, "4. correctness" section, is to have a better metric than PCC for checking LLM generation correctness.

FYI/thoughts? @odjuricicTT @umalesTT @pglusacTT

Issue #3331, opened 2026-02-18
vllm_tt: presence_penalty, frequency_penalty, repetition_penalty silently ignored — not applied on device or CPU


kmabeeTT:
## Issue

When a user sets `presence_penalty`, `frequency_penalty`, or `repetition_penalty` in `SamplingParams`, the vLLM TT plugin accepts the values but silently discards them. Token selection is done from completely unpenalized logits. The parameters have zero effect.

This is a P0 regression relative to the GPU path, which applies all three penalties correctly via `apply_all_penalties` before sampling.

## Root Cause

Three gaps in `vllm_tt/metadata.py` and `vllm_tt/sampler.py`, inherited from the upstream TPU sampler:

1. **`no_penalties` hardcoded `True`** — `XLASupportedSamplingMetadata.no_penalties` was never set to `False`. Because this bool is captured at `torch.compile(dynamic=False)` trace time, the penalty branch was completely absent from every compiled TTIR graph regardless of user-provided values.

2. **Penalty fields never populated** — `frequency_penalties`, `presence_penalties`, `repetition_penalties` were untyped class-level `None` vars (not dataclass fields). `from_input_batch` never read them from `InputBatch` or transferred them to device.

3. **No `apply_penalties` in the sampler** — `vllm_tt/sampler.py` (forked from `vllm/v1/sample/tpu/sampler.py`) had no penalty math. There is also no CPU fallback; `apply_penalties` exists only in the GPU sampler paths and is never called from the TT/TPU model runner.

## Repro

Capture TTIR from a penalty test with and without the fix:

```bash
# WITH fix — penalty test produces a distinct 8th compiled graph with 30 extra ops
TTXLA_LOGGER_LEVEL=DEBUG pytest tests/integrations/vllm_plugin/sampling/test_sampling_params.py \
  -m "nightly and single_device" -k "test_sampling_param_sweep and presence_penalty" -v -s \
  2>&1 | tee penalty_withfix.log
python3 scripts/extract_mlir_graphs.py penalty_withfix.log --type ttir --output-dir /tmp/penalty_withfix

# WITHOUT fix — penalty test produces identical IR to no-penalty baseline
git checkout <pre-fix SHA> -- integrations/vllm_plugin/vllm_tt/sampler.py \
  integrations/vllm_plugin/vllm_tt/metadata.py integrations/vllm_plugin/vllm_tt/model_runner.py
[...]

PR #3370, opened 2026-02-19
[vLLM plugin] Add presence/frequency/repetition penalty support to Sampler + update tests + remove asserts

kmabeeTT:
### Ticket
#3331

### Problem description
`presence_penalty`, `frequency_penalty`, and `repetition_penalty` were silently ignored. Two separate issues:

1. `apply_penalties()` was not yet implemented in the vllm_tt Sampler.
2. Even after implementation, greedy decoding (`temperature=0`) bypassed the entire penalty path via two independent shortcuts that did not account for active penalties. This meant penalties were broken for all greedy sampling regardless of values set by the caller.

### What's changed
- **`sampler.py`**: Implements `apply_penalties()` using element-wise masking ops only (no sort/scatter/gather). Matches the vLLM GPU spec: repetition penalty covers prompt ∪ output tokens; frequency and presence are output-only. `test_sampling_params_synthetic.py` adds an on-device `test_penalties` test and removes the xfail from `test_sampling_feature_graph_coverage` that was blocking on this implementation.
- **`metadata.py`**: Adds `prompt_token_mask` field and `_compute_prompt_mask()`. Fixes greedy early-return that skipped penalty tensor setup when `all_greedy=True` and penalties were active.
- **`model_runner.py`**: Fixes `sample_from_logits()` argmax shortcut that bypassed the sampler entirely for greedy requests with active penalties.
- **`test_penalties_correctness.py`** (new): Pure CPU unit tests — no TT hardware, no torch_xla, runs in ~3s. Validates the penalty math against an independent reference implementation and pins the prompt-token scope invariant (`rep_mask = occurred_output | prompt_token_mask`) with hardcoded expected values.
 - **`test_sampling_params.py`**:  Unskips presence_penalty, frequency_penalty, and repetition_penalty in test_sampling_param_sweep. Adds test_additive_penalties_end_to_end and test_repetition_penalty_end_to_end with shared helpers; the repetition penalty test gets its own timeout budget since it triggers a separate lazily-compiled on-device
  graph.
 - **`conftest.py`**:  Fixes a hang when a test fails and the LLM is recreated for the next test. Root cause: deleting the LLM object didn't guarantee the EngineCore subprocess exited before the new LLM was created, leaving the TT device held by two processes simultaneously. Fix: _flush_llm_cache() now calls engine_core.shutdown() explicitly,
  which blocks until the subprocess exits (terminate() + join()). Also adds a pytest_runtest_makereport hook that sets a _needs_recreate flag on any call-phase failure, and increases TEST_TIMEOUT_SECONDS 120→180 for headroom on slower targets.

### Checklist
 - New tests passing locally, onPush/onPR and in nightly link: https://github.com/tenstorrent/tt-xla/actions/runs/22259069181

### DIFF
--- a/tests/integrations/vllm_plugin/sampling/conftest.py
+++ b/tests/integrations/vllm_plugin/sampling/conftest.py
@@ -28,10 +28,28 @@ def get_or_create_llm(name: str, **llm_args) -> vllm.LLM:
     return _llm_cache[name]
 
 
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    """Flag the LLM engine for recreation after any test failure."""
+    global _needs_recreate
+    outcome = yield
+    report = outcome.get_result()
+    if report.when == "call" and report.failed:
+        _needs_recreate = True
+
+
 def _flush_llm_cache():
     """Delete all cached LLM instances and force GC to terminate their subprocesses."""
     for name in list(_llm_cache):
-        del _llm_cache[name]
+        llm = _llm_cache.pop(name)
+        # Explicitly shut down the engine core subprocess before deleting the
+        # LLM object.  Without this, the subprocess may still hold the TT
+        # device when the next LLM is created, causing a hang.
+        try:
+            llm.llm_engine.engine_core.shutdown()
+        except Exception:
+            pass
+        del llm
     _llm_cache.clear()
     gc.collect()
 
--- a/tests/integrations/vllm_plugin/sampling/test_sampling_params.py
+++ b/tests/integrations/vllm_plugin/sampling/test_sampling_params.py
@@ -18,7 +18,6 @@ complementary:
 
 import signal
 
-import conftest
 import pytest
 import vllm
 from conftest import TEST_TIMEOUT_SECONDS, get_or_create_llm
@@ -126,8 +125,6 @@ def _test_timeout(llm):
     """
 
     def _handler(signum, frame):
-        # Flag the engine for recreation — a timeout means it stopped responding.
-        conftest._needs_recreate = True
         raise TimeoutError(
             f"Test exceeded {TEST_TIMEOUT_SECONDS}s — vLLM engine likely dead"
         )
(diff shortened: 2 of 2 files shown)


kmabeeTT:
## Implementation and testing notes

### The greedy bypass bug

The `apply_penalties` implementation exposed a pre-existing bug: any request with `temperature=0` silently skipped penalties entirely, regardless of the values set by the caller. Two code paths conspired:

1. `metadata.py` `from_input_batch()` had an early return for `all_greedy=True` that skipped building `output_token_counts`, `prompt_token_mask`, and all penalty tensors — returning `no_penalties=True` even when e.g. `repetition_penalty=50.0` was set.

2. `model_runner.py` `sample_from_logits()` short-circuited to `argmax` for any greedy request, bypassing the sampler (and thus `apply_penalties`) entirely.

The existing `test_sampling_param_sweep` didn't catch this because it uses `temperature=0.8` — with temperature > 0, requests go through the random sampling path, not the greedy shortcut.

### Test strategy

Three test types, each covering something the others can't:

**CPU unit tests (`test_penalties_correctness.py`)** — validate the penalty math in isolation. An independent reference implementation (using plain `/` instead of `reciprocal`) catches algebraic bugs. A hardcoded-value invariant test directly exercises the `rep_mask = occurred_output | prompt_token_mask` fix: a prompt-only token with logit 3.0 and rep=2.0 must produce exactly 1.5, not 3.0 (which would mean prompt scope was ignored).

**Synthetic on-device (`test_penalties`)** — validates that penalties compile and execute correctly on TT hardware with controlled inputs, without model loading.

**End-to-end greedy (`test_penalties_applied_end_to_end`)** — validates the full pipeline wiring: that `output_token_counts` and `prompt_token_mask` are correctly built from live generation state and fed into `apply_penalties`. This test could not be written until after the greedy bypass bug was fixed — before the fix, penalized output was always identical to baseline regardless of penalty value.

One subtlety in testing `frequency_penalty` and `presence_penalty` end-to-end with greedy decoding: their API cap of 2.0 is often not large enough to flip the greedy choice at positions where the model is mid-phrase, because contextual logit advantages at those positions can exceed the penalty magnitude. A count-based assertion (max token frequency in a 64-token output decreases from 11 → 3 / 4) is more reliable than an output-change assertion for these two types. `repetition_penalty` is uniquely testable with a single-token output-change assertion because it is uncapped and applies to prompt tokens from step 1.

Issue #3366, opened 2026-02-18
[vLLM] logprobs_tensors always None in vllm_tt Sampler — logprobs requests silently return no data


kmabeeTT:
## Issue

Requesting `logprobs` via `vllm.SamplingParams(logprobs=N)` silently returns no logprob data on TT hardware. `Sampler.forward()` unconditionally returns `logprobs_tensors=None` regardless of whether the request asked for logprobs:

```python
# vllm_tt/sampler.py
def forward(self, logits, sampling_metadata):
    # TODO(houseroad): Add support for logprobs_mode.
    sampled = self.sample(logits, sampling_metadata)
    return SamplerOutput(
        sampled_token_ids=sampled.unsqueeze(-1),
        logprobs_tensors=None,  # ← always None
    )
```

`test_logprobs` in `test_sampling_params.py` is currently `@pytest.mark.skip` due to a separate numpy serialization crash (see #3310). Once that is resolved, this issue will cause the test to fail on the `assert result.logprobs is not None` assertion.

## Root Cause

The infrastructure to compute and gather logprobs already exists and is partially wired:

- `Sampler.compute_logprobs` computes log-softmax over logits
- `Sampler.gather_logprobs` gathers top-k logprobs and the sampled token rank
- `model_runner.py` calls `self.gather_logprobs(logits, selected_token_ids)` when `tpu_sampling_metadata.logprobs` is True and assembles the result into `LogprobsLists`
- `_precompile_gather_logprobs` warms up the `gather_logprobs` path at startup

The missing piece is that `Sampler.forward()` is not consulted for logprobs at all — the logprobs assembly happens in `model_runner.py` directly, bypassing `forward()`. The `logprobs_tensors=None` in `SamplerOutput` is therefore a stub left over from the TPU upstream fork. The `TODO(houseroad)` comment refers to `logprobs_mode` (a future extension for controlling which logprobs to return), not to basic logprob support.

## What Needs to Happen

1. Verify that the existing `model_runner.py` logprobs path (`gather_logprobs` → `LogprobsLists`) is correctly serialized into the `SamplerOutput` returned to the vLLM engine.

2. Verify XLA correctness of `gather_logprobs` under compilation — specifically `torch.topk`, `torch.gather`, and the rank computation:

```python
topk_logprobs, topk_indices = torch.topk(logprobs, num_logprobs, dim=-1)
token_logprobs = logprobs.gather(-1, token_ids)
[...]

Issue #3310, opened 2026-02-13
VLLM test_logprobs : TypeError: Object of type <class 'numpy.int64'> is not serializableSet


kmabeeTT:
Hitting this in new test after vllm v0.13.0 uplift... needs debug and fix or env-var workaround.

```
tests/integrations/vllm_plugin/sampling/test_sampling_params.py::test_logprobs 2026-02-13 21:07:04.289611: W torch_xla/csrc/runtime/pjrt_computation_client.cpp:703] Failed to deserialize executable: UNIMPLEMENTED: Deserializing serialized executable not supported.
2026-02-13 21:07:04.458756: W torch_xla/csrc/runtime/pjrt_computation_client.cpp:703] Failed to deserialize executable: UNIMPLEMENTED: Deserializing serialized executable not supported.
2026-02-13 21:07:04.554575: W torch_xla/csrc/runtime/pjrt_computation_client.cpp:703] Failed to deserialize executable: UNIMPLEMENTED: Deserializing serialized executable not supported.
(EngineCore_DP0 pid=51725) Exception in thread Thread-6 (process_output_sockets):
(EngineCore_DP0 pid=51725) Traceback (most recent call last):
(EngineCore_DP0 pid=51725)   File "/usr/lib/python3.11/threading.py", line 1045, in _bootstrap_inner
(EngineCore_DP0 pid=51725)     self.run()
(EngineCore_DP0 pid=51725)   File "/usr/lib/python3.11/threading.py", line 982, in run
(EngineCore_DP0 pid=51725)     self._target(*self._args, **self._kwargs)
(EngineCore_DP0 pid=51725)   File "/localdev/kmabee/tt-xla/venv/lib/python3.11/site-packages/vllm/v1/engine/core.py", line 1126, in process_output_sockets
(EngineCore_DP0 pid=51725)     buffers = encoder.encode_into(outputs, buffer)
(EngineCore_DP0 pid=51725)               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=51725)   File "/localdev/kmabee/tt-xla/venv/lib/python3.11/site-packages/vllm/v1/serial_utils.py", line 152, in encode_into
(EngineCore_DP0 pid=51725)     self.encoder.encode_into(obj, buf)
[...]

PR #3416, opened 2026-02-23
[vLLM plugin] Enable logprobs: fix serialization, gather correctness, and TT precision workarounds

kmabeeTT:
### Ticket
#3310 #3366 #3463 #3464

### Problem description
Three issues were blocking a working logprobs implementation:

1. `LogprobsLists` assembly in `model_runner.py` used `list.extend()` on numpy arrays, producing scalars/sub-arrays that crashed vLLM's msgpack serializer with "Object of type numpy.int32/int64 is not serializable" (#3310).

2. `gather_logprobs()` had dtype inconsistencies in the topk indices cat.

3. After enabling the logprobs path, two TT hardware precision bugs were uncovered in `gather_logprobs()` (#3463, #3464):
   - `torch.cat` on integer tensors with non-32-aligned shapes rounds large token IDs (e.g. 19585 → 19584), collapsing distinct top-k entries in the logprob dict and causing `expected >= N logprob entries, got N-1` failures.
   - `count_tokens_ge` can return rank=0 when the gathered logprob is fractionally above all stored values; `clamp(min=1)` — the natural fix — returns -1 instead of 1 on TT because booleans are represented as bf16 (tt-metal has no bool type) and the fused sum+clamp kernel mishandles the bf16 zero case. `torch.maximum` is used as a workaround.

### What's changed
- **`model_runner.py`**: Replace `concat_lists` with `np.concatenate` for all three `LogprobsLists` numpy fields, fixing the msgpack serialization crash (#3310).
- **`metadata.py`**: Remove the `NotImplementedError` guard that blocked logprob requests before any sampler or model_runner code ran.
- **`sampler.py`**: Extract `count_tokens_ge` as a named function for testability, and apply two TT workarounds: use `torch.maximum` instead of `clamp(min=1)` (#3464), and build the indices tensor on CPU to avoid integer cat rounding (#3463). TODO comments mark both for removal once tt-mlir is fixed.
- **`test_sampling_params.py`**: Remove `@pytest.mark.skip`. Sharpen assertions: count matches token count, each dict has ≥ requested entries, values ≤ 0, ranks ≥ 1.
- **`test_sampling_params_synthetic.py`**: Add `test_gather_logprobs` (on-device, production vocab sizes) and regression tests for both precision artifacts.
- **`test_logprobs_correctness.py`** (new): CPU-only push tests for `compute_logprobs` and `gather_logprobs` covering correctness, rank semantics, dtype conventions, and both precision artifact regression scenarios.
- **`tests/torch/ops/test_concat.py`** (new) and **`test_eltwise_unary_ops.py`**: xfailed op-level tests documenting two TT integer precision bugs: concat int32 rounding (TTNNWorkaroundsPass, #3463) and fused sum+clamp on bf16 booleans (#3464).

### Checklist
- [x] Correctness and synthetic tests passing locally on single_device
- [x] Nightly VLLM tests run on CI: https://github.com/tenstorrent/tt-xla/actions/runs/22411816310
- [x] Nightly VLLM tests after PR feedback: https://github.com/tenstorrent/tt-xla/actions/runs/22504280385

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/sampler.py
+++ b/integrations/vllm_plugin/vllm_tt/sampler.py
@@ -18,13 +18,13 @@ def count_tokens_ge(logprobs: torch.Tensor, threshold: torch.Tensor) -> torch.Te
 
     Workaround for https://github.com/tenstorrent/tt-xla/issues/3464:
     tt-metal does not support boolean tensors, so ElementTypeNormalization
-    converts i1 (bool) to bfloat16 early in the TTIR pipeline.  The
+    converts i1 (bool) to bfloat16 early in the TTIR pipeline. The
     comparison (logprobs >= threshold) produces bf16 1.0/0.0 values.
     When sum(-1).clamp(min=1) is fused into a single kernel, the result
     is -1 instead of 1 on TT (each op is correct in isolation).
     torch.maximum with an explicit ones tensor avoids the broken fusion.
 
-    Returns int64 (natural sum dtype).  Callers that need int32 — e.g.
+    Returns int64 (natural sum dtype). Callers that need int32 — e.g.
     gather_logprobs for the LogprobsTensors convention — must cast after.
     """
     counts = (logprobs >= threshold).sum(-1)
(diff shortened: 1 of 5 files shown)


mmanzoorTT (approved):
Thanks for the updates.

PR #3428, opened 2026-02-23
[pjrt] simplify event instance

### LINKED ISSUE #2604: Threading issue when running Llama 70B single batch on Galaxy with KV Cache on 79 or more layers
### Describe the bug

The `meta-llama/Meta-Llama-3.1-70B` (which has 80 total hidden layers) successfully runs E2E for a single batched input when using only 78 or less layers. However it fails with 79 and 80 layers.

- With 79 layers:
Image

- With 80 layers:
Image

Both results seem to imply an error caused due to hitting the max thread limit. The 79 layers one happens after `ClientInstance::PJRT_Client_Compile` is called while the 80 layers one happens before.

1. Checkout the `hshah/llama-70b-demo` of tt-xla and build.
2. Change `num_hidden_layers` in [llama_70b_cache.py](https://github.com/tenstorrent/tt-xla/blob/0852ee380df8ebb85d65baa4908e9205e02206a3/examples/pytorch/multichip/galaxy/llama_70b_cache.py#L41) to either 78, 79, or 80 (depending on what type of error you want to reproduce)
3. Run `python examples/pytorch/multichip/galaxy/llama_70b_cache.py`.

Note you need a Galaxy machine for repro.

### Expected behavior

Both 79 and 80 layered Llama should run E2E

_No response_

pilkicTT:
We are currently spawning a thread per event instance to initiate callbacks when the event is ready. This is a large hammer and can cause problems, since threads take physical resources. There are also other places in the codebase where we are pretty relaxed with spawning threads, which will also need to be addressed in the future.

Modifying the code so that the callbacks are executed on the same thread on which `markAsReady` is called. If we see any issues with this, we will then introduce asynchrony.

Since the callbacks can initiate event destroy, the code is refactored so that the method `markAsReadyAndCallback()` is a static method. In this method, we mark the event as ready, take out the callbacks from the passed event instance and call them.

The case when there are multiple awaiters and also a destructive callback registered is now detected and we terminate the execution. The reason for this is that i would like to know if this is a valid scenario in XLA - judging by the code in `openxla` it seems that it is not. Previously, this scenario triggered undefined behaviour. Test is added which triggers this.

Closes #2604

### DIFF
--- a/pjrt_implementation/src/api/event_instance.cc
+++ b/pjrt_implementation/src/api/event_instance.cc
@@ -102,6 +102,11 @@ void EventInstance::onReady(PJRT_Event_OnReadyCallback callback_function,
   // For now, to keep things simple, we will execute the callbacks on the same
   // thread which is marking the event as ready. Or in this thread, if the
   // event is already ready.
+  //
+  // TODO: In the future we might consider having one dedicated thread per
+  // client for executing all event callbacks. However, executing all callbacks
+  // on a single thread could lead to congestion if callbacks are slow or
+  // numerous, so a smarter solution (e.g. a thread pool) might be needed.
 
   {
     std::lock_guard<std::mutex> ready_lock(m_ready_mutex);


sgligorijevicTT (commented):
Maybe run nightly on this branch just to be sure nothing breaks?

pjrt_implementation/src/api/event_instance.cc
```diff
@@ -148,25 +99,41 @@ void EventInstance::onReady(PJRT_Event_OnReadyCallback callback_function,
...
   //
-  // Based on this we decide to invoke the callback immediately on the calling
-  // thread if the event is ready, otherwise to add it to the list so it can be
-  // executed on a separate thread once the event is ready. We don't execute the
-  // pending callbacks on a thread which marks event as ready and instead do it
-  // on a separate thread to avoid doing expensive computations on a low-latency
-  // thread (as the XLA comment warns).
+  // For now, to keep things simple, we will execute the callbacks on the same
+  // thread which is marking the event as ready. Or in this thread, if the
+  // event is already ready.
```
mrakitaTT: Please add here a note that in the future we might consider having a one thread per client dedicated for all event callbacks, and also that executing all event callbacks on a separate thread might lead to congestion so we might need some smarter solution.

PR #3445, opened 2026-02-24
Replace empty sharding annotations with `{replicated}`

### LINKED ISSUE #3302: Unneeded graph retracing in multi-chip scenarios
When running llama 70b benchmark on qb/lb, graph retracing is triggered on each iteration. This was observed during investigation of #2897.

Namely, during execution of `optimized_mod` in `torch_xla` we enter `extract_graph_helper()` on each iteration of forward. Relevant snippet - from `dynamo_bridge.py`:
```python
    # If input sharding has changed from the previous program, dynamo current can
    # not detect this. It will mistakenly believe the program is the same. We need
    # to retrace it here.
    if xr.is_spmd():
      # if the input sharding was the same for skip_checking_input_sharding_threashold times
      # we will skip checking the input sharding since it can be expensive.
      if skip_checking_input_sharding_threashold > 0:
        if torch_xla._XLAC._get_xla_sharding_specs(
            xla_args_tensor_only) != xla_args_sharding_spec:
          # update the xla_args with the input with new sharding and retrace
          xla_model.xla_args = args
          (xla_args_sharding_spec, args_and_out_copy, graph_hash,
           arg_index_to_need_update_index, none_remover, graph_input_matcher,
           special_return_handler, xla_args_need_update) = extract_graph_helper(
               xla_model, sym_constants_to_graph_vars)
          skip_checking_input_sharding_threashold = xu.getenv_as(
              'XLA_DYNAMO_INPUT_SHARDING_CHECK_THRESHOLD', int, 5)
        else:
          skip_checking_input_sharding_threashold -= 1
```

There is something which triggers the condition that the new sharding specs are different from what we've already cached.

sgligorijevicTT:
### Ticket
Fixes #3302

### Problem description
Multichip models were being retraced every step, incurring python interpreter overhead. 

Low level detail is that apparently tensors would enter with blank sharding annotations and the dispatch code would store those as having `{replicated}` annotations, which would then differ under a simple equals comparison, causing a retrace.

### What's changed
Tensors with no sharding annotations are explicitly annotated as `{replicated}` before being sent to the dispatch code, removing the dispatch. Note that this assume unannotated tensors are replicated, but that should be a relatively reasonable assumptions.

To be honest this is a bit of a hack.

### Checklist

### DIFF
--- a/python_package/tt_torch/backend/backend.py
+++ b/python_package/tt_torch/backend/backend.py
@@ -198,7 +198,8 @@ class XLAExecutor:
 
         full_args = self.params_and_consts + args
         # Ensure unsharded tensors are marked REPLICATED in SPMD mode so their
-        # sharding spec matches what was recorded at graph capture time.
+        # sharding spec matches what was recorded at graph capture time. This is a temporary workaround
+        # until a change is made in torch-xla to automatically mark unsharded tensors as REPLICATED at runtime in SPMD mode.
         _mark_unsharded_args_replicated(full_args)
         return self.compiled_graph(*full_args)
 


jameszianxuTT:
edit: Actually, I think it wouldn't be that hard in torch xla. I think we could just patch [this comparator](https://github.com/pytorch/xla/blob/c04e61c3424142c0eebcc9e59984b9d8fced18c0/torch_xla/_dynamo/dynamo_bridge.py#L534-L535) to alias the None and replicated case. 

I now feel like upstream patch would be more suitable than a w/a in our compiler, though others are welcome to comment.

AleksKnezevic (approved):

Issue #3451, opened 2026-02-25
[pjrt] fix retrieving tensor size


pilkicTT:
When `PJRT_Buffer_ToHostBuffer` is called to query the host buffer size (without performing a transfer), the size is currently computed from the runtime tensor.

This introduces two issues:
- Requires the runtime tensor to be available, forcing synchronization with any in-flight device-to-host transfers.
- May overestimate the required size when the tensor is stored in tile layout on device, since host buffers always use row-major layout and the two sizes may differ.

Ideally, we would want to compute the tensor size via its logical shape and the data type of the elements, without even accessing the runtime tensor.

PR #3458, opened 2026-02-25
[pjrt] PJRT_Buffer_ToHostBuffer size query fix

mstojkovicTT:
### Ticket
fixes: #3451

### Problem description
Explained in the issue.

### What's changed
  - Fix `PJRT_Buffer_ToHostBuffer` size query (`dst == nullptr`) to return the logical buffer size instead of the tile-padded physical size     
  - Add `logicalTensorSize()` method to `BufferInstance` that computes size from dimensions and data type without requiring a runtime tensor                                                                            
  - Update `copyToHost` assert to use logical size since data is untilized before copy

### DIFF
--- a/pjrt_implementation/src/api/buffer_instance.cc
+++ b/pjrt_implementation/src/api/buffer_instance.cc
@@ -486,7 +486,7 @@ PJRT_Error *onBufferToHostBuffer(PJRT_Buffer_ToHostBuffer_Args *args) {
   // This API function can be used with null `dst` to query the required size.
   if (!args->dst) {
     ZoneScopedN("QueryHostBufferSize");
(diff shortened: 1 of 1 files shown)


pilkicTT (commented):
@jameszianxuTT 
> 1. Has this call been observed from anywhere?

i would hope that it is, since the logic was already existing... but that was also a part of motivation for this effort - @mstojkovicTT will confirm whether or not this is calculated anywhere, that's still work-in-progress

additional motivation is to implement it properly (so that we don't get overestimation) and to avoid using runtime tensor, since that causes additional complexity because we will need to synchronize this access with any in-flight host<->device transfers... beacuse of this, we intentionally want to avoid using `getTensorLogicalVolume()` on the runtime tensor

Issue #3386, opened 2026-02-19
Pass shard status for results in stablehlo pipeline


tapspatel:
From this PR: https://github.com/tenstorrent/tt-mlir/pull/7120

We need to pass the result shard status as a pipeline option to stablehlo-pipeline.

Context in this issue: https://github.com/tenstorrent/tt-mlir/issues/7002
See `include/ttmlir/Dialect/StableHLO/Pipelines/StableHLOPipelines.h` specifically `ListOption<int64_t> resultPresharded`. It's a list of shard status for each result, True = presharded. False = unsharded.

jameszianxuTT:
It occurs to me that maybe it would be better if we make torch_xla emit output shardings than to do this frontend detection / pipeline forwardng in #3459 to fix the problem of torch_xla and jax having different default behaviour

As a next step - we could probably alter torch_xla behaviour to just attach output sharding properly by default, in the same way it default-annotates inputs as replicated. We've already seen that torchxla just doesn't handle output shardings properly so we already have a hack #2731 - and maybe that could be patched upstream.

PR #3459, opened 2026-02-25
Pass shard status for results in stablehlo pipeline

btsoiTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3386

### What's changed
Pass the result shard status as a pipeline option to stablehlo-pipeline. It is represented as a vector of int64, where 1 = presharded, 0 = unsharded.

### DIFF
--- a/pjrt_implementation/src/api/module_builder/module_builder.cc
+++ b/pjrt_implementation/src/api/module_builder/module_builder.cc
@@ -766,10 +766,11 @@ std::vector<int64_t> ModuleBuilder::collectResultPresharded(
       !module.get()->getAttrOfType<mlir::IntegerAttr>("mhlo.num_replicas");
 
   for (mlir::func::FuncOp &func_op : publicFuncOps) {
-    for (unsigned i = 0; i < func_op.getNumResults(); ++i) {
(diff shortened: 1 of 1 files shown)


pjrt_implementation/src/api/module_builder/module_builder.cc
```diff
@@ -749,13 +751,46 @@ mlir::LogicalResult ModuleBuilder::createShardingsFromShardy(
...
+std::vector<int64_t> ModuleBuilder::collectResultPresharded(
+    const mlir::OwningOpRef<mlir::ModuleOp> &module) {
+
+  std::vector<mlir::func::FuncOp> publicFuncOps = getPublicFuncOps(module);
+
+  std::vector<int64_t> result_presharded;
+
+  // Detect torch-xla with a heuristic mentioned in collectNumDevicesToUtilize.
+  // torch-xla never sets mhlo.num_partitions or mhlo.num_replicas on the
+  // module, whereas JAX always sets at least one.
```
acicovicTT: if this is used somewhere else as well I would suggest extracting it to a helper function, since it's basically a guess based on the behavior of torch vs jax

PR #3461, opened 2026-02-25
Introduce accuracy testing for LLMs (top1/top5 metrics)

dgolubovicTT:
Closes tenstorrent/tt-xla#3355 

## Overview: LLM Accuracy Testing with TOP1/TOP5 Metrics                                                                                                                                                                                                                                          
   
  This PR introduces token-level accuracy testing for LLM benchmarks, measuring how well TT device predictions match CPU reference outputs.                                                                                                                                                         
                                  
  **Accuracy Metrics:**
  - **TOP1 Accuracy**: Percentage of tokens where the device's prediction matches the CPU reference model's top prediction (argmax)
  - **TOP5 Accuracy**: Percentage of tokens where the device's prediction appears in the CPU reference model's top-5 predictions

  **Test Corpus:**
  We use "A Tale of Two Cities" by Charles Dickens as our reference text. Sample of text is split into two parts:
  - **Prefill phase** (first half): Input context fed to the model
  - **Decode phase** (second half): Ground truth tokens for validation with teacher forcing

  **Critical Requirement - Sequence Length Matching:**
  The CPU reference outputs (.refpt files) **must** be generated with the same `total_length` as the `input_sequence_length` used during device testing. Mismatched sequence lengths cause accuracy degradation even with teacher forcing, because CPU model and compiled model need to have same input sequence size (context).

NOTE: We currently use almost identical approach to accuracy testing in tt-metal for 1-1 comparison. Same text corpus (Tale of Two Cities) and same top1/top5 metrics. However, since we test accuracy on llms only on n150, we had to use total sequence length 128 (for memory constraints). That means 64 tokens in prefil, and 64 for decode.

## Implementation

### Introduce accuracy testing to test_llms.py
- Add --accuracy-testing argument to test_llms.py tests that we track in benchmark testing.
- Add --batch-size argument to accuracy tests in test_llms.py because we can't fit default batch 32 on device
  due to larger input sequence length required in accuracy testing.
  With batch size 32 and input sequence length, 7B and 8B models failed with OOM issues.
- Run accuracy tests in separate job called: run-n150-accuracy-benchmarks of the perf-benchmark-experimental workflow

###On-demand golden reference generation
                                                                                                                                                                                                                                                                                                                                                                    
  Reference .refpt files are generated on-demand during accuracy tests instead of being committed to git. Only the source text corpus (tale-of-two-cities.txt.bz2) is tracked in the repository.

How it works:

- Before running an accuracy test, init_accuracy_testing() calls TokenAccuracy.needs_regeneration() to check if the .refpt file is missing or stale (torch/transformers version mismatch).
- If regeneration is needed, generate_reference_outputs() (reference_generator.py) runs a CPU-based teacher-forced decode using the HuggingFace model and saves top-1/top-5 token predictions to a .refpt file atomically.
- On subsequent runs the cached .refpt file is reused, skipping generation entirely.

### Adding shared utility for decode (decode_utils.py)
  Centralize LLM decode operations used by reference output generation and accuracy testing:
  - generate_and_benchmark method - Unified decode loop for benchmarks, accuracy testing, and reference generation. Supports both regular and teacher forced decode.
  - Static cache and accuracy testing initialization
  - top-k extraction

  **Prevents implementation drift between reference generation and benchmark paths
  by sharing the same decode logic, tokenization, and cache semantics.**

**Note on decode path safety:**

The timed decode loop inside generate_and_benchmark() that produces benchmark numbers is unchanged. All accuracy-testing logic is gated behind if accuracy_testing / if ground_truth_tokens is not None and never executes during regular perf benchmarks. The only non-gated changes are: (1) eval-mode/dropout assertions that run once before the timed loop, (2) init_static_cache() helper that performs the exact same StaticCache() + early_initialization() calls previously inlined in construct_inputs(), (3) removal of unused is_multichip/mesh parameters that were dead code. None of these touch the measured decode path.


### TokenAccuracy class for validating LLM inference quality(token_accuracy.py)
  - Loads precomputed reference data from .refpt files (tokens, top1/top5 predictions)
  - Validates torch/transformers versions match reference file for reproducibility
  - Splits reference tokens into prefill (input) and decode (ground truth) windows
  - Computes TOP1/TOP5 accuracy by comparing model predictions against reference
  - Provides teacher forcing tokens for deterministic decode loops
  - needs_regeneration() detects missing or stale .refpt files (torch/transformers version mismatch), triggering on-demand regeneration

### Slight refactoring
  - Simplify static cache initialization using init_static_cache helper
  - Remove unused variables is_multichip and mesh from generate_and_benchmark function.

### DIFF
--- a/examples/vllm/TinyLlama-1.1B-Chat-v1.0/responses_client.py
+++ b/examples/vllm/TinyLlama-1.1B-Chat-v1.0/responses_client.py
@@ -1,0 +1,78 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Example client for the OpenAI /v1/responses API with a TT-backed vLLM server.
+
+Usage:
+    1. Start the vLLM server:
+       $ bash examples/vllm/TinyLlama-1.1B-Chat-v1.0/service.sh
+
+    2. Run this client:
+       $ python examples/vllm/TinyLlama-1.1B-Chat-v1.0/responses_client.py
+"""
+
+import json
+
+import requests
+
+
+def main():
+    url = "http://localhost:8000/v1/responses"
+    model = "TinyLlama/TinyLlama-1.1B-Chat-v1.0"
+
+    conversation = []
+
+    while True:
+        input_text = input("Enter a message ('c' to clear history, 'q' to quit): ")
+        if input_text == "q":
+            break
+        if input_text == "c":
+            conversation.clear()
+            print("Conversation history cleared.")
+            continue
+
+        conversation.append({"role": "user", "content": input_text})
+
+        data = {
+            "model": model,
+            "input": conversation,
+            "stream": True,
+        }
+
+        try:
+            full_response = ""
+            with requests.post(url, json=data, stream=True) as response:
+                if response.status_code != 200:
+                    error = response.json().get("error", {})
+                    print(f"Server error: {error.get('message', response.text)}")
+                    conversation.pop()
+                    continue
+                for line in response.iter_lines(decode_unicode=True):
+                    if not line or not line.startswith("data: "):
+                        continue
+                    event_data = line[len("data: ") :]
+                    if event_data.strip() == "[DONE]":
+                        break
+                    try:
+                        event = json.loads(event_data)
+                        if event.get("type") == "response.output_text.delta":
+                            delta = event.get("delta", "")
+                            print(delta, end="", flush=True)
+                            full_response += delta
+                    except json.JSONDecodeError:
+                        pass
+
+            print()
+            conversation.append({"role": "assistant", "content": full_response})
+        except requests.exceptions.ConnectionError:
+            print(
+                "Server returned a connection error. This usually occurs when a "
+                "request is made before the service is ready. Please wait for the "
+                "service to be ready and try again."
+            )
+            conversation.pop()
+
+
+if __name__ == "__main__":
+    main()
--- a/examples/vllm/TinyLlama-1.1B-Chat-v1.0/service.sh
+++ b/examples/vllm/TinyLlama-1.1B-Chat-v1.0/service.sh
@@ -2,9 +2,10 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-TTXLA_LOGGER_LEVEL=DEBUG vllm serve TinyLlama/TinyLlama-1.1B-Chat-v1.0 \
-    --max-model-len 2048 \
-    --max-num-batched-tokens 2048 \
+vllm serve TinyLlama/TinyLlama-1.1B-Chat-v1.0 \
+    --max-model-len 1024 \
+    --max-num-batched-tokens 1024 \
     --max-num-seqs 1 \
     --no-enable-prefix-caching \
+    --gpu-memory-utilization 0.1 \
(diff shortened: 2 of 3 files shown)


sgligorijevicTT (changes requested):

nsumrakTT (approved):

tests/benchmark/decode_utils.py
```diff
@@ -0,0 +1,315 @@
...
+            next_token_ids = logits[:, -1].argmax(dim=-1)
+            predicted_tokens.append(next_token_ids[0].item())
+
+            # Top-k extraction (for reference generation; batch_size must be 1).
+            # On step 0 logits has shape [1, prompt_len, vocab] so extract_topk
+            # naturally covers all prefill positions; on step 1+ it covers one.
+            if collect_topk:
+                top1, top5 = extract_topk(logits, k=topk_k)
+                all_top1.append(top1.squeeze(0).cpu())
+                all_top5.append(top5.squeeze(0).cpu())
```
odjuricicTT: This should not be a part of the benchmarking loop (what is measured with the timer). This function should only return logits and itteration times. Extracting topK is post processing of the logics and should be somewhere in the accuracy calculation.

tests/benchmark/decode_utils.py
```diff
@@ -0,0 +1,315 @@
...
+                    1, 1
+                )  # CPU [1,1]
+                input_args["input_ids"] = (
+                    next_tok_host.expand(batch_size, 1).contiguous().to(device)
+                )
+                # XLA-safe cache_position: normalize to stable [1] shape
+                host_cache_pos = (
+                    input_args["cache_position"].to("cpu").reshape(-1)[-1:]
+                )  # CPU [1]
+                input_args["cache_position"] = (host_cache_pos + 1).to(device)
```
odjuricicTT: Is this the same as in the else branch below? Seems that this if can only set what the next token on host is and the rest of the logic stays the same?

tests/benchmark/test_llms.py
```diff
@@ -69,7 +74,12 @@ def test_llm(
         read_logits_fn: Function to extract logits from model output
         required_pcc: Required PCC threshold
         num_layers: Number of layers to override
+        accuracy_testing: Enable token accuracy testing with reference data
     """
+    # Set default batch size if None
+    if batch_size is None:
+        batch_size = DEFAULT_BATCH_SIZE
+
```
odjuricicTT: Why is this needed? It should already be set to default if not passed into the function.

vkovacevicTT (approved):
Nice!

tests/benchmark/reference_outputs/Falcon3-1B-Base.refpt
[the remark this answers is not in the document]
mrakitaTT: I think these are too slow to generate for every run so we probably want to have them checked in somewhere, but @dgolubovicTT if you have some numbers let's decide based on that. If we decide to have them checked in somewhere, I doubt we want to introduce git lfs again but I am also unaware of better options, that's why my initial reaction was to unblock this PR if the files are indeed going to stay this small. In any case, I'll leave @vmilosevic to decide because I'm impartial.

dgolubovicTT:
Opened an issue on tt-mlir to investigate L1 clash issues on accuracy models: https://github.com/tenstorrent/tt-mlir/issues/7364
For now, we will just use smaller batch size to mitigate this

Issue #3363, opened 2026-02-18
[vLLM] bad_words silently ignored in vllm_tt Sampler


kmabeeTT:
## Issue

`bad_words` in `vllm.SamplingParams` has no effect on TT hardware. Words specified in `bad_words` appear freely in generated output. Tracked in the test suite as:

```python
@pytest.mark.xfail(reason="bad_words not enforced in vllm_tt Sampler")
def test_bad_words(llm, prompt): ...
```

## Root Cause

`XLASupportedSamplingMetadata` (forked from `vllm/v1/sample/tpu/metadata.py`) has `bad_words_token_ids = None` as an untyped class-level stub — it is never populated in `from_input_batch`. `vllm_tt/sampler.py` has no `apply_bad_words` method.

The upstream GPU implementation (`vllm/v1/sample/ops/bad_words.py`) applies bad words by iterating over the batch with Python loops, checking whether the tail of generated tokens matches a bad word prefix, then setting the next-token logit to `-inf`. This involves per-batch Python iteration and runtime token history access — neither compiles to a static XLA graph.

**Complication**: `bad_words` supports multi-token sequences. Suppressing `["New York"]` requires checking whether the last token was `"New"` before banning `"York"` — this prefix matching needs generated token history per request, which is engine-level state. Multi-token bad words are fundamentally harder to vectorize. Single-token bad words are tractable: they reduce to a dense `[batch, vocab]` bool mask set to `-inf`, the same pattern used by `apply_penalties` (#3331).

## Repro

```python
params = vllm.SamplingParams(temperature=0.0, max_tokens=16, bad_words=["the"])
output = llm.generate(prompt, params)[0].outputs[0].text
assert "the" not in output.lower()  # FAILS — "the" appears freely
```

## Suggested Fix

**Phase 1 — single-token bad words:**
- In `from_input_batch`, extract single-token entries from `bad_words_token_ids` and build a dense `[batch, vocab]` float32 additive mask (0 or `-inf`).
- Add `no_bad_words: bool = True` gate (same pattern as `no_penalties`) so the path is absent from compiled graphs when unused.
- Add `apply_bad_words` to `vllm_tt/sampler.py`: `logits += bad_words_mask`.

**Phase 2 — multi-token bad words:**
Requires prefix matching against generated token history — engine-level, not straightforward to vectorize for XLA. Defer or apply on CPU post-sampling.

kmabeeTT:
[...]

PR #3482, opened 2026-02-26
[vLLM plugin] Add multi-token bad_words support (#3363)

kmabeeTT:
### Ticket
#3363

### Problem description
Multi-token `bad_words` sequences raised `NotImplementedError` — the initial bad_words implementation (#3415) only supported single-token bans. Multi-token bad words (e.g. "New York" → `[188, 469]`) require prefix matching against generated token history to ban the final token only when the preceding tokens match.

### What's changed
- **`metadata.py`**: New `_compute_bad_words_mask()` static method replaces the inline single-token loop + `NotImplementedError`. Performs prefix matching on CPU: a bad word `[t0, t1, ..., tN]` sets `mask[req, tN] = -inf` only when the most recent N output tokens equal `[t0, ..., tN-1]`. Single-token bad words are always banned (same behavior as before). The resulting `[batch, vocab]` float32 mask is transferred to device and applied via the existing `Sampler.apply_bad_words` path — no sampler or model_runner changes needed.
- **`test_logit_bias_bad_words_correctness.py`**: Replaces 3 old mask-building tests (including the multi-token-raises test) with 12 new `_compute_bad_words_mask` tests: prefix match, no-match, insufficient history, empty history, two-token, mixed single+multi, multi-request independence, padding rows, out-of-vocab, empty sequence, and exact match against upstream `vllm.v1.sample.ops.bad_words._apply_bad_words_single_batch`.
- **`test_sampling_params.py`**: Removes the single-token-only skip guard from `test_bad_words`; adds multi-token assertion path for models where the banned word tokenizes to multiple tokens.

### Checklist
- [x] 24/24 correctness tests passing (logit_bias + bad_words)
- [x] E2E `test_bad_words[single_device]` passing on device (OPT-125m) locally and n300/n300-llmbox in onPR
- [x] Running vLLM nightly on CI: https://github.com/tenstorrent/tt-xla/actions/runs/22456218467

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/metadata.py
+++ b/integrations/vllm_plugin/vllm_tt/metadata.py
@@ -215,7 +215,7 @@ class XLASupportedSamplingMetadata:
             no_logit_bias = True
 
         # Build bad_words_mask tensor (single-token AND multi-token).
-        # Single-token bad words are always banned.  Multi-token bad words
+        # Single-token bad words are always banned. Multi-token bad words
         # use prefix matching against output token history: a sequence
         # [t0, t1, ..., tN] bans tN only when the most recent N output
         # tokens equal [t0, ..., tN-1].
--- a/tests/integrations/vllm_plugin/sampling/test_sampling_params.py
+++ b/tests/integrations/vllm_plugin/sampling/test_sampling_params.py
@@ -345,8 +345,8 @@ def test_seed(llm, prompt):
 def test_bad_words(llm, prompt):
     """Test that bad_words bans are enforced in the generated output.
 
-    Uses "the" as a banned word.  For models where this tokenizes to a single
-    token, asserts the token ID never appears.  For multi-token tokenizations,
+    Uses "the" as a banned word. For models where this tokenizes to a single
+    token, asserts the token ID never appears. For multi-token tokenizations,
     asserts the full contiguous sequence never appears (prefix matching
     suppresses the final token whenever the prefix is generated).
     """


kmabeeTT:
## Architecture walkthrough

### The big picture

When vLLM generates text, it works token-by-token in a loop called "decode steps." Each step:
1. The model produces **logits** — a score for every token in the vocabulary (e.g. 50,272 scores for OPT-125m)
2. The **sampler** picks the next token based on those scores
3. The chosen token is appended to the output, and the loop repeats

`bad_words` says: "never generate this word/phrase." The job is to suppress the banned tokens' logits to `-inf` so the sampler can never pick them.

### Single-token vs multi-token

**Single-token** (e.g. banning "the" → token `5`): Simple — set `logits[5] = -inf` every step.

**Multi-token** (e.g. banning "New York" → tokens `[188, 469]`): You can't always ban token `469` ("York") — it's a valid word on its own. You only ban `469` when the **previous** token was `188` ("New"). This requires checking the output history each step.

### The flow

```
  ┌─────────────────────────────────────────────────────────────┐
  │                    DECODE LOOP (per step)                   │
  │                                                             │
  │  ┌──────────────┐                                           │
  │  │  Model runs   │   logits: [batch, vocab_size]             │
  │  │  on TT device │──────────────────────┐                   │
  │  └──────────────┘                       │                   │
  │                                         ▼                   │
  │                              ┌─────────────────────┐        │
  │                              │  logits on device    │        │
  │                              └─────────┬───────────┘        │
  │                                        │                    │
  │  ┌─ CPU ──────────────────────────┐    │                    │
  │  │                                │    │                    │
  │  │  _compute_bad_words_mask()     │    │                    │
  │  │                                │    │                    │
  │  │  For each request:             │    │                    │
  │  │    bad_words = [[188, 469]]    │    │                    │
  │  │    output_so_far = [... 188]   │    │                    │
  │  │                                │    │                    │
  │  │    "Last token is 188,         │    │                    │
  │  │     prefix matches [188]!      │    │                    │
  │  │     Ban token 469."            │    │                    │
  │  │                                │    │                    │
  │  │  Result: mask[req, 469] = -inf │    │                    │
  │  │          mask[req, *]   = 0    │    │                    │
  │  │                                │    │                    │
  │  │  mask shape: [batch, vocab]    │    │                    │
  │  └──────────────┬─────────────────┘    │                    │
  │                 │ .to(device)           │                    │
  │                 ▼                       ▼                    │
  │         ┌──────────────────────────────────────┐            │
  │         │  Sampler (on TT device)              │            │
  │         │                                      │            │
  │         │  1. logits += bad_words_mask  ◄─ -inf │            │
  │         │  2. logits += logit_bias             │            │
  │         │  3. apply penalties                  │            │
  │         │  4. apply temperature                │            │
  │         │  5. top-k / top-p filtering          │            │
  │         │  6. sample token                     │            │
  │         └──────────────┬───────────────────────┘            │
  │                        │                                    │
  │                        ▼                                    │
  │               selected_token = 302  (not 469!)              │
  │                        │                                    │
  │                        ▼                                    │
  │              output_so_far.append(302)                      │
  │              (history grows for next step's prefix matching) │
  │                                                             │
  └──────────────────── next decode step ───────────────────────┘
```

### Why the CPU/device split?

The prefix matching logic is inherently dynamic:
- Different requests have different bad_words lists
- Token histories grow each step and vary per request
- Prefix comparison is a variable-length list equality check

None of this can be expressed as fixed-shape tensor ops for XLA compilation. But the **result** — "which tokens are banned right now" — is a fixed-shape `[batch, vocab]` tensor. So we do the dynamic work on CPU, produce a static tensor, and send it to the device.

This is the same pattern used throughout the TT sampler:
- `_compute_token_counts` → penalty mask (CPU build, device apply)
- `_compute_prompt_mask` → repetition penalty mask (CPU build, device apply)
- `_compute_bad_words_mask` → bad words mask (CPU build, device apply)

### Concrete example

Prompt: `"The capital of"`, bad_words: `["New York"]` → tokens `[188, 469]`

| Step | Output so far | Prefix check | Mask | Sampled |
|------|--------------|--------------|------|---------|
| 1 | `[]` | Need 1 token for prefix, have 0 → skip | all zeros | `188` ("New") |
| 2 | `[188]` | Last 1 token `[188]` == prefix `[188]` → **ban 469** | `mask[469]=-inf` | `312` (not "York"!) |
| 3 | `[188, 312]` | Last 1 token `[312]` != prefix `[188]` → skip | all zeros | `...` |

The model wanted to say "New York" but step 2 banned "York" because "New" had just been generated. The model picks a different continuation.

Issue #3480, opened 2026-02-26
Add ability to load system descriptor


btsoiTT:
We currently pull the system descriptor from runtime but it would be useful to inject system descriptors. This will enable us to compile for a different system than the one we are currently on.

PR #3483, opened 2026-02-26
Implement system descriptor loading for compile only mode

btsoiTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3480

### What's changed
Adds the ability to load system descriptors and enable compile-only mode so that it is possible to compile for a different system than the one the program is on.

### Checklist


pjrt_implementation/src/api/loaded_executable_instance.cc
```diff
@@ -319,9 +323,16 @@ onLoadedExecutableExecute(PJRT_LoadedExecutable_Execute_Args *args) {
   ZoneScoped;
   DLOG_F(LOG_DEBUG, "LoadedExecutableInstance::PJRT_LoadedExecutable_Execute");
 
-  tt_pjrt_status status =
-      LoadedExecutableInstance::unwrap(args->executable)->execute(args);
+  LoadedExecutableInstance *instance =
+      LoadedExecutableInstance::unwrap(args->executable);
+
+  if (instance->isCompileOnly()) {
+    LOG_F(ERROR, "Execution is not supported in compile-only mode "
```
jameszianxuTT: Phrasing this as not supported is a bit odd 😅. Are you using early abort as "control flow" to interrupt the normal execution path?

Issue #3365, opened 2026-02-18
[vLLM] Per-request seed in SamplingParams not supported on TT hardware (XLA limitation)


kmabeeTT:
## Issue

`seed` in `vllm.SamplingParams` has no effect on TT hardware. The same prompt with the same seed produces different outputs on successive calls. Tracked in the test suite as:

```python
@pytest.mark.xfail(reason="Torch XLA does not support per-request seed (platform.py)")
def test_seed(llm, prompt): ...
```

## Root Cause

The GPU sampler seeds sampling per request by overwriting individual rows of the Gumbel noise tensor `q` with a per-request `torch.Generator`:

```python
q = torch.empty_like(probs).exponential_(1)  # unseeded base
if generators:
    for i, generator in generators.items():
        q[i].exponential_(generator=generator)  # per-request seed
```

XLA traces the full computation graph ahead of execution. Per-element in-place operations driven by a Python `torch.Generator` object cannot be captured in a static XLA trace. The metadata documents this explicitly:

```python
# Generator not supported by torch/xla. This field must be immutable.
_generators: dict[int, torch.Generator] = field(default_factory=lambda: dict())
```

`_generators` is always empty, so the seeding loop in `random_sample` is never reached. This is an upstream XLA limitation, not specific to TT-XLA.

## Repro

```python
params = vllm.SamplingParams(seed=42, temperature=0.8, top_p=0.9, max_tokens=16)
out1 = llm.generate(prompt, params)[0].outputs[0].text
out2 = llm.generate(prompt, params)[0].outputs[0].text
assert out1 == out2  # FAILS — seed is ignored, outputs differ
```

## Suggested Fix

No straightforward fix within the current XLA compilation model. Options:

1. **Upstream XLA/PJRT support** — add support for seeded random ops in the XLA trace. Requires upstream PyTorch/XLA changes.
2. **Compile-time seed** — bake the seed scalar as a graph constant at compile time. Each unique seed triggers recompilation, which is expensive and impractical for variable user seeds.
3. **PRNG key injection** — use a deterministic PRNG (e.g. Philox/counter-based) that accepts a seed tensor as a graph input rather than a Python `torch.Generator`. Would require XLA-compatible random number generation support.

Recommend tracking for upstream PyTorch/XLA support. Not actionable within TT-XLA alone.

PR #3487, opened 2026-02-26
[vLLM plugin] Enable SamplingParams seed support (#3365)

kmabeeTT:
### Ticket
#3365

### Problem description
`SamplingParams(seed=N)` was silently rejected at the platform level — `validate_request()` raised `ValueError` for `RANDOM_SEED`, and the generator was hardcoded to `None` at request creation. Two calls with the same seed produced different outputs.

### What's changed
- **`platform.py`**: Remove `RANDOM_SEED` rejection guard.
- **`model_runner.py`**: Create CPU `torch.Generator` from `SamplingParams.seed` at request creation (matching the GPU model runner path). Add `no_generators` to the greedy fast-path check. Remove stale GPU-specific generator rewind code (CPU generators don't support `get_offset`/`set_offset`).
- **`metadata.py`**: Build `q_samples` exponential noise tensor on CPU in `from_input_batch()` using per-request generators, transfer to XLA device. Add `no_generators` and `q_samples` fields.
- **`sampler.py`**: When `q_samples` is provided, use it directly in `random_sample()` instead of generating noise on-device.
- **`test_seed_correctness.py`** (new): 4 CPU-only unit tests for q_samples logic.
- **`test_sampling_params_synthetic.py`**: On-device determinism and mixed-batch tests at production vocab sizes.
- **`test_sampling_params.py`**: Enable E2E `test_seed` (was skipped).

### Checklist
- [x] New tests (CPU-only, synthetic, e2e) passing locally for single_device
- [x] Tests run in VLLM Nightly passing: https://github.com/tenstorrent/tt-xla/actions/runs/22551275489

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/metadata.py
+++ b/integrations/vllm_plugin/vllm_tt/metadata.py
@@ -61,12 +61,23 @@ class XLASupportedSamplingMetadata:
 
     allowed_token_ids_mask = None
 
-    # Generator not supported by xla
+    # Per-request generators live in InputBatch, not here. from_input_batch()
+    # drains them to build q_samples; this field is always constructed empty.
     _generators: dict[int, torch.Generator] = field(default_factory=lambda: dict())
+
+    # True (default) when no per-request seeds are present.
+    # False when seeds exist: prevents the greedy fast-path so that
+    # the seeded q_samples tensor flows through the compiled sampling graph.
+    no_generators: bool = True
+    # Pre-computed exponential noise [padded_num_reqs, vocab_size] on device
+    # for the Gumbel-max trick, or None when no seeds are present (common
+    # case — noise is generated on device inside the compiled graph).
+    q_samples: torch.Tensor | None = None
 
     @property
     def generators(self) -> dict[int, torch.Generator]:
-        # Generator not supported by torch/xla. This field must be immutable.
+        # Generators are consumed in from_input_batch() to build q_samples.
+        # After construction this dict is always empty.
         return self._generators
 
     @staticmethod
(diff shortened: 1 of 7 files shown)


kmabeeTT:
## Design notes

**Why CPU-side pre-sampling?**
`torch.Generator` only works on CPU/CUDA — it can't run on the TT XLA device. The approach follows the pattern from commit 98a7c425c (CPU-side exponential pre-sampling on TPU): build the noise tensor on CPU using per-request generators, then pass it as a regular input to the compiled XLA graph. The compiled graph just does `probs / q_samples` — no on-device RNG needed for seeded requests.

**Common case (no seeds) is unchanged.**
When no requests have seeds, `q_samples` is `None` and `no_generators` is `True`. The sampler takes the existing on-device `exponential_()` path. This does mean seeded and non-seeded requests compile separate graph variants (due to the `if q_samples is not None:` branch in `random_sample()`), but the seeded graph is compiled lazily on first use — no extra startup cost.

**No precompile for the seeded graph.**
We intentionally skip pre-compiling the seeded graph variant during warmup. Most deployments never use `seed=N`, so the extra compilation time at startup isn't justified. The first seeded request triggers lazy compilation. A precompile pass can be added later if the latency spike matters.

**Generator rewind removed.**
The existing GPU-inherited code at model_runner.py:1333 tried to call `generator.set_offset()` to rewind state for partial (prefill) requests. This is CUDA-specific and crashes on CPU generators. On our path, generators are consumed during `from_input_batch()` (not during `random_sample()`), so no rewind is needed — `q_samples` are rebuilt fresh each decode step.

Issue #3221, opened 2026-02-09
 KeyError: c_lifted_tensor_1


devisettymahidhar608:
**Describe the bug:**

- While running the mixtral/pixtral model, I encountered the following error
```
x = c_lifted_tensor_1

>   env[cur] = new_graph.node_copy(cur, lambda x: env[x])
                                                  ^^^^^^
E   KeyError: c_lifted_tensor_1
```
Issue originate from the [generate_block_attention_mask](https://github.com/huggingface/transformers/blob/main/src/transformers/models/pixtral/modeling_pixtral.py#L519)- 

**Repro:**

```
git checkout[devisettym/pixtral_1](https://github.com/tenstorrent/tt-xla/tree/devisettym/pixtral_1)
git submodule update --init —recursive

# sanity run
pytest -svv  tests/torch/graphs/test_pixtral.py 
```

**Observations:**

When function containing multiple torch.tensor() calls, the PyTorch XLA CPU fallback partitioning fails with KeyError: c_lifted_tensor_1 during graph legalization. The error occurs specifically when the second (or subsequent) lifted tensor node is being copied to the new graph environment.

compiled graph before failure:
```
graph():
    %lifted_tensor_0 : [num_users=1] = get_attr[target=lifted_tensor_0]
    %mark_argument_attributes : [num_users=1] = call_function[target=torch.ops.tt.mark_argument_attributes](args = (%lifted_tensor_0,), kwargs = {argument_type: constant, name: lifted_tensor_0})
    %lifted_tensor_1 : [num_users=1] = get_attr[target=lifted_tensor_1]
    %mark_argument_attributes_1 : [num_users=1] = call_function[target=torch.ops.tt.mark_argument_attributes](args = (%lifted_tensor_1,), kwargs = {argument_type: constant, name: lifted_tensor_1})
    %clone : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%mark_argument_attributes,), kwargs = {})
    %cumsum : [num_users=1] = call_function[target=torch.ops.aten.cumsum.default](args = (%clone, -1), kwargs = {})
    %clone_1 : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%mark_argument_attributes_1,), kwargs = {})
    %cumsum_1 : [num_users=1] = call_function[target=torch.ops.aten.cumsum.default](args = (%clone_1, -1), kwargs = {})
    return (cumsum, cumsum_1)

```

stack trace:

```
[...]

PR #3529, opened 2026-03-02
Add a call to legalize_graph in new compile

sgligorijevicTT:
### Ticket
Addresses one part of #3221 

### Problem description
Prev and next pointers seem to get mangled somehow before reaching the low overhead compile code. The graph partitioning and merging back logic relies on these pointers being valid and up to date while processing the graph. As in our case they aren't this leads to deletion of nodes that shouldn't be deleted, which subsequently causes values required for execution to go missing.

### What's changed
Added a call to `legalize_graph` from torch fx. It seems like a decently light weight function. It in place rebuilds the graph to be in topological order and with valid pointers.

### Checklist

### DIFF
--- a/python_package/tt_torch/backend/backend.py
+++ b/python_package/tt_torch/backend/backend.py
@@ -199,10 +199,12 @@ class XLAExecutor:
             # inlined in the function signature (torch calls this "lifting" the arguments). Exporting does this.
             program = torch.export.export(self.module, tuple(args), strict=False)
 
-            # torch.export may leave erased "zombie" nodes in the graph's linked
-            # list with inconsistent prev/next pointers.  legalize_graph rebuilds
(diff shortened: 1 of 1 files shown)


python_package/tt_torch/backend/backend.py
```diff
@@ -198,6 +199,12 @@ def _call_experimental_compile(self, *args):
             # inlined in the function signature (torch calls this "lifting" the arguments). Exporting does this.
             program = torch.export.export(self.module, tuple(args), strict=False)
 
+            # torch.export may leave erased "zombie" nodes in the graph's linked
```
pilkicTT: do we know what creates these "zombie" nodes? what is the purpose of them? can we add some assertions (under a flag) to make sure we are not creating them?

Issue #3158, opened 2026-02-04
Create benchmarking flow throught vLLM


AleksKnezevic:
Create a benchmark flow through vLLM, similar to what's done in tt-forge. We need the ability to:

- set custom compile options for the test
- consistently run decode with a given batch
- parameterize on model

AleksKnezevic:
completed

PR #3554, opened 2026-03-03
Implement vLLM benchmarks

btsoiTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3158

### What's changed
Implement vLLM benchmarks. The benchmarks are configured through the `VLLMBenchmarkConfig` object.
2 tests are added: llama-3.2-3b and llama-3.2-3b-batch32.

### Checklist

### DIFF
--- a/tests/benchmark/benchmarks/vllm_benchmark.py
+++ b/tests/benchmark/benchmarks/vllm_benchmark.py
@@ -256,4 +256,5 @@ def benchmark_vllm(
         input_sequence_length=config.max_model_len,
         device_count=device_count,
         mesh_shape=mesh_shape,
+        vllm=True,
     )
--- a/tests/benchmark/test_vllm_benchmarks.py
+++ b/tests/benchmark/test_vllm_benchmarks.py
@@ -30,7 +30,9 @@ SINGLE_DEVICE_CONFIGS = [
 
 
 def _run_vllm_benchmark(config, output_file, request):
-    display_name = resolve_display_name(request=request, fallback=config.model)
+    display_name = "vllm_" + resolve_display_name(
+        request=request, fallback=config.model
+    )
 
     print(f"\n{'='*60}")
     print(f"vLLM Benchmark: {display_name}")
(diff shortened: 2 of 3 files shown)


integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2007,6 +2013,24 @@ def sample_from_logits(
...
+            sampling_metadata.all_greedy
+            and sampling_metadata.no_penalties
+            and sampling_metadata.no_logit_bias
+            and sampling_metadata.no_bad_words
+        ):
+            out_tokens = torch.argmax(logits, dim=-1, keepdim=True)
+        else:
+            out_tokens = self.sampler(logits, sampling_metadata).sampled_token_ids
+        return out_tokens
+
```
kmabeeTT: (Edit: Typo fix)

Cool workaround, should prepend it with an issue comment here or somewhere.... but this is something else for us to maintain while it exists... by that I mean a merge a couple days ago #3487 updated the original `sample_from_logits()` function to add new condition that this function is missing (and more PR's may add new conditions):

```
            and sampling_metadata.no_generators
```

So I'm thinking we could simplify this to just use torch.argmax() greedy sampling always and don't invoke self.sampler() or support it in the benchark flow for now until this OOM bug gets debugged+fixed and we can return back to on-device sampling for high batches.  Discussed with Aleks and Asif and they agree.  

```
    def sample_from_logits_cpu(
        self, logits: torch.Tensor, sampling_metadata: XLASupportedSamplingMetadata
    ) -> torch.Tensor:
        """
        Sample on CPU instead of compiling a device sampling graph.
        """
        logits = logits.cpu()
        # Only support greedy sampling on CPU for now while this temporary function exists 
        # to reduce maintenance overhead.
        out_tokens = torch.argmax(logits, dim=-1, keepdim=True)
        return out_tokens
```

For completeness, we can also make this change in `tests/benchmark/benchmarks/vllm_benchmark.py` so it would naturally go down the greedy path anyways now and in the future, for deterministic behavior:

```
-    sampling_params = vllm.SamplingParams(max_tokens=config.max_tokens, ignore_eos=True)
+    sampling_params = vllm.SamplingParams(
+        max_tokens=config.max_tokens, ignore_eos=True, temperature=0.0
+    )
```

It doesn't solve the OOM failure though, so still need the `sample_from_logic_cpu()` temporary workaround function.

vkovacevicTT (approved):

Issue #3577, opened 2026-03-04
Add fusion graph tests


sgligorijevicTT:
Initial batch of patterns that will get unit tests:

Patterns from `TTIRFusing.cpp`:
- [x] ConvAddBias 
- [x] ReductionWithReshapePattern
- [x] SoftmaxFusionPattern 
- [x] NumericStableSoftmaxFusionPattern
- [x] ReluFusionPattern 
- [x] Relu6FusionPattern 
- [x] HardsigmoidFusionPattern 
- [x] SiluFusionPattern 
- [x] MishFusingPattern 
- [x] ConvWithMultiply 
- [ ] ConvTagWeights (won't test, doesn't replace op with another op, only adds an attr, that might end up as a different layout on other things which is too hard to test)
- [x] BatchNormDecomposition
- [x] RepVGGConvSumFusionPattern 
- [x] MatmulWithBiasFusionPattern 
- [x] ScaledSumToMeanPattern 
- [x] SpatialMeanOptimizationPattern 
- [x] ConcatenateHeadsUpdatePattern 
- [x] SplitQueryKeyValueAndSplitHeadsUpdatePattern
- [x] GeluFusionPattern 
- [x] RMSNormFusionPattern 
- [x] PermuteMatmulFusionPattern 
- [x] ReshapeBroadcastReshapeToRepeatPattern

Patterns from `TTNNFusing.cpp`:
- [x] TTNNConv2dWithActivation
- [x] TTNNMatmulAndLinearWithActivation
- [x] SDPAFusing

Issue text will be updated as patterns get tests added and if the the issue is rescoped

Scope was slightly expended to add rudimentary filechecks to most important models, and add some separate graph/"semi-integration" tests

PR #3600, opened 2026-03-05
Add fusion tests

sgligorijevicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-xla/issues/3577

### Problem description
Fusion patterns are fragile, least we could do is add tests to be aware when they inevitably break

### What's changed
Added some fusion unit tests.
Added a realworld attention test that exercises all attention relevant patterns
Added a few more filecheck directives
Also tweaked the testinfra to allow compile options in graph tests(required for graph tests with optimizer, itself required for exercising SDPA fusion)

Most tests are nightly only, but I marked a few with push, it shouldn't exacerbate the onPR duration problems and should offer adequate coverage. 

### Context
(thanks to @acicovicTT for suggesting to add this)
We already have llvm FileCheck plumbed trough our repo, for both end to end model tests and individual tests. The infrastructure takes in a filename from the corresponding pytest marker(calling it exactly `*.ttir.mlir` or `*.ttnn.mlir` is load bearing) and uses them to find the corresponding files with FileCheck directives and check them against the `.mlir` files emitted by that test. We use this to check that sequences of ops that should have been fused away don't appear, and that ops only achievable trough fusion do appear.

Most tests are barebones unit tests, that is: just enough (commonsense) torch code so that the pattern triggers. 
There is a separate subfolder with a "realworld" test, I have extracted the entire attention block from 2 models and check for occurrences of all relevant fusion patterns. 
Code had to be duplicated as one specific pattern, SDPA, requires optimizer to be on and I didn't want to turn it on in the existing attention tests. This is also why the testinfra had to be tweaked.
In addition to that, I have added filecheck markers to some existing graph tests where we do observe certain fusions.(bikeshedding point: I only marked a few most important ones, just enough to catch ~every break IMO, should I have added filechecks to all tests?)

Furthermore, I expanded filechecks applied to one end to end model test and filecheck directives for a few more(bikeshedding point: should I have added filechecks to more models?)

Reviewer guide:
See python files directly in `tests/torch/graph/fusion_patterns/` first. Filecheck `.mlir` files in `tests/filecheck` should be up next. Next the realworld test and modifications to existing graph test. Then infra tweak and test config changes.

### Checklist

### DIFF
--- a/tests/torch/graphs/fusion_patterns/test_concatenate_heads.py
+++ b/tests/torch/graphs/fusion_patterns/test_concatenate_heads.py
@@ -9,6 +9,7 @@ from infra import Framework, run_graph_test_with_random_inputs
 from utils import Category
 
 
+@pytest.mark.extended
 @pytest.mark.nightly
 @pytest.mark.single_device
 @pytest.mark.record_test_properties(category=Category.GRAPH_TEST)
@@ -28,6 +29,7 @@ def test_concatenate_heads_3d(request):
     )
 
 
+@pytest.mark.extended
 @pytest.mark.nightly
 @pytest.mark.single_device
 @pytest.mark.record_test_properties(category=Category.GRAPH_TEST)
--- a/tests/torch/graphs/fusion_patterns/realworld/test_attn.py
+++ b/tests/torch/graphs/fusion_patterns/realworld/test_attn.py
@@ -37,6 +37,7 @@ SEQ_LEN = 1024
 
 
 @pytest.mark.push
+@pytest.mark.extended
 @pytest.mark.nightly
 @pytest.mark.single_device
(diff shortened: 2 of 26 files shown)


tests/torch/graphs/fusion_patterns/test_concatenate_heads.py
```diff
@@ -0,0 +1,47 @@
...
+import pytest
+import torch
+from einops import rearrange
+from infra import Framework, run_graph_test_with_random_inputs
+from utils import Category
+
+
+@pytest.mark.nightly
+@pytest.mark.single_device
+@pytest.mark.record_test_properties(category=Category.GRAPH_TEST)
```
acicovicTT: Should we have a new category for unit tests?

PR #3606, opened 2026-03-05
Use experimental_enable_weight_bfp8_conversion in inference testers

### LINKED ISSUE #3163: Torch resnet inference [optimization_level_0-bfp8] fail
**Note:**
On new main error is:
```
Always | TT_FATAL: batch_norm only supports bfloat16, float32. Got: DataType::BFLOAT8_B
```

Repro:
```
pytest -svv tests/torch/models/resnet/test_resnet.py::test_torch_resnet_inference[optimization_level_0-bfp8]
```

### ...
never finishes executing if tt-mlir is on commit or newer:
Remove the generic ttir::PoolingOp and update SHLO-TTIR conversion pass (https://github.com/tenstorrent/tt-mlir/pull/6573)

Workflow runs that tested this:
https://github.com/tenstorrent/tt-xla/actions/runs/21684122324
https://github.com/tenstorrent/tt-xla/actions/runs/21684139211

### LINKED ISSUE #3595: Use experimental_enable_weight_bfp8_conversion instead of enable_bfp8_conversion in create_torch_inference_tester and create_jax_inference_tester
enable_bfp8_conversion flag causes whole graph (including activations and all intermediate tensors) to be in bfp8 data format. This creates bunch of issues since bfp8 data format requres tensors to be in tiled layout, and many ops require row major... Additionally, having all activations and intermediate tensors in lower precision  doesn't have much sense since it regresses the accuracy significantly.

On the other hand, experimental_enable_weight_bfp8_conversion flag triggers only weghts casting to bfp8, and this is used on many places already in the stack.

It is worth noting that `experimental_enable_weight_bfp8_conversion` currently casts only matmul/linear weights, and not convolution weights. Therefore, switching to bfp8 weights on convolutional models won't have any practical effect. Nevertheless, wegiht only casting is becomming default path, so we should start cleaning up `enable_bfp8_conversion` flag where we can.

TODO:
- See how create_torch_inference_tester and create_jax_inference_tester are used in testing infra.
- Exchange flag enable_bfp8_conversion for experimental_enable_weight_bfp8_conversion
- Remove xfail flags from tests that now pass.

kdimicTT:
### Ticket
Closes #3595
Closes #3163

### Problem description
Inference testers used `enable_bfp8_conversion`, causing the full-graph to be in bfp8 format. This leads to issues with ops which don't support block formats, while also causing accuracy regression.

### What's changed
Switched inference testers to use `experimental_enable_weight_bfp8_conversion` that triggers only weights casting to bfp8. Enabled bfp8 in MNIST MLP and ResNet tests.

## Note
`experimental_enable_weight_bfp8_conversion` currently casts only matmul/linear weights. It does not support convolution weights casting because of issue https://github.com/tenstorrent/tt-mlir/issues/5995

### Checklist


dgolubovicTT (approved):
Cool

Issue #3619, opened 2026-03-06
GLM graph tests


AleksKnezevic:
Please add attention and MoE graphs for GLM. Attention should be added immediately and MoE once @sshonTT MoE branch lands.

PR #3633, opened 2026-03-06
Add GLM attention graph tests

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3619

### Problem description
We want to add GLM attention tests.

### What's changed
Added tests for GLM attention prefill, attention decode and attention following the format of existing attention tests. Differences for GLM are:
- prefill doesn't fit on a single device and so it's only run on an llmbox
- attention_bias=True, so the shard spec includes q/k/v_proj.bias entries
- GLM uses partial rotary embeddings (`partial_rotary_factor=0.5`), so `cos_sin` size uses `rope_dim = head_dim × partial_rotary_factor` rather than `head_dim`

### Checklist

### DIFF
--- a/tests/torch/graphs/test_attention.py
+++ b/tests/torch/graphs/test_attention.py
@@ -2466,7 +2466,7 @@ def test_gpt_oss_attention_decode(variant, variant_config, arch):
 
 @pytest.mark.nightly
 @pytest.mark.llmbox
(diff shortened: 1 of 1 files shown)


gengelageTT:
Tests are passing locally and running in CI here: https://github.com/tenstorrent/tt-xla/actions/runs/22784820997

Issue #3186, opened 2026-02-05
How to deal with sliding window attention


AleksKnezevic:
Current implementation of gpt-oss uses static attention for all layers, however the model supports sliding window. We need to figure out how to support sliding window in order to run larger sequence lengths.

PR #3688, opened 2026-03-11
Adding support for sliding window attention

jazpurTT:
### Ticket
#3186 

### Problem description
Lack of support for models using sliding window attention. When running a model (e.g. GPT-OSS, Mistral) that supports sliding attention the cache uses the `StaticSlidingWindowLayer` class for the decode layers using sliding attention. The current version of `StaticSlidingWindowLayer` triggers torch.compile() to recompiles on every decode step. The recompile limit is hit quickly and generation is effectively uncompiled or very slow.
`StaticSlidingWindowLayer` keeps `cumulative_length` as a Python `int` and updates it each step `self.cumulative_length += key_states.shape[-2]`. It is used in:
- `update` – for branching (is_full, cumulative_length + key_states.shape[-2] > self.max_cache_len, etc.).
- `get_mask_sizes` – which returns (kv_length, kv_offset) as Python ints derived from self.cumulative_length.

`torch._dynamo` treats these as constants and installs guards on their concrete values. Each decode step changes cumulative_length (e.g. 71 → 72 → 73…), so the guards fail and the graph recompiles for every model forward pass.

### What's changed
I opened an Issue/Feature Request to address this https://github.com/huggingface/transformers/issues/44609 in the hugginface repo but there seems to be no intention to change it. From their perspective making the compilation dynamic `torch.compile(mode, dynamic=True)` solves this but that doesn't work for our `tt` backend compilation. 

The solution I am proposing is:
1. Replace `StaticSlidingWindowLayer` with our own custom class where`get_mask_sizes()` returns constants `(max_cache_len, 0)` and the `update()` function:
     -  When decoding (1 new token) always rolls the KV buffer left by 1 positions and writes the new token to the rightmost slots,
    - and on pre-fill/new prompt, appends the new tokens to the back of the cache and return that tensor for attention, we slice out needed portion from that tensor and store it in the cache. This is necessary in order to ensure we don't lose any context in the sliding window. This entails using a tensor on attention that is larger than necessary when the cache is not yet full (leading empty slots of the cache).
2. The changes made to `StaticSlidingWindowLayer` forces us to replace `create_sliding_window_causal_mask` with a custom mask building function to accommodate the changes.

There's a potential alternative on pre-fill//new prompt where we slice out the leading zeros and we have a tensor that's not larger than necessary, but in order to do so we'd need to re-compile (ONLY) on every pre-fill step until the cache is full.

### Checklist
- [x] Compiled and ran examples/pytorch/gpt_oss_20B.py with sliding attention for 180+ forward passes with only two compiled graphs, one for the pre-fill pass and one for decode passes.
- [x] Compiled and ran mistralai/Ministral-8B-Instruct-2410 200+ forward passes with only two compiled graphs, one for the pre-fill pass and one for decode passes.


pglusacTT (approved):
For training this looks okay 👍

AleksKnezevic (approved):

PR #3716, opened 2026-03-13
Fix sfpi option passing to mlir

nsumrakTT:
### Problem description
Sfpi option is not passed to tt-mlir, where default is changed

### What's changed
Add option to tt-xla and pass it to tt-mlir

### Checklist


nsumrakTT:
This has to go in or release will be ruined, force merging.

Issue #1714, opened 2025-10-16
How do we evaluate top-k output


AleksKnezevic:
Using PCC to validate top-k output does not work well. 

From this [PR](https://github.com/tenstorrent/tt-xla/pull/1702#issuecomment-3410712871) we see:

```
Golden output:

tensor([[26., 31., 37., 21.,  7.]])
Device output:

tensor([[26., 31.,  5., 21.,  7.]])
```
produces a PCC of 0.22 even though the results are reasonable. We should see if there's another potential way to validate top-k (or perhaps all integer) outputs. There could be schemes such as contains(1-N), same top1 value, etc. 

The task is to investigate what is used in industry and recommend a solution.

PR #3730, opened 2026-03-15
Add custom comparator for testing topk op

hshahTT:
### Ticket
Fixes https://github.com/tenstorrent/tt-xla/issues/1714

### Problem description
The default evaluation method (just PCC) doesn't work well for evaluating the topk op's indices tensor since both the CPU and TT device can give different valid top K indices.

### What's changed
- Added support for a custom comparator function to the OpTester, GraphTester, and ModelTester classes. This allows for more advanced control over how to define success criteria for ops.
- Added a custom comparator for the topk op that does the following:
    - Calculates the PCC between the golden and device value tensors returned by the topk op. This should always have good PCC
    - Asserts that the returned device indices tensor does not have duplicate values.
    - Gathers the input tensor based on the returned device indices tensor compares the cosine similarity of the result with the golden values tensor. This ensures that the returned indices tensor is valid.

### Checklist

### DIFF
--- a/tests/infra/testers/single_chip/model/torch_model_tester.py
+++ b/tests/infra/testers/single_chip/model/torch_model_tester.py
@@ -5,7 +5,7 @@
 import collections
 import os
 from contextlib import contextmanager
-from typing import Any, Dict, Mapping, Sequence, Set, Tuple
+from typing import Any, Callable, Dict, Mapping, Optional, Sequence, Set, Tuple
 
 import torch
 import torch_xla
@@ -77,6 +77,7 @@ class TorchModelTester(ModelTester):
         compiler_config: CompilerConfig = None,
         parallelism=None,
         dtype_override=None,
+        custom_comparator: Optional[Callable] = None,
     ) -> None:
 
         self._input_activations: Dict | Sequence[Any] = None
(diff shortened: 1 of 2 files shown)


tests/infra/testers/single_chip/op/op_tester.py
```diff
@@ -41,12 +41,14 @@ def __init__(
         framework: Framework = Framework.JAX,
         compiler_config: CompilerConfig = None,
         torch_options: dict = None,
+        custom_comparator: Optional[Callable] = None,
```
AleksKnezevic: can this be added to the `BaseTester` class and have op, model and graph inherit it? Seems like it would be useful.

tests/torch/ops/test_topk.py
```diff
@@ -4,10 +4,67 @@
 
 import pytest
 import torch
+from benchmark.utils import compute_pcc
 from infra import Framework, run_op_test_with_random_inputs
 from utils import Category
 
 
+def topk_indices_comparator(device_output, golden_output, args, kwargs):
```
jameszianxuTT: I have CPU-only unit tests at tests/infra_tests/single_chip/test_result_assertion.py that mock device/golden input to check that the comparators themselves are working as expected with both positive and negative tests.

It might be nice to add some tests there for the custom comparator to make sure no regressions creep in if this ever gets changed.

PR #3760, opened 2026-03-17
Adding script for building torch_xla and linking with tt-xla

ajakovljevicTT:
Adding a script for automatically cloning and building (along with all required dependencies) `torch_xla` in `tt_xla`, based on what is written here: https://gist.github.com/vkovinicTT/dce22ed9d94065d3c64c61cf1805ef14 by @vkovinicTT, and updating it for the newest versions of python (3.12) and torch (2.9.1) that we use. That doc is now also added as a prt of the documentation. The cloned torch and torch_xla repo will be in the `temp` folder in root, which is now added as part of .gitignore.

### DIFF
--- a/docs/src/torch_xla_build.md
+++ b/docs/src/torch_xla_build.md
@@ -11,260 +11,33 @@ This guide covers building PyTorch and PyTorch-XLA (Tenstorrent fork) from sourc
 - **Bazel**: 7.4.1
 - **Total Time**: ~2-2.5 hours (first build)
 
-## Automated Build
+## Build
 
-A script automates the entire process:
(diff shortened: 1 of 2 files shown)


AleksKnezevic (approved):

ajakovljevicTT:
> > I like the idea, did you mean it as a standalone, manually triggered job? As this build takes a long time, I don't think it should be in nightly or onPR.
> 
> Yes, exactly. I also don't think we should do it regularly. But since we have the script, I feel that we are 90% there to having a manual workflow.
> 
> Just a note if this does get implemented (maybe in a later PR) - Github doesn't allow you to preview or run new actions that don't exist in main. They won't show up in the actions sidebar. Instead, you can hijack an existing action (like manual-update-durations) and overwrite it for testing, you can add options to the manual dispatch and they will show up when you switch to your branch :)

Thanks @jameszianxuTT , will talk with @vvukomanTT about this. I forgot that I set this on auto merge, so will be in some next PR!

Issue #3507, opened 2026-02-27
Host memory leak between torch models due to experimental compile (250GB in gpt-oss-120B on galaxy) 


kmabeeTT:
A general problem seen even in smaller models too, but notably seen here when trying to run models on WH Galaxy, where gpt-oss-120b finishes nicely, but reports as memory usage on system still using 366 GB, then next test (llama3.1 70B) dies due to oom Exit Code 137:

https://github.com/tenstorrent/tt-xla/actions/runs/22492600763/job/65161236444

```
PASSED
2026-02-27 16:05:26.275 +00:00 | INFO     | tests.conftest:memory_usage_tracker:400 - Test memory usage:
2026-02-27 16:05:26.280 +00:00 | INFO     | tests.conftest:memory_usage_tracker:401 -     By test: 231360.66 MB
2026-02-27 16:05:26.280 +00:00 | INFO     | tests.conftest:memory_usage_tracker:402 -     Minimum: 146316.31 MB
2026-02-27 16:05:26.280 +00:00 | INFO     | tests.conftest:memory_usage_tracker:403 -     Maximum: 377676.97 MB
2026-02-27 16:05:26.280 +00:00 | INFO     | tests.conftest:memory_usage_tracker:404 -     Average: 312522.54 MB
2026-02-27 16:05:27.363 +00:00 | INFO     | tests.conftest:memory_usage_tracker:416 - Memory usage after garbage collection: 375250.23 MB

tests/runner/test_models.py::test_all_models_torch[llama/causal_lm/pytorch-3.1_70B-tensor_parallel-inference] Running tests/runner/test_models.py::test_all_models_torch[llama/causal_lm/pytorch-3.1_70B-tensor_parallel-inference] - pytorch_Llama_3.1_70B_nlp_causal_lm_huggingface
`torch_dtype` is deprecated! Use `dtype` instead!
WARNING:transformers.configuration_utils:`torch_dtype` is deprecated! Use `dtype` instead!

Loading checkpoint shards:   0%|          | 0/30 [00:00<?, ?it/s]
Loading checkpoint shards:  90%|█████████ | 27/30 [00:00<00:00, 262.64it/s]
Loading checkpoint shards: 100%|██████████| 30/30 [00:00<00:00, 270.09it/s]
[...]

Issue #2821, opened 2026-01-13
[Experimental Compile Torch] Understand and handle `InputKind`: `CUSTOM_OBJ` and `TOKEN`


sgligorijevicTT:
Agreed to be split up into a separate issue in https://github.com/tenstorrent/tt-xla/pull/2734#discussion_r2682112717

The two InputKinds into torch fx graph are entirely untested and we don't yet understand what they represent.

PR #3795, opened 2026-03-18
[torch] AOTAutograd on by default

### LINKED ISSUE #3077: Enable AOTAutograd for torch compile
To merge training with the "experimental" compile flow, we need to add AOTAutograd to the backend.

### LINKED ISSUE #3572: Usage of `fake_tensor_unsupported`
More of an issue documenting a workaround than an... issue.

AOTAutograd under the hood uses FakeTensor mode, as does torch.export(which we invoke as part of our fw_compiler), and those two will conflict unless we signal to torch to defakeify the tensors that reach the fw_compiler. We should probably investigate the implications of this.

### LINKED ISSUE #3746: Default PyTorch backed using AOTAutograd

sgligorijevicTT:
HAS PERF IMPLICATIONS, FYI.

### Ticket
closes https://github.com/tenstorrent/tt-xla/issues/3746
closes https://github.com/tenstorrent/tt-xla/issues/3077
closes https://github.com/tenstorrent/tt-xla/issues/3572

### Problem description
We want to support as much user code as we can, and ideally not behind any flags. For that we need AOTAutograd on by default.

### What's changed
**BIG CHANGE** torch.export was ripped out on the AOTAutograd path. In general the torch_pass_pipeline was implicitly branched into two, in one path it accepts metadata from AOTAutograd and avoids torch.export entirely(AOTAutograd handles functionalization, decomposition and mutation detection) and the other is basically our old logic where torch.export was handling everything.
`use_aot_autograd` flag now defaults to on
An extra hack was added, namely to allow mixed fake and real tensors inside the FakeMode passed to us from Dynamo. Despite being more code and uglier on the surface, this is much less of a hack than having a `fake_mode_unsupported` decorator on our fw_compiler. This has the potential to remove the root cause of https://github.com/tenstorrent/tt-xla/issues/3507 but I haven't verified that yet, FYI @kmabeeTT

There are also misc changes regarding name demangling in torch. I was getting extra errors about failed demangling, but I noticed the names seemed to be mangled on main too, therefore I changed the code.

One consequence of adding AOTAutograd is that graphs can change, so far I mostly noticed differences in reshapes, usually extra unsqueeze-squeeze or flattening-unflattening pairs appearing or disappearing.

### Checklist
- [ ] Potentially another once-over on the code, PR is heavily AI assisted. 

PERF STATUS:
https://github.com/tenstorrent/tt-mlir/pull/7632 
https://github.com/tenstorrent/tt-mlir/pull/7599
https://github.com/tenstorrent/tt-mlir/pull/7566 
https://github.com/tenstorrent/tt-mlir/pull/8647
https://github.com/tenstorrent/tt-mlir/pull/8448
https://github.com/tenstorrent/tt-mlir/pull/8383
and a couple of other mlir prs were all results of trying to not regress perf
To the best of my knowledge, no regressions will happen after landing this, final run still pending


python_package/tt_torch/backend/decompositions.py
```diff
@@ -434,6 +434,14 @@ def populate_decompositions() -> DecompositionTable:
     decompositions.pop(torch.ops.aten.sum.default)
     decompositions.pop(torch.ops.aten.sum.out)
 
+    # The core decomposition for aten.detach rewrites it as aten.alias. AOTAutograd
+    # inserts detach() around saved-for-backward tensors as markers; rewriting them
+    # to alias() (a view) confuses the downstream FunctionalTensor / torch-xla
+    # extract_compiled_graph machinery under SPMD multichip and causes user outputs
+    # to be mapped to the wrong tensors (loss/logits resolve to attention
+    # intermediates). Keep detach as a no-op aten op instead.
+    decompositions.pop(torch.ops.aten.detach.default)
```
jameszianxuTT, line 577: wtf indeed

python_package/tt_torch/backend/backend.py
```diff
@@ -26,64 +41,264 @@
...
+                InputSpec(
+                    kind=InputKind.BUFFER,
+                    arg=TensorArgument(name=node.name),
+                    target=target,
+                    persistent=True,
+                )
+            )
+            if is_mutated:
+                mutated_buffer_targets.add(target)
+        else:
```
pglusacTT, line 235: How do we handle [other input kinds](https://docs.pytorch.org/docs/2.11/user_guide/torch_compiler/export/api_reference.html#torch.export.graph_signature.InputKind)?

python_package/tt_torch/backend/backend.py
```diff
@@ -301,26 +464,83 @@ def aot_backend(
     # THIS IS A HACK https://github.com/tenstorrent/tt-xla/issues/3549
     gm = rewrite_adaptive_avgpool_to_mean(gm)
 
+    # Rewrite F.interpolate(bilinear/nearest) to matmul-based implementation.
+    # AOTAutograd seems to force decompose F.interpolate no matter what, and that op doesn't reach our decomposition table.
+    # The standard decomposition works, but the rest of the stack is overfit on the old form and the new behavior would introduce a perf regression.
+    # You guessed it, THIS IS A HACK (again) https://github.com/tenstorrent/tt-xla/issues/3912.
+    gm = rewrite_interpolate_to_matmul(gm)
```
pglusacTT, line 712: What's stopping us from finding this decomposition?

pglusacTT: Also, should it be handled on torch dispatch level, the same as [matmul/einsum](https://github.com/tenstorrent/tt-xla/blob/242a57151a33b9b85f3d7cc72388a8dde9925e0d/python_package/tt_torch/torch_overrides.py#L8)?

python_package/tt_torch/custom_ops.py
```diff
@@ -5,9 +5,38 @@
...
+
+    @staticmethod
+    def forward(ctx, x, name, pos, id, is_input, attr):
+        # Use _AutoDispatchBelowAutograd to call the real op
+        # without re-entering our Autograd registration and recursing.
+        with torch._C._AutoDispatchBelowAutograd():
+            return torch.ops.xla.mark_tensor(x, name, pos, id, is_input, attr)
+
+    @staticmethod
+    def backward(ctx, grad):
```
pglusacTT: This backward registration process is not the same as for other custom bwd ops, can we unify them?

tests/torch/training/test_moe.py
```diff
@@ -91,15 +91,13 @@ def test_gpt_oss_moe_multichip_backward():
...
 
-    loss = outputs.loss.mean()
+    loss = outputs.loss
 
     loss.backward()
 
     loss_value = loss.item()
     assert loss_value > 0, f"Loss should be positive, got {loss_value}."
 
-    torch_xla.sync(wait=True)
```
pglusacTT, line 100: Are we sure that gradients actually get computed? They can populated with lazy tensors without actual values.

sgligorijevicTT: I'm pretty sure they are. I see >2 graphs being executed without this, one of them returning a lot of intermediates and arguments(that has to be the fwd with saved intermediates) the other returning a lot of intermediates(that has to be the bwd). Besides, dynamobridge and our backend both already explicitly sync. WIth this sync, I see 4 graphs, two of those, one trivial to layout, and one that looks like a joint fwd-bwd. The real question here is where is the extra graph coming from. 
For the sake of testing, I can do an .item() in one of the parameters in the verification below, does that work?

python_package/tt_torch/backend/backend.py
sgligorijevicTT: Try as I might, I could not get TOKEN or CUSTOM_OBJ to ever appear, there is an issue tracking that https://github.com/tenstorrent/tt-xla/issues/2821
Good catch on the CONSTANT_TENSOR, I don't think I have seen that appear either, but that might be worth handling. I can push that later.

python_package/tt_torch/backend/backend.py
sgligorijevicTT: FYI one of the bugs I ran into is that the torch dispatch wasn't (properly) triggering under AOTAutograd so I had to duplicate that exact workaround in this form too. We can investigate why that is, but I don't see the problem with explicit fx passes.

sgligorijevicTT: Another tidbit of important context is that AOTAutograd uses no dispatch in at least one place, and also has arcane rules for when what gets captured vs traced trough, and for when it gets decomposed. It's worth investigating, but do you want to make this PR wait for that?

tests/torch/ops/test_fusion_ops.py
```diff
@@ -42,7 +42,7 @@ def test_llama_rms_norm_fusion(batch_size, seq_len, hidden_size, dtype, request)
 
 @pytest.mark.single_device
 @pytest.mark.push
-@pytest.mark.filecheck(["rms_norm.ttir.mlir"])
+@pytest.mark.filecheck(["rms_norm.ttnn.mlir"])
```
sgligorijevicTT: I got an IRL review here, this is stale and will be reverted

third_party/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 option(USE_CUSTOM_TT_MLIR_VERSION "Flag to use TT_MLIR_VERSION set by the user" OFF)
 
 if (NOT DEFINED TT_MLIR_VERSION OR NOT USE_CUSTOM_TT_MLIR_VERSION)
-    set(TT_MLIR_VERSION "eb9005fa360a80e44607e2dfd4404137b510092e")
+    set(TT_MLIR_VERSION "3a57078757454e8fd68111fc359aaabc38bcb14b")
```
sgligorijevicTT: mark
this should be removed once https://github.com/tenstorrent/tt-mlir/pull/8448 gets into xla main

python_package/tt_torch/custom_ops.py
sgligorijevicTT: Important detail why this can't be exactly like all the others up to the last detail is that mark_tensor is not our custom op, however we can make it close*r* to the current style, I will push a tweak later

python_package/tt_torch/backend/backend.py
```diff
@@ -26,64 +42,333 @@
...
+    # TTIRFusing from matching. Register them as CONSTANT_TENSOR specs here
+    # so the marker pass tags them ``constant``, matching the non-AOT path.
+    for node in gm.graph.nodes:
+        if node.op != "get_attr":
+            continue
+        if not str(node.target).startswith("_tensor_constant"):
+            continue
+        input_specs.append(
+            InputSpec(
+                kind=InputKind.CONSTANT_TENSOR,
```
mstojkovicTT: Discussed offline, review generality of this.

python_package/tt_torch/backend/backend.py
```diff
@@ -26,64 +42,333 @@
...
+                arg=TensorArgument(name=node.name),
+                target=str(node.target),
+            )
+        )
+
+    # AOTAutograd functionalises all mutations, so the forward graph itself
```
mstojkovicTT, line 249: Some composite ops that were `nn.modules` and needed to be "functionalized" could make use out of this

python_package/tt_torch/backend/backend.py
```diff
@@ -26,64 +42,333 @@
...
     handle_composite_ops,
     insert_argument_type_markers,
     rewrite_adaptive_avgpool_to_mean,
+    rewrite_interpolate_to_matmul,
     run_fusion_passes,
 )
 
 
+def _relocate_lifted_constants_to_xla(gm: torch.fx.GraphModule) -> None:
+    """Move the CPU constant buffers AOTAutograd lifted onto ``gm`` to XLA.
```
vkovinicTT, line 61: How does current backend handles this? It makes sense that this should cause graph break as `math.sqrt(hidden_size)` is not torch op.

python_package/tt_torch/backend/backend.py
```diff
@@ -26,64 +42,333 @@
...
+        # resulting host op. See _relocate_lifted_constants_to_xla.
+        _relocate_lifted_constants_to_xla(compiled_graph)
+
+        # AOTAutograd's matmul lowering reshapes (..., M, K) @ (..., K, N) to
+        # (B*H, M, K) @ (B*H, K, N) via a view->bmm->view sandwich.  Under TP
+        # with the head dim sharded, the collapsed bmm form destroys head
+        # sharding and forces SPMD to insert f32 all_gathers on the 4-D
+        # Q/K/V.  Re-expand to a rank-preserving einsum here, after AOTA
+        # has produced the sandwich.
+        fold_view_bmm_view_to_einsum(compiled_graph)
```
vkovinicTT, line 445: Something similar is done in `TorchFunctionOverride` right? Can we add this pass instead of `TorchFunctionOverride` in the regular compile as well then?

python_package/tt_torch/backend/backend.py
```diff
@@ -26,64 +42,333 @@
...
+        # Nodes in the fx graph can have inconsistent prev/next pointers after
+        # torch.export.export.  legalize_graph rebuilds the graph in topological
+        # order from usage information, fixing up the pointers.  This must happen
+        # right after export, before any downstream consumer (e.g. DynamoBridge
+        # Partitioner) can be confused by wrong next-node links.
+        legalize_graph(compiled_graph)
+
+        # Extract params and constants from the ExportedProgram so they can be
+        # passed to bridge.extract_compiled_graph later (experimental compile).
+        params_and_consts = _extract_params_and_consts(program)
```
vkovinicTT: Do we need `_extract_params_and_consts` if this is regular compile path and not experimental?

python_package/tt_torch/backend/metadata_propagation.py
```diff
@@ -313,6 +313,12 @@ def _extract_source_and_module_hierarchy_info(
         for path, class_name in modules:
             DBG_LOC and print(f"    path: {path}")
             DBG_LOC and print(f"    class_name: {class_name}")
+            # AOTAutograd may store actual type objects rather than strings
+            # in nn_module_stack; normalise to a qualified name string.
+            if isinstance(class_name, type):
+                class_name = f"{class_name.__module__}.{class_name.__qualname__}"
+            else:
+                class_name = str(class_name)
```
vkovinicTT: ?? What does this even do? If AOT saves the name of the op, then we might not even need this metadata propagation?

python_package/tt_torch/backend/passes.py
```diff
@@ -11,6 +11,78 @@
...
+            new_node = graph.call_function(
+                replacement_fn,
+                args=new_args,
+            )
+            node.replace_all_uses_with(new_node)
+            graph.erase_node(node)
+            modified = True
+
+    if modified:
+        gm.recompile()
```
vkovinicTT, line 81: Can we use `recompile` after all of the passes are done instead of recompiling in each pass?

tests/torch/graphs/test_mlp.py
```diff
@@ -224,6 +224,7 @@ def test_llama_mlp(seq_len, variant, variant_config, arch, request):
     loader = LlamaModelLoader(variant=variant)
     config = loader.load_config()
     mlp = LlamaMLP(config).to(torch.bfloat16)
+    mlp.forward = torch.inference_mode()(mlp.forward)
```
vkovinicTT, line 230: Does this mean that from now on I will always need to use `torch.inference_mode()` when running forward pass?

sgligorijevicTT: Only if the intermediate saving causes issues. This is exactly how torch works on every other device too

python_package/tt_torch/backend/passes.py
sgligorijevicTT: Why? `recompile` is cheap and I'd strongly prefer to have functions leave graphs in self consistent state, that makes that more reuseable and testable

python_package/tt_torch/backend/backend.py
sgligorijevicTT: Most likely but perhaps that should be left for followup work

python_package/tt_torch/backend/backend.py
sgligorijevicTT: It graph breaks(verified based on logs). Only with AOTAutograd is that constant load bearing for fusions.

python_package/tt_torch/backend/backend.py
sgligorijevicTT: We do. There was a latent correctness bug with legacy and aotautograd(that's a nonsense combination, but technically still supported), will push a fix in a bit

python_package/tt_torch/backend/metadata_propagation.py
sgligorijevicTT: torch export attaches a string, aotautograd attaches a type, so we need to branch here to handle both.
As for the second part of your comment, I'm kind of confused as this is about the module hierarchy not the op name...

vkovinicTT: @svuckovicTT can you please check whether this still looks correct as you are the one who uses these names :))

python_package/tt_torch/backend/backend.py
sgligorijevicTT: I reviewed and found some room for improvement, pushed a (decently sized) change

python_package/tt_torch/backend/metadata_propagation.py
svuckovicTT: Can we keep the old behaviour for when it is not a type?

```suggestion
            # in nn_module_stack; normalise to a qualified name string.
            if isinstance(class_name, type):
                class_name = f"{class_name.__module__}.{class_name.__qualname__}"
```

sgligorijevicTT: Alright!

vkovinicTT (approved):

tests/infra/workloads/workload.py
```diff
@@ -110,7 +111,7 @@ def serialize(self, output_prefix: str, compiler_options=None) -> None:
                         f.unlink()
 
             xr.clear_computation_cache()
-
+            torch._dynamo.reset()
```
mrakitaTT, line 114: It would be good to add comment here to explain why these are needed

mrakitaTT (approved):
Signing off on test infra changes

tests/infra/workloads/workload.py
sgligorijevicTT: dynamo's `reset` and `clear_computation_cache` should always go in pairs, we already have them paired up in multiple other places, otherwise the code runs into a cache miss inside dynamobridge. The real question here is how did this ever work on main, and as far as I can tell, it was because of gradient tracking being misconfigured

python_package/tt_torch/backend/backend.py
sgligorijevicTT: After another comment I further expanded constant tensor handling so I think this basically exhausive now

python_package/tt_torch/custom_ops.py
sgligorijevicTT: rendered invalid anyway by somebody else's more elegant code picked up in a rebase(also I was partially wrong here)

PR #3865, opened 2026-03-23
Static model info fallback writing in case of segfault

ndrakulicTT:
### Ticket
Link to Github Issue

### Problem description
  When model tests run with `--forked`, each test executes in a subprocess. If that subprocess crashes (segfault, abort hang, os._exit()) before the test body completes, pytest-forked generates a crash report with `when="???"` and empty user_properties. This means the junitxml output for crashed tests contains no `tags`(model_name, bringup_status,...) - making them invisible in result dashboards and impossible to triage.

The root cause is a two-part gap:                                                                                                                                                                                                                                                                                                                          
1. record_model_test_properties() never runs because the subprocess dies first, so report.user_properties stays empty.
2. Even if properties were injected onto the crash report, junitxml's finalize() — which transfers user_properties into XML <properties> — is only called for when="teardown" reports. A when="???" crash report never reaches that branch.

### What's changed

Crash-report fallback hook in conftest.py: 
A new pytest_runtest_logreport hookwrapper intercepts crash reports (when="???") and reconstructs model metadata from collection-time data. 
During pytest_collection_modifyitems, a nodeid -> item mapping is built so the parent process can look up the original test item. 

The hook builds the full tags dict (model name, run mode, parallelism, arch, ...), appends it to report.user_properties, and re-classifies report.when from "???" to "teardown" so that junitxml's finalize() writes the properties to XML.

  ModelLoader init moved outside RequirementsManager in test_models.py: ModelLoader() and get_model_info() are now called before entering the RequirementsManager context manager. This ensures these calls are safe to invoke from the parent process (in the conftest hook) without requiring per-model pip dependencies to be installed.

### Checklist
- [ ] To check the PR I have inserted by hand `os._exit(1)` into test to check it crashing, here are three files
1. normal execution of the test with `--forked` [test_output_normal.xml](https://github.com/user-attachments/files/26555021/test_output_normal.xml)
2. same as above but with inserted `os._exit(1)` in the middle of the test [test_output_crash.xml](https://github.com/user-attachments/files/26555022/test_output_crash.xml)
3. same as above but with changes from this PR [test_output_crash_fix.xml](https://github.com/user-attachments/files/26555023/test_output_crash_fix.xml)

### DIFF
--- a/tests/runner/conftest.py
+++ b/tests/runner/conftest.py
@@ -10,9 +10,18 @@ from tests.runner.test_config.constants import ALLOWED_ARCHES
 from tests.runner.test_config.jax import test_config as jax_test_config
 from tests.runner.test_config.torch import test_config as torch_test_config
 from tests.runner.test_config.torch_llm import test_config as torch_llm_test_config
-from tests.runner.test_utils import ModelTestConfig, ModelTestStatus
+from tests.runner.test_utils import (
+    ModelTestConfig,
+    ModelTestStatus,
+    RunPhase,
+    to_marshal_safe,
+)
+from tests.utils import BringupStatus, Category
 
 _BRINGUP_STAGE_FILE = "._bringup_stage.txt"
+
+# Maps nodeid -> item, populated during collection for crash-report fallback.
+_item_by_nodeid: dict = {}
 
 
 def _get_model_group_from_item(item):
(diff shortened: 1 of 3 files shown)


tests/runner/conftest.py
```diff
@@ -177,3 +186,94 @@ def pytest_collection_modifyitems(config, items):
...
+        "failing_reason": {
+            "name": None,
+            "description": None,
+            "component": None,
+            "summary": None,
+        },
+        "guidance": [],
+    }
+
+    report.user_properties.append(("tags", _to_marshal_safe(tags)))
```
kmabeeTT: consider renaming to to_marshal_safe since it's no longer module-private

tests/runner/test_models.py
```diff
@@ -79,14 +79,14 @@ def _run_model_test_impl(
     loader_path = test_entry.path
     variant, ModelLoader = test_entry.variant_info
 
+    # Get the model loader and model info from desired model, variant.
+    loader = ModelLoader(variant=variant)
+    model_info = ModelLoader.get_model_info(variant=variant)
+    print(f"Running {request.node.nodeid} - {model_info.name}", flush=True)
+
```
kmabeeTT: Hmm this part makes me nervous... I can't gaurantee that all models in tt-forge-models with requirements*.txt files can work with this change enough to call `ModelLoader.get_model_info()`... if there are imports in top of loader.py that are not lazily loaded when used (kind of preferred) then I think this would break... can you try it on some sample models, or try modifying one as an example to see?

vzeljkovicTT:
Should we also capture the raw crash message here even unparsed, it would give more context for triage than all-None fields.
Discussion started offline.

Issue #3942, opened 2026-03-25
BERT perf regression with Transformers 5.x: TTNN SDPA verifier rejection and fallback to manual attention


ssaliceTT:
### Describe the bug

When running BERT with the transformers 5.x uplift (`4.57.1` → `5.2.0`), throughput drops from ~65 SPS to ~27 SPS. The root cause is a shape mismatch in the SDPA fusing pass in tt-mlir.

**What changed in transformers 5.x:**
In 4.57.1, `_prepare_4d_attention_mask` helpers built the attention mask using `torch.where`, producing a fully materialized `[B, 1, S, S]` tensor (lowered to a `WhereOp` in the IR). In 5.x, the new `sdpa_mask()` function in `masking_utils.py` instead builds the mask by broadcasting index tensors via `_non_vmap_expansion_sdpa`, producing a `[B, 1, 1, S]` result, then calling `.expand(B, -1, S, S)` ([masking_utils.py:509](https://github.com/huggingface/transformers/blob/main/src/transformers/masking_utils.py#L509)) to get the logical `[B, 1, S, S]` shape. Because `.expand()` is a zero-stride broadcast, PyTorch lowers it to a `RepeatOp` in the IR rather than a `WhereOp`.

**Why the fusing pass breaks:**
`SDPAFusingPattern::prepareMask` unconditionally strips `RepeatOp`s. Thus, the seq-len dim `RepeatOp` introduced by `.expand()` also gets stripped, leaving a `[B, 1, 1, S]` tensor instead of `[B, 1, S, S]`. The TTNN SDPA op verifier then rejects it:

```
loc("dot.4870"): error: 'ttnn.scaled_dot_product_attention' op Attention mask at dim 2 must match query sequence length
```

This causes silent fallback to decomposed manual f32 attention, explaining the ~2.4× perf regression.

**Perf numbers and nightly runs for reference:**

| Scenario | SPS | Nightly Runs |
|---|---|---|
| Before uplift (transformers 4.57.1) | ~32.99 | [Nightly #496](https://github.com/tenstorrent/tt-xla/actions/runs/23222601518/job/67526606911#step:15:94) |
| After uplift (transformers 5.2.0) | ~27.15 | [Nightly #497](https://github.com/tenstorrent/tt-xla/actions/runs/23273415455/job/67717732800#step:15:526)
[...]

PR #3943, opened 2026-03-25
Transformers 5.2 Re-uplift

ssaliceTT:
### Problem description
Originally this PR had merged in a Transformers 5.2.0 uplift: #3371 
Due to some issues with gpt_oss benchmarks it had to be reverted: #3824 
Now that it is resolved, this PR is meant to reuplift

### What's changed
Reverted two PR's that reverted original uplift related changes.

Aside from that made some more changes since original PR to keep up with nightly and fix perf issues:
- added push marker for qwen_2_5 JAX test as onPR has no tests supported with Transformers uplift and this test is in nightly working already. 
- Also set shared_runners to true for `n150 and expected_passing and push and large` to not have non deterministic host OOM issues onPR
- `n300 and nightly and tensor_parallel and expected_passing and not large` in nightly set to 2 parallel groups to avoid any host OOM failures (error code 137 like [here](https://github.com/tenstorrent/tt-xla/actions/runs/23518680501/job/68461660094#step:17:1213)). When tested with two groups there were no such failures consistently, while one would fail non-deterministically but frequently.
- Skipping bge custom encode test as FlagEmbedding version doesnt support Transformers 5.x yet.
- Patch for gpt_oss models: Monkey patch the the gpt oss forward to transformers==4.57.1 which had worked before so perf benchmarking is not broken anymore. The patch will be removed once the root cause of the previous failure is identified and fixed.
- Skipping gpt oss 120B single device mlp test

### Checklist

### DIFF
--- a/python_package/tt_torch/torch_overrides.py
+++ b/python_package/tt_torch/torch_overrides.py
@@ -56,7 +56,7 @@ def _experts_forward(self, hidden_states, router_indices=None, routing_weights=N
     num_experts = routing_weights.shape[1]
     hidden_states = hidden_states.reshape(-1, self.hidden_size)
 
-    if hidden_states.device.type == "cpu" or self.training:
+    if hidden_states.device.type == "cpu":
         next_states = torch.zeros_like(hidden_states)
         with torch.no_grad():
             expert_mask = torch.nn.functional.one_hot(


ssaliceTT:
> @ssaliceTT
> 
> Can you compare it vs latest nightly using https://github.com/rpavlovicTT/gh-perf-report ? And post SS here for reference.

n150-perf
n300-llmbox-perf

The Bert regression was noted earlier and an issue has already been created for it #3942 as Ognjen had requested. 

For the n300-llmbox regressions, I didn't see it before so they must have popped up more recently. I did some initial analysis on the graphs and they're all basically identical aside from a few extra ops here and there in const eval functions. Tried running each locally and there were no regressions either. 
So, I rebased and ran just the regressed tests [here](https://github.com/tenstorrent/tt-xla/actions/runs/23764594730). Here is the report for it. 

rebase-rerun-llmbox-perf

The regressions are gone except in one. I manually verified the sps values too and they are correct. 
So I ran just those tests with the pre rebase commit (same as the 5 regressions) and ran [here](https://github.com/tenstorrent/tt-xla/actions/runs/23767389951). Here is the report for it too. 

og-rerun-llmbox-perf

Also manually verified the sps to be safe. Everything is accurate. Now there's only one regression as well but for a different model. 
Locally I ran some of the regressed models multiple times with and without the uplift and I always got consistent positive results. 
So, I suspect these regressions that showed up are just device/run specific issue and not related to the uplift itself?  
So I would like to go through with the uplift if it's all good @odjuricicTT @rpavlovicTT .

python_package/tt_torch/torch_overrides.py
```diff
@@ -24,3 +24,115 @@ def __torch_function__(self, func, types, args, kwargs=None):
...
+    Args:
+        hidden_states: [T, H] or [B, S, H]
+        router_indices: [T, K]
+        routing_weights: [T, E] full sparse routing weights
+    """
+    batch_size = hidden_states.shape[0]
+    num_experts = routing_weights.shape[1]
+    hidden_states = hidden_states.reshape(-1, self.hidden_size)
+
+    if hidden_states.device.type == "cpu" or self.training:
```
pglusacTT: For the training path we do not want to have this for loop, we want the other case.

PR #3975, opened 2026-03-26
Add per-tensor weight dtype overrides

### LINKED ISSUE #2322: Per parameter dtype
Add support for per-parameter dtype conversion. 

Often times, we want to have attention weights in bf16 while having MoE weights in bfp8 or even 4 or 2. We need a mechanism, to be able to mark inputs with particular dtypes.

### LINKED ISSUE #3659: Support weight dtype overrides on single tensor level in pytorch
As a part of a Manual Mixed Precision effort, we need to have support for overriding weight dtypes in pytorch.
Currently, we only have support for setting global weight dtype override in compiler config. 

Current proposition is to do this the same way sharding is currently passed from pytorch. 
To override the dtype of some tensor user would need to specify it like this: torch.ops.mark_argument_attributes(weight, "bfp4")
This function will annotate weight tensor and we can later pick this up in MLIR and cast that weight to specified data type. Currently, this seems like a pretty clean way of doing it, but I am open to any suggestions. Here is the reference to the[ mark_argument_attributes ](https://github.com/tenstorrent/tt-xla/blob/c5007c0dff9e662ff54161874dbcffa9b0c8ab0b/python_package/tt_torch/custom_ops.py#L14)method.

We should consider supporting loading these kinds of overrides from some json.

dgolubovicTT:
This PR introduces per-tensor weight dtype overriding. We are now able to easily
- create  templated .json per-tensor overrides
- edit them manually
- apply them to the model automatically

### Ticket
Closes #2322. Closes #3659

### Problem description
As part of Mixed Precision effort, we needed per-tensor overrides for weight dtype.

### What's changed
Introduced a `tt::weight_dtype_override` custom op and Python API (`weight_dtype.py`) that enables per-tensor weight dtype control via torch parametrize.

Changes:
- New torch.ops.tt.weight_dtype_override custom op with fake/autograd support
- Added a new C++ frontend pass (`PopulateWeightDtypeFromCustomCall`) that extracts dtype annotations from StableHLO custom calls and propagates them as function argument attributes to be later consumed by tt-mlir.

Python API: 
- dump_weight_names - Generate a JSON template of all weight parameters in a model.
- apply_weight_dtype_overrides - Applies per-tensor weight dtype overrides (passed via .json/dict/string) to a model using torch.nn.utils.parametrize.
- remove_weight_dtype_overrides - Removes  WeightDtypeParametrization instances from a model, preserving any other parametrizations on the same parameters.

- `tt-gen-weight-template` CLI that generates a .JSON config template (using dump_weight_names) used as a starting point for fine-tuning per-tensor weight dtype overrides.
- Rename weight dtype values: bfp8 -> bfp_bf8, bfp4 -> bfp_bf4 across codebase. This aligns with tt-mlir and tt-metal block format notation.
- Auto-apply per-tensor overrides in LLM benchmarks and model test runner
- Multi-chip tests for matmul/linear with weight dtype override + CCL
- Add per-tensor mixed precision documentation to performance guide

### DIFF
--- a/docs/src/performance.md
+++ b/docs/src/performance.md
@@ -19,7 +19,7 @@ This guide covers best practices and techniques for optimizing the performance o
 
 1. **[Optimization Levels](#1-optimization-levels)** - Compiler optimization levels (0, 1, 2) to balance compile and runtime performance
 2. **[Device Warmup](#2-device-warmup)** - Eliminate first-run overhead by performing warmup iterations
-3. **[Data Formats](#3-data-formats)** - Use bfloat16 and bfloat8_b for faster computation and reduced memory usage
+3. **[Data Formats](#3-data-formats)** - Use bfloat16 and bfloat8_b for faster computation and reduced memory usage, including manual mixed precision via per-tensor weight dtype overrides
 4. **[Runtime Trace](#4-runtime-trace)** - Reduce host-device communication overhead by recording and replaying command sequences
 5. **[Batch Size Tuning](#5-batch-size-tuning)** - Find the optimal batch size to maximize throughput for your model
 
@@ -124,14 +124,14 @@ bfloat16 (Brain Floating Point 16-bit) provides:
 
 ### bfloat8_b
 
-Enable bfp8 weight conversion using compile options. The model **MUST** be cast to bfloat16 before compilation.
+Enable bfp_bf8 weight conversion using compile options. The model **MUST** be cast to bfloat16 before compilation.
 ```python
 torch_xla.set_custom_compile_options({
-    "experimental_weight_dtype": "bfp8",  # Cast matmul weights to bfloat8_b
+    "experimental_weight_dtype": "bfp_bf8",  # Cast matmul weights to bfloat8_b
 })
 ```
 
-bfloat8_b (Block Float 8-bit) weight conversion casts matmul weights to bfp8 format, providing faster computation and reduced memory usage.
+bfloat8_b (Block Float 8-bit) weight conversion casts matmul weights to bfp_bf8 format, providing faster computation and reduced memory usage.
 
 #### Notes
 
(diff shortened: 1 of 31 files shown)


docs/src/performance.md
```diff
@@ -140,6 +140,84 @@ bfloat8_b (Block Float 8-bit) weight conversion casts matmul weights to bfp8 for
...
 
+### Per-Tensor Weight Dtype Overrides (Mixed Precision)
+
+When uniform weight conversion causes accuracy degradation in specific layers, you can specify dtype overrides on a per-tensor basis. This lets you keep sensitive layers at higher precision (e.g. `bf16`) while converting the rest to a lower format (e.g. `bfp_bf8` or `bfp_bf4`).
+
+> **Note:** Currently only matmul/linear layer weight overrides are propagated and respected. Convolution weights on lower data types are not yet supported through the compiler.
+
+#### Step 1: Generate a template JSON
+
+Use the `tt-gen-weight-template` CLI to generate a JSON config listing all weight parameters in a model:
```
odjuricicTT: It seems overly complicated to me that an additional CLI tool is needed in order to use basic MP. And this particular CLI tool is tied to tt-forge-models infrastructure which makes it hard to use for users. If possible i think that there should be a way to use MP annotations:
* without tt-forge-modles
* without an additional config file (json)
* with a simple python api similar to how weight sharding is provided

e.g. this is how it looks for sharding:
1. define per weight type mapping in a python dict:
```python
    # Sharded column-wise (head-parallel): [2880, 4096/num_devices]
    shard_specs[layer.self_attn.q_proj.weight] = ("model", "batch")
    shard_specs[layer.self_attn.q_proj.bias] = ("model",)

    # k_proj weight shape: [2880, 512]
    # Sharded column-wise (head-parallel): [2880, 512/num_devices]
    shard_specs[layer.self_attn.k_proj.weight] = ("model", "batch")
    shard_specs[layer.self_attn.k_proj.bias] = ("model",)
```
2. apply the dict mapping
```python
    for tensor, shard_spec in shard_specs.items():
        xs.mark_sharding(tensor, mesh, shard_spec)
```

This is independant of tt-forge-models infra. A full examples for gpt-oss can be found here:
https://github.com/tenstorrent/tt-xla/blob/main/tests/benchmark/test_llms.py#L1097

Issue #3958, opened 2026-03-26
gpt-oss 20b pcc and mlp pcc failure


ddilbazTT:
The following tests failed in nightly:
```
tests/runner/test_models.py::test_all_models_torch[gpt_oss/pytorch-20B-tensor_parallel-inference]
tests/runner/test_models.py::test_all_models_torch[gpt_oss/pytorch-20B-tensor_parallel-inference]
tests/torch/graphs/test_mlp.py::test_gpt_oss_mlp[20B-llmbox-standard]
tests/torch/graphs/test_mlp.py::test_gpt_oss_mlp[20B-llmbox-sparse]
tests/torch/graphs/test_mlp.py::test_gpt_oss_mlp[120B-llmbox-standard]
tests/torch/graphs/test_mlp.py::test_gpt_oss_mlp[120B-llmbox-sparse]
tests/torch/graphs/test_mlp.py::test_gpt_oss_mlp[20B-llmbox-standard]
tests/torch/graphs/test_mlp.py::test_gpt_oss_mlp[20B-llmbox-sparse]
tests/torch/graphs/test_mlp.py::test_gpt_oss_mlp[120B-llmbox-standard]
tests/torch/graphs/test_mlp.py::test_gpt_oss_mlp[120B-llmbox-sparse]
tests/torch/graphs/test_mlp.py::test_gpt_oss_mlp[120B-single_device-standard]
```

Bad commit analysis points to https://github.com/tenstorrent/tt-xla/commit/4c565d84e711a08f9ff354f87a0a3393173fa82a

Locally, I made a branch which reverts this commit and tested `pytest -s tests/torch/graphs/test_mlp.py::test_gpt_oss_mlp[120B-single_device-standard]` which ran. I launched runs to confirm this commit is the culprit, will update this issue when I have more confirmation that this commit is the culprit.

- https://github.com/tenstorrent/tt-xla/actions/runs/23614100661: tests/runner/test_models.py::test_all_models_torch[gpt_oss/pytorch-20B-tensor_parallel-inference] on galaxy
 - https://github.com/tenstorrent/tt-xla/actions/runs/23614186983: tests/torch/graphs/test_mlp.py::test_gpt_oss_mlp[20B-llmbox-standard] on llmbox

Issue #3728, opened 2026-03-15
Add a composite op for `torch.topk`


hshahTT:
### Is your feature request related to a problem? Please describe

Currently, `torch.topk` gets replaced with a series of StableHLO ops (eg. `sort`->`slice`) despite the fact that we have a direct [TTNN replacement](https://docs.tenstorrent.com/tt-metal/latest/ttnn/ttnn/api/ttnn.topk.html).

### Describe the solution you'd like

1. Replace `torch.topk` with a correct stablehlo composite op variant within our tt-xla pass pipelines.
2. Handle these composite ops correctly within tt-mlir's `StableHLOLegalizeCompositePass`

PR #4010, opened 2026-03-30
Add composite op handling for multi-output torch ops, and emit torch.topk as composite

hshahTT:
### Ticket
Fixes https://github.com/tenstorrent/tt-xla/issues/3728
Fixes https://github.com/tenstorrent/tt-xla/issues/3958

### Problem description
Our current composite op pass cannot handle the case where the torch op being replaced has multiple outputs but only a subset of them are actually used within the graph.

### What's changed
- Add composite op handling for multi-output torch ops in which not all outputs are used in the graph.
- Support call_method nodes (when the composite op is used as `input.function(args)` instead of `function(input, args)`.
- Add composite op support for the torch.topk op

### Checklist
- [x] https://github.com/tenstorrent/tt-mlir/pull/7721 is submitted and uplifted to tt-xla


hshahTT:
Note: this PR was previously submitted as https://github.com/tenstorrent/tt-xla/pull/3729, but had to be reverted due to causing PCC regressions in gpt-oss model tests (https://github.com/tenstorrent/tt-xla/issues/3958). 

I've determined the source of those issues and have a PR up in tt-mlir: https://github.com/tenstorrent/tt-mlir/pull/7721. This PR will only be submitted once the fix PR is submitted and the corresponding changes are uplifted to tt-xla.

PR #4020, opened 2026-03-31
[Composite OPs] Add SDPA composite op and tests

mstojkovicTT:
This is the initial push to model `SDPA` trough frontend, which means some features are not supported, primarily modeling of the `attention_sink` that is supported in MLIR ([issue](https://github.com/tenstorrent/tt-xla/issues/4030)) and `dropout_p` which is not modeled (in ttir) for now.

### What's changed

- Added `composite_scaled_dot_product_attention` wrapping for `F.scaled_dot_product_attention`                                                                                                     
- Log a debug for `dropout_p != 0` since dropout is not modeled in TT-MLIR for this op, and fallback to native impl.      
- Added tests
- Added a generic constraint-checking mechanism (`can_apply_composite` in `composite_ops.py`) that `handle_composite_ops` calls before replacing an FX node target with its composite equivalent. This is needed because sdpa kernel in metal currently supports only `bf16` and lower formats, and to account for user wanting to use `fp32`. This can also be used for new composites in the future, and is similar to `match_pattern` functionality in fusing.
- Replaced `torch.rand(...)` attention masks with proper additive causal masks (`0` for attend, `-inf` for mask out) across all attention tests in `test_attention.py`.

### Checklist
- [x] Wait for mlir uplift of https://github.com/tenstorrent/tt-mlir/pull/7728
- [x] Wait for torch_xla uplift of https://github.com/tenstorrent/tt-xla/pull/4033
- [x] Run on-pr with mlir and torch_xla uplift https://github.com/tenstorrent/tt-xla/actions/runs/23874716442

### DIFF
--- a/python_package/tt_torch/composite_ops.py
+++ b/python_package/tt_torch/composite_ops.py
@@ -246,12 +246,6 @@ def composite_scaled_dot_product_attention(
     if scale is not None:
         attr["scale"] = scale
 
-    if dropout_p > 0:
-        logger.warning(
-            "Dropout is not supported for composite scaled_dot_product_attention, setting dropout_p to 0"
-        )
-        dropout_p = 0
-
     builder = StableHLOCompositeBuilder(
         name="tenstorrent.scaled_dot_product_attention", attr=attr
     )
@@ -394,9 +388,19 @@ def replace_group_norm_module(
 def _check_sdpa_constraints(node: torch.fx.Node) -> bool:
     """
     Check that SDPA inputs are bfloat16, the only dtype our composite supports.
-    If not, skip the composite and use the native implementation.
+    Also, check for dropout_p > 0, which is not supported in composite SDPA.
+    If either of these conditions are met, skip the composite and use the native implementation.
     """
-    # Check all positional args and tensor kwargs (e.g. attn_mask)
+    # Dropout is not supported in composite SDPA
+    dropout_p = node.kwargs.get("dropout_p", 0.0)
+    if dropout_p is not None and dropout_p > 0:
+        logger.debug(
+            "composite scaled_dot_product_attention does not support dropout = {dropout_p}, "
+            "skipping composite and using native implementation."
+        )
+        return False
+
+    # Check all inputs are bfloat16
     tensor_args = list(node.args) + [
         v for v in node.kwargs.values() if isinstance(v, torch.fx.Node)
     ]
--- a/python_package/tt_torch/composite_ops.py
+++ b/python_package/tt_torch/composite_ops.py
@@ -7,6 +7,7 @@ from typing import List, Optional, Union
 import torch
 from torch import Tensor
 from torch_xla.experimental.mark_pattern_utils import StableHLOCompositeBuilder
+from ttxla_tools.logging import logger
 
 """
 From XLA documentation: '(Composite operation) Encapsulates an operation made up (composed) of
@@ -244,6 +245,12 @@ def composite_scaled_dot_product_attention(
     attr = {"is_causal": is_causal}
     if scale is not None:
         attr["scale"] = scale
+
+    if dropout_p > 0:
+        logger.warning(
+            "Dropout is not supported for composite scaled_dot_product_attention, setting dropout_p to 0"
+        )
+        dropout_p = 0
 
     builder = StableHLOCompositeBuilder(
         name="tenstorrent.scaled_dot_product_attention", attr=attr
--- a/python_package/tt_torch/composite_ops.py
+++ b/python_package/tt_torch/composite_ops.py
@@ -227,6 +227,48 @@ def composite_topk_indices(
     return indices
 
 
+def composite_scaled_dot_product_attention(
+    query: Tensor,
+    key: Tensor,
+    value: Tensor,
+    attn_mask: Optional[Tensor] = None,
+    dropout_p: float = 0.0,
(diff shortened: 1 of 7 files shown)


mmanzoorTT (dismissed):
We are modelling `SDPA` op as CustomCall [here](https://github.com/tenstorrent/tt-xla/blob/9aff95a55d9808004a231ef43d51f3f6e24d1bc7/python_package/tt_torch/custom_ops.py#L133). Why do we need another definition of this op?

python_package/tt_torch/composite_ops.py
```diff
@@ -339,6 +388,48 @@ def replace_group_norm_module(
...
+    """
+    # Check all positional args and tensor kwargs (e.g. attn_mask)
+    tensor_args = list(node.args) + [
+        v for v in node.kwargs.values() if isinstance(v, torch.fx.Node)
+    ]
+    for arg in tensor_args:
+        if hasattr(arg, "meta"):
+            val = arg.meta.get("example_value", None)
+            if val is not None and val.dtype != torch.bfloat16:
+                logger.debug(
```
nvukobratTT, line 411: We logged a warning for dropout support, but erroring out for bfp16 support. Any specific reason why?

tests/torch/graphs/test_attention.py
```diff
@@ -304,7 +309,12 @@ def get_shard_spec(attention, args, kwargs):
     )
     cos_sin = torch.rand(batch_size, seq_len, config.head_dim, dtype=torch.bfloat16)
     position_embeddings = (cos_sin, cos_sin)
-    attention_mask = torch.rand(batch_size, 1, seq_len, seq_len, dtype=torch.bfloat16)
+    # Additive causal mask: 0 for attend, -inf for mask out
+    # TTIR requires 4D mask: [batch, heads, seq_len, seq_len]
+    attention_mask = torch.zeros(batch_size, 1, seq_len, seq_len, dtype=torch.bfloat16)
```
nvukobratTT: How did this work beforehand?

python_package/tt_torch/composite_ops.py
mstojkovicTT: Not really, i will change both to fallback to native implementation for consistancy

tests/torch/graphs/test_attention.py
mstojkovicTT: Not sure, my best guess is that it didnt use sdpa kernel at all. Also, having random mask doesnt really make sense in this context.

tests/torch/ops/test_composite_ops.py
```diff
@@ -573,3 +574,114 @@ def forward(self, x):
...
+
+
+@pytest.mark.nightly
+@pytest.mark.single_device
+@pytest.mark.parametrize("is_causal", [True, False])
+@pytest.mark.parametrize("use_attn_mask", [True, False])
+@pytest.mark.parametrize(
+    "batch_size, num_heads, seq_len, head_dim",
+    [(1, 1, 32, 32), (1, 8, 64, 64), (2, 4, 128, 64)],
+)
```
sgligorijevicTT, line 586: These tests could use a filecheck directive. Also probably can go in extended too

tests/torch/ops/test_composite_ops.py
```diff
@@ -573,3 +574,114 @@ def forward(self, x):
...
+        attn_mask.masked_fill_(
+            ~torch.ones(seq_len, seq_len).bool().tril(), float("-inf")
+        )
+    else:
+        attn_mask = None
+
+    model = SDPAModel(is_causal)
+    inputs = [query, key, value, attn_mask] if use_attn_mask else [query, key, value]
+
+    with torch._inductor.config.patch({"inplace_buffers": False}):
```
sgligorijevicTT, line 625: What does this do? Why is it needed? What happens if you remove this?

python_package/tt_torch/composite_ops.py
```diff
@@ -227,6 +228,54 @@ def composite_topk_indices(
...
+    Creates composite scaled_dot_product_attention operation for torch xla
+    """
+
+    attr = {"is_causal": is_causal}
+    if scale is not None:
+        attr["scale"] = scale
+
+    if dropout_p > 0:
+        logger.warning(
+            "Dropout is not supported for composite scaled_dot_product_attention, setting dropout_p to 0"
```
sgligorijevicTT: I am not okay with warning and performing incorrect behavior when it's not hard to avoid it. I would strongly prefer to fail to replace instead. Crash would be acceptable too.

tests/torch/ops/test_composite_ops.py
mstojkovicTT: This just tells torch to not reuse input buffers for output, because we are using same inputs in test infra for both golden and tt model

python_package/tt_torch/composite_ops.py
mstojkovicTT: Agreed. Changed to using native implementation if dropout > 0

tests/torch/ops/test_composite_ops.py
mstojkovicTT: Done

sgligorijevicTT (approved):
Code LGTM. We synced offline regarding the broader point of custom call vs composite and duplication of work, but short term this is okay to go in

nvukobratTT (approved):

PR #4060, opened 2026-04-01
Add xla_commit input for XLA checkout in workflow

pdeviTT:
### Problem description
Not able to test specific commit in CI pipeline

### What's changed
Added Commit feature in Run Single Test Pipeline

cc: @nvukobratTT

### DIFF
--- a/.github/workflows/manual-test-single.yml
+++ b/.github/workflows/manual-test-single.yml
@@ -54,12 +54,16 @@ on:
           description: 'Git SHA of commit in tenstorrent/tt-mlir'
           required: false
           type: string
+      xla_commit:
+          description: 'tt-xla commit SHA or branch to checkout, build and test (leave empty to use the branch this workflow runs on)'
+          required: false
+          type: string
 
 permissions:
   packages: write
   checks: write
 
-run-name: 'Test ( Dir: ${{ inputs.dir }} - Mark: ${{ inputs.mark }} - Run-on: ${{ inputs.runs_on }} - Shared runner: ${{ inputs.shared_runners }} )'
+run-name: 'Test ( Dir: ${{ inputs.dir }} - Mark: ${{ inputs.mark }} - Run-on: ${{ inputs.runs_on }} - Shared runner: ${{ inputs.shared_runners }} - XLA commit: ${{ inputs.xla_commit}} )'
 
 jobs:
   build-image:
@@ -67,6 +71,7 @@ jobs:
     secrets: inherit
     with:
       mlir_override: ${{ inputs.mlir_override }}
+      xla_sha_override: ${{ inputs.xla_commit || github.sha }}
 
   build-ttxla:
     needs: build-image
@@ -75,6 +80,7 @@ jobs:
     with:
       docker_image: ${{ needs.build-image.outputs.docker-image }}
       mlir_override: ${{ inputs.mlir_override }}
+      xla_sha_override: ${{ inputs.xla_commit || github.sha }}
 
   generate-test-config:
     needs: build-ttxla


nsumrakTT (commented):
why is this feature needed?

jameszianxuTT (approved):
Per my comment, don't do this please. 

Also, these xla_sha_override inputs are not wired up to anything so this PR as it is does nothing but change the name of the run

jameszianxuTT (changes requested):
sorry accidental approve

nsumrakTT (changes requested):
I still don't understand on which commits you would like to run single test and why. Please give me your use case.

pdeviTT:
**USE CASE:**

- Bisecting the regressed cases, which are not able to replicate in local run, but facing issue on CI run.
- When i try to test it on other commits, It will be very helpful to switch on to other xla commits (like we have in tt-mlir). Since branch name for the specific commits, it's not visible once the patch got merged
Example: let's assume, I want to test in 81df2377fa1d162b9fcd49d327c95dcd97028bb5 commit. How can i test in Run Test Single pipeline ? How can i get the branch name of it ? If commit switch feature is enabled, we can easily switch it to their respective commit and test it

Screenshot 2026-04-13 at 10 14 30 PM

jameszianxuTT:
> **USE CASE:**
> 
> * Bisecting the regressed cases, which are not able to replicate in local run, but facing issue on CI run.
> * When i try to test it on other commits, It will be very helpful to switch on to other xla commits (like we have in tt-mlir). Since branch name for the specific commits, it's not visible once the patch got merged
>   Example: let's assume, I want to test in [81df237](https://github.com/tenstorrent/tt-xla/commit/81df2377fa1d162b9fcd49d327c95dcd97028bb5) commit. How can i test in Run Test Single pipeline ? How can i get the branch name of it ? If commit switch feature is enabled, we can easily switch it to their respective commit and test it
> 
> Screenshot 2026-04-13 at 10 14 30 PM

I think the normal way this is done is by doing a `git checkout <commit you want to inspect> && git checkout -b <branch name>`, pushing the branch to origin and running tests on that branch.

Basically, create a new branch based on the commit you want to inspect. There probably won't be an existing branch pointing at each commit.

This has the added advantage of not needing to rebuild the wheel, since the wheel build caching infra should match on the commit you're pushing. To make the feature you're describing work, you'd need to make more changes than what's currently in this PR, and bypass existing github actions built in behaviour, which will be tangly. For example, overriding checkout action behaviour.

If you really want to try this kind of workflow, you can also directly try on-branch changes inside github actions GUI (just switch the branch for Run Test Single to your branch and you'll see your new tt-xla commit input field), but I suggest to just make a branch from the commit you want to test directly.

vvukomanTT:
You could create a new branch as James said or you could create a tag on that commit and reference the tag in the workflow dispatch dropdown.

pdeviTT:
> > **USE CASE:**
> > 
> > * Bisecting the regressed cases, which are not able to replicate in local run, but facing issue on CI run.
> > * When i try to test it on other commits, It will be very helpful to switch on to other xla commits (like we have in tt-mlir). Since branch name for the specific commits, it's not visible once the patch got merged
> >   Example: let's assume, I want to test in [81df237](https://github.com/tenstorrent/tt-xla/commit/81df2377fa1d162b9fcd49d327c95dcd97028bb5) commit. How can i test in Run Test Single pipeline ? How can i get the branch name of it ? If commit switch feature is enabled, we can easily switch it to their respective commit and test it
> > 
> > Screenshot 2026-04-13 at 10 14 30 PM
> 
> I think the normal way this is done is by doing a `git checkout <commit you want to inspect> && git checkout -b <branch name>`, pushing the branch to origin and running tests on that branch.
> 
> Basically, create a new branch based on the commit you want to inspect. There probably won't be an existing branch pointing at each commit.
> 
> This has the added advantage of not needing to rebuild the wheel, since the wheel build caching infra should match on the commit you're pushing. To make the feature you're describing work, you'd need to make more changes than what's currently in this PR, and bypass existing github actions built in behaviour, which will be tangly. For example, overriding checkout action behaviour.
> 
> If you really want to try this kind of workflow, you can also directly try on-branch changes inside github actions GUI (just switch the branch for Run Test Single to your branch and you'll see your new tt-xla commit input field), but I suggest to just make a branch from the commit you want to test directly.

I understand your concern. Thanks for Your input :)). 
I already tried that method 1 (checkout to specific branch, pushed and tested it). May be I will give a try to checkout to my branch and switch to tt-xla commit in input field. I prefer the workflow that can be completed in the least time :))
Shall i make this branch as draft and use it ?

PR #4077, opened 2026-04-02
SDXL Performance tests

ppadjinTT:
## TL;DR

Adding a benchmark for running performance analysis for SDXL pipeline through forge. This requires a custom implementation since tt-forge-models doesn't support non-nn.Module models. 

## Testing Logic
We are asserting against the average UNet denoising total time and average UNet perf-step time. Since we are running VAE decoder and text encoder on CPU, it is risky to assert against that, since that performance is really dependent on the actual host's compute.

Additionally, I ran both 512x512 and 1024x1024 resolution tests on a single p100 locally. The thresholds for perf regression are 110% of those baselines (accounting for some hardware variability, lmk if this is too big of a margin).


vvukomanTT (approved):

tests/benchmark/benchmarks/sdxl_pipeline_benchmark.py
```diff
@@ -0,0 +1,428 @@
...
+    loop_count,
+    data_format,
+    ttnn_perf_metrics_output_file,
+    cfg_scale=SDXLConstants.CFG_SCALE,
+    prompt=SDXLConstants.PROMPT,
+    negative_prompt=SDXLConstants.NEGATIVE_PROMPT,
+    seed=SDXLConstants.SEED,
+):
+    if resolution == 1024:
+        model_id = "stabilityai/stable-diffusion-xl-base-1.0"
```
odjuricicTT: Same as the first comment, these should be defined in the model test and passed here.

odjuricicTT (commented):
Thanks for adding this! A few minor structural comments.

.github/workflows/call-perf-test.yml
```diff
@@ -487,3 +487,49 @@ jobs:
         fi
 
         echo "Performance check passed. Diff: ${perf_diff}%"
+
+    - name: Check latency metrics regression
```
odjuricicTT, line 491: @vvukomanTT Not sure if you reviewed these changes, do take another look.

odjuricicTT (approved):

.github/workflows/perf-bench-matrix.json
```diff
@@ -294,6 +294,18 @@
         "pyreq": "accelerate datasets diffusers==0.36.0 loguru pytest requests torch==2.9.0 tqdm transformers==5.2.0",
         "pytest": "tests/benchmark/test_encoders.py::test_unet_for_conditional_generation"
       },
+      {
+        "name": "sdxl_pipeline_512",
+        "pyreq": "accelerate diffusers==0.36.0 loguru pytest requests torch==2.9.0 tqdm transformers==4.57.1",
+        "pytest": "tests/benchmark/test_sdxl_pipeline.py::test_sdxl_pipeline_512",
+        "runs-on": "p100"
```
nsumrakTT, line 301: Note: we don't have p100 in CIv1, so just shared runs will run (the other will wait indefinitely).

nsumrakTT (approved):

.github/workflows/call-perf-test.yml
vvukomanTT: Since this PR won't be going in soon, I'll implement a more comprehensive perf regression check system in the meantime. (Darko's request for accuracy metrics regression checks)
So @ppadjinTT, I'll help you integrate this into the new system when you have time to get back to this PR.

thanks @odjuricicTT

PR #4140, opened 2026-04-07
Add KV cache dtype conversion option

kdimicTT:
### Problem description                                                                                                                   
                                                                                                                                        
KV cache tensors consume significant device memory. Storing them in lower-precision block floating-point formats (BFP8, BFP4) reduces memory usage and can improve performance. 
                                                                                                                                        
### What's changed                                                                                                                      

  - Added `experimental_kv_cache_dtype` compile option to allow overriding the KV cache tensor dtype
  - Extended `llm_benchmark.py` to accept and pass through target kv cache dtype parameter                          
  - Extended tests/benchmark/test_llms.py with dedicated bfp_bf8 KV cache benchmark tests          

### Comparing accuracy results


| Model | Baseline TOP1 | bfp_bf8 TOP1 | Baseline TOP5 | bfp_bf8 TOP5 |
| ----- | ------------- | ------------ | ------------- | ------------ |
| Llama-3.2-3B-Instruct | 92.19% | 90.62% | 100.00% | 100.00% |
| Llama-3.1-8B-Instruct | 79.69% | 81.25% | 90.62% | 92.19% |
| Mistral-7B-Instruct-v0.3 | 100.00% | 100.00% | 100.00% | 100.00% |
| Phi-1 | 95.31% | 95.31% | 100.00% | 100.00% |
| Phi-1.5 | 93.75% | 96.88% | 100.00% | 100.00% |
| Phi-2 | 93.75% | 93.75% | 100.00% | 100.00% |
| Falcon3-1B-Base | 98.44% | 100.00% | 100.00% | 100.00% |
| Falcon3-3B-Base | 96.88% | 96.88% | 96.88% | 100.00% |
| Falcon3-7B-Base | 98.44% | 98.44% | 100.00% | 100.00% |
| Qwen2.5-1.5B-Instruct | 84.38% | 78.12% | 96.88% | 92.19% |
| Qwen3-4B | 68.75% | 71.88% | 90.62% | 92.19% |
| Ministral-8B-Instruct | 93.75% | 95.31% | 100.00% | 100.00% |
| Qwen2.5-0.5B-Instruct | 96.88% | 90.62% | 100.00% | 98.44% |



### Note

- Adds explicit `experimental_kv_cache_dtype` per test for now, setting it as a default is tracked in 
#4343 
- This PR depends on changes in [tenstorrent/tt-mlir#7859](https://github.com/tenstorrent/tt-mlir/pull/7859)

### DIFF
--- a/pjrt_implementation/src/api/module_builder/module_builder.cc
+++ b/pjrt_implementation/src/api/module_builder/module_builder.cc
@@ -60,6 +60,7 @@
 #include "ttmlir/Dialect/TTIR/Transforms/Passes.h"
 #include "ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h"
 #include "ttmlir/Dialect/TTNN/Transforms/Passes.h"
+#include "ttmlir/Dialect/TTNN/Utils/BFPDtypeParser.h"
 #include "ttmlir/RegisterAll.h"
 #include "ttmlir/Target/Python/PythonEmitter.h"
 #include "ttmlir/Target/TTNN/TTNNToFlatbuffer.h"
@@ -993,11 +994,11 @@ tt_pjrt_status ModuleBuilder::convertFromTTIRToTTNN(
   }
 
   options.optimizationLevel = compile_options.optimization_level;
-  // Map user-facing dtype names to WeightDtype enum values.
+  // Map user-facing dtype names to BFPDtype enum values.
   if (compile_options.experimental_weight_dtype == "bfp_bf8") {
-    options.experimentalWeightDtype = mlir::tt::ttnn::WeightDtype::BFP_BFloat8;
+    options.experimentalWeightDtype = mlir::tt::ttnn::BFPDtype::BFP_BFloat8;
   } else if (compile_options.experimental_weight_dtype == "bfp_bf4") {
-    options.experimentalWeightDtype = mlir::tt::ttnn::WeightDtype::BFP_BFloat4;
+    options.experimentalWeightDtype = mlir::tt::ttnn::BFPDtype::BFP_BFloat4;
   } else if (!compile_options.experimental_weight_dtype.empty()) {
     LOG_F(ERROR,
           "Unknown experimental_weight_dtype: '%s'. Valid values: 'bfp_bf8', "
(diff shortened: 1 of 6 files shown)


pjrt_implementation/src/api/module_builder/module_builder.cc
```diff
@@ -993,11 +994,11 @@ tt_pjrt_status ModuleBuilder::convertFromTTIRToTTNN(
   }
 
   options.optimizationLevel = compile_options.optimization_level;
-  // Map user-facing dtype names to WeightDtype enum values.
+  // Map user-facing dtype names to BFPDtype enum values.
   if (compile_options.experimental_weight_dtype == "bfp_bf8") {
-    options.experimentalWeightDtype = mlir::tt::ttnn::WeightDtype::BFP_BFloat8;
+    options.experimentalWeightDtype = mlir::tt::ttnn::BFPDtype::BFP_BFloat8;
```
sgligorijevicTT, line 999: Arguably some of this code is better suited for `compile_options.cc` rather than this function. 
I think we should merge as is and plan to move this code sometime later. @dgolubovicTT @kdimicTT what do you think?

sgligorijevicTT (approved):

sgligorijevicTT:
Heads up, this change needs to land in lockstep with the newest uplift. ~~I'm going to cherry pick the commit onto the uplift branch https://github.com/tenstorrent/tt-xla/pull/4450~~ Change of plans, I will take only the minimal set of changes, this PR should still land, you will have a minor merge conflict

vkovacevicTT (commented):
Some benchmark questions that need addressing:
- What is the idea of adding two new tests with kv_cache_dtype config? Do we want bfp_bf8 to be the default kv_cache_dtype for these two llama models?
- Looking at the topk results in the description, will this be run with accuracy testing in the benchmark, or a regular PCC run?

kdimicTT:
> Some benchmark questions that need addressing:
> 
> * What is the idea of adding two new tests with kv_cache_dtype config? Do we want bfp_bf8 to be the default kv_cache_dtype for these two llama models?
> * Looking at the topk results in the description, will this be run with accuracy testing in the benchmark, or a regular PCC run?

The idea is to put it as a default later. The tests are added temporarily to prevent regression until the issue in tt-mlir linked below is resolved. The issue describes why  `optimization_level=2` (which is set as default) causes PCC failure with the KV cache conversion.  https://github.com/tenstorrent/tt-mlir/issues/8030

Accuracy testing makes more sense in this case since PCC is computed only for the first token, which cannot catch KV cache issues.

tests/benchmark/test_llms.py
```diff
@@ -317,6 +320,36 @@ def test_llama_3_2_3b(
     )
 
 
+def test_llama_3_2_3b_kv_cache(
```
mvasiljevicTT: Quick note - adding tests here will not automatically add them to the perf benchmark workflow, so currently they could only be used manually and won't catch regressions. To do that, you need to list tests in the [perf-benchmark-matrix](https://github.com/tenstorrent/tt-xla/blob/main/.github/workflows/perf-bench-matrix.json). 

[Here](https://github.com/tenstorrent/tt-xla/actions/workflows/manual-benchmark.yml) you can run perf benchmark with your branch and custom mlir commit, the safest way is to rebase your mlir change on top of the mlir used in xla you are testing (in this case it is ok to just wait for mlir uplift and rebase xla branch). If you add new tests to perf benchmark job, please run it (you can filter it to just those tests) and attach the link of the CI job for future reference

kdimicTT: Thanks for the heads up. I'll leave the tests out for now until I can add this option by default

vkovacevicTT (approved):

PR #4145, opened 2026-04-07
Add golden pip environment restore for crash recovery in forked tests

ajakovljevicTT:
When running model tests with `--forked`, each test gets its own child process where RequirementsManager installs per-model pip dependencies and rolls them back on exit — but if a child is OOM-killed mid-test, `__exit__` never runs and the environment is left dirty, causing cascading failures in subsequent tests that have nothing to do with the models themselves. This PR fixes that by capturing a "golden" snapshot of the pip environment once at session start (before any forking), and restoring it in the begging of the next test.

### DIFF
--- a/tests/runner/requirements.py
+++ b/tests/runner/requirements.py
@@ -107,26 +107,17 @@ class RequirementsManager:
 
         if to_restore:
             _dbg(f"[Requirements] Restoring: restoring {len(to_restore)} packages")
-            restore_file = None
-            try:
-                with tempfile.NamedTemporaryFile(
-                    mode="w",
-                    suffix=".txt",
-                    delete=False,
-                    prefix="tt_xla_restore_golden_",
-                ) as f:
-                    f.write("\n".join(to_restore) + "\n")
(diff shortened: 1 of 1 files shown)


tests/runner/requirements.py
```diff
@@ -43,6 +43,94 @@ class RequirementsManager:
...
+            except Exception as e:
+                warnings.warn(
+                    f"[Requirements] Restoring: uninstall failed: {e}",
+                    RuntimeWarning,
+                    stacklevel=2,
+                )
+
+        if to_restore:
+            _dbg(f"[Requirements] Restoring: restoring {len(to_restore)} packages")
+            # Install one-by-one so that a single unresolvable package
```
sgligorijevicTT: Non blocking but I want to start a discussion. 
If a restore of a single package fails we're probably screwed anyway so it might make sense to die at this point already, instead of merely logging a warning and praying(pointlessly) that a venv in inconsistent state will magically work later, or at least crash loundly enough for us to understand this was the real issue.

PR #4190, opened 2026-04-09
Fix KV cache bleed: logits_indices overflow and prefix cache block overwrite 

### LINKED ISSUE #3899: Potential KV Cache issue with batch_size>1 served models sharing state
Details coming soon -  but saw a case last night where llama3.1-8b-instruct model I was hosting via tt-xla -> vllm -> tt-inference server in a particular config (4 parallel chat sessions) having text leak between different "user" sessions sharing the same chip and KV cache. 

Image

Edit: Able to reproduce on p100a with stress testing.

kmabeeTT:
### Ticket
  Closes https://github.com/tenstorrent/tt-xla/issues/3899

  ### Problem description
  Three bugs in the vLLM plugin cause cross-user response contamination (KV cache bleed) when serving multiple concurrent users:

  1. **logits_indices overflow** — batched prefill used cumulative offsets from `query_start_loc` that overflow past each user's padded slot, reading wrong hidden states
  2. **Prefix cache block overwrite** — `paged_fill_cache` writes suffix KV data starting at the shared prefix block, corrupting it for other requests
  3. **Mixed-batch prefix cache bleed** — the initial fix for Bug 2 used a single `min(num_computed_tokens)` roll offset, which is 0 whenever the batch mixes cached and non-cached requests

  ### What's changed

  **`integrations/vllm_plugin/vllm_tt/model_runner.py`**
  - Replace cumulative `query_start_loc` logits_indices with per-user `num_scheduled_tokens - 1`
  - Compute per-row `fill_page_table` by rolling each user's page_table row by their own `num_computed_tokens // block_size`, handling mixed-batch prefix cache states
  - Skip roll and device transfer when no prefix offset needed

  **`integrations/vllm_plugin/vllm_tt/attention.py`**
  - Add `fill_page_table` field to `TTMetadata` (defaults to `page_table`)
  - Use `fill_page_table` instead of `page_table` in `paged_fill_cache` calls

  **`integrations/vllm_plugin/vllm_tt/pooling_runner.py`**
  - Same `logits_indices` fix (no prefix caching fix needed — pooling doesn't use paged attention)

  **`tests/integrations/vllm_plugin/generative/test_kv_cache_bleed.py`** (new)
  - `test_kv_cache_no_cross_user_bleed`: server-based regression test with TinyLlama-1.1B-Chat
  - Phase 1: 3 rounds of concurrent requests (triggers Bug 1)
  - Phase 2: 5 rounds of staggered overlapping requests (triggers Bugs 2/3)
  - Keyword contamination detection across 4 distinct topic prompts with greedy decoding
  - Fails without fix, passes with fix

  ### Checklist
  - [x] `test_kv_cache_no_cross_user_bleed` passes (fails without fix)
  - [x] `test_opt_generation` passes (single-device, batched generation)
  - [x] `test_batched_inference` passes (pooling path)

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/attention.py
+++ b/integrations/vllm_plugin/vllm_tt/attention.py
@@ -173,6 +173,9 @@ class TTMetadata:
     attn_mask: torch.Tensor
     page_table: torch.Tensor
     is_causal: bool
+    # Block offset for prefix caching: skip this many blocks in page_table
+    # when filling the cache during prefill (prefix blocks are already filled).
+    prefill_block_offset: int
 
     def __init__(
         self,
@@ -180,11 +183,13 @@ class TTMetadata:
         attn_mask: torch.Tensor | None = None,
         page_table: torch.Tensor | None = None,
         is_causal: bool = True,
+        prefill_block_offset: int = 0,
     ):
         self.cache_position = cache_position
         self.attn_mask = attn_mask
         self.page_table = page_table
         self.is_causal = is_causal
+        self.prefill_block_offset = prefill_block_offset
 
 
 class TTAttentionBackendImpl(AttentionImpl):
@@ -470,11 +475,20 @@ class TTAttentionBackendImpl(AttentionImpl):
             key_for_update = inputs.key.transpose(1, 2)
             value_for_update = inputs.value.transpose(1, 2)
 
+            # With prefix caching, some blocks are already filled from the
+            # cached prefix. Offset the page_table to skip those blocks so
+            # paged_fill_cache only writes to the new (non-prefix) blocks.
+            fill_page_table = attn_metadata.page_table
+            if attn_metadata.prefill_block_offset > 0:
+                fill_page_table = attn_metadata.page_table[
+                    :, attn_metadata.prefill_block_offset :
+                ]
+
             for batch_idx in range(inputs.users):
                 k_cache = torch.ops.tt.paged_fill_cache(
                     k_cache,
                     key_for_update[batch_idx : batch_idx + 1],
-                    attn_metadata.page_table,
+                    fill_page_table,
                     batch_idx=torch.tensor(
                         [batch_idx], dtype=torch.int32, device=k_cache.device
                     ),
@@ -482,7 +496,7 @@ class TTAttentionBackendImpl(AttentionImpl):
                 v_cache = torch.ops.tt.paged_fill_cache(
                     v_cache,
                     value_for_update[batch_idx : batch_idx + 1],
-                    attn_metadata.page_table,
+                    fill_page_table,
                     batch_idx=torch.tensor(
                         [batch_idx], dtype=torch.int32, device=v_cache.device
                     ),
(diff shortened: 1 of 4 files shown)


kmabeeTT:
### Investigation details

  **Bug 1 (logits_indices):** In vLLM TT's padded-per-user layout, each user gets a `[padded_len]` slot. The old code used `query_start_loc[1:] - 1` which produces cumulative offsets like `[31, 63, 95, 127]` for 4 users — indices 63+ overflow past `padded_len=32`. The fix uses per-user `num_scheduled_tokens - 1` which stays within each slot.

  **Bug 2 (prefix cache overwrite):** When prefix caching is enabled and requests share a common prefix (e.g., same chat template), `paged_fill_cache` receives the full `page_table` including shared prefix blocks but only has suffix KV data to write. It writes starting at `page_table[0]` (the shared prefix block), overwriting cached prefix data. Fixed by rolling
  prefix blocks to the end of the page_table before passing to `paged_fill_cache`.

  **Bug 3 (mixed-batch variant, found during test development):** The initial fix for Bug 2 used `min(num_computed_tokens)` across all users to compute a single roll offset. The scheduler batches users with different prefix cache states together — when one user has `num_computed=0` and another has `num_computed=32`, the min is 0 and the roll doesn't happen. The per-row roll fix handles each user independently. This was caught by the regression test (`test_kv_cache_no_cross_user_bleed`) which failed on concurrent r1 in 6/6 runs even with the original fix. See https://github.com/tenstorrent/tt-xla/issues/3899#issuecomment-4219429304.

  **Recompilation avoidance:** The page_table manipulation is done outside the compiled graph using `torch.roll` (constant tensor shape) to avoid torch.compile recompilation that previously caused ~24s TTFT regressions.

  **Test details:** `test_kv_cache_no_cross_user_bleed` starts a vLLM server with TinyLlama-1.1B-Chat and sends concurrent + staggered requests with 4 distinct topics (penguin, submarine, dinosaur, chocolate). Checks for foreign keyword contamination in responses. Runs in ~3:45 with `--enforce-eager`. Without fix: 7 failures across both phases. With fix: 0 failures.

  Bleed repro scripts and stress tests used for debugging are on branch `kmabee/issue_3899_kv_cache_bleed_debug`.

integrations/vllm_plugin/vllm_tt/attention.py
```diff
@@ -470,19 +477,21 @@ def _handle_paged_attention(
...
             value_for_update = inputs.value.transpose(1, 2)
 
+            fill_page_table = attn_metadata.fill_page_table
+
             for batch_idx in range(inputs.users):
                 k_cache = torch.ops.tt.paged_fill_cache(
                     k_cache,
                     key_for_update[batch_idx : batch_idx + 1],
-                    attn_metadata.page_table,
+                    fill_page_table,
```
mmanzoorTT: Q: Why not to use `attn_metadata.fill_page_table` directly here?

mmanzoorTT (approved):

integrations/vllm_plugin/vllm_tt/attention.py
kmabeeTT: fair point, updated.

PR #4219, opened 2026-04-12
Add training mode and LoRA backward tests for LLM torch models

### LINKED ISSUE #4107: Add LoRA backward tests for LLMs in test_llms_torch
`test_llms_torch` currently only runs inference (`RunMode.INFERENCE`). Add LoRA-based backward tests:

- Add `RunMode.TRAINING` to `test_llms_torch`'s `run_mode` parametrization (same as `test_all_models_torch` already does).
- Create a LoRA wrapper that applies LoRA adapters to the LLM, freezing base weights and only training LoRA parameters. This reduces memory vs full backward, making it feasible on single-device for larger LLMs.
- The existing `_test_training` flow in `TorchModelTester` (forward on CPU, forward on TT, backward on both, compare gradients) applies as-is.
- These LoRA backward tests run in addition to the full backward tests.

agobeljicTT:
### Ticket
Closes #4107

### Problem description

LLM torch tests only covered inference. This PR adds training mode support and initial LoRA backward pass coverage to validate fine-tuning workflows on Tenstorrent hardware.

### What's changed

| File | Change |
|------|--------|
| `tests/runner/test_config/constants.py` | Adds `"training"` to `RUN_MODES_LLM` |
| `tests/runner/test_models.py` | Parametrizes `test_llms_torch` with `RunMode.TRAINING`; skips training on decode phase |
| `tests/runner/test_config/torch_llm/test_config_training_single_device.yaml` | New training configs for llama, llama_lora, gemma, gemma_lora, phi1, phi1_lora, qwen_2_5, qwen_2_5_lora |
| `tests/torch/training/test_llama_lora.py` | Standalone backward test verifying LoRA params receive gradients pre- and post-compile |
| `tests/infra/testers/single_chip/model/torch_model_tester.py` | Switches training path to `tt_legacy_compile`; unconditionally disables fx fusion pass; adds gradient assertion guards |
| `tests/infra/runners/torch_device_runner.py` | Gates shard spec computation on `is_multichip`; simplifies `mark_sharding` guard |
| `tests/infra/evaluators/torch_comparison_evaluator.py` | Fixes dim check to require 3D+ tensors before applying token mask (2D gradient matrices have no sequence dim) |
| `third_party/tt_forge_models` | Bumps submodule to include LoRA model loaders |

### Checklist


tests/runner/validate_test_config.py
```diff
@@ -443,17 +444,19 @@ def _generate_test_ids(self, torch_models, jax_models, llm_models) -> set:
...
-                            f"{base}-{phase}-seq_1-batch_1" f"-{parallelism}-{run_mode}"
-                        )
-                    else:
-                        for seq in LLM_SEQUENCE_LENGTHS:
-                            for batch in LLM_BATCH_SIZES:
-                                ids.add(
-                                    f"{base}-{phase}-seq_{seq}-batch_{batch}"
-                                    f"-{parallelism}-{run_mode}"
-                                )
+                    for adapter_mode in ADAPTER_MODES_LLM:
```
agobeljicTT: use "hide whitespaces" for better visuals

tests/runner/testers/torch/dynamic_torch_model_tester.py
```diff
@@ -110,10 +115,18 @@ def _remove_weight_dtype_overrides(self):
...
+        adapter_mode (e.g. LORA reduces memory via low-rank matrices).
+
         Returns:
             Model instance loaded from the loader
         """
-        return self.dynamic_loader.load_model()
+        model = self.dynamic_loader.load_model()
+
+        if self._adapter_mode == AdapterMode.LORA and self._run_mode == RunMode.TRAINING:
+            model = apply_lora_adapters(model)
```
agobeljicTT: this function is a part of `tt-forge-models`
that PR is linked to this one and should be merged first.

tests/runner/testers/torch/dynamic_torch_model_tester.py
```diff
@@ -37,7 +38,8 @@ def __init__(
         compiler_config: CompilerConfig = None,
         parallelism: Parallelism = Parallelism.SINGLE_DEVICE,
         run_phase: RunPhase = RunPhase.DEFAULT,
-        test_metadata=None,
+        test_metadata = None,
+        adapter_mode: AdapterMode = AdapterMode.NONE,
```
ndrakulicTT: Can we maybe store adapter_mode inside the test_metadata, as it is LLM specific that way we dont have new param in `_run_model_test_impl`, as that function handles both the LLMs and other models?

agobeljicTT: yes

tests/runner/test_config/torch_llm/test_config_training_single_device.yaml
```diff
@@ -0,0 +1,59 @@
...
+  llama/causal_lm/pytorch-3.2_1B-llm_prefill-seq_128-batch_1-single_device-training-no_adapter:
+    status: EXPECTED_PASSING
+    required_pcc: 0.88
+
+  llama/causal_lm/pytorch-3.2_1B-llm_prefill-seq_128-batch_1-single_device-training-lora:
+    status: EXPECTED_PASSING
+    lora_r: 16
+    lora_alpha: 32.0
+    lora_target_modules: ["q_proj", "k_proj", "v_proj"]
+    lora_dropout: 0.05
```
agobeljicTT: demonstration sake

tests/runner/validate_test_config.py
```diff
@@ -443,17 +444,19 @@ def _generate_test_ids(self, torch_models, jax_models, llm_models) -> set:
...
+                            ids.add(
+                                f"{base}-{phase}-seq_1-batch_1"
+                                f"-{parallelism}-{run_mode}-{adapter_mode}"
+                            )
+                        else:
+                            for seq in LLM_SEQUENCE_LENGTHS:
+                                for batch in LLM_BATCH_SIZES:
+                                    ids.add(
+                                        f"{base}-{phase}-seq_{seq}-batch_{batch}"
+                                        f"-{parallelism}-{run_mode}-{adapter_mode}"
```
agobeljicTT: imma need extra validation here, seems that some tests fail due to this:

Image

> [!NOTE]
> This shouldn't be the case, but this error wasn't encountered on other branches I made, and this diff is the only validation diff I made, so this could be the culprit...

tests/runner/test_config/torch_llm/test_config_training_single_device.yaml
ndrakulicTT: About setting custom LoRA configs for the model not sure which repo should have ownership over that `tt-xla` or `tt-forge-models`. 
One downside of this approach is that we cant have multiple test with different LoRA configs. But that is most likely not needed. Still maybe it is better to have LoRA config in `tt-forge-models`, as in future we could just pull that config for `tt-blacksmith` also
@pglusacTT what are your thoughts on this?

pglusacTT: LoRA is, in a way, similar to sharding patterns; you choose some pattern of weights and apply some function to it. That could potentially be a generic function in `tt-forge-models` model. Since shardings are there I'm in favor of moving LoRA there as well.

tests/runner/test_config/torch_llm/test_config_inference_single_device.yaml
```diff
@@ -8,175 +8,175 @@
 #   - bringup_status (if present): BringupStatus names (e.g., FAILED_TTMLIR_COMPILATION) or values (e.g., failed_ttmlir_compilation)
 
 test_config:
-  falcon/pytorch-3_1B_Base-llm_decode-seq_1-batch_1-single_device-mesh_default-inference:
+  falcon/pytorch-3_1B_Base-llm_decode-seq_1-batch_1-single_device-mesh_default-inference-no_adapter:
```
acicovicTT: Is it possible not to pollute the name of the test by adding lora configs in the body of the test config below, e.g. `lora: true` would indicate lora adapter, similar to how you added `lora_r`, `lora_alpha` etc? if there is no `lora: true` then it will default to no adapter behavior, and you wouldn't need to change inference test configs at all

agobeljicTT: we need a way to test both versions, by adding a metadata entry that would mean only one version will be tested.
additionally, we want to split the two versions of the test, if we put `lora: true` in metadata, that'd mean that we need to run both versions in the same test

tests/runner/test_config/torch_llm/test_config_training_single_device.yaml
agobeljicTT: like having separate model for lora? that could work, we could have lora only for models we want and we won't need to pollute the configs @acicovicTT

tests/infra/evaluators/torch_comparison_evaluator.py
```diff
@@ -119,7 +119,7 @@ def apply_real_token_mask(x: torch.Tensor, y: torch.Tensor):
             ), f"Expected pcc_mask to have shape [batch, seq], got {tuple(pcc_mask.shape)}"
             batch_size, seq_len = pcc_mask.shape
 
-            if x.dim() < 2 or y.dim() < 2:
+            if x.dim() < 3 or y.dim() < 3:
```
agobeljicTT, line 125: Some models fail due to this, claude pointed out that this should be < 3 as rank 2 should be escaped as well. @ndrakulicTT your thoughts?

tests/torch/training/test_llama_lora.py
```diff
@@ -0,0 +1,55 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```
agobeljicTT, line 1: this test is perhaps redundant, maybe just to check the grads are indeed returned properly?

tests/infra/evaluators/torch_comparison_evaluator.py
mmilosevicTT: This does not seem right, the flow of the code here is:
- Check if x and y have at least batch_size, seq_len to continue
- Check if first dim is batch
- Fail if we do not have dim 3 or 4 for x

More important question here IMO should be, why do would we pass a mask if we do not have what to filter?

umalesTT: @agobeljicTT yes, dimension check was there for ensuring we apply pcc mask on logits&kv cache only, so we could measure pcc on non-padding tokens. This fix seems fine, just wanted to ask, are you returning only logits beside gradients?

tests/infra/testers/single_chip/model/torch_model_tester.py
```diff
@@ -266,18 +266,19 @@ def _test_training(self) -> Tuple[ComparisonResult, ...]:
...
+            "tt_enable_torch_fx_fusion_pass": False,
+        }
 
         self._compile_for_tt_device(self._workload, compile_options)
         tt_res = self._run_on_tt_device(self._workload)
         tt_res = self._unpack_forward_output(tt_res)
 
         # Force graph break so we can differentiate between forward and backward
-        torch_xla.sync(wait=True)
+        # TODO(agobeljicTT): Decide on if we want to keep this after: https://github.com/tenstorrent/tt-xla/issues/3234
```
umalesTT: I'd advise not to, for larger models all intermediates will be kept at the end of forward and later will become inputs to backward graph and we won't be able to deallocate them, you will get OOM for larger models

tests/infra/testers/compiler_config.py
```diff
@@ -99,8 +99,9 @@ def to_jax_compiler_options(self) -> Dict[str, str]:
         if self.experimental_enable_fusing_conv2d_with_multiply_pattern:
             options["experimental_enable_fusing_conv2d_with_multiply_pattern"] = "true"
 
-        if not self.experimental_enable_permute_matmul_fusion:
-            options["experimental_enable_permute_matmul_fusion"] = "false"
+        options["experimental_enable_permute_matmul_fusion"] = (
```
umalesTT: Why not stick to the format of options above, with `if` outside options definition?

tests/torch/training/test_llama_lora.py
umalesTT: @agobeljicTT I think we should keep this on-push, it is short and simple and will test lora testing code-path per PR

tests/infra/testers/single_chip/model/torch_model_tester.py
agobeljicTT: reverting

tests/torch/training/test_llama_lora.py
agobeljicTT: this runs on PR and on merge

tests/infra/evaluators/torch_comparison_evaluator.py
agobeljicTT: so I spent some time and here is what I found:
1. why this was crashing with < 2: _when lora rank == batch size_ the check `if x.shape[0] != batch_size or y.shape[0] != batch_size:` is `False` so we go ahead and hit the assertion at the bottom.
2. qwen returns both outputs and gradients here
3. we do need to use the mask here, otherwise we would compare padded outputs
4. the correct fix here is to leave < 3 (or <= 2)

umalesTT (approved):

tests/infra/evaluators/torch_comparison_evaluator.py
```diff
@@ -119,7 +119,10 @@ def apply_real_token_mask(x: torch.Tensor, y: torch.Tensor):
             ), f"Expected pcc_mask to have shape [batch, seq], got {tuple(pcc_mask.shape)}"
             batch_size, seq_len = pcc_mask.shape
 
-            if x.dim() < 2 or y.dim() < 2:
+            # Mask only applies to rank-3/4 tensors (logits [B,S,V] or KV [B,H,S,D]).
+            # Rank-1/2 tensors (e.g. 2D gradient matrices in training) have no sequence
+            # dimension to mask; skip rather than crash on the assert below.
+            if x.dim() < 3 or y.dim() < 3:
```
acicovicTT, line 125: Can this change somehow affect inference tests?

tests/infra/runners/torch_device_runner.py
```diff
@@ -129,9 +129,16 @@ def _safely_put_workload_on_device(
...
+        is_multichip = (
+            hasattr(workload, "mesh")
+            and workload.mesh
+            and len(workload.mesh.device_ids) > 1
+        )
+
         shard_specs = None
         if (
-            device.type != "cpu"
+            is_multichip
```
acicovicTT, line 140: Is there a specific reason to move the `is_multichip` check here?

acicovicTT (approved):
Approved with a couple of questions

tests/infra/runners/torch_device_runner.py
agobeljicTT: yes, so that sharding is None from the start and we do not even load the sharding specs if it isn't multichip

tests/infra/evaluators/torch_comparison_evaluator.py
agobeljicTT: it shouldn't, it is expected that we never mask tensors which are 2 or less. there is one more assert downstairs.

PR #4236, opened 2026-04-14
Fix EventInstance by using a worker thread for callbacks

### LINKED ISSUE #3838: [Regression] maskformer_swin_b hangs
maskformer_swin_b test
```
tests/runner/test_models.py::test_all_models_torch[maskformer_swin_b/pytorch-Swin_Base_Coco-single_device-inference]
```
never finishes execution.
It hangs a little after TTNN graph print. This is possibly related to our implementation of `EventInstance`.

[swin_log.txt](https://github.com/user-attachments/files/26136395/swin_log.txt)

### LINKED ISSUE #3852: Dense unet hanging in nightly 24 Mar 2026
### Describe the bug

`dense_unet_3d` is hanging (see here for an example): 
https://github.com/tenstorrent/tt-xla/actions/runs/23518462642/job/68534831265
`tests/runner/test_models.py::test_all_models_torch[dense_unet_3d/pytorch-Base-single_device-inference]`

### LINKED ISSUE #4004: gemma fails on single device
### Describe the bug

`gemma/pytorch-1.1_2B_IT-llm_decode-seq_1-batch_1-single_device-inference` hangs in nightly:
https://github.com/tenstorrent/tt-xla/actions/runs/23722501898/job/69152060398

needs further investigation

### LINKED ISSUE #4094: llm tests hang in nightly
```
qwen_3/causal_lm/pytorch-0_6B-llm_decode-seq_1-batch_1-single_device-inference
gemma/pytorch-1.1_7B_IT-llm_decode-seq_1-batch_1-single_device-inference
```

sgligorijevicTT:
### Ticket
fixes #3838 fixes #4004 fixes #3852 fixes #4094

### Problem description
Event Instance callbacks seem to be deadlocking. Our best understanding is that calling the callbacks from the main thread can sometimes somehow deadlock with the python GIL.

### What's changed
Added a MPSC queue. Added a worker thread that drains the queue.
Changed callback dispatch to always be done on the worker thread.
Added PJRT unit tests.

### Checklist

### DIFF
--- a/pjrt_implementation/inc/utils/callback_worker.h
+++ b/pjrt_implementation/inc/utils/callback_worker.h
@@ -1,0 +1,69 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TT_XLA_PJRT_IMPLEMENTATION_INC_UTILS_CALLBACK_WORKER_H_
+#define TT_XLA_PJRT_IMPLEMENTATION_INC_UTILS_CALLBACK_WORKER_H_
+
+// c++ standard library includes
+#include <atomic>
+#include <semaphore>
+#include <thread>
+
+// PJRT C API includes
+#include "xla/pjrt/c/pjrt_c_api.h"
+
+// tt-xla includes
+#include "utils/mpsc_queue.h"
+
+namespace tt::pjrt::utils {
+
+// Work item representing a single PJRT event callback to be executed
+// asynchronously on the worker thread.
+struct CallbackWorkItem {
+  // Callback function to invoke.
+  PJRT_Event_OnReadyCallback callback_function = nullptr;
+
+  // Caller-supplied context pointer passed to the callback.
+  void *user_arg = nullptr;
+
+  // Pre-created PJRT error (or nullptr for success). Ownership is transferred
+  // to the callback function.
+  PJRT_Error *error = nullptr;
+};
+
+// A single worker thread that drains event callbacks from a lock-free MPSC
+// queue and executes them. This breaks the GIL + device lock deadlock by
+// ensuring callbacks run on a thread that holds neither lock.
+//
+// Thread-safe for multiple concurrent producers (enqueue). The consumer is the
+// internal worker thread.
+class CallbackWorker {
+public:
+  explicit CallbackWorker(size_t queue_capacity = 1024);
+
+  // Signals shutdown, joins the worker thread, and drains any remaining items.
+  ~CallbackWorker();
+
+  // Non-copyable, non-movable.
+  CallbackWorker(const CallbackWorker &) = delete;
+  CallbackWorker &operator=(const CallbackWorker &) = delete;
+
+  // Enqueue a callback for asynchronous execution on the worker thread.
+  // Lock-free for producers. Spins briefly if the queue is momentarily full.
+  void enqueue(PJRT_Event_OnReadyCallback callback_function, void *user_arg,
+               PJRT_Error *error);
+
+private:
+  // Worker thread main loop.
+  void workerLoop();
+
+  MPSCQueue<CallbackWorkItem> m_queue;
+  std::thread m_worker_thread;
+  std::atomic<bool> m_shutdown{false};
+  std::binary_semaphore m_work_available{0};
+};
+
+} // namespace tt::pjrt::utils
+
+#endif // TT_XLA_PJRT_IMPLEMENTATION_INC_UTILS_CALLBACK_WORKER_H_
--- a/pjrt_implementation/inc/utils/callback_worker.h
+++ b/pjrt_implementation/inc/utils/callback_worker.h
@@ -50,7 +50,8 @@ public:
   CallbackWorker &operator=(const CallbackWorker &) = delete;
 
   // Enqueue a callback for asynchronous execution on the worker thread.
-  // Lock-free for producers. Spins briefly if the queue is momentarily full.
+  // Lock-free for producers. If the queue is full, this call sleeps briefly
+  // and retries until the item is successfully enqueued.
   void enqueue(PJRT_Event_OnReadyCallback callback_function, void *user_arg,
                PJRT_Error *error);
 
--- a/pjrt_implementation/inc/utils/callback_worker.h
+++ b/pjrt_implementation/inc/utils/callback_worker.h
@@ -16,7 +16,7 @@
 // tt-xla includes
 #include "utils/mpsc_queue.h"
 
-namespace tt::pjrt::utils {
+namespace tt::pjrt {
 
 // Work item representing a single PJRT event callback to be executed
 // asynchronously on the worker thread.
@@ -59,12 +59,12 @@ private:
   // Worker thread main loop.
   void workerLoop();
 
-  MPSCQueue<CallbackWorkItem> m_queue;
+  internal::MPSCQueue<CallbackWorkItem> m_queue;
   std::thread m_worker_thread;
   std::atomic<bool> m_shutdown{false};
   std::binary_semaphore m_work_available{0};
 };
 
-} // namespace tt::pjrt::utils
+} // namespace tt::pjrt
 
 #endif // TT_XLA_PJRT_IMPLEMENTATION_INC_UTILS_CALLBACK_WORKER_H_
--- a/pjrt_implementation/inc/utils/callback_worker.h
+++ b/pjrt_implementation/inc/utils/callback_worker.h
@@ -62,7 +62,7 @@ private:
   internal::MPSCQueue<CallbackWorkItem> m_queue;
   std::thread m_worker_thread;
   std::atomic<bool> m_shutdown{false};
-  std::binary_semaphore m_work_available{0};
+  std::counting_semaphore<> m_work_available{0};
 };
 
 } // namespace tt::pjrt
--- a/pjrt_implementation/inc/utils/callback_worker.h
+++ b/pjrt_implementation/inc/utils/callback_worker.h
@@ -42,7 +42,7 @@ class CallbackWorker {
 public:
   explicit CallbackWorker(size_t queue_capacity = 1024);
 
-  // Signals shutdown, joins the worker thread, and drains any remaining items.
+  // Signals shutdown and joins the worker thread.
   ~CallbackWorker();
 
   // Non-copyable, non-movable.
--- a/pjrt_implementation/inc/utils/mpsc_queue.h
+++ b/pjrt_implementation/inc/utils/mpsc_queue.h
@@ -15,7 +15,7 @@
 // tt-xla includes
 #include "utils/assert.h"
 
-namespace tt::pjrt::utils {
+namespace tt::pjrt::internal {
 
 // Lock-free bounded Multi-Producer Single-Consumer (MPSC) ring buffer.
 //
@@ -127,6 +127,6 @@ private:
   alignas(64) size_t m_read_pos{0};
 };
 
-} // namespace tt::pjrt::utils
+} // namespace tt::pjrt::internal
 
 #endif // TT_XLA_PJRT_IMPLEMENTATION_INC_UTILS_MPSC_QUEUE_H_
--- a/pjrt_implementation/inc/utils/mpsc_queue.h
+++ b/pjrt_implementation/inc/utils/mpsc_queue.h
@@ -1,0 +1,131 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TT_XLA_PJRT_IMPLEMENTATION_INC_UTILS_MPSC_QUEUE_H_
+#define TT_XLA_PJRT_IMPLEMENTATION_INC_UTILS_MPSC_QUEUE_H_
+
+// c++ standard library includes
+#include <atomic>
+#include <cassert>
+#include <cstddef>
+#include <memory>
+#include <new>
+#include <utility>
+
+namespace tt::pjrt::utils {
+
+// Lock-free bounded Multi-Producer Single-Consumer (MPSC) ring buffer.
+//
+// Uses the Vyukov bounded MPSC queue algorithm. Each cell carries an atomic
+// sequence number that acts as a per-slot state machine:
+//   sequence == pos            -> slot is free for a producer to claim
+//   sequence == pos + 1        -> slot contains data ready for the consumer
+//   sequence == pos + capacity -> slot has been consumed and recycled
+//
+// Producers compete via CAS on a shared write position. The single consumer
+// advances its own read position without contention.
+//
+// Capacity must be a power of 2 (enforced by assert).
+template <typename T> class MPSCQueue {
+public:
+  explicit MPSCQueue(size_t capacity)
+      : m_capacity(capacity), m_mask(capacity - 1),
+        m_cells(std::make_unique<Cell[]>(capacity)) {
+    assert(capacity >= 2 && (capacity & (capacity - 1)) == 0 &&
+           "Capacity must be a power of 2");
+
+    for (size_t i = 0; i < m_capacity; ++i) {
+      m_cells[i].sequence.store(i, std::memory_order_relaxed);
+    }
+  }
+
+  // Non-copyable, non-movable (contains atomics).
+  MPSCQueue(const MPSCQueue &) = delete;
+  MPSCQueue &operator=(const MPSCQueue &) = delete;
+
+  // Try to push an element. Lock-free, safe for multiple concurrent producers.
+  // Returns false if the queue is full.
+  bool tryPush(T &&item) {
+    size_t pos = m_write_pos.load(std::memory_order_relaxed);
+
+    for (;;) {
+      Cell &cell = m_cells[pos & m_mask];
+      size_t seq = cell.sequence.load(std::memory_order_acquire);
+
+      intptr_t diff = static_cast<intptr_t>(seq) - static_cast<intptr_t>(pos);
+
+      if (diff == 0) {
+        // Slot is free at this position. Try to claim it.
+        if (m_write_pos.compare_exchange_weak(pos, pos + 1,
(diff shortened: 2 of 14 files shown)


pjrt_implementation/src/utils/callback_worker.cc
```diff
@@ -0,0 +1,56 @@
...
+}
+
+void CallbackWorker::enqueue(PJRT_Event_OnReadyCallback callback_function,
+                             void *user_arg, PJRT_Error *error) {
+  CallbackWorkItem item{callback_function, user_arg, error};
+
+  while (!m_queue.tryPush(std::move(item))) {
+    DLOG_F(WARNING, "CallbackWorker queue is full, spinning...");
+    std::this_thread::yield();
+  }
```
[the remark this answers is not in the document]
sgligorijevicTT: More of a theoretical point but fair enough, a sleep for a constant time ought to be enough to alleviate concerns

pjrt_implementation/inc/utils/mpsc_queue.h
```diff
@@ -0,0 +1,130 @@
...
+// advances its own read position without contention.
+//
+// Capacity must be a power of 2 (enforced by assert).
+template <typename T> class MPSCQueue {
+public:
+  explicit MPSCQueue(size_t capacity)
+      : m_capacity(capacity), m_mask(capacity - 1),
+        m_cells(std::make_unique<Cell[]>(capacity)) {
+    assert(capacity >= 2 && (capacity & (capacity - 1)) == 0 &&
+           "Capacity must be a power of 2");
```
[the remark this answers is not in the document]
sgligorijevicTT: Okay catch, in this repo we use TT_ASSERT or TT_FATAL anyway, please switch to that

sgligorijevicTT:
@copilot apply changes based on the comments in [this thread](https://github.com/tenstorrent/tt-xla/pull/4236#pullrequestreview-4106084435)

pjrt_implementation/src/api/event_instance.cc
```diff
@@ -132,13 +118,12 @@ void EventInstance::markAsReadyAndCallback(EventInstance *event_instance,
...
-  // Release the lock before executing callbacks.
+  // Release the lock before dispatching callbacks.
   ready_lock.unlock();
 
-  // Execute callbacks without holding lock (event may be destroyed in callback)
   for (OnReadyCallback &callback : callbacks_to_execute) {
-    callback.callback_function(*ErrorInstance::makeError(status).release(),
-                               callback.user_arg);
+    getCallbackWorker().enqueue(callback.callback_function, callback.user_arg,
+                                *ErrorInstance::makeError(status).release());
```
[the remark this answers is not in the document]
sgligorijevicTT: @copilot apply changes based on [this feedback](https://github.com/tenstorrent/tt-xla/pull/4236#discussion_r3079621804)

pjrt_implementation/src/utils/unit_tests/mpsc_queue_tests.cc
```diff
@@ -0,0 +1,192 @@
...
+  producer.join();
+  consumer.join();
+
+  ASSERT_EQ(received.size(), num_items);
+  for (size_t i = 0; i < num_items; ++i) {
+    EXPECT_EQ(received[i], i);
+  }
+}
+
+TEST(MPSCQueueUnitTests, MultiProducerSingleConsumer) {
```
jameszianxuTT, line 124: I think another MPSC test covering many producers with fewer items per producer is also a useful case to test

pjrt_implementation/src/utils/callback_worker.cc
```diff
@@ -0,0 +1,57 @@
...
+
+void CallbackWorker::workerLoop() {
+  for (;;) {
+    m_work_available.acquire();
+
+    // Drain all available items.
+    CallbackWorkItem item;
+    while (m_queue.tryPop(item)) {
+      item.callback_function(item.error, item.user_arg);
+    }
```
jameszianxuTT: Serializing these event instance callbacks behind single worker thread does simplify the concurrency situation a lot but does it affect perf? Is there a risk of us being bottlenecked behind this singular worker thread?

My knowledge of what actually happens in these callbacks is pretty limited, ik for bufferInstance doneWithHostBuffer just destructs the backing transferred tensor which shouldn't take long. However, I'm not sure if there are other callbacks that might block for longer

jameszianxuTT (approved):
I am a little wary of adding this much complexity to handle an issue that's not super well understood, given the PR description:

> Event Instance callbacks seem to be deadlocking. Our best understanding is that calling the callbacks from the main thread can sometimes somehow deadlock with the python GIL.

I do appreciate the test coverage though, and the implementation here of singleton MPSC queue and callback executor seems reasonable. 

So a cautious approve from me, though I would prefer that something be added to the backlog to root cause the actual failure

pjrt_implementation/src/utils/callback_worker.cc
```diff
@@ -0,0 +1,57 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
jameszianxuTT: 2026?

pjrt_implementation/src/utils/unit_tests/mpsc_queue_tests.cc
```diff
@@ -0,0 +1,192 @@
...
+  constexpr size_t num_producers = 4;
+  constexpr size_t items_per_producer = 25000;
+  constexpr size_t total_items = num_producers * items_per_producer;
+  MPSCQueue<size_t> queue(1024);
+
+  // Each producer writes values encoded as: producer_id * items_per_producer +
+  // sequence. This lets us verify per-producer ordering after collection.
```
jameszianxuTT, line 131: Nice trick :)

pjrt_implementation/src/utils/unit_tests/mpsc_queue_tests.cc
sgligorijevicTT: Fair point, will add

pjrt_implementation/src/utils/callback_worker.cc
sgligorijevicTT: A comment we have in our code mentions that XLA code comments mention that execution the callback on the caller thread is safe(contrary to what we observed), so I will take it as a sign that callbacks are relatively light. My mental model of events is that callbacks are just ways for you to observe the event, and doing any heavy lifting is supposed to be done elsewhere.
P.S. This made me realize I never pushed a commit that restored that comment after claude deleted it, fixed now.

pjrt_implementation/src/utils/unit_tests/callback_worker_tests.cc
```diff
@@ -0,0 +1,202 @@
...
+      worker.enqueue(incrementCallback, &counter, nullptr);
+    }
+    // Worker destroyed here — should drain all remaining items.
+  }
+
+  EXPECT_EQ(counter.load(std::memory_order_acquire), num_items);
+}
+
+// Callback that records the thread ID it runs on.
+static void recordThreadIdCallback(PJRT_Error *error, void *user_arg) {
```
acicovicTT: It seems like this is not used anywhere

pjrt_implementation/inc/utils/callback_worker.h
```diff
@@ -0,0 +1,70 @@
...
+#include <semaphore>
+#include <thread>
+
+// PJRT C API includes
+#include "xla/pjrt/c/pjrt_c_api.h"
+
+// tt-xla includes
+#include "utils/mpsc_queue.h"
+
+namespace tt::pjrt::utils {
```
acicovicTT: I think `tt::pjrt` is a bettern namespace for this, it is a core mechanic of the plugin and not a small helper like the rest of stuff in `utils`

pjrt_implementation/inc/utils/mpsc_queue.h
```diff
@@ -0,0 +1,132 @@
...
+        // Another producer already claimed this slot but hasn't finished
+        // writing yet. Reload write position and retry.
+        pos = m_write_pos.load(std::memory_order_relaxed);
+      }
+    }
+  }
+
+  // Try to pop an element. Single consumer only.
+  // Returns false if the queue is empty.
+  bool tryPop(T &item) {
```
acicovicTT, line 83: Just thinking out loud, I like the idea of `CallbackWorker` extending `MPSCQueue`, then this method can be `protected` and callable only from the context of `CallbackWorker`, actually ensuring one consumer. Other methods like `isEmpty` could also be `protected`.

pjrt_implementation/src/api/unit_tests/event_instance_tests.cc
```diff
@@ -246,13 +245,19 @@ TEST(EventInstanceUnitTests, API_PJRT_Event_Test_Await_Callbacks_Combination) {
     }
 
     // Spawn a "worker" thread which will wait and then mark the event
-    // as ready. This will trigger callback execution in the same
-    // thread.
+    // as ready. Then immediately destroy the event (simulating what the
+    // XLA on-ready callback does — which we've observed happens in XLA).
     std::thread work_thread = std::thread([&]() {
       std::this_thread::sleep_for(std::chrono::seconds(1));
       std::cerr << "Worker thread marking event as ready and executing "
```
acicovicTT: Just noticed this, not part of your changeset, but can you change it to use loguru?

pjrt_implementation/inc/utils/mpsc_queue.h
```diff
@@ -0,0 +1,132 @@
...
+#include <atomic>
+#include <cstddef>
+#include <memory>
+#include <new>
+#include <utility>
+
+// tt-xla includes
+#include "utils/assert.h"
+
+namespace tt::pjrt::utils {
```
acicovicTT: Same as `CallbackWorker`, I think `tt::pjrt` makes more sense, or at least `tt::pjrt::internal`

pjrt_implementation/src/api/unit_tests/event_instance_tests.cc
```diff
@@ -173,63 +187,48 @@ TEST(EventInstanceUnitTests, API_PJRT_Event_OnReady_ready) {
...
 // ready, but there is also an on-ready callback which destroys the event as
 // soon as it is marked ready. Destroying the event will trigger the execution
 // of an event destructor, which should terminate the execution since there are
 // still awaiters waiting on the event.
+//
+// NOTE: ASSERT_DEATH uses fork(), and that interferes with our callback worker.
+// To work around this, we directly simulate what the
+// callback would do (destroy the event) on the thread that marks the event
+// as ready.
 TEST(EventInstanceUnitTests, API_PJRT_Event_Test_Await_Callbacks_Combination) {
```
acicovicTT: One interesting comment I found produced by `/code-reviewer` skill, copying it verbatim and leaving it up to you to decide:
> However, the original test existed specifically to verify the "callback destroys event" pattern — which is the core scenario that was deadlocking. The new test no longer validates that a callback dispatched through the worker   
  thread can safely destroy the event that triggered it. The test now only proves that destroying an event with active awaiters terminates — which was already working before this PR. Recommendation: Add a separate (non-death) test that verifies the callback-destroys-event pattern works without deadlock

pjrt_implementation/src/utils/unit_tests/callback_worker_tests.cc
```diff
@@ -0,0 +1,202 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```
acicovicTT, line 1: Q: Is there a test for a scenario where the callback thread itself (i.e. a callback function) produces a new task for the callback thread? This should work because produce is a lock-free operation, but should we add a test to check?

acicovicTT (approved):
Approved with some suggestions. Thanks for adding new tests! :)

pjrt_implementation/src/utils/unit_tests/callback_worker_tests.cc
sgligorijevicTT: leftover from an earlier iteration of a test, will remove

pjrt_implementation/inc/utils/callback_worker.h
sgligorijevicTT: yeah that was a mistake, will fix

pjrt_implementation/inc/utils/mpsc_queue.h
sgligorijevicTT: Fair point, I think internal makes the most sense here

pjrt_implementation/inc/utils/mpsc_queue.h
sgligorijevicTT: I don't think I agree. A worker is not a queue, logically this is composition, and I prefer to implement composition as class members. Besides, with protected this would violate Liskov.

acicovicTT: > Besides, with protected this would violate Liskov.

How so?

I'm not pushing this idea, it's up to you and an implementation detail, it's not likely someone will try to call `tryPop`, but it is not protected from having multiple callers (it's possible I'm missing some knowledge here)

pjrt_implementation/src/utils/unit_tests/callback_worker_tests.cc
sgligorijevicTT: I don't expect this to happen in practice, but it's not much effort to add so sure.

pjrt_implementation/src/api/unit_tests/event_instance_tests.cc
sgligorijevicTT: fair point, addressed

pjrt_implementation/inc/utils/mpsc_queue.h
```diff
@@ -0,0 +1,132 @@
...
+//   sequence == pos + 1        -> slot contains data ready for the consumer
+//   sequence == pos + capacity -> slot has been consumed and recycled
+//
+// Producers compete via CAS on a shared write position. The single consumer
+// advances its own read position without contention.
+//
+// Capacity must be a power of 2 (enforced by TT_FATAL).
+template <typename T> class MPSCQueue {
+public:
+  explicit MPSCQueue(size_t capacity)
```
pilkicTT, line 34: please run perf benchmarks - we are now funnelling the callbacks through a single thread + the queue capacity is fixed which can cause further blocking...

pjrt_implementation/src/api/event_instance.cc
```diff
@@ -132,13 +130,14 @@ void EventInstance::markAsReadyAndCallback(EventInstance *event_instance,
...
 
-  // Release the lock before executing callbacks.
+  // Release the lock before dispatching callbacks.
   ready_lock.unlock();
 
-  // Execute callbacks without holding lock (event may be destroyed in callback)
   for (OnReadyCallback &callback : callbacks_to_execute) {
-    callback.callback_function(*ErrorInstance::makeError(status).release(),
-                               callback.user_arg);
+    getCallbackWorker().enqueue(
```
pilkicTT, line 137: since we are now executing callbacks in a different thread, please check `BufferInstance::deleteData()` - we are deleting the event after calling `markAsReadyAndCallback`, is this safe?

pjrt_implementation/inc/utils/callback_worker.h
```diff
@@ -0,0 +1,70 @@
...
+// ensuring callbacks run on a thread that holds neither lock.
+//
+// Thread-safe for multiple concurrent producers (enqueue). The consumer is the
+// internal worker thread.
+class CallbackWorker {
+public:
+  explicit CallbackWorker(size_t queue_capacity = 1024);
+
+  // Signals shutdown, joins the worker thread, and drains any remaining items.
+  ~CallbackWorker();
```
pilkicTT, line 46: why do we even bother with destroying this worker?

pilkicTT: also, does it really drain the remaining items on shutdown? we just return from the worker loop...

pjrt_implementation/inc/utils/mpsc_queue.h
```diff
@@ -0,0 +1,132 @@
...
+
+private:
+  struct Cell {
+    std::atomic<size_t> sequence;
+    T data;
+  };
+
+  const size_t m_capacity;
+  const size_t m_mask;
+  std::unique_ptr<Cell[]> m_cells;
```
pilkicTT, line 121: i think you can align the cells also, if you worry about cache thrashing...

pilkicTT (commented):
please consider trying out the simplest solution (queue + mutex) as well... all this lock-free complexity seems like an overkill...

pjrt_implementation/inc/utils/callback_worker.h
sgligorijevicTT: I would prefer a clean shutdown over risking a segfault(in case the thread isn't joined) or not completing all work(in case thread is detached). The destructor code is short anyway.
As for the draining comment, good catch that's a leftover from an earlier iteration, as it is right now it's not really draining and there shouldn't ever be a need for draining the queue. I will update the comment.

pjrt_implementation/inc/utils/mpsc_queue.h
sgligorijevicTT: I don't think it's worth the tradeoff, positions are just two fields while cells are an array. Also as only writer_pos is used in a tight loop with a CAS, perhaps not even reader_pos needs to be aligned, but one more field is not much, as opposed to an array of aligned structs.

pjrt_implementation/src/api/event_instance.cc
sgligorijevicTT: We synced offline, it's not safe, it will be addressed with a hack, specifically by moving the delete into another callback.

sgligorijevicTT:
This was merged hastily to fix P0s


http://github.com/tenstorrent/tt-xla/issues/4289
Issue for any followup.
Re James's comment: I agree we should get more clarity, tracked in the issue
Re Pedja's comment: The problem setup is a textbook usecase MPSC and AI generated a standard solution so I think it's fine to stay as is, but at least considering a barebones queue is a fair point, will be done at some point.

Issue #4230, opened 2026-04-13
Multihost Call Test Integration


jameszianxuTT:
Reduce CI fragmentation by moving the dual t3k specialized call-test script into the main call-test 

How? Using docker socket container escape to launch a sibling container and conduct multihost operations from that sibling container.

<img width="955" height="667" alt="Image" src="https://github.com/user-attachments/assets/12c888c3-dace-4661-93e8-aca26cfe01b4" />

PR #4241, opened 2026-04-14
Add docker cli install in tt-xla-ci container for multihost call-test integration

jameszianxuTT:
### Ticket
Needed for #4230 

### Problem description
Need to use docker run/exec/stop/rm commands from inside our `container:`  to facilitate multihost integration without requiring a container escape. 

### What's changed
Install latest docker:cli image and copy into local. Based on precedent that metal-internal-workflows provisioning doesn't pin the docker version either.

There is a risk of version skew between host daemon from provisioning and the version installed inside the container, but I expect that to fail loudly, and will verify that it works before merging anything currently. 

### Checklist
- [] New/Existing tests provide coverage for changes

### DIFF
--- a/.github/Dockerfile
+++ b/.github/Dockerfile
@@ -1,4 +1,5 @@
 ARG MLIR_TAG=latest
+ARG DOCKER_CLI_IMAGE=docker:27.5.1-cli
 
 #============================================================================
 # Base image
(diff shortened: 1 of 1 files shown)


nsumrakTT (approved):

vvukomanTT:
Could we try to conditinally mount the docker.sock, then we only need the docker cli in the multihost case?

vvukomanTT (approved):

PR #4243, opened 2026-04-14
Vzeljkovic/nightly

vzeljkovicTT:


sdjukicTT (approved):

tests/torch/models/whisper/test_whisper.py
```diff
@@ -28,6 +29,14 @@
     ModelVariant.WHISPER_LARGE_V3,
     ModelVariant.WHISPER_LARGE_V3_TURBO,
 ]
+# Actual measured PCC per variant (n150: 0.533, p150: 0.535)
+_WHISPER_LARGE_V3_PCC = 0.5
+# Actual measured PCC per variant (n150: 0.720, p150: 0.729)
+_WHISPER_LARGE_V3_TURBO_PCC = 0.7
```
sgligorijevicTT, line 35: For values as low as these, we usually don't do comparisons at all.

Issue #4245, opened 2026-04-14
Failure to run any model: "host channel does not exist"


abrown:
### Describe the bug

On a Blackhole p150b, I am attempting to run any of the models supported by tt-forge-models using tt-forge and tt-xla. All models fail, and the current reason looks like:

```
# python demos/tt-xla/nlp/pytorch/opt_demo.py
...
RuntimeError: TT_ASSERT @ /project/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/tt_metal/third_party/umd/device/chip/local_chip.cpp:241: channel < get_num_host_channels()
info:
Querying size for a host channel that does not exist.
backtrace:
 --- tt::umd::LocalChip::get_host_channel_size(unsigned int)
 --- tt::tt_metal::SystemMemoryManager::SystemMemoryManager(int, unsigned char)
 --- tt::tt_metal::Device::init_command_queue_host()
 --- tt::tt_metal::CommandQueueInitializer::init(std::vector<tt::tt_metal::Device*, std::allocator<tt::tt_metal::Device*> > const&, std::unordered_set<tt::tt_metal::InitializerKey, std::hash<tt::tt_metal::InitializerKey>, std::equal_to<tt::tt_metal::InitializerKey>, std::allocator<tt::tt_metal::InitializerKey> > const&)
 --- tt::tt_metal::DeviceManager::init_firmware_on_active_devices()
...
```

This is likely not a tt-xla issue but I'm posting this here first because I am following tt-xla's [Getting Started](https://docs.tenstorrent.com/tt-xla) guide. Feel free to move this to the right place for resolution.

### Steps to reproduce the issue

I first attempted the [install a wheel](https://docs.tenstorrent.com/tt-xla/#installing-a-wheel-and-running-an-example) path:

```shell-session
$ uv venv --python 3.12 --prompt tt --seed .venv
$ source .venv/bin/activate
$ pip install pjrt-plugin-tt --extra-index-url https://pypi.eng.aws.tenstorrent.com/
$ pip install tt-torch --extra-index-url https://pypi.eng.aws.tenstorrent.com/
$ pip install flax transformers
$ pip install -r demos/tt-xla/nlp/jax/requirements.txt
$ PYTHONPATH=. python demos/tt-xla/nlp/jax/gpt_demo.py
```

[...]

PR #4248, opened 2026-04-14
Clean up getting started docs to use mnist + tinyllama model demos

jameszianxuTT:
### Ticket
Related: #4245 

### Problem description
Current getting started docs are too complex, and don't work due to several issues.

### What's changed
- Provide mnist model as user entrypoint - requiring no dependency on tt-forge-models or external libs
- Remove mentions of gpt model and opt model
- Improve mnist.py model output when run without pytest
- Suggest use of tt-forge-install CLI tool to sync system SFPI version
- Add @ddilbazTT's tinyllama prefill demo as an LLM example

### Checklist
- Tested locally, described in PR comments.

### DIFF
--- a/docs/src/getting_started.md
+++ b/docs/src/getting_started.md
@@ -36,8 +36,14 @@ To install a wheel and run an example model, do the following:
 
 #### Step 1. Install the Latest Wheel:
 
+Download the latest wheel with pip.
 ```bash
 pip install pjrt-plugin-tt --extra-index-url https://pypi.eng.aws.tenstorrent.com/
+```
+
+Run the tt-forge-install script to install missing system dependencies.
+```
+tt-forge-install
 ```
 
 #### Step 2. Run some models:
--- a/docs/src/getting_started.md
+++ b/docs/src/getting_started.md
@@ -42,23 +42,15 @@ pip install pjrt-plugin-tt --extra-index-url https://pypi.eng.aws.tenstorrent.co
 
 #### Step 2. Run a Model:
 
-- Navigate to the section of the [TT-Forge repo that contains TT-XLA demos](https://github.com/tenstorrent/tt-forge/tree/main/demos/tt-xla)
-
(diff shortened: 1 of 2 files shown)


docs/src/getting_started.md
```diff
@@ -42,23 +42,15 @@ pip install pjrt-plugin-tt --extra-index-url https://pypi.eng.aws.tenstorrent.co
...
-   ```
-
-- Download the [**gpt_demo.py** file ](https://github.com/tenstorrent/tt-forge/blob/main/demos/tt-xla/nlp/jax/gpt_demo.py ) The demo you are about to run takes a piece of text and tries to predict the next word that logically follows.
+- Download the [**mnist.py** example](https://github.com/tenstorrent/tt-xla/blob/main/examples/pytorch/mnist.py) from the TT-XLA repo. This demo runs a simple CNN model on a Tenstorrent device and compares the output against a CPU reference.
 
 - Run the model:
 
    ```bash
-   python gpt_demo.py
+   python mnist.py
```
AleksKnezevic: Is there any demo we can use to provide an actual output? the jax demos are all probably broken due to transformers dropping flax support? Do any of the torch ones provide something reasonable?

jameszianxuTT: @ddilbazTT found that tinyllama 1B does work and is ungated + merged the demo into tt-forge, so I will add it. 

However, I think having a very simple baseline example where all the torch-level source code is visible is more approachable for new devs, so I would like to include both demos here.

ddilbazTT: tinyllama looks like this: https://github.com/tenstorrent/tt-forge/commit/20173cee51bf8543685bf7b543048b8c3d4e4695

AleksKnezevic: sounds good.

AleksKnezevic (approved):

jameszianxuTT:
What's tested:
###  Wheel install:
1. make a blank venv - python -m venv venv && source venv/bin/activate
2. pip install the wheel `pip install pjrt-plugin-tt --extra-index-url https://pypi.eng.aws.tenstorrent.com/` and `tt-forge-install` to sync up SFPI system dependency
3. wget test files and run
4. confirm tests passing and output as described

### Docker install:
1. on local BH QB - start tt-xla-slim as described
2. inside container, wget tests and immediately execute by copying code blocks provided
3. confirm tests passing and output as described

PR #3921, opened 2026-03-25
EmitPy testing

sgligorijevicTT:
### Ticket
closes https://github.com/tenstorrent/tt-xla/issues/1760

### Problem description
Leadership wants EmitPy to be tested

### What's changed
Added PythonModelRunner. CodegenPy backend now has the ability to actually execute the generated code trough the frontend.
Added dry_run flag. Previously FB didn't have a dry run flag, and codegen behaved as if dry run was always on. Defaults for dry_run are consistent with old behavior.
Added testinfra changes to test

### Checklist
- [X] New/Existing tests provide coverage for changes

Issue #2139, opened 2025-11-13
Experimental compile splits models in two


odjuricicTT:
With experimental compile enabled https://github.com/tenstorrent/tt-xla/pull/2029, most models we benchmark seem to be split in two. One is usually sam random single op and the other is the model.

Here is what we get in shlo for resnet:

[Archive 2.zip](https://github.com/user-attachments/files/23524291/Archive.2.zip)

Not sure if impact to performance is significant yet, but this should be solved anyway.

svuckovicTT:
Hey @sgligorijevicTT I see this issue is closed. It is mentioned as the reason we use `legacy_compile` in the codegen path - does this mean we can now use the "default" path for codegen?

https://github.com/tenstorrent/tt-xla/blob/c9192ef558b79b3f6091dcf3c266e45766fa003e/python_package/tt_torch/codegen.py#L38

PR #4261, opened 2026-04-15
Reland EmitPy testing

### LINKED ISSUE #4100: Codegen Python standalone execution fails for TP models (tensor sharding inconsistency)
## Summary

Generated Python codegen (`codegen_py` backend) for tensor-parallel (TP) models cannot be executed standalone. The exported tensors have inconsistent sharding — some tensors are TP-sharded while others are not, causing shape mismatches at runtime.

## Repro Steps

```bash
cd /localdev/rpavlovic/tt-xla
git checkout rpavlovic/test-greedy-optimizer

# 1. Enable codegen_py and export_tensors in tests/benchmark/benchmarks/llm_benchmark.py
#    Add to the options dict (~line 387):
#        "backend": "codegen_py",
#        "export_tensors": True,

# (added by svuckovicTT)
# Also make sure to set the optimization level to 0 (currently set to 2)
#   file: tests/benchmark/test_llms.py
#   line: 17
# DEFAULT_TP_OPTIMIZATION_LEVEL = 0

# 2. Generate codegen (4 layers, optimization_level=0)
tt-smi -r 0,1,2,3
source venv/activate
timeout 600 python3 -m pytest -svv tests/benchmark/test_llms.py::test_qwen_3_14b_tp --num-layers 4

# 3. After "Python codegen successful" messages appear, kill the test.
#    Generated files are in modules/:
#      modules/main.py, consteval.py, utils.py, tensors/

# 4. Run standalone
tt-smi -r 0,1,2,3
source venv/activate
cd modules
python3 main.py
```

## Observed Behavior

**With optimization_level=0 (no greedy optimizer):**
```
TT_FATAL: bmm (non-bcast matmul) expects input tensors of shapes BCMK*BCKN=BCMN or equivalent
```
Batch dimensions of matmul operands don't match — one operand is TP-sharded, the other is not.

**With optimization_level=2, greedy optimizer enabled, 16 layers:**
```
TT_FATAL: GQA expects Q to have a multiple of K heads, but got 5 and 8
  @ sdpa_decode_device_operation.cpp:319
```
Q tensor is sharded to 5 heads per device (40/8 TP), but K tensor retains all 8 KV heads (unsharded).

## Root Cause

The codegen tensor export does not consistently handle TP sharding. When a model runs on a 1x8 mesh (TP=8), some tensors are exported in their sharded form while others are exported unsharded. Running standalone, these shape mismatches cause `TT_FATAL` assertions in matmul and SDPA ops.

## Environment

- Model: Qwen/Qwen3-14B (40 Q heads, 8 KV heads)
- Mesh: 1x8 (8 wormhole_b0 devices, 4x n300)
- tt-xla branch: `rpavlovic/test-greedy-optimizer`
- tt-mlir SHA: per tt-xla submodule

## Expected Behavior

Standalone execution of codegen Python should produce a runnable program — all exported tensors should be consistently sharded (or unsharded) so that op shape validations pass.

sgligorijevicTT:
closes #4100

Relands https://github.com/tenstorrent/tt-xla/pull/3921

For now without any scheduled jobs or actions on CI, until EmitPy is stabilized, it will be added to nightly at a later date

Also fixes one testinfra issue I spotted in the meantime

Also piggybacking changes to output shardings when calling codegen path, this fixes another issue


sdjordjevicTT (approved):

nsumrakTT (approved):

vvukomanTT (approved):

svuckovicTT (approved):

pjrt_implementation/src/api/loaded_executable_instance.cc
```diff
@@ -208,6 +211,122 @@ tt_pjrt_status LoadedExecutableInstance::getInputRuntimeTensors(
...
+      output_sharding.getShardType() ==
+          mlir::tt::ttcore::MeshShardType::Replicate) {
+    return output_shape;
+  }
+
+  llvm::SmallVector<int64_t> shard_shape = output_sharding.getShardShape();
+  TT_FATAL(shard_shape.size() == output_shape.size(),
+           "Output sharding shape doesn't match the output shape: "
+           "shard_shape.size()={}, output_shape.size()={}",
+           shard_shape.size(), output_shape.size());
```
mrakitaTT: We should return `tt_pjrt_status` instead of killing the process

pjrt_implementation/src/api/loaded_executable_instance.cc
```diff
@@ -208,6 +211,122 @@ tt_pjrt_status LoadedExecutableInstance::getInputRuntimeTensors(
...
+
+  for (size_t i = 0; i < output_shape.size(); ++i) {
+    TT_FATAL(output_shape[i] % shard_shape[i] == 0,
```
mrakitaTT: Same here

pjrt_implementation/src/api/so_loaded_executable_instance.cc
```diff
@@ -121,10 +121,36 @@ SOLoadedExecutableInstance::execute(PJRT_LoadedExecutable_Execute_Args *args) {
...
+  } else {
+#if !TTXLA_ENABLE_EMITPY_EXECUTION
+    LOG_F(ERROR, "EmitPy execution requested, but this build does not include "
+                 "PythonModelRunner support");
+    return tt_pjrt_status::kInternal;
+#else
+    // Execute the generated Python code via PythonModelRunner.
+    tt::alchemist::PythonModelRunner runner;
+    runner.addToSysPath(options.export_path.value());
+    runner.loadModule("main", "main_for_test");
```
mrakitaTT: Can we read this string constant from the alchemist lib? Or at least put it into some static const string variable.

tests/runner/test_models.py
```diff
@@ -160,7 +160,9 @@ def _run_model_test_impl(
                         pytest.mark.filecheck(test_metadata.filechecks)
                     )
 
-                comparison_result = tester.test(request=request)
+                emitpy_enabled = request.config.getoption("--emitpy", default=False)
```
mrakitaTT: You don't use this until line 174, should be moved below

tests/infra/testers/single_chip/model/model_tester.py
```diff
@@ -150,17 +150,26 @@ def _get_forward_method_name(self) -> str:
         """
         return "__call__"
 
-    def test(self, request=None) -> Tuple[ComparisonResult, ...] | None:
-        """Tests the model depending on test type with which tester was configured."""
+    def test(self, request=None) -> Tuple[Tuple[ComparisonResult, ...] | None, Any]:
```
mrakitaTT, line 153: I really don't like this change, it makes API a bit messy. What about cashing last device result in tester local variable and exposing a getter for it which you will call from emitpy test path?

tests/runner/test_models.py
```diff
@@ -169,6 +171,18 @@ def _run_model_test_impl(
                 #     fallthrough to finally block on failure.
                 Evaluator._assert_on_results(comparison_result)
 
+                # EmitPy verification: re-run via codegen_py and compare
```
mrakitaTT, line 175: We just keep piling up code into this function, should consider extracting some things into separate functions, for example running jax tester, running pytorch tester, running emitpy, etc..

mrakitaTT (approved):
Non-blocking comments, but please consider fixing

jameszianxuTT:
Just a heads up, once #4271 is merged, we will see some merge conflicts

pjrt_implementation/inc/api/compile_options.h
```diff
@@ -115,6 +115,12 @@ struct CompileOptions {
   // default name.
   std::string ttnn_perf_metrics_output_file = "";
 
+  // Perform everything as if the graph was executed, but don't actually execute
+  // it. Instead, return zero-filled output buffers. Useful for just getting the
+  // IRs out, or input tensors, or codegen code. Default true for codegen
+  // backends, false for flatbuffer backend.
+  bool dry_run = true;
```
jameszianxuTT, line 122: I am seeing more overlap with #4271, which I think should land first as it's a smaller change. 

Semantically I'm totally clear what the overlap is between this dry_run and the compile_only path but I think it's a bit messy. 

Compile only is set via env variable and _also_ returns zero-filled output buffers, but also blocks any runtime API calls that touch the device (eg. system descriptor query, transfer of tensors to device in BufferInstance bufferFromHostBuffer). 

It's probably okay to leave both these paths in for now as some tech debt but I think they could be merged in a coherent way at some point.

tests/infra/testers/single_chip/model/model_tester.py
jameszianxuTT: Personally, I prefer fewer side effects on tester API, and adding more statefulness to tester seems worse than complicating the return type

However, I don't really understand the purpose of piping this up through the API in the first place? If it's just for verify_emitpy, is it possible to 
- compare FB and emitpy results separately against CPU? And then compare their comparison result to establish equality via transitive property
- embed the tt_result inside ComparisonResult object under IDENTITY type or something like that

tests/infra/testers/single_chip/model/torch_model_tester.py
```diff
@@ -310,6 +313,76 @@ def _test_training(self) -> Tuple[ComparisonResult, ...]:
...
+            emitpy_options = {
+                **original_options,
+                "backend": "codegen_py",
+                "export_path": emitpy_export_path,
+                "export_tensors": "true",
+                "dry_run": "false",
+            }
+            torch_xla.set_custom_compile_options(emitpy_options)
+
+            # Recompile with legacy compile to avoid _run_cached_graph.
```
jameszianxuTT: I'm not totally clear on why emitpy needs to use legacy compile. It seems like a hack based on comment in codegen.py


```
# Using legacy compile is a temporary hack to 1) Make MetaDataProp work 2) Decrease odds of codegenning for graphs that never need to get executed.
# New compile serves to primarily reduce execution overhead, which we don't really care about if executing only once for codegen.
# TODO(sgligorijevic): Clean this up

```


Maybe it's worth making an issue here and referencing it in both places? Not totally necessary though, just to avoid confusion from people reading this comment

jameszianxuTT (approved):
lgtm, a few minor nits and suggestions especially regarding overlap with #4271

tests/infra/testers/single_chip/model/model_tester.py
sgligorijevicTT: My preference is generally towards eliminating state therefore I did it this way, but I don't have a strong opinion one way or another.

I am decently confident that we can't get this comparison result transitively. The ask from compiler team was to check for bit exact match, meanwhile our accuracy standards for CPU vs device are **very** loose, and both pcc and max difference can stay identical for significant changes to underlying values of the result because of that. Besides, that forces me to pass the CPU result into the function, and we're back to square one.
As for packing the value in the ComparisonResult, I don't really like that idea as it's not a result of any comparison.

tests/runner/test_models.py
sgligorijevicTT: In principle I agree but I would strongly prefer that a testinfra refactor stays separate from landing emitpy testing

pjrt_implementation/inc/api/compile_options.h
sgligorijevicTT: Totally agree.

tests/infra/testers/single_chip/model/torch_model_tester.py
sgligorijevicTT: Actually that comment you pinned your review to is stale, I will remove it. 

I know I opened an issue for this before but I can't find it right now. Basically, set_custom_compile_options is global state, that is ignored by both Dynamo, DynamoBridge and TorchXLA, and that can come back to bite us hard. That's why I aggressively clean caches inside verify_emitpy, why I'm careful with what I send into that function, and why in a previous version I reran the CPU and FB again inside this fn.

As for the comment you copy pasted, 1) might be out of date IDK, 2) sure I can link an old issue for that ( https://github.com/tenstorrent/tt-xla/issues/2139 )

PR #4269, opened 2026-04-15
Handle empty tensor copy-to-host by skipping runtime memcpy for zero-size buffers

### LINKED ISSUE #4260: ViLT(QA & MLM) : Copy to host buffer failed with error: Buffer pointers must not be null
### Describe the bug

- Vilt model (both QA & Mlm) fails with `Copy to host buffer failed with error: Buffer pointers must not be null`
- In QA, Issues originates from this [nonzero](https://github.com/huggingface/transformers/blob/7d9754a05193eb79b1d86aa744b622b8068008cd/src/transformers/models/vilt/modeling_vilt.py#L142) operation
- Input x_mask is ones tensors, so `(1 - x_mask).nonzero(as_tuple=False)` leads to creation of empty tensor and it crashes on host transfer 

### Repro

```
git checkout kkannan/apr15_empty_tensor_crash
git submodule update --init --recursive

pytest tests/runner/test_models.py::test_all_models_torch[vilt/question_answering/pytorch-Vqa-single_device-inference] -svv
pytest tests/runner/test_models.py::test_all_models_torch[vilt/masked_lm/pytorch-Mlm-single_device-inference] -svv

pytest tests/torch/ops/test_non_zero.py -svv
```

### Logs

- [apr15_non_zero_before_fix.log](https://github.com/user-attachments/files/26757159/apr15_non_zero_before_fix.log)
- [apr15_vilt_mlm_before_fix.log](https://github.com/user-attachments/files/26757157/apr15_vilt_mlm_before_fix.log)
- [apr15_vilt_qa_before_fix.log](https://github.com/user-attachments/files/26757158/apr15_vilt_qa_before_fix.log)

kamalrajkannan78:
### Ticket

- closes https://github.com/tenstorrent/tt-xla/issues/4260

### Problem description

- Empty tensors crash during device-to-host copy. The tt-mlir runtime's getRawHostDataPtr() returns nullptr for zero-size buffers, and handleBufferCast() throws `Buffer pointers must not be null`, causing a fatal abort.

### What's changed

- Added an early return in onBufferToHostBuffer() when `logicalTensorSize() == 0`. Mirrors the existing `!args->dst` early return pattern. Since there is no data to copy, we skip the runtime memcpy and return a ready event immediately. This prevents the null pointer from reaching tt-mlir's `handleBufferCast()`.

### Checklist

### Logs

**Before fix**

- [apr15_empty_before_fix.log](https://github.com/user-attachments/files/26757205/apr15_empty_before_fix.log)
- [apr15_vilt_mlm_before_fix.log](https://github.com/user-attachments/files/26757157/apr15_vilt_mlm_before_fix.log)
- [apr15_vilt_qa_before_fix.log](https://github.com/user-attachments/files/26757158/apr15_vilt_qa_before_fix.log)

**After fix**

- [apr15_empty_after_fix.log](https://github.com/user-attachments/files/26757194/apr15_empty_after_fix.log)
- [apr15_vilt_mlm_after_fix.log](https://github.com/user-attachments/files/26757195/apr15_vilt_mlm_after_fix.log)
- [apr15_vilt_qa_after_fix.log](https://github.com/user-attachments/files/26757196/apr15_vilt_qa_after_fix.log)

### DIFF
--- a/pjrt_implementation/src/api/buffer_instance.cc
+++ b/pjrt_implementation/src/api/buffer_instance.cc
@@ -513,6 +513,17 @@ PJRT_Error *onBufferToHostBuffer(PJRT_Buffer_ToHostBuffer_Args *args) {
     return nullptr;
   }
 
+  // Empty tensors (e.g. shape [0]) have zero bytes and no device-side
+  // allocation, so there is nothing to copy. Return a ready event immediately
+  // to avoid calling into the runtime with null buffer pointers.
+  if (buffer->logicalTensorSize() == 0) {
+    std::unique_ptr<EventInstance> event = EventInstance::createInstance();
+    EventInstance::markAsReadyAndCallback(event.get(),
+                                          tt_pjrt_status::kSuccess);
+    args->event = *event.release();
+    return nullptr;
+  }
+
   return *ErrorInstance::makeError(
               buffer->copyToHost(
                   args->dst, args->dst_size,
--- a/tests/torch/ops/test_empty.py
+++ b/tests/torch/ops/test_empty.py
@@ -1,0 +1,39 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Minimal torch.empty empty-tensor host-transfer repro."""
+
+import pytest
+import torch
+import torch_xla.core.xla_model as xm
+import torch_xla.runtime as xr
+from utils import Category
+
+
+@pytest.mark.nightly
+@pytest.mark.single_device
+@pytest.mark.record_test_properties(
+    category=Category.OP_TEST,
+    torch_op_name="torch.empty",
+)
+def test_empty():
+    class Empty(torch.nn.Module):
+        def forward(self, x):
+            return torch.empty((0,), device=x.device, dtype=torch.int64)
+
+    xr.set_device_type("TT")
+
+    model = torch.compile(Empty(), backend="tt")
+    device = xm.xla_device()
+    model = model.to(device)
+    input_tensor = torch.tensor(0, dtype=torch.int64).to(device)
+
+    with torch.no_grad():
+        output = model(input_tensor)
+
+    host_output = output.to("cpu")
+
+    assert host_output.numel() == 0
+    assert host_output.shape == torch.Size([0])
+    assert host_output.dtype == torch.int64


saiarthiraguram (approved):

ppadjinTT:
@pilkicTT @mstojkovicTT can someone take a look?

pjrt_implementation/src/api/buffer_instance.cc
```diff
@@ -501,6 +501,17 @@ PJRT_Error *onBufferToHostBuffer(PJRT_Buffer_ToHostBuffer_Args *args) {
...
+  // allocation, so there is nothing to copy. Return a ready event immediately
+  // to avoid calling into the runtime with null buffer pointers.
+  if (buffer->logicalTensorSize() == 0) {
+    std::unique_ptr<EventInstance> event = EventInstance::createInstance();
+    EventInstance::markAsReadyAndCallback(event.get(),
+                                          tt_pjrt_status::kSuccess);
+    args->event = *event.release();
+    return nullptr;
+  }
+
```
jameszianxuTT, line 526: This is an interesting problem and I'm surprised we haven't seen it come up before. 

> Input x_mask is ones tensors, so (1 - x_mask).nonzero(as_tuple=False) leads to creation of empty tensor and it crashes on host transfer

Does it make sense is going on in the IR? The non_zero_before log shows that the graph should return a 
`tensor<1x192xi64>`, is this entity actually the [0] shape tensor (... in which case, are we not modelling something correctly in compiler)

My point is that this fix does look like it resolves the problem at hand, of a crash when returning an empty tensor, but I wonder if something else is going wrong to try to return an empty tensor in the first place.

jameszianxuTT: Though, from the testcase you added, I guess it is possible to induce this failure easily...

kamalrajkannan78: - In the non_zero sanity log attached [here](https://github.com/tenstorrent/tt-xla/issues/4260), tensor<1x192xi64> in the compiled graph is the output of (1 - x_mask), the subtraction step before nonzero.
- The nonzero op is not present in the StableHLO IR. The result of nonzero on an all-zeros tensor is a valid empty tensor [0, 2], which is expected PyTorch behavior. 
- The crash is only in the PJRT layer not handling this valid empty tensor during host transfer. Apologies if my test name test_non_zero created confusion — the compiled graph only captures the subtraction.

```
class nonzero(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x_mask ):
        non_valid_idx = (1 - x_mask).nonzero(as_tuple=False)
        return non_valid_idx
```

kamalrajkannan78:
@acolicTT could you pls review this PR?

jameszianxuTT (approved):

tests/torch/ops/test_empty.py
```diff
@@ -0,0 +1,35 @@
...
+
+    model = torch.compile(Empty(), backend="tt")
+    device = xm.xla_device()
+    model = model.to(device)
+    input_tensor = torch.tensor(0, dtype=torch.int64).to(device)
+
+    with torch.no_grad():
+        output = model(input_tensor)
+
+    host_output = output.to("cpu")
```
acolicTT, line 35: Can you add some sanity assert, so we know that we got 0 tensor on host?

acolicTT (approved):

PR #4271, opened 2026-04-15
Compile only improved

AleksKnezevic:
### Problem description
We want to be able to run a compilation on a CPU only machine. 

### What's changed
Don't query device at all if `TT_COMPILE_ONLY_SYSTEM_DESC` is set. Instead of erroring out if `TT_COMPILE_ONLY_SYSTEM_DESC` is set, return uninitialized data and don't validate it.


tests/infra/evaluators/evaluation_config.py
```diff
@@ -53,6 +54,11 @@ class ComparisonConfig:
     allclose: AllcloseConfig = field(default_factory=lambda: AllcloseConfig(False))
     assert_on_failure: bool = True  # Default to True for backwards compatibility
 
+    def __post_init__(self):
+        if os.environ.get("TT_COMPILE_ONLY_SYSTEM_DESC"):
+            self.disable_all()
+            self.assert_on_failure = False
```
jameszianxuTT, line 63: This makes sense but I think you should log a warning here - if someone gets as far as constructing a comparison config in CPU only mode, they might be doing something wrong and should be notified

Also:
- if someone wanted to just use ComparisonConfig for CPU-CPU comparisons on a compile only script for some reason, they might get confused why the ComparisonConfig doesn't work
- the way that `TT_COMPILE_ONLY_SYSTEM_DESC` is currently set through enable_compile_only() is pretty unsafe now that I look at it. It contaminates the env for other tests run in the same process, which is currently not an issue due to subprocess wrapping of example tests, but still, the risk of leaking this env var and breaking future comparators is there

jameszianxuTT (approved):

jameszianxuTT:
I would also like to put up some regression tests for CPU-only workflow since it has a large surface area for regressions (eg. someone adds a runtime call that touches device somewhere in PJRT).

Either by moving the meta infra tests to ubuntu-latest (though that seems to have a few issues) or making a separate testgroup running on CPU only runners.

That can go in after this PR though

tests/infra/evaluators/evaluation_config.py
AleksKnezevic: agreed. I added a warning and removed the enable helper. Users can set env var as desired. It's still set in the examples, but, as you said, those are run in a single process.

.github/workflows/call-test.yml
```diff
@@ -425,7 +410,7 @@ jobs:
 
       - name: Show consolidated test report
         continue-on-error: true
-        uses: mikepenz/action-junit-report@v6
+        uses: mikepenz/action-junit-report@v5
```
nsumrakTT: Please do not downgrade actions

.github/workflows/call-test.yml
```diff
@@ -179,16 +174,6 @@ jobs:
...
-    - name: Install custom torch_xla from workflow run
-      if: ${{ inputs.torchxla_build_run_id != '' }}
-      shell: bash
-      run: |
-        echo "Using custom torchxla install from run ${{ inputs.torchxla_build_run_id }}"
-        gh run download ${{ inputs.torchxla_build_run_id }} --repo ${{ github.repository }}
-        source venv/activate
-        uv pip install --force-reinstall xla-torch-wheel-release/*.whl
-        uv pip show torch_xla
-
```
nsumrakTT, line 200: how is this removal connected with the topic of this PR?
It seems as merge error to me.

AleksKnezevic: you are correct.

acicovicTT (approved):

nsumrakTT (approved):

pjrt_implementation/src/api/buffer_instance.cc
```diff
@@ -322,7 +323,17 @@ tt_pjrt_status BufferInstance::copyToHost(void *host_buffer,
                                           size_t host_buffer_size,
                                           EventInstance **out_copy_done_event) {
   ZoneScoped;
-  TT_FATAL(m_pjrt_tensor, "Copy from buffer without an associated tensor.");
+
+  // In compile-only mode, output buffers have no associated tensor (no
+  // hardware to compute on). Zero-fill the host buffer and signal success.
+  if (!m_pjrt_tensor) {
```
hshahTT, line 341: We should add a check to make sure that we are in compile-only mode before zero-filling the host buffer. So something like:
```C++
if (!m_pjrt_tensor) {
    if (!is_compile_only) {
        // Throw error similar to the previous TT_FATAL 
    } else {
        // zero-fill buffer
    }
}
```

hshahTT (changes requested):
Overall looks good to me. Just have a suggestion to add a compile mode check in `copyToHost` to prevent any unintended copying from buffer without an associated tensor **outside** of compile only mode.

pjrt_implementation/src/api/buffer_instance.cc
AleksKnezevic: Good call. Addressed.

hshahTT (approved):

vvukomanTT (approved):

sshonTT (approved):

kmabeeTT (approved):
Looks good but did you mean to include `examples/pytorch/n150_system_desc.ttsys` binary file? Seemed like https://github.com/tenstorrent/tt-xla/pull/4271/commits/c65faae5dac4846371056c66f78fc2cad60d6128 removed one but not both :)

vmilosevic:
Should we add a workflow covering this case, running compile "cpu" on ubuntu-latest machine?

AleksKnezevic:
@vmilosevic agreed, I'll add it in a subsequent PR.

PR #4279, opened 2026-04-16
[Accuracy benchmark] Fix reference generator and bump transformers to 5.2.0 in accuracy tests

### LINKED ISSUE #4146: Benchmark accuracy tests failing on CI due to mismatch in transformers version.
`check_transformers_version` in `tests/benchmark/benchmarks/llm_benchmark.py` expects latest transformers version when running any benchmark test (5.2.0). Turns out `.github/workflows/perf-bench-matrix.json` contains old version of transformers for all accuracy tests: `transformers==4.57.1`

TODO: check if accuracy tests run on new transformers version. Expectation is that they should run since they don't differ that much from regular benchmark tests (just input sequence length is bigger). If accuracy tests work with new transformers version there is no reason to keep old version.

dgolubovicTT:
### Ticket
Closes #4146 

### Problem description
The reference generator was using LLMSamplingWrapper which only returns the last position's logits, discarding all prefill predictions. This produced 64 predictions instead of the expected 127, but was hidden because existing .refpt files predated the breaking change.

### What's changed
Solving minor bugs in reference generation, and cleanup after transformers version uplift.
- Replace LLMSamplingWrapper with direct model calls that capture all position logits during prefill. Bump transformers from 4.57.1 to 5.2.0 across all accuracy test configs.
- Remove unnecessary Qwen rope_scaling override in reference generator.
- Remove batch-size override from perf-benchmark-matrix because now tests pass with default batch size 32

### DIFF
--- a/.github/workflows/perf-bench-matrix.json
+++ b/.github/workflows/perf-bench-matrix.json
@@ -322,15 +322,13 @@
         "name": "llama_3_1_8b_instruct_accuracy",
         "pyreq": "datasets loguru pytest requests tabulate timm torch==2.9.0 tqdm transformers==5.2.0",
         "pytest": "tests/benchmark/test_llms.py::test_llama_3_1_8b",
-        "accuracy-testing": true,
-        "batch-size": 16
+        "accuracy-testing": true
       },
       {
         "name": "mistral_7b_accuracy",
         "pyreq": "datasets loguru pytest requests torch==2.9.0 tqdm transformers==5.2.0 protobuf sentencepiece",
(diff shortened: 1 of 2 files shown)


vvukomanTT (approved):

nsumrakTT (approved):

vkovacevicTT (approved):
Thanks for fixing this :)

vkovacevicTT:
We will bring back full prefill logits in tt-xla benchmark after this lands: https://github.com/tenstorrent/tt-mlir/pull/7961

PR #4285, opened 2026-04-16
[vLLM] Add batch=32 test coverage for on-device sampling and p150 vllm tests

### LINKED ISSUE #3610: [vLLM] large batch size cause OOM when sampling with device
### Describe the bug

Running batch 32 with opt-125m on vLLM gives OOM. The issue is related to device sampling, a temporary workaround is to use CPU sampling instead.

With branch in https://github.com/tenstorrent/tt-xla/pull/3554, 

Comment out cpu sampling
```
diff --git a/tests/benchmark/benchmarks/vllm_benchmark.py b/tests/benchmark/benchmarks/vllm_benchmark.py
index 62280ea2..66af26ae 100644
--- a/tests/benchmark/benchmarks/vllm_benchmark.py
+++ b/tests/benchmark/benchmarks/vllm_benchmark.py
@@ -39,7 +39,7 @@ class VLLMBenchmarkConfig:
 def _create_llm(config: VLLMBenchmarkConfig) -> vllm.LLM:
     """Build engine args from config and create a vLLM LLM instance."""
     additional_config = dict(config.additional_config)
-    additional_config.setdefault("cpu_sampling", True)
+    # additional_config.setdefault("cpu_sampling", True)

     llm_args: Dict[str, Any] = {
         "model": config.model,
```
Then run
```
pytest -svv tests/benchmark/test_vllm_benchmarks.py::test_vllm_benchmark[opt-125m-batch32]
```
gives the OOM error
```
2026-03-05 18:05:21.318 | critical |          Always | TT_FATAL: Out of Memory: Not enough space to allocate 6589251584 B DRAM buffer across 12 banks, where each bank needs to store 549105664 B, but bank size is 1071821792 B (allocated: 582463392 B, free: 489358400 B, largest free block: 480051168 B) (assert.hpp:104)
2026-03-05 18:05:21.358 (  96.425s) [        B4FF9640]                utils.h:62     ERR| Exception:
{TT_FATAL @ /localdev/btsoi/tt-xla/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/tt_metal/impl/allocator/bank_manager.cpp:434: false
info:
Out of Memory: Not enough space to allocate 6589251584 B DRAM buffer across 12 banks, where each bank needs to store 549105664 B, but bank size is 1071821792 B (allocated: 582463392 B, free: 489358400 B, largest free block: 480051168 B)
```
I tried to have Claude look into the mlir graphs and how memory usage increases as batch size increases. It is saying that it is moreh cumsum using lots of memory.

### Expected behavior

_No response_

_No response_

kmabeeTT:
### Ticket
Closes https://github.com/tenstorrent/tt-xla/issues/3610

### Problem description
- Running opt-125m with batch size 32 caused an OOM crash on TT hardware. The root cause was the MoreH cumsum op during on-device top-p filtering over-allocating DRAM proportional to batch size, exhausting per-bank space despite sufficient total DRAM. The underlying issue is now resolved, but there was no test coverage at batch=32 to guard against regression.
- vllm tests like llama were failing on blackhole due to SDPA op issues and not tested, recently.

### What's changed

**`tests/integrations/vllm_plugin/sampling/test_sampling_params_synthetic.py`**:
- Add `test_greedy_batch32` (push) — greedy argmax at batch=32 using the largest production vocab size (gpt-oss-120b, 201088)
- Add `test_non_greedy_batch32` (push) — stochastic (top-k/top-p) sampling at batch=32; directly exercises the cumsum path that previously OOM'd

**`tests/integrations/vllm_plugin/generative/test_opt_generation.py`**:
- Add `test_opt_generation_large_batch` (nightly) — parametrized over batch sizes 8, 16, and 32

**`.github/workflows/test-matrix-presets/vllm-model-tests.json`, `vllm-model-tests-nightly.json`**:
- Add p150 (Blackhole) to vLLM push and nightly test matrices, mirroring the existing n150 single_device jobs
- Add run_vllm_p150_tests to call-test.yml like run_vllm_n150_tests so nightly tests work

### Checklist
- [x] New tests in push/onPR and vllm nightly passing on branch: https://github.com/tenstorrent/tt-xla/actions/runs/24700367061

### DIFF
--- a/tests/integrations/vllm_plugin/generative/test_opt_generation.py
+++ b/tests/integrations/vllm_plugin/generative/test_opt_generation.py
@@ -67,9 +67,9 @@ def test_opt_generation_large_batch(batch_size):
     sampling_params = vllm.SamplingParams(temperature=0.8, top_p=0.95, max_tokens=16)
     llm_args = {
         "model": "facebook/opt-125m",
-        "max_num_batched_tokens": batch_size * 128,
+        "max_num_batched_tokens": batch_size * 32,
         "max_num_seqs": batch_size,
-        "max_model_len": 128,
+        "max_model_len": 32,
         "gpu_memory_utilization": 0.001,
         "additional_config": {
             "enable_const_eval": False,
--- a/tests/integrations/vllm_plugin/generative/test_opt_generation.py
+++ b/tests/integrations/vllm_plugin/generative/test_opt_generation.py
@@ -55,3 +55,29 @@ def test_opt_generation_multibatch():
     output_text2 = output[1].outputs[0].text
     print(f"prompt: {prompts[0]}, output: {output_text1}")
     print(f"prompt: {prompts[1]}, output: {output_text2}")
+
+
+@pytest.mark.nightly
+@pytest.mark.single_device
+@pytest.mark.parametrize(
+    "batch_size", [8, 16, 32], ids=["batch8", "batch16", "batch32"]
+)
+def test_opt_generation_large_batch(batch_size):
+    prompts = ["The quick brown fox"] * batch_size
+    sampling_params = vllm.SamplingParams(temperature=0.8, top_p=0.95, max_tokens=16)
+    llm_args = {
+        "model": "facebook/opt-125m",
+        "max_num_batched_tokens": batch_size * 128,
+        "max_num_seqs": batch_size,
+        "max_model_len": 128,
+        "gpu_memory_utilization": 0.001,
+        "additional_config": {
+            "enable_const_eval": False,
+            "min_context_len": 32,
+        },
+    }
+    llm = vllm.LLM(**llm_args)
+    outputs = llm.generate(prompts, sampling_params)
+    assert len(outputs) == batch_size
+    for i, out in enumerate(outputs):
+        print(f"[{i}] {out.outputs[0].text}")
--- a/.github/workflows/call-test.yml
+++ b/.github/workflows/call-test.yml
@@ -183,6 +183,7 @@ jobs:
           --dir wheels \
           --name ${{ steps.strings.outputs.wheel_artifact_name }}
         if [[ "${{ matrix.build.name }}" == "run_vllm_n150_tests" || \
+              "${{ matrix.build.name }}" == "run_vllm_p150_tests" || \
               "${{ matrix.build.name }}" == "run_vllm_n300_tests" || \
               "${{ matrix.build.name }}" == "run_vllm_llmbox_tests" ]]; then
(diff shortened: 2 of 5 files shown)


kmabeeTT:
Using this PR to also add missing p150 coverage (which was blocked due to a bh specific sdpa op issue 3 weeks ago), will update title and desc.

vvukomanTT (approved):

nsumrakTT (approved):

tests/integrations/vllm_plugin/generative/test_opt_generation.py
```diff
@@ -55,3 +55,29 @@ def test_opt_generation_multibatch():
...
     print(f"prompt: {prompts[0]}, output: {output_text1}")
     print(f"prompt: {prompts[1]}, output: {output_text2}")
+
+
+@pytest.mark.nightly
+@pytest.mark.single_device
+@pytest.mark.parametrize(
+    "batch_size", [8, 16, 32], ids=["batch8", "batch16", "batch32"]
+)
+def test_opt_generation_large_batch(batch_size):
```
AleksKnezevic, line 65: Can we add this for all tests? Maybe just 1 & 32?

kmabeeTT: Hmm.. all vllm models? We could, but would prefer to keep batch32 in vllm benchmark tests only and beef that up, in general, due to nightly CI run time impact, and to have less overlap between benchmark/non-benchmark tests (both would end up running batch32).   Original oom bug was purely sampler specific (reproducible in opt-125m or even the synthetic model-less tests), so this PR more than covers that hole.

AleksKnezevic: That makes sense, I'll create an issue for the benchmarks.

tests/integrations/vllm_plugin/generative/test_opt_generation.py
```diff
@@ -55,3 +55,29 @@ def test_opt_generation_multibatch():
...
+    prompts = ["The quick brown fox"] * batch_size
+    sampling_params = vllm.SamplingParams(temperature=0.8, top_p=0.95, max_tokens=16)
+    llm_args = {
+        "model": "facebook/opt-125m",
+        "max_num_batched_tokens": batch_size * 128,
+        "max_num_seqs": batch_size,
+        "max_model_len": 128,
```
mmanzoorTT: You are using small prompt so you can reduce the lengths as below or make `min_context_len=128`
```suggestion
        "max_num_batched_tokens": batch_size * 32,
        "max_num_seqs": batch_size,
        "max_model_len": 32,
```

You will be precompiling one graph instead of 3 for prefill with this change.

kmabeeTT: fair point, done updated.

mmanzoorTT (approved):

PR #4286, opened 2026-04-16
Multihost CI - call-test integration

jameszianxuTT:
### Ticket
#4230 

### Problem description
CI is fragmented and existing multihost CI is not fully integrated into existing workflows.

### What's changed
Multihost test execution in call-test.yml:
  - Added multihost flag support in test matrix configuration
  - Split Run Test step into 3 phases for env setup/teardown and actual pytest execution
  - Added helper scripts: filter_multihost_worker_hostfile.sh and start_multihost_container.sh for multihost container setup and run multihost sibling container startup / teardown 
  - Add `basic-multihost-test.json` preset for dual BH loudbox and dual T3K topologies

Multihost Topology configuration refactored:
  - Moved tests/torch/multi_host/conftest.py → scripts/multihost_topology.py as a standalone CLI tool
  - Topology env vars are now set in CI before pytest runs (via eval $(python scripts/multihost_topology.py ...))
  - Added dual_bh_loudbox_1x16 topology configuration
  - Added to manual-test.yml dropdown options

  Test improvements:
  - Simplified test_multihost_basic.py to use get_mesh_shape_for_device_count() from infra utilities
  - Added push marks to multihost basic tests for integration with standard CI
  - Cleaned up remote_docker.sh script (removed sudo, parameterized username/container name)

  Additional workflow fixes:
  - Optional -m and -k pytest arguments (only added when non-empty)

### Checklist
- [x] Example run: https://github.com/tenstorrent/tt-xla/actions/runs/24677656583


.github/workflows/call-test.yml
```diff
@@ -81,6 +81,7 @@ jobs:
       image: ${{ !matrix.build.shared-runners && inputs.docker_image || format('harbor.ci.tenstorrent.net/{0}', inputs.docker_image) }}
       options: --device /dev/tenstorrent
       volumes:
+        - /var/run/docker.sock:/var/run/docker.sock
```
jameszianxuTT, line 84: tbd - consider security implications. Is there a nice way to variably mount this...

.github/workflows/call-test.yml
```diff
@@ -72,7 +72,7 @@ jobs:
       matrix:
         build: ${{ fromJson(needs.generate-matrix-test.outputs.test_matrix) }}
 
-    runs-on: ${{ !matrix.build.shared-runners && fromJson(format('["{0}", "in-service"]', matrix.build.runs-on)) || matrix.build.runs-on  }}
+    runs-on: ${{ matrix.build.runs-on }}
```
jameszianxuTT, line 75: TBD - revert; put 2xbhlb in service

.github/workflows/call-test-multihost-dualt3k.yml
```diff
@@ -10,9 +10,8 @@ name: Multi-host Test - Dual T3K (call)
 #   3. Installs the wheel artifact inside the controller container
 #   4. Runs pytest inside the controller container
 #
-# Environment variables for distributed execution (TT_DISTRIBUTED_*) are set
-# automatically by the configure_topology pytest fixture in the test's conftest,
-# using the "dual_t3k" topology from tests/torch/multi_host/conftest.py.
+# Distributed-related env is provided by the runner/container; this job sets
+# dual-t3k-specific variables (e.g. TT_RUNTIME_USING_DUALT3K) before pytest.
```
jameszianxuTT: TBD remove this file

.github/workflows/call-test.yml
```diff
@@ -229,6 +231,84 @@ jobs:
           fi
         fi
 
+    - name: Install Docker CLI for multihost
+      if: ${{ matrix.build.multihost }}
+      run: |
+        curl -fsSL https://get.docker.com | sh
```
jameszianxuTT, line 238: I think we can bake this into the tt-xla-base image too, I'm not sure if that's desirable

#4241

.github/workflows/call-test.yml
```diff
@@ -229,6 +231,84 @@ jobs:
...
+            --tag-output \
+            bash -lc "
+              echo '\''$SCRIPT_B64'\'' | base64 -d | bash -s -- \
+                '\''$DOCKER_IMAGE'\'' \
+                '\''$HOST_WORKSPACE_PATH'\'' \
+                '\''$IN_CONTAINER_WORKSPACE'\'' \
+                ubuntu-host-mapped
+            "
+        '
+        echo "All multihost containers ready"
```
jameszianxuTT, line 290: There is some unfortunate messiness around doing everything with docker exec wrap -> mpirun wrap -> shell scripts. It is pretty reason about additionally nested shell scripts, and I thought this the least bad compromise by reusing a simple .github/scripts/start_multihost_container.sh and wrapping that with docker exec + mpirun

.github/workflows/call-test.yml
```diff
@@ -341,8 +432,7 @@ jobs:
       run: |
         source venv/activate
         echo "Collecting tests for group ${{ matrix.build.group-id || 1}}..."
-        python -m pytest --collect-only -q --disable-warnings ${{ env.BASE_PYTEST_CMD }} > collected_tests.txt
-        cat collected_tests.txt
+        python -m pytest --collect-only -q --disable-warnings ${{ env.BASE_PYTEST_CMD }} |& tee collected_tests.txt
```
jameszianxuTT, line 436: I added tee results here since test collection fails sometimes and it's easier to debug

.github/workflows/call-test.yml
nsumrakTT, line 75: This will allow out of service hosts to run the test workflows (all), please revert

nsumrakTT (commented):
I understand logic to keep call-test.yml as a single test workflow, however:
* There are too many ifs inside (including Run test!) which make this test workflow really hard to comprehend.
* if we are letting docker.sock in the same workflow we are introducing security problem to ALL test machines.
* It seems to me that you've joined these two on someone's request (I see deleted multihost workflow) but IMO there are too much problems that this implementation can persist. Maybe, we could have call-test that call call-test-singlehost/call-test-multihost workflows depending on multihost flag?

.github/workflows/call-test.yml
jameszianxuTT: Sorry, meant to revert this, it was a testing change for experimental OOS multihost cluster

.github/workflows/call-test.yml
```diff
@@ -229,6 +231,84 @@ jobs:
...
+            bash -c "
+              echo \"\$(hostname): hostname OK\"
+              if [ -f /.dockerenv ]; then
+                echo \"\$(hostname): /.dockerenv EXISTS (running inside container)\"
+              else
+                echo \"\$(hostname): /.dockerenv NOT FOUND (running on bare metal)\"
+              fi
+            "
+          echo "=== End SSH connectivity check (via remote_docker.sh) ==="
+        '
```
jameszianxuTT, line 311: Can be removed

.github/workflows/call-test.yml
jameszianxuTT: >There are too many ifs inside (including Run test!) which make this test workflow really hard to comprehend.

I think we can debate about how much complexity is acceptable in call-test. There is already a lot of design effort and complexity in call-test to make it this monolithic entry point into tt-xla test infra. I have tried to minimize the additional complexity in this change - In this PR are 3 branch points added to call-test, and I don't think it is that complex:

1. Multihost container setup
2. Run Test (wrapped in docker exec for multihost)
3. Multihost container teardown

If it helps, I can abstract these even further by hiding the complexity in bash launch/teardown scripts, and go over the design decisions and constraints for multihost CI with infra team.

> if we are letting docker.sock in the same workflow we are introducing security problem to ALL test machines.

I understand this is a security concern, and we can make this variable, at the cost of more pipeline complexity to prepare docker container arguments in a job prior to call-test. It is not necessary to mount globally.

> It seems to me that you've joined these two on someone's request (I see deleted multihost workflow) but IMO there are too much problems that this implementation can persist. Maybe, we could have call-test that call call-test-singlehost/call-test-multihost workflows depending on multihost flag?

Multihost comes with some inherent environment complexity and we have to pay for that somehow. 

- Maintaining essentially duplicate call-test pipelines to avoid the three branch described points above is also a maintenance burden as these pipelines have similar execution flows, but different environment setup
- I see that call-test is already fragmented for benchmarks in call-perf-tests, and that is more justifiable based on the huge difference in execution flow between existing call test. In my opinion, the overlap between singlehost and multihost is much larger and don't justify further fragmenting the call-test space

vmilosevic:
Great work James! Thanks for setting this up.

Biggest question here is do we really need "--network=host". What kind of errors are you getting without this, maybe we can just open some ports on container that Github spins with "container" options for mpi to work? Can you please explore this?
Making this work with standard container can significantly simplify the workflow, so please explore this more and tell us the errors you are seeing.

For the reset of change, if ther there is no way around this and we must use seccond container I would like to see this in separate "call-test-multihost" workflow as it complicates standard call-test.

Other feedback is for the workflow structure

  1. call-test.yml needs to be rebased on main — stale vLLM steps will re-introduce removed code, the vLLM wheel is now handled by the Download and install wheels step via extra-wheel.

  2. runs-on labels needs fixing (as Nebojsa suggested) we should have "in-service" label always
  PR removes:
  runs-on: ${{ !matrix.build.shared-runners && fromJson(format('["{0}", "in-service"]', matrix.build.runs-on)) || matrix.build.runs-on }}
  and replaces with:
  runs-on: ${{ matrix.build.runs-on }}

  3. Docker image expression duplicated 4× in run steps
  ${{ !matrix.build.shared-runners && inputs.docker_image || format('harbor.ci.tenstorrent.net/{0}', inputs.docker_image) }}. 
  Compute it once in the existing "Set reusable strings" step, replace all occurrences with ${{ steps.strings.outputs.docker-image }}.

  4. The base64-over-SSH pattern in "Start multi host worker containers" looks unnecessary
  Workers mount HOST_WORKSPACE_PATH:IN_CONTAINER_WORKSPACE — the script is already on disk on every worker host. Call it directly:
  mpirun ... bash -lc "bash '$HOST_WORKSPACE_PATH/.github/scripts/start_multihost_container.sh' \
    '$DOCKER_IMAGE' '$HOST_WORKSPACE_PATH' '$IN_CONTAINER_WORKSPACE' ubuntu-host-mapped"
  (Claude suggested this one, please double check)

  5. Extract multi host setup/teardown to composite actions
  The 4 setup steps (install Docker CLI, start controller, start workers, SSH check) and 3 teardown steps are ~120 lines that could live in 
  .github/actions/multihost-setup/action.yml and .github/actions/multihost-teardown/action.yml. 
  This leaves call-test.yml with two clean "uses" blocks.
  
  6. Match topology keys should match runner labels. Runner labels we will use for wormhole systems are:
    multi-host-galaxy (2x galaxy)
    multi-host-glx-4x (4x galaxy)
    multi-host-t3000 (2x WH Loudbox)
  Currently basic-multihost-test.json has "runs-on": "multi-host-t3000" paired with "topology": "dual_t3k" — two names for the same hardware
  Can we avoid specifying host names and just use labels? Host "master" should have the "worker" names already configured in rank file, can you check this please.

jameszianxuTT:
Hi Vladimir, thanks for your comments. I think the suggestions are good, and my response to some of them are below.

> Biggest question here is do we really need "--network=host". What kind of errors are you getting without this, maybe we can just open some ports on container that Github spins with "container" options for mpi to work? Can you please explore this?

1. I will do some more pathfinding around --network=host today and check in with metal infra to see if have since found a way around it. 

In my previous experience, MPI would hang somewhere in initialization without forwarding host network into container, and combined with precedent that metal infra was also using --network=host, I did not think it was worth tracing through mechanics of MPI implementation and docker networking implementation, but I can do so. 

> For the reset of change, if ther there is no way around this and we must use seccond container I would like to see this in separate "call-test-multihost" workflow as it complicates standard call-test.

Okay. In my [previous comment](https://github.com/tenstorrent/tt-xla/pull/4286#discussion_r3117868414) to @nsumrakTT, I explained why I thought it was better to integrate into call-test. If you think that the additional complexity of second container is really untenable, then we can split it off and pay the maintenance burden down the road. 

> 6. Match topology keys should match runner labels. Runner labels we will use for wormhole systems are: [...]

Maybe we can avoid this another way, it may no longer be necessary to specify `topology` at all due to underlying tt-run changes, but that may be better put in a different PR.

I also do not think the CI labels are well named - we are using the "topology" strings to index into a dictionary of multihost config files in metal based on the cluster topology for ease of use. I can see the argument for replacing existing key names with multihost labels (to reduce extra indirection step), but it will force user to understand that our CI label "multi-host-galaxy" means 2x, and that adds a layer of confusion.

.github/scripts/start_multihost_container.sh
```diff
@@ -0,0 +1,50 @@
...
+
+if [[ "$(docker ps --filter "name=^/${CONTAINER_NAME}$" --format '{{.Names}}')" == "${CONTAINER_NAME}" ]]; then
+  echo "$(hostname): Container ${CONTAINER_NAME} already running, skipping"
+  exit 0
+fi
+
+docker_args=(
+  run --rm -d
+  --name "${CONTAINER_NAME}"
+  --pid=host --network=host
```
jameszianxuTT, line 29: Hey @nsumrakTT @vvukomanTT 

Per @vmilosevic's comment
> Biggest question here is do we really need "--network=host". What kind of errors are you getting without this, maybe we can just open some ports on container that Github spins with "container" options for mpi to work? Can you please explore this?

Metal infra has done [some research here](https://tenstorrent.slack.com/archives/C094BP47GK1/p1778163453435829?thread_ts=1753241964.673999&cid=C094BP47GK1) and given up to workaround this --network=host issue. I think we can treat it as a constraint for now without spending more time looking for more w/a they have missed. 

Repeating something I mentioned offline to @vmilosevic:

If we cannot get around requirement of --network=host, which is looking like a major pathfinding effort, are we sure we want to go with a duplicate yaml for call-test-multihost path?

What I am worried about is that in the future, we will see that the deviations between the near-duplicate call-test and call-test-multihost could cause too much maintenance burden and lead us to try to integrate them together, after more complexity is added. In my opinion, it is not that much complexity to add to call-test, but maybe that's because I have multihost context 😅 

Also, we will want to do multihost benchmarking too at some point, so if we go down this path there will be more fragmentation. Two dimensions of specialization is acceptable I guess - `singlehost/multihost x benchmark/non-benchmark`, but we should consider the proliferation of yaml files that may be required in future if we end up with more dimensions to create new yamls for.

nsumrakTT:
@jameszianxuTT I think we should go with separate workflow while multihost is still in development, as it is easier to develop without constrains of single test workflow and also keep test workflow protected from multihost experiments. When multihost becomes stable, we can reconsider and join these into the single wf.
Anyway, whatever the decision is today, it is not carved in stone.

nsumrakTT:
https://github.com/tenstorrent/tt-xla/pull/4971 is merged which supersedes this PR

jameszianxuTT:
Resolved by #4971  - more fixes are made at provisioning level to avoid needing to expose docker sock and --network host bypass path by modifying GHA container launcher.

Issue #4171, opened 2026-04-08
Remove use "full_attention" workaround in several models after "sliding window attention" support added


kmabeeTT:
After https://github.com/tenstorrent/tt-xla/pull/3688 is merged remove this workaround from few models in tt-forge-models ([olmo3](https://github.com/tenstorrent/tt-forge-models/pull/484/files), [gemma](https://github.com/tenstorrent/tt-forge-models/pull/563), mistral):

```
        if getattr(model.config, "use_cache", True):
            model.config.text_config.layer_types = [
                "full_attention"
            ] * model.config.text_config.num_hidden_layers
```

PR #4287, opened 2026-04-17
Testing models using sliding attention

devisettymahidhar608:
### Ticket

### Problem description
Since PR  https://github.com/tenstorrent/tt-xla/pull/3688 was merged into main, I’ve been testing models with sliding attention, replacing the previously hardcoded full attention.

Model variants yet to be tested
allenai/Olmo-3-1025-7B
allenai/Olmo-3-1125-32B
mistralai/Ministral-8B-Instruct-2410

Taken this branch as reference: https://github.com/tenstorrent/tt-xla/compare/main...jazpur/gemma-4

### What's changed
Added a test file in the examples folder; tests passed successfully

logs:
[mistral.log](https://github.com/user-attachments/files/27126975/mistral.log)
[olmo3_7b.log](https://github.com/user-attachments/files/27126979/olmo3_7b.log)
[olmo3_32b.log](https://github.com/user-attachments/files/27126989/olmo3_32b.log)

### Checklist

### DIFF
--- a/examples/pytorch/mistral_8b.py
+++ b/examples/pytorch/mistral_8b.py
@@ -1,0 +1,184 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import argparse
+import os
+from typing import List
+
+import torch
+import torch_xla
+import torch_xla.runtime as xr
+from transformers import AutoModelForCausalLM, AutoTokenizer, PreTrainedTokenizer
+from transformers.cache_utils import StaticCache
+from transformers.configuration_utils import PretrainedConfig
+from tt_torch.transformers_overrides import (
+    override_cache_sliding_window_layers,
+    override_ministral_sliding_window_causal_mask,
+)
+
+MODEL_NAME = "mistralai/Ministral-8B-Instruct-2410"
+DEFAULT_PROMPT = "Who would win in a fight - a dinosaur or a cow named Moo Moo?"
+MAX_LENGTH = 256
+BATCH_SIZE = 1
+
+
+# --------------------------------
+# Mistral 8B Generation Loop Example
+# --------------------------------
+def mistral_8b():
+    # Set up config variables.
+    model_name = "mistralai/Ministral-8B-Instruct-2410"
+
+    # Connect the device
+    device: torch.device = torch_xla.device()
+    model, tokenizer = _load_model_and_tokenizer(model_name)
+
+    input_args, formmatted_prompts = _prepare_inputs(
+        [DEFAULT_PROMPT], tokenizer, model.config, BATCH_SIZE, MAX_LENGTH
+    )
+    max_tokens_to_generate = 20
+
+    for layer in input_args["past_key_values"].layers:
+        layer.keys = layer.keys.to(device)
+        layer.values = layer.values.to(device)
+        if isinstance(getattr(layer, "cumulative_length", None), torch.Tensor):
+            layer.cumulative_length = layer.cumulative_length.to(device)
+        if hasattr(layer, "device"):
+            layer.device = device
+
+    input_args["input_ids"] = input_args["input_ids"].to(device)
+    input_args["cache_position"] = input_args["cache_position"].to(device)
+    input_args["attention_mask"] = input_args["attention_mask"].to(device)
+    model = model.to(device)
+
+    torch_xla.set_custom_compile_options({"experimental_weight_dtype": "bfp_bf8"})
+    compiled_model = torch.compile(model, backend="tt")
+    output_tokens: List[List[str]] = [[] for _ in range(1)]
+    with torch.no_grad():
+        for step in range(max_tokens_to_generate):
+            output = compiled_model(**input_args)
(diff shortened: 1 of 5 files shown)


kmabeeTT:
@jazpurTT what do you think about this and the associated tt-forge-models PR https://github.com/tenstorrent/tt-forge-models/pull/604? Seems consistent with previous changes you made? In meeting earleir, Aleks was asking if these changes should go into tt-forge-models instead... I suggested maybe that can be a follow up task, if it makes sense.

python_package/tt_torch/transformers_overrides.py
```diff
@@ -43,8 +69,12 @@ def override_cache_sliding_window_layers(
         tt_layer.dtype = layer.dtype
         tt_layer.max_batch_size = layer.max_batch_size
         tt_layer.num_heads = layer.num_heads
-        tt_layer.v_head_dim = layer.v_head_dim
-        tt_layer.k_head_dim = layer.k_head_dim
+        tt_layer.k_head_dim = getattr(
+            layer, "k_head_dim", getattr(layer, "head_dim", None)
```
jazpurTT: is this for backward compatibility of transformers or do the mistral/olmo models use `head_dim` in their model/cache config?

examples/pytorch/mistral_8b.py
```diff
@@ -0,0 +1,174 @@
...
+        padding_side="left",
+        return_attention_mask=True,
+    )
+    seq_len = inputs.input_ids.shape[1]
+
+    static_cache = StaticCache(
+        config=model_config,
+        max_cache_len=max_cache_len,
+    )
+    _init_static_cache(static_cache, model_config, BATCH_SIZE)
```
jazpurTT: does mistral have mismatching head_dims? do we want to make `_init_static_cache` a single entry point for static cache init? regardless of mismatching head_dims?

jazpurTT:
> @jazpurTT what do you think about this and the associated tt-forge-models PR [tenstorrent/tt-forge-models#604](https://github.com/tenstorrent/tt-forge-models/pull/604)? Seems consistent with previous changes you made? In meeting earleir, Aleks was asking if these changes should go into tt-forge-models instead... I suggested maybe that can be a follow up task, if it makes sense.

Left some questions/feedback, but it looks good to me. To clarify, is the idea would be to move the `transformers_override.py` functions to tt_forge_models? we could yeah, it would just make the dependency for tt-forge-models stronger cause now if a customer wants support for sliding window they'd need tt-forge-models.

examples/pytorch/mistral_8b.py
devisettymahidhar608: Hi @jazpurTT , I’ve addressed this—could you please review it now,

jazpurTT (approved):
Thanks for addressing the comments, LGTM

kmabeeTT (approved):

python_package/tt_torch/transformers_overrides.py
```diff
@@ -21,6 +21,32 @@ def override_gpt_oss_sliding_window_causal_mask():
     gpt_oss_mod.create_sliding_window_causal_mask = tt_create_sliding_window_causal_mask
 
 
+def override_olmo3_sliding_window_causal_mask():
```
AleksKnezevic, line 24: @kmabeeTT and @jazpurTT. I don't think specific model overrides belong here, I think this should be in the examples, and the loader should have some override_sliding_attention flag, which if true would get overriden by the test infra. What are your thoughts?

kmabeeTT: I agree that would be cleaner, it's just worth noting that this is different from existing precedent so we'd be doing the cleanup now rather than later.

Just to double check... So with your idea, this existing function `override_gpt_oss_sliding_window_causal_mask()` would go away and example `examples/pytorch/gpt_oss_20b.py` would just inline the 2 lines (import + override) directly in their source? 

```
examples/pytorch/gpt_oss_20b.py:    override_gpt_oss_sliding_window_causal_mask,
examples/pytorch/gpt_oss_20b.py:    override_gpt_oss_sliding_window_causal_mask()
python_package/tt_torch/transformers_overrides.py:def override_gpt_oss_sliding_window_causal_mask():
```

...and new optional flag in tt_forge_models per model like `override_sliding_attention`would trigger a generic (not per-model) code in test infra that walked the model and found modules to replace, something like this?

```
  import sys

  _PATCHABLE_FNS = {
      "create_sliding_window_causal_mask": tt_create_sliding_window_causal_mask,
      # add future patchable functions here
  }

  def apply_tt_compat_overrides(model):
      """Walk the model, patch any transformers modeling module it touches."""
      seen = set()
      for sub in model.modules():
          mod_name = sub.__class__.__module__
          if mod_name in seen or not mod_name.startswith("transformers.models."):
              continue
          seen.add(mod_name)
          mod = sys.modules.get(mod_name)
          if mod is None:
              continue
          for fn_name, replacement in _PATCHABLE_FNS.items():
              if hasattr(mod, fn_name):
                  setattr(mod, fn_name, replacement)
```

AleksKnezevic (approved):
As discussed, approving for now, refactor to come in subsequent PR.

PR #4288, opened 2026-04-17
Specify artifacts to use when running manual tets

vmilosevic:
### Ticket
slack

### Problem description
Specify artifacts to use when running manual tests

### What's changed
Add input to manual test workflows that specifies commit from which to take artifacts

### Checklist

### DIFF
--- a/.github/workflows/call-build.yml
+++ b/.github/workflows/call-build.yml
@@ -3,6 +3,9 @@ name: Check and Build tt-xla
 # Implements artifact caching to avoid unnecessary rebuilds.
 # Artifacts are reused only when: not a PR, no mlir_override, and exact name match exists.
 # Naming: "xla-whl-{release|explorer}[-commit_sha]" (SHA added for individual commits only).
+#
+# Artifact reuse: when artifact_sha is provided, artifacts built for that commit are reused
+# without rebuilding. Falls back to a normal build if any artifact is missing or expired.
 
 on:
   workflow_call:
@@ -33,6 +36,10 @@ on:
         required: false
         type: string
         default: ${{ github.sha }}
+      artifact_sha:
+        description: 'tt-xla commit SHA to reuse pre-built artifacts from. If artifacts are not found the workflow builds normally.'
+        required: false
+        type: string
     outputs:
       artifacts_run_id:
         description: 'Run ID of the workflow that produced the existing artifact(s)'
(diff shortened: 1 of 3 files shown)


.github/workflows/call-inspect-changes.yml
```diff
@@ -170,6 +175,30 @@ jobs:
...
+            # Check if any build-affecting files changed.
+            # Files outside these patterns (tests, docs, .github, examples) do not require a rebuild.
+            build_patterns=(
+              "CMakeLists.txt"
+              "**/CMakeLists.txt"
+              "CMakePresets.json"
+              "pjrt_implementation/*"
+              "python_package/*"
+              "third_party/*"
+              "integrations/*"
```
nsumrakTT: I think the glob search should be `**/* `to include all of the subdirectory tree eg.
`pjrt_implementation/**/*`

.github/workflows/call-inspect-changes.yml
```diff
@@ -170,6 +175,30 @@ jobs:
...
+              "scripts/build_torch_xla.sh"
+            )
```
nsumrakTT: Should we add .github/workflows/pr-main.yml, call-build.yml and similar wf files?

vmilosevic: I'm not sure about this list, we should adjust this to which changes actually require build

.github/workflows/pr-main.yml
```diff
@@ -43,9 +43,10 @@ jobs:
       docker_image: ${{ needs.build-image.outputs.docker-image }}
       docker_image_manylinux: ${{ needs.build-image.outputs.docker-image-manylinux }}
       build_types: ${{ needs.inspect-changes.outputs.wheel-changes == 'true' && 'release,explorer,manylinux' || 'release,explorer' }}
+      needs_build: ${{ needs.inspect-changes.outputs.needs-build }}
```
nsumrakTT: maybe we should add a way to force build, eg. through pr label or something like that?
Also, if docker is changed (rebuilt) should we force rebuilding?

.github/workflows/call-inspect-changes.yml
vvukomanTT: +1 Let's add some files from `/.github` in this list as well.

.github/workflows/pr-main.yml
vvukomanTT: +1 If docker is getting rebuilt we should rebuild the wheel as well.

nsumrakTT (approved):

vvukomanTT (approved):

Issue #4206, opened 2026-04-10
Bad PCC on GPT-OSS 120B TP8-DP4-EP32 Galaxy benchmark


odjuricicTT:
## Description
The new GPT-OSS 120B TP8-DP4-EP32 benchmark on Galaxy (4x8 mesh, batch size 128) produces bad PCC accuracy results that need investigation.

## Context
- Introduced in #4205
- Test: `test_gpt_oss_120b_tp_dp_galaxy_batch_size_128`
- Configuration: TP=8, DP=4, EP=32
- Perf benchmark run: https://github.com/tenstorrent/tt-xla/actions/runs/24247948813

odjuricicTT:
## PCC Investigation Plan

Starting investigation of bad PCC on prefill logits for the GPT-OSS 120B TP8-DP4-EP32 Galaxy benchmark (`--num-layers 1`).

### Analysis of TP-DP shard spec vs TP-only shard spec

Comparing `_moe_throughput_galaxy_shard_spec_fn` (TP-DP, used by this test) vs `ModelLoader.load_shard_spec()` (TP-only, used by `test_gpt_oss_120b_tp_galaxy_batch_size_64`):

| Parameter | TP-only (loader) | TP-DP (custom fn) |
|-----------|-------------------|-------------------|
| `batch_axis` | `"batch"` | `None` (replicated) |
| `q/k/v_proj.bias` | `("model",)` sharded | **Missing entirely** |
| `o_proj.bias` | `("batch",)` sharded | **Missing entirely** |
| `lm_head.weight` | `("model", "batch")` | `(None, None)` replicated |
| `sinks` | `(None,)` replicated | `("model",)` sharded |
| Expert sharding | `("model", "batch", None)` | `(("model", "batch"), None, None)` |

### Key hypotheses
1. **Missing attention bias shardings** in TP-DP shard spec — compiler propagation may assign incorrect shardings
2. **Stale KV cache** — accuracy benchmark reuses cache from warmup+perf runs
3. **MoE expert routing sensitivity** — numerical differences flipping top-4 expert selection

### Phase 0: Characterizing the failure
Added diagnostics to capture: actual PCC value, NaN/Inf checks, shape/dtype/range, argmax token agreement, and per-position PCC breakdown. Test running now.

odjuricicTT:
## Phase 0: Failure Characterization

Test: `test_gpt_oss_120b_tp_dp_galaxy_batch_size_128 --num-layers 1`

[...]

PR #4316, opened 2026-04-20
Change monkey patch for GptOssTopKRouter

mvasiljevicTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/4206)

### Problem description
GPT OSS 120b currently gives garbage outputs for all users except 1st. One of the reasons is wrong lowering of the scatter op, which is leading to all indices being in the first 1/4 of the range. More about this in the [discussion](https://github.com/tenstorrent/tt-mlir/discussions/7334#discussioncomment-16558816). 

### What's changed
Use `values*one_hot(indices)` instead of `zeros.scatter_(indices,values)` to bypass scatter op. This is temporary fix until we start using MoE ops directly.

With additional mlir change, we get good outputs for all 128 users.

### Checklist
- [x] [perf benchmark on galaxy](https://github.com/tenstorrent/tt-xla/actions/runs/24666905318/job/72129144995) - with [mlir fix](https://github.com/tenstorrent/tt-mlir/pull/8028)


mvasiljevicTT:
cc @odjuricicTT

mstojkovicTT (approved):

PR #4317, opened 2026-04-20
Fix Deformable DETR DRAM OOM by wrapping grid_sample as StableHLO composite op

### LINKED ISSUE #3805: Deformable_detr : Out of Memory issue
### Describe the bug

- Deformable_detr fails due to Out of memory issue caused by [grid_sample](https://github.com/huggingface/transformers/blob/8cb5963cc22174954e7dca2c0a3320b7dc2f4edc/src/transformers/models/deformable_detr/modeling_deformable_detr.py#L88) operation
- Full model run - [deformable_detr.log](https://github.com/user-attachments/files/26087115/deformable_detr.log)

```
Out of Memory: Not enough space to allocate 18686672896 B DRAM buffer across 12 banks, where each bank needs to store 1557225472 B, but bank size is 1071821792 B
```

- Sanity run - [grid_sample.log](https://github.com/user-attachments/files/26087166/grid_sample.log)

```
Out of Memory: Not enough space to allocate 18686672896 B DRAM buffer across 12 banks, where each bank needs to store 1557225472 B, but bank size is 1071821792 B
```

```
git checkout 
git submodule update --init --recursive

# Model run
pytest tests/runner/test_models.py::test_all_models_torch[deformable_detr/pytorch-Deformable_Detr-single_device-inference] -svv

# Sanity run
pytest tests/torch/ops/test_grid_sample.py -svv
```

### Expected behavior

_No response_

_No response_

saiarthiraguram:
## Describe the bug

- Deformable DETR model crashes with DRAM OOM during `torch_xla.sync()` in the deformable attention encoder layer:
  `Out of Memory: Not enough space to allocate 18686672896 B DRAM buffer across 12 banks, where each bank needs to store 1557225472 B, but bank size is only 1073741792 B`
- The OOM was previously traced to `ttnn.reshape` trying to allocate the ~18.6 GB buffer (#3805, tt-metal [#40266](https://github.com/tenstorrent/tt-metal/issues/40266)).
- The same OOM occurs in the standalone grid_sample op test.

## Root Cause Analysis

- `aten.grid_sampler_2d` was in the default decomposition list in `decompositions.py`.
- When PyTorch decomposes `F.grid_sample`, it breaks it into 114 primitive ops including coordinate computation (`floor`, `mul`, `sub` for bilinear weights), bounds checking (`logical_and`, `where`), and `aten.index` (fancy indexing) to gather values at computed coordinates. verified by dumping the decomposed graph - [grid_sample_decomposed_graph.log](https://github.com/userattachments/files/26810533/grid_sample_decomposed_graph.log).
- For Deformable DETR shapes (`batch=8, channels=32, spatial=17821, points=4`), these intermediate tensors flow into `ttnn.reshape` which tries to allocate ~18.6 GB — far exceeding the ~1 GB per-bank DRAM limit.
- Meanwhile, `ttnn.grid_sample` already exists as a native kernel in tt-metal that handles the same operation on-chip without materializing these intermediate tensors.
- The pipeline never reaches `ttnn.grid_sample` because PyTorch decomposes the op before tt-mlir can see it.

## Fix

1. **Added `composite_grid_sample()` in `composite_ops.py`** — wraps `F.grid_sample` as a `"tenstorrent.grid_sample"` StableHLO composite using `StableHLOCompositeBuilder` (same pattern as `composite_gelu`, `composite_rms_norm`, `composite_layer_norm`). This preserves the high-level grid_sample semantic so tt-mlir can lower it directly to `ttnn.grid_sample`.

2. **Removed `aten.grid_sampler_2d` from the decomposition list in `decompositions.py`** — prevents PyTorch from decomposing grid_sample into the massive primitive ops. Both changes are needed: removing from the decomposition list stops the decomposition, and the composite wrapper gives tt-mlir a named op to recognize and lower.

3. **Added comprehensive grid_sample op tests in `test_grid_sample.py`** — 20 test cases covering all mode/padding_mode/align_corners parameter combinations, real model shapes (Deformable DETR, DETR3D, BEVFormer), with strict pcc + allclose + equal checks enabled. All 17 runnable tests pass with `equal=True, pcc=1.0, max_diff=0.0`.

## Steps to reproduce
```bash
# grid_sample op test (OOM before fix)
pytest tests/torch/ops/test_grid_sample.py -svv

# Deformable DETR full model test (OOM before fix)
pytest tests/runner/test_models.py::test_all_models_torch[deformable_detr/pytorch-Deformable_Detr-single_device-inference] -svv
```
## Test plan
- [x] pytest tests/torch/ops/test_grid_sample.py — 17 passed, 3 skipped (all pcc=1.0, equal=True)
- [x] Standalone bfloat16 comparison (CPU vs TT): equal=True, pcc=1.0, max_diff=0.0 across deformable_detr, detr3d, bevformer shapes.
- [x] Deformable DETR model test runs to completion (previously DRAM OOM crash)
- [x] No regression: no currently passing model uses grid_sample
- [x] Resolves #3805; may also reduce similar OOM in other grid_sample models (OFT, DETR3D)
Note: The Deformable DETR full model test currently reports pcc=nan due to bfloat16 numerical overflow in the CPU golden reference (not related to grid_sample) — to be addressed in a separate follow-up PR.

[test_grid_sample.py](https://github.com/user-attachments/files/26896743/test_grid_sample.py)
[grid_sample_decomposed_graph.log](https://github.com/user-attachments/files/26896754/grid_sample_decomposed_graph.log)
[grid_sample_strict.log](https://github.com/user-attachments/files/26896755/grid_sample_strict.log)


mstojkovicTT:
Composite looks good, can you please add tests for this composite inside `tests/torch/ops/test_composite_ops.py`. Also, how do you match this composite inside [ttir](https://github.com/tenstorrent/tt-mlir/blob/main/lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp), because i dont see anything that could match `grid_sample`.

saiarthiraguram:
> Composite looks good, can you please add tests for this composite inside `tests/torch/ops/test_composite_ops.py`. Also, how do you match this composite inside [ttir](https://github.com/tenstorrent/tt-mlir/blob/main/lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp), because i dont see anything that could match `grid_sample`.

grid_sample is not present in ttir, but we have in ttnn(ttnn.grid_sample), what is happening in our implementation is 
aten.grid_sampler_2d is NOT in the decomposition list(removed)
The composite wrapper preserves it as a single high-level op
torch-xla has no lowering for aten.grid_sampler_2d itself
→ CPU fallback kicks in → grid_sample runs on CPU → no OOM
Now upon further investigation I found that  need to write a pattern to match stablehlo.composite @tenstorrent.grid_sample op and lower it to ttnn.grid_sample.

I thought its is working fine but it is due to cpu fallback, need to look further into this.

saiarthiraguram:
Adding PR in tt-mlir for resolution [PR Link](https://github.com/tenstorrent/tt-mlir/pull/8261)

PR #4326, opened 2026-04-20
Use distinct per-user prompts for GPT-OSS MoE benchmarks

odjuricicTT:
### Ticket
N/A

### Problem description
LLM benchmarks replicate a single prompt across every user in the batch. For MoE models this collapses expert routing to one pattern for all users and understates realistic workloads, so reported throughput is not representative.

### What's changed
- New optional `per_user_prompts: list[str]` on `test_llm` / `benchmark_llm_torch_xla`. When provided, the benchmark tokenizes each prompt, asserts every prompt has the same token count, and stacks them into a `[batch_size, seq_len]` tensor used as the per-user input.
- Hand-tuned list of 129 distinct questions (`_GPT_OSS_PER_USER_QUESTIONS` in `test_llms.py`) that all tokenize to exactly 40 tokens under the `openai/gpt-oss-20b` tokenizer — verified locally. No padding or truncation needed at benchmark time; the assert enforces the invariant.
- Wired into the 6 gpt-oss tests (`test_gpt_oss_20b_tp`, `_batch_size_1`, `_galaxy_batch_size_64`, `test_gpt_oss_120b_tp_galaxy_batch_size_64`, `_dp_galaxy_batch_size_128`, `_tp_qb2`). Skipped when `accuracy_testing=True` since accuracy testing needs its reference prompt.
- `construct_inputs` accepts a 2D `input_prompt_tokens` `[B, S]` passthrough (previous 1D replication path unchanged).

### Checklist

### Perf runs
- Performance Benchmark (`test_filter=gpt_oss`, `runs-on-filter=All`, `sh-runner=false`): https://github.com/tenstorrent/tt-xla/actions/runs/24677562336

### DIFF
--- a/tests/benchmark/benchmarks/llm_benchmark.py
+++ b/tests/benchmark/benchmarks/llm_benchmark.py
@@ -98,14 +98,15 @@ def construct_inputs(
         Dictionary containing input_ids, past_key_values, cache_position, and use_cache
     """
     if input_prompt_tokens is not None:
-        if input_prompt_tokens.ndim != 1:
-            raise ValueError(
-                f"input_prompt_tokens must be 1D token IDs, got shape {tuple(input_prompt_tokens.shape)}"
+        if input_prompt_tokens.ndim == 2:
+            input_ids = input_prompt_tokens[:, :max_cache_len].contiguous()
+        else:
+            input_ids = (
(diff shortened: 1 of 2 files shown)


odjuricicTT:
This PR should probably wait for PCC check for all users to land first

PR #4332, opened 2026-04-21
[vLLM] Fix cpu sampling

### LINKED ISSUE #4331: [vLLM plugin][Bug] CPU samling failure
vLLM plugin create sampling metadata on `cpu` when `cpu_sampling` is enabled. However, model runner is always using same `sample_from_logits` function. This cause failure as the sampling data is created on cpu whereas logits are still on device. 
`sample_from_logits_cpu` moves data to cpu but keep the output on cpu which may cause failure later.

mmanzoorTT:
### Ticket
closes #4331 

### Problem description
model runner always process sampling on device even if cpu_sampling is enabled. 

### What's changed
Update the function call to use appropriate functionality.

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -1380,7 +1380,9 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                     require_struct_decoding, grammar_bitmask_padded, logits, arange
                 )
 
-            selected_token_ids = self.sample_from_logits(logits, tpu_sampling_metadata)
+            selected_token_ids = self.sample_from_logits_func(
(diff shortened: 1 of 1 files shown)


integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1380,7 +1380,9 @@ def sample_tokens(
                     require_struct_decoding, grammar_bitmask_padded, logits, arange
                 )
 
-            selected_token_ids = self.sample_from_logits(logits, tpu_sampling_metadata)
+            selected_token_ids = self.sample_from_logits_func(
+                logits, tpu_sampling_metadata
```
AleksKnezevic: Doesn't need to be in this PR, but we should rename tpu_sampling_metadata

AleksKnezevic (approved):

PR #4346, opened 2026-04-21
Enable prefill testing on wh-galaxy and qb2-blackhole architectures

umalesTT:
### Ticket
None

### Problem description
Until now, extensive prefill testing was supported only on n150/p150/llmbox architectures.
We need to extend this support to other supported architectures on CI.

### What's changed
Added support for prefill testing on wh-galaxy and qb2-blackhole.
Updated some tests that no longer OOM when run on wh-galaxy.

### Checklist


vvukomanTT (approved):

nsumrakTT (approved):
My worry is that we currently have 1 (one) galaxy.
Are these tests scheduled or these are ment to be run manually?
What is the length of the runs?

tests/runner/test_models.py
```diff
@@ -424,6 +431,14 @@ def _validate_llm_sharding_mesh_combination(sharding_config, mesh_shape):
...
 
+    if mesh_shape not in _EXPLICIT_STRATEGY_MESH_SHAPES:
+        return (
+            f"Unsupported mesh_shape={mesh_shape} for explicit sharding strategy. "
+            f"Supported: {sorted(_EXPLICIT_STRATEGY_MESH_SHAPES)}"
+        )
+
+    # When the non-model mesh dim is 1, FSDP collapses to Megatron, so we keep
+    # only megatron-no_dp to avoid redundant tests.
     if mesh_shape == (1, 8):
```
acicovicTT: How about (1, 4), this is new when looking at test parameters, should it be handled like this?

acicovicTT (approved):

umalesTT:
> My worry is that we currently have 1 (one) galaxy. Are these tests scheduled or these are ment to be run manually? What is the length of the runs?

These tests are scheduled for nightly. For wh-galaxy, only 3 tests are running due to DRAM OOM issues and each of them lasted several minutes when I tried.

tests/runner/test_models.py
umalesTT: Yes, thanks for the catch!

Issue #4296, opened 2026-04-18
decode only vLLM tests


AleksKnezevic:
Enable decode only tests for vLLM.

PR #4354, opened 2026-04-22
Add vLLM decode tests with prefill length fixed to 1

jackzhangTT:
Use prefill length 1 so the run does not compile across varying prefill sizes; the coverage target is decode, not prefill compilation.

### Ticket
https://github.com/tenstorrent/tt-xla/issues/4296

### Problem description
Add a QOL test that focuses on decode by specifying a vLLM test with prefill length fixed to 1. Users of this test is expected to change the model themselves but keep min_context_len and max_model_len fixed to 1.

### What's changed
add a new test with prefill length fixed to 1. 

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -290,6 +290,12 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
             min_token_size=self.tt_config.min_context_len,
             max_token_size=self.max_model_len,
         )
+        if self.tt_config.decode_only:
+            # Restrict all num_tokens bucketing to the decode shape so every
+            # precompile pass (and anything else that reads
+            # self.num_tokens_paddings) only compiles num_tokens=1.
+            self.num_tokens_paddings = [1]
+            logger.info("decode_only=True; restricting num_tokens_paddings to [1].")
 
         # In case `max_num_tokens < max(num_tokens_paddings)` use the actual
         # padded max value to pre-allocate data structures and pre-compile.
@@ -1899,9 +1905,12 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         Precompile all the subgraphs with possible input shapes.
         """
         torch._dynamo.config.dynamic_shapes = False
+        decode_only = self.tt_config.decode_only
         with self.maybe_setup_dummy_loras(self.lora_config):
-            self._precompile_mm_encoder()
             self._precompile_backbone()
+            if self.tt_config.decode_only:
+                return
+            self._precompile_mm_encoder()
             self._precompile_select_hidden_states()
             self._precompile_compute_logits()
             self._precompile_structured_decoding()
@@ -1917,6 +1926,8 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         self,
         num_tokens: int,
     ) -> None:
+        if self.tt_config.decode_only:
+            return
         logger.info(f"Profiling run with num_tokens={num_tokens}.")
         torch._dynamo.config.dynamic_shapes = False
         # Profile with multimodal encoder & encoder cache.
--- a/tests/integrations/vllm_plugin/generative/test_decode.py
+++ b/tests/integrations/vllm_plugin/generative/test_decode.py
@@ -1,0 +1,24 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import pytest
+import vllm
+
+
+@pytest.mark.nightly
+@pytest.mark.single_device
+def test_decode():
+    llm_args = {
+        "model": "meta-llama/Llama-3.2-3B",
+        "max_num_batched_tokens": 128,
+        "max_num_seqs": 1,
+        "max_model_len": 16,
+        "gpu_memory_utilization": 0.002,
+        "additional_config": {
+            "enable_const_eval": False,
+            "min_context_len": 1,
+            "num_hidden_layers": 1,
+            "decode_only": True,
+        },
+    }
+    llm = vllm.LLM(**llm_args)
--- a/integrations/vllm_plugin/vllm_tt/platform.py
+++ b/integrations/vllm_plugin/vllm_tt/platform.py
@@ -72,6 +72,17 @@ class TTConfig:
     # Perform token sampling on CPU instead of compiling a sampling graph for device
     cpu_sampling: bool = False
 
+    # When True, `capture_model` precompiles only the graphs needed for decode
+    # (num_tokens == 1):
+    #   - `_precompile_backbone` is restricted to the decode shape.
+    #   - `_precompile_select_hidden_states` is restricted to the decode shape.
+    #   - `_precompile_mm_encoder` and `_precompile_structured_decoding` are
+    #     skipped entirely (multimodal and structured-output decoding are not
+    #     exercised in a plain decode-only run).
+    # Useful for speeding up startup when only decode performance matters
+    # (e.g. local debugging of decode-only tests).
+    decode_only: bool = False
+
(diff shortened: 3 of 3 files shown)


tests/integrations/vllm_plugin/generative/decode/test_decode.py
```diff
@@ -0,0 +1,29 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
mmanzoorTT:
```suggestion
# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```

tests/integrations/vllm_plugin/generative/decode/test_decode.py
```diff
@@ -0,0 +1,29 @@
...
+        "gpu_memory_utilization": 0.002,
+        "additional_config": {
+            "enable_const_eval": False,
+            "min_context_len": 1,
+        },
+    }
+    llm = vllm.LLM(**llm_args)
+
+    output_text = llm.generate(prompts, sampling_params)[0].outputs[0].text
+    print(f"prompt: {prompts[0]}, output: {output_text}")
```
mmanzoorTT: I think, we don't these. `llm = vllm.LLM(**llm_args)` will precompile the model with input-seq-len=1 and attention layer differentiate between prefill and decode based on input-seq-len (1 token = decode; > 1 tokens = prefill). vLLM may also crash due to the conflicting configs here. `max_model_len=1` will not allow any text generation.

mmanzoorTT (commented):
I thought about it and this approach will not work. We are always creating at least two token lengths (token-length=1; for decode and token-length=32; for prefill); have a look at `model_runner.py::_get_token_paddings`. 
I'll suggest to consider adding a flag (For reference: https://github.com/tenstorrent/tt-xla/pull/2094). If flag enabled; then we should limit `capture_model` to only run `_precompile_backbone` and limit `_precompile_backbone` to run `dummy_run` for decode only (token-length=1). You will only create graph(s) for decode only.

tests/integrations/vllm_plugin/generative/decode/test_decode.py
```diff
@@ -0,0 +1,29 @@
...
+        "gpu_memory_utilization": 0.002,
+        "additional_config": {
+            "enable_const_eval": False,
+            "min_context_len": 1,
+            "precompile_backbone_decode_only": True,
+        },
+    }
+    llm = vllm.LLM(**llm_args)
+
+    output_text = llm.generate(prompts, sampling_params)[0].outputs[0].text
```
mmanzoorTT: Remove this line. this will trigger compilation for prefill as well.

tests/integrations/vllm_plugin/generative/decode/test_decode.py
```diff
@@ -0,0 +1,29 @@
...
+def test_decode():
+    prompts = [
+        "Enter custom prompt here",
+    ]
+    sampling_params = vllm.SamplingParams(temperature=0.8, top_p=0.95, max_tokens=32)
+    llm_args = {
+        "model": "Specify your model here",
+        "max_num_batched_tokens": 128,
+        "max_num_seqs": 1,
+        "max_model_len": 1,
```
mmanzoorTT: You don't need to use `1` for max_model_len and min_context_len with this flag.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1721,7 +1728,14 @@ def _precompile_mm_encoder(self) -> None:
     def _precompile_backbone(self) -> None:
         logger.info("Compiling the model with different input shapes.")
         start = time.perf_counter()
-        for num_tokens in self.num_tokens_paddings:
+        num_tokens_paddings = self.num_tokens_paddings
```
mmanzoorTT: You can also update `capture_model` if you do not want to compile other components/graphs.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2446,6 +2470,8 @@ def _get_token_paddings(min_token_size: int, max_token_size: int) -> list[int]:
     """
     # Adjust min_token_size to be power of 2 and >=32 (if required)
     num = _adjust_min_token(min_token_size)
+    print("============= MADCOW =================")
```
AleksKnezevic: Please remove debug print

tests/integrations/vllm_plugin/generative/test_decode.py
```diff
@@ -0,0 +1,27 @@
...
+import vllm
+
+
+@pytest.mark.nightly
+@pytest.mark.single_device
+def test_decode():
+    prompts = [
+        "Hello",
+    ]
+    sampling_params = vllm.SamplingParams(temperature=0.8, top_p=0.95, max_tokens=8)
```
AleksKnezevic: Can you please run with env var `TTXLA_LOGGER_LEVEL=DEBUG` and validate that only a single graph is generated.

AleksKnezevic (changes requested):

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1721,7 +1728,14 @@ def _precompile_mm_encoder(self) -> None:
...
         logger.info("Compiling the model with different input shapes.")
         start = time.perf_counter()
-        for num_tokens in self.num_tokens_paddings:
+        num_tokens_paddings = self.num_tokens_paddings
+        if self.tt_config.decode_only:
+            num_tokens_paddings = [1]
+            logger.info(
+                "decode_only=True; restricting backbone precompile to decode "
+                "only (num_tokens=1)."
+            )
```
mmanzoorTT: You can move this logic to `__init__` and overwrite `self.num_tokens_padding = [1]`. You will not have to repeat it everywhere and it will be safe/consistent if we add something else relying on `self.num_tokens_padding` in future.

tests/integrations/vllm_plugin/generative/test_decode.py
mmanzoorTT: @AleksKnezevic This will also generate graphs for `select_hidden_states`, `compute_logits` etc. Do we only need `decode` graph?

tests/integrations/vllm_plugin/generative/test_decode.py
```diff
@@ -0,0 +1,27 @@
...
+import pytest
+import vllm
+
+
+@pytest.mark.nightly
+@pytest.mark.single_device
+def test_decode():
+    prompts = [
+        "Hello",
+    ]
```
mmanzoorTT: you can remove `prompts` and `sampling_params`; both are unused.

tests/integrations/vllm_plugin/generative/test_decode.py
AleksKnezevic: Yeah, we should remove this, and have it do decode only.

AleksKnezevic (approved):

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1893,12 +1899,15 @@ def capture_model(self) -> None:
...
         with self.maybe_setup_dummy_loras(self.lora_config):
-            self._precompile_mm_encoder()
+            if not decode_only:
+                self._precompile_mm_encoder()
             self._precompile_backbone()
             self._precompile_select_hidden_states()
             self._precompile_compute_logits()
-            self._precompile_structured_decoding()
+            if not decode_only:
+                self._precompile_structured_decoding()
```
mmanzoorTT: You can reorder these pre-compilation; always generate model backbone graphs and skip others if `decode_only` is enabled.  
```suggestion
        with self.maybe_setup_dummy_loras(self.lora_config):
            self._precompile_backbone()
            if self.tt_config.decode_only:
                return
            self._precompile_mm_encoder()
            self._precompile_select_hidden_states()
            self._precompile_compute_logits()
            self._precompile_structured_decoding()
```

mmanzoorTT (approved):

PR #4362, opened 2026-04-22
Switch accuracy metrics to quantile-based (p5) + mean over full batch

### LINKED ISSUE #4363: Average top1/top5 metrics over batch, not only on index 0
Currently we have: TOP1/TOP5 accuracy is computed using only user 0's predictions.
We want: Accuracy computed per-user and averaged across the entire batch.

dgolubovicTT:
### Ticket
Closes #4363

### Problem description
**Before:** TOP1/TOP5 accuracy was computed using only user 0's predictions. This is fragile — a single user's result isn't representative of the whole batch.

Plain averaging hides problems too: if 95% of users have 0.95 accuracy and 5% have 0.2, the average is 0.91 — but the worst users are essentially broken.

### What's changed

**Accuracy metrics (`llm_benchmark.py`):**

Accuracy is computed per-user across the entire batch, then reported using both **p5** (5th percentile) and **mean** as tracked metrics:

| Measurement | Meaning |
|---|---|
| `top1_accuracy_p5` / `top5_accuracy_p5` | "95% of users are at or above this" — catches broken users |
| `top1_accuracy_mean` / `top5_accuracy_mean` | Average across all users — catches overall degradation |

The print output shows a full distribution (min / p5 / median / mean / max) for both TOP1 and TOP5:
```
Token accuracy over 32 users:
  TOP1 — min=84.38%  p5=84.38%  median=84.38%  mean=84.38%  max=84.38%
  TOP5 — min=98.44%  p5=98.44%  median=98.44%  mean=98.44%  max=98.44%
```

Some models exhibit per-user accuracy divergence even with identical inputs due to numerical precision differences across batch positions on device. The p5 metric is specifically designed to catch these cases.

**Regression checker (`check_regression.py`):**

Updated `check_accuracy()` to check all 4 metrics with distinct failure labels ("Top-1 accuracy (p5) regression" vs "Top-1 accuracy (mean) regression"), so CI logs pinpoint *what* degraded.

**Transition handling:** The measurement keys are renamed (`top1_accuracy` → `top1_accuracy_p5`, etc.). On the first run after merge, Superset still has the old keys, so none of the new keys match — all 4 checks see `prev=None` and skip gracefully. No false failures. Full regression checking activates automatically on the second run once Superset has the new keys.

### Validation results (same build, same machine, batch_size=32)

Comparing the old user[0]-only metric against the new per-batch metrics:

| Model | Old TOP1 (user 0) | New TOP1 (p5) | New TOP1 (mean) | Old TOP5 (user 0) | New TOP5 (p5) | New TOP5 (mean) |
|-------|------|------|------|------|------|------|
| Llama-3.2-1B | 84.38% | 84.38% | 84.38% | 98.44% | 98.44% | 98.44% |
| Qwen-2.5-0.5B | 90.62% | 90.62% | 90.62% | 100.00% | 100.00% | 100.00% |
| Gemma-1.1-2B | 90.62% | 90.62% | 90.62% | 100.00% | 100.00% | 100.00% |
| Phi-2 | 87.50% | 87.50% | 87.50% | 100.00% | 100.00% | 100.00% |
| Mistral-7B | 100.00% | 100.00% | 100.00% | 100.00% | 100.00% | 100.00% |

> On single chip tests all 32 users produce identical accuracy (min = max) for these 5 models. The p5 metric will diverge from mean on models that exhibit per-user numerical precision variance across batch positions.

### DIFF
--- a/.github/scripts/check_regression.py
+++ b/.github/scripts/check_regression.py
@@ -31,19 +31,24 @@ def check_accuracy(prev_report, current_report):
     for m in current_report.get("measurements", []):
         current[m["measurement_name"]] = float(m["value"])
 
-    print(
-        f"Previous: top1={prev.get('top1_accuracy')}  top5={prev.get('top5_accuracy')}"
-    )
-    print(
-        f"Current:  top1={current.get('top1_accuracy')}  top5={current.get('top5_accuracy')}"
-    )
+    # Each metric is checked independently.  If the previous report lacks a key
+    # (e.g. first run after the metric rename), check_regression() sees prev=None
+    # and skips gracefully — no false failures during transition.
+    metrics = [
+        ("Top-1 accuracy (p5)", "top1_accuracy_p5"),
+        ("Top-5 accuracy (p5)", "top5_accuracy_p5"),
+        ("Top-1 accuracy (mean)", "top1_accuracy_mean"),
+        ("Top-5 accuracy (mean)", "top5_accuracy_mean"),
+    ]
 
-    failed = check_regression(
-        "Top-1 accuracy", prev.get("top1_accuracy"), current.get("top1_accuracy")
-    )
-    failed |= check_regression(
-        "Top-5 accuracy", prev.get("top5_accuracy"), current.get("top5_accuracy")
-    )
+    failed = False
+    for label, key in metrics:
+        prev_val = prev.get(key)
+        current_val = current.get(key)
+
+        print(f"{label}: prev={prev_val}  current={current_val}")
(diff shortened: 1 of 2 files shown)


odjuricicTT (approved):

dgolubovicTT:
For sanity check I just ran single chip tests, and as expected all users within batch have the same top1/top5 accuracy. However, this is not the case for all models. And that is what we want to catch with p5 metrics...

nsumrakTT (approved):

vvukomanTT (approved):

Issue #4387, opened 2026-04-23
[vLLM] gather_logprobs graph fails trace-insertion at opt_level=1


kmabeeTT:
## Summary

vLLM's `gather_logprobs` graph fails compilation under metal trace (`enable_trace=True`) at `optimization_level=1`. Upcoming PR will hang the parent metal-trace feature and works around this one graph by (a) skipping the startup precompile and (b) routing sampled-token logprob post-processing through CPU when `enable_trace=True`. `optimization_level=0` compiles on-device correctly. The CPU fallback is covered by the regression test `tests/integrations/vllm_plugin/sampling/test_trace_logprobs.py::test_opt125m_trace_logprobs` at `opt_level=1`.

## Details

`gather_logprobs` uses `torch.gather(logits, -1, selected_token_ids)`, which lowers to `ttir.embedding`. With trace insertion enabled at `opt_level=1`, the pass fails:

```
error: 'ttnn.capture_or_execute_trace' op All output tensors of trace function must be on device.
%22 = "ttnn.from_device"(%arg1) : (tensor<1x1xsi32, ...dram>) ->
      tensor<1x1xsi32, ...system_memory> is not on device.
```

Same shape of failure as the RoPE case in #4220 — `from_device` is inserted on a runtime index tensor (here `selected_token_ids`, shape `1x1xsi32`), and the resulting host-side tensor becomes a trace output, violating the trace verifier.

## Repro

The test below passes today because of the CPU fallback in `model_runner.py`:

```bash
pytest -sv tests/integrations/vllm_plugin/sampling/test_trace_logprobs.py::test_opt125m_trace_logprobs
```

To reproduce the underlying compiler failure, remove the `enable_trace` branch that routes through CPU (the `TODO(#4387)` block at the `gather_logprobs` call site in `model_runner.py`) and rerun the test. The on-device path will then hit the trace-insertion error shown above.

## Desired fix

Either:
1. Trace-insertion pass handles `ttir.embedding` with an on-device index (generalizes the RoPE fix from #4220).
2. `torch.gather` with an on-device index lowers to something other than `ttir.embedding`.

## Follow-ups in tt-xla once fixed

- Remove the CPU fallback branch in `model_runner.py` (inline `TODO(#XXXX)` at the `gather_logprobs` call site).
- Remove the `enable_trace` guard around `_precompile_gather_logprobs` in `model_runner.py::capture_model`.
[...]

PR #4388, opened 2026-04-23
[vLLM] Support metal trace for Llama/RoPE models

### LINKED ISSUE #4220: [vLLM] Enable metal trace for Llama models — blocked by from_device on position_ids
### Problem description

Metal trace (`enable_trace`) gives a **2.14x decode speedup** on OPT-125M through vLLM (75 → 158 tok/s), but fails on all Llama/RoPE models with:

```
error: 'ttnn.capture_or_execute_trace' op All output tensors of trace function must be on device.
%15 = "ttnn.from_device"(%arg6) : (tensor<1x128xsi32, ...dram>) -> tensor<1x128xsi32, ...system_memory> is not on device.
```

The failing tensor is `position_ids` (`tensor<1x128xsi32>`, used for RoPE). During TTIR→TTNN lowering, a `from_device` is inserted on position_ids to support the gather from the cos/sin cache, and that result ends up as a trace output — which violates the trace constraint that all outputs must remain on-device.

### Root cause

vLLM's `LlamaForCausalLM` uses a pre-built cos/sin cache (`131072x64xbf16`) and gathers from it dynamically using position_ids. The TTNN lowering of this gather requires position_ids on host → `from_device` → trace verifier rejects it.

The non-vLLM benchmark path (torch.compile) works because it computes rotary embeddings from `inv_freq` via `const_eval` functions that run before the trace — position_ids never enter the traced graph.

`enable_const_eval=True` and `optimization_level=2` do not help in the vLLM path because const_eval cannot hoist a dynamic gather from a pre-built cache.

### Repro

Branch: `kmabee/vllm_enable_trace`

Simplest repro (fails in ~2 min):
```bash
pytest -sv tests/integrations/vllm_plugin/generative/test_trace_generation.py::test_llama_3_2_1b_trace
```

OPT-125M comparison (passes):
```bash
pytest -sv tests/integrations/vllm_plugin/generative/test_trace_generation.py::test_opt_125m_trace
```

Benchmark tests (greedy × trace on/off for both models):
```bash
pytest -sv "tests/benchmark/test_vllm_benchmarks.py::test_vllm_trace[opt-125m-greedy-trace]"
pytest -sv "tests/benchmark/test_vllm_benchmarks.py::test_vllm_trace[llama-1b-greedy-trace]"
```

Interactive server+client demo (toggle with `ENABLE_TRACE=true`):
```bash
cd examples/vllm/OPT-125M && ENABLE_TRACE=true ./service.sh   # works
cd examples/vllm/Llama-3.2-1B && ENABLE_TRACE=true ./service.sh  # fails
```

### Investigation details

Full analysis with IR dumps and graph comparisons: [trace_investigation_2026-04-12.md](https://github.com/tenstorrent/tt-xla/blob/kmabee/vllm_enable_trace/trace_investigation_2026-04-12.md) on the branch.

Key finding: the benchmark Llama graph has **zero `from_device` ops** and uses `const_eval` to pre-compute RoPE. The vLLM Llama graph has a `from_device` on position_ids that becomes a trace output.

### Performance impact (OPT-125M, batch=1, greedy, device sampling)

| Metric | No trace | Trace | Improvement |
|--------|----------|-------|-------------|
| Decode tok/s (benchmark) | 75.6 | 158.3 | 2.09x |
| Decode tok/s (server) | 73.3 | 157.0 | 2.14x |
| TTFT (server) | 27ms | 17ms | 1.6x |

### Next steps

1. Fix the TTNN gather lowering so position_ids indexing works on-device without `from_device`, or restructure the trace insertion pass to keep the gather outside the traced region.
2. Validate with Llama-3.2-1B, then target Llama-3.1-8B for production perf gains.
3. Once working, enable trace by default for vLLM Llama serving.

kmabeeTT:
## Ticket

Closes #4220
Related: #4387 (follow-up compiler bug for `gather_logprobs` under trace)

## Problem description

Metal trace (`enable_trace`) gives a decode-side speedup through vLLM (small to large depending on model) but fails on all Llama/RoPE models because vLLM's `RotaryEmbedding` gathers from a pre-built `cos_sin_cache` using `position_ids`. That lowers to `ttir.embedding`, which requires `from_device` on the indices and violates the trace constraint that all outputs must remain on-device. A second graph — `gather_logprobs` — hits the same class of failure at `opt_level=1`.

Both are worked around at the vLLM integration layer following the existing `TTRMSNorm` precedent in `overrides.py`; a compiler-side fix (#4387) would make vLLM's native paths work and is a candidate follow-up. See follow-up comment for full design rationale and perf numbers.

## What's changed

- `integrations/vllm_plugin/vllm_tt/platform.py`: add `enable_trace` field to `TTConfig` (default `False`), wired through `get_pjrt_compile_config()`.
- `integrations/vllm_plugin/vllm_tt/overrides.py`: add `TTRotaryEmbedding` replacing vLLM's `RotaryEmbedding` and its subclasses (e.g. `Llama3RotaryEmbedding`). Introduces an `ISINSTANCE_OVERRIDES` dispatch alongside the existing fqn-based map so subclasses inherit the override.
- `integrations/vllm_plugin/vllm_tt/model_runner.py`: (a) skip `_precompile_gather_logprobs` at startup when `enable_trace=True`; (b) route `gather_logprobs` through CPU at the runtime call site when `enable_trace=True`. Both tagged `TODO(#4387)`.
- `tests/integrations/vllm_plugin/sampling/test_trace_logprobs.py`: new regression test exercising the CPU-fallback path at `opt_level=1`.
- `tests/integrations/vllm_plugin/generative/test_llama3_3b_generation.py`: add `test_llama3_3b_generation_trace` variant (greedy + device sampling + bfp8 + const-eval + opt_level=1) to exercise the metal-trace path end-to-end on a RoPE model.

## Checklist

- [x] Unit / integration tests added
- [x] Benchmark numbers verified locally (posted as follow-up comment)

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -1390,11 +1390,23 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
             # temperature scaling) for the top-k logprobs. We can't enforce it
             # due to recompilations outside torch.compiled code, so just make
             # sure `sample_from_logits` does not modify the logits in-place.
-            logprobs = (
-                self.gather_logprobs(logits, selected_token_ids)
-                if sampling_metadata.logprobs
-                else None
-            )
+            if not sampling_metadata.logprobs:
+                logprobs = None
+            elif self.tt_config.enable_trace:
+                # TODO(#4387): remove once trace-insertion pass handles
+                # ttir.embedding on on-device indices. The on-device
+                # gather_logprobs graph fails trace at opt_level=1, so
+                # fall back to CPU — post-processing is cheap and only
+                # the sampled-token id + logits need to move to host.
+                logits_cpu = logits.cpu()
+                tokens_cpu = selected_token_ids.cpu()
+                logprobs = self.sampler.gather_logprobs(
+                    self.sampler.compute_logprobs(logits_cpu),
+                    self.model_config.max_logprobs,
+                    token_ids=tokens_cpu.squeeze(-1),
+                )
+            else:
+                logprobs = self.gather_logprobs(logits, selected_token_ids)
 
             # Remove padding on cpu and keep dynamic op outside of xla graph.
             selected_token_ids = selected_token_ids.cpu()[:num_reqs]
@@ -1920,7 +1932,11 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                 logger.warning(
                     "cpu_sampling=True: skipping device sampling precompilation"
                 )
-            self._precompile_gather_logprobs()
+            # TODO(#4387): precompile fails trace-insertion at opt_level=1;
+            # skip when trace is on. Paired with the CPU fallback at the
+            # runtime call site. Remove both once the compiler bug is fixed.
+            if not self.tt_config.enable_trace:
+                self._precompile_gather_logprobs()
 
     def profile_run(
         self,
--- a/integrations/vllm_plugin/vllm_tt/overrides.py
+++ b/integrations/vllm_plugin/vllm_tt/overrides.py
@@ -7,6 +7,8 @@ from typing import OrderedDict
 import torch
 import torch.nn as nn
 from vllm.model_executor.layers.layernorm import RMSNorm
+from vllm.model_executor.layers.rotary_embedding.base import RotaryEmbedding
+from vllm.model_executor.layers.rotary_embedding.common import ApplyRotaryEmb
 
 from .logger import tt_init_logger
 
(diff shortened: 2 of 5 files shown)


kmabeeTT:
**Design rationale**

Two workarounds at the vLLM integration layer rather than compiler fixes:

- **RoPE**: `TTRotaryEmbedding` computes cos/sin on-the-fly from `inv_freq` (outer + cos/sin), mathematically equivalent to vLLM's cached-gather path but staying entirely on-device. Follows the precedent of `TTRMSNorm` already in `overrides.py`, which replaces vLLM's `RMSNorm` for an analogous TT-compile-path incompatibility (FakeTensor `.data` access that breaks torch.compile tracing). The `replace_modules` infrastructure exists for exactly this kind of integration-layer override.
- **gather_logprobs**: two-part — skip the startup precompile when trace is on (otherwise engine init fails), and route runtime logprob post-processing through CPU when `enable_trace=True` (cheap — only the selected-token id and current-step logits move to host). Tracked in #4387; both branches are gated by `enable_trace` and are removed together when the compiler bug is fixed.

A compiler-side fix (avoiding `from_device` for on-device embedding indices in tt-mlir) would make vLLM's native paths work for both cases.

**Benchmark numbers**

Decode tok/s with trace on vs. off, batch=1, greedy + device sampling + bfp8 + const-eval + opt_level=1, measured on p100a (bh-30):

| Model | No trace | Trace | Speedup |
|---|---|---|---|
| OPT-125M | 71.5 | 180.8 | 2.53x |
| Llama-3.2-1B | 35.4 | 70.6 | 2.00x |
| Llama-3.2-3B | 15.4 | 26.7 | 1.73x |
| Llama-3.1-8B | 17.0 | 19.8 | 1.17x |

Speedup tapers at larger models — decode is increasingly matmul-bound, so eliminating dispatch overhead via trace has less relative impact.

**Note**

- Trace's benefit also depends on host-dispatch cost: these p100a numbers are from a host with an older server CPU. On a host with a faster CPU + newer PCIe, decode is already host-unbound for these model sizes and trace shows ~flat speedup — expected, not a regression.

mmanzoorTT (approved):

PR #4390, opened 2026-04-23
Add support for new ForgePrefillModel class in testing infra

umalesTT:
### Ticket
None

### Problem description
The test runner needs a clear way to know which loaders support the focused prefill path. Inheritance from a dedicated base class is the cleanest signal for this, replacing the previous signature-inspection approach.

### What's changed
- Test discovery in `dynamic_loader.py` is extended so that `get_model_variants` no longer assumes one class per loader. It now also picks up any `ForgePrefillModel` subclass defined alongside the regular `ModelLoader`, registering each as its own variant set. To make this work, the helper that imports a `loader.py` was renamed `import_model_loader → import_loader_module` and changed to return the loaded module itself rather than the `ModelLoader` class — discovery can then iterate the module's attributes directly (`mod.ModelLoader` plus any sibling prefill class) instead of trying to look the module up later in `sys.modules`. That lookup would have failed: dynamic loader files are registered under a deliberately invalid dashed tt-forge-models... key, while attributes' `__module__` carries the underscore form, so the lookup raises KeyError and would silently drop most loaders.
- For `validate_test_config.py`, same generalization at AST level. Helpers now take a class name (new. `_extract_variants_for_class`, `_has_method`); a new `_find_prefill_class_name` returns whichever class subclasses `ForgePrefillModel`. `_has_method` ignores inherited methods so a prefill subclass doesn't double-claim its parent's phases. Phase routing mirrors the runtime: prefill is validated against the prefill class, decode and other phases against `ModelLoader.` The prefill class reuses `ModelLoader's _VARIANTS ` when it doesn't declare its own.
- Dispatch in `dynamic_loader.py` and `dynamic_torch_model_tester.py` switches from signature-based checks to inheritance check against ForgePrefillModel
- In `test_models.py`, the parametrization for `test_all_models_torch` and `test_all_models_op_by_op` is taken from a filtered list (now `test_entries_torch_regular`) that excludes prefill loaders, keeping those suites identical to before. The unfiltered list feeds `_llm_test_params`, which routes prefill loaders to `LLM_PREFILL` and regular `ModelLoaders` with `load_inputs_decode` to `LLM_DECODE`, eliminating the previous signature-based fallback path.
- Relevant `tt_forge_models` [PR](https://github.com/tenstorrent/tt-forge-models/pull/612)

### Checklist

### DIFF
--- a/tests/runner/test_models.py
+++ b/tests/runner/test_models.py
@@ -1,7 +1,6 @@
 # SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
 #
 # SPDX-License-Identifier: Apache-2.0
-import inspect
 import os
 import shutil
 import socket
@@ -58,7 +57,23 @@ MODELS_ROOT_TORCH, test_entries_torch = TorchDynamicLoader.setup_test_discovery(
     PROJECT_ROOT
 )
 MODELS_ROOT_JAX, test_entries_jax = JaxDynamicLoader.setup_test_discovery(PROJECT_ROOT)
-all_test_entries = test_entries_torch + test_entries_jax
+
+# setup_test_discovery adds the models root to sys.path, so ForgeFocusModel
+# can now be imported from the same namespace as the dynamically loaded loaders.
+from tt_forge_models.base import ForgeFocusModel  # noqa: E402
+
+
+def _is_focus_loader(entry) -> bool:
+    """Return True if the entry's loader class is a ForgeFocusModel subclass."""
+    return issubclass(entry.variant_info[1], ForgeFocusModel)
+
+
+# Regular (non-focus) torch entries drive test_all_models_torch. Focus loaders
+# only participate in LLM-phase tests (see _llm_test_params below).
+test_entries_torch_regular = [
+    entry for entry in test_entries_torch if not _is_focus_loader(entry)
+]
+all_test_entries = test_entries_torch_regular + test_entries_jax
 
 
 def _run_model_test_impl(
@@ -297,7 +312,7 @@ def _run_model_test_impl(
 )
 @pytest.mark.parametrize(
     "test_entry",
-    test_entries_torch,
+    test_entries_torch_regular,
     ids=DynamicLoader.create_test_id_generator(MODELS_ROOT_TORCH),
 )
 def test_all_models_torch(
@@ -383,10 +398,15 @@ def test_all_models_jax(
 # LLM Specific tests for decode and prefill phases. Separate test to avoid impacting
 # original test names in test_all_models_torch and no need for collection-time deselection logic.
 
-# Build list of (test_entry, run_phase) pairs based on loader capabilities
+# Build list of (test_entry, run_phase) pairs based on loader capabilities.
+# Focus loaders are dedicated to the prefill phase; regular (non-focus) loaders
+# provide decode (and, for legacy non-focus loaders, prefill via hasattr).
 _llm_test_params = []
 for entry in test_entries_torch:
     ModelLoader = entry.variant_info[1]
+    if _is_focus_loader(entry):
+        _llm_test_params.append((entry, RunPhase.LLM_PREFILL))
+        continue
     if hasattr(ModelLoader, "load_inputs_decode"):
         _llm_test_params.append((entry, RunPhase.LLM_DECODE))
     if hasattr(ModelLoader, "load_inputs_prefill"):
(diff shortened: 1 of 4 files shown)


tests/runner/utils/dynamic_loader.py
```diff
@@ -236,28 +240,48 @@ def import_model_loader(cls, loader_path: str, models_root: str):
         sys.modules[module_path] = mod
         spec.loader.exec_module(mod)
 
-        return mod.ModelLoader
+        return mod, mod.ModelLoader
```
pglusacTT: If this returns mod and model loader then it could only return module? If so, consider renaming it.

tests/runner/utils/dynamic_loader.py
```diff
@@ -236,28 +240,48 @@ def import_model_loader(cls, loader_path: str, models_root: str):
...
+                    continue
+                loader_classes.append(attr)
+
+            for LoaderCls in loader_classes:
+                variants = LoaderCls.query_available_variants()
+                if variants:
+                    for variant_name in variants.keys():
+                        loader_paths[loader_path].append((variant_name, LoaderCls))
+                else:
+                    loader_paths[loader_path].append((None, LoaderCls))
```
pglusacTT, line 287: Why append the None?

tests/runner/test_models.py
```diff
@@ -58,7 +57,23 @@
...
-all_test_entries = test_entries_torch + test_entries_jax
+
+# setup_test_discovery adds the models root to sys.path, so ForgeFocusModel
+# can now be imported from the same namespace as the dynamically loaded loaders.
+from tt_forge_models.base import ForgeFocusModel  # noqa: E402
+
+
+def _is_focus_loader(entry) -> bool:
+    """Return True if the entry's loader class is a ForgeFocusModel subclass."""
+    return issubclass(entry.variant_info[1], ForgeFocusModel)
```
pglusacTT: `variant_info[1]` is easily breakable – it's a pure list? How does it look like?

tests/runner/test_models.py
```diff
@@ -58,7 +57,23 @@
...
+
+
+# Regular (non-focus) torch entries drive test_all_models_torch. Focus loaders
+# only participate in LLM-phase tests (see _llm_test_params below).
+test_entries_torch_regular = [
+    entry for entry in test_entries_torch if not _is_focus_loader(entry)
```
pglusacTT: Did you check the number of tests before and after your change? Do they match?

tests/runner/test_models.py
```diff
@@ -400,15 +417,13 @@ def _generate_mesh_shape_id(mesh_shape):
 
 
 def _supports_strategy_shard_spec(model_loader_cls) -> bool:
```
pglusacTT, line 436: Where is this function used?

tests/runner/utils/dynamic_loader.py
agobeljicTT: I vote to `return mod` only

umalesTT: Yes, now it's returning only `mod`

tests/runner/utils/dynamic_loader.py
umalesTT: Some models have no explicit variants in their loader

tests/runner/test_models.py
umalesTT: It's a tuple of `(ModelVariant, loader class)`
We could do type annotations to raise a potential warning, but I guess it's out of scope for this PR - as `variant_info[1]` is already used throughout infra

tests/runner/test_models.py
umalesTT: Main:
792 ::test_all_models_jax 
1568 ::test_all_models_op_by_op
4000 ::test_all_models_torch
30231 ::test_llms_torch
18 ::test_placeholder_models

Branch:
792 ::test_all_models_jax
1568 ::test_all_models_op_by_op
4000 ::test_all_models_torch
24615 ::test_llms_torch
18 ::test_placeholder_models

Drop for `test_llms_torch` is expected, prefill used to be generated for every `ModelLoader` with `load_inputs_prefill`  after the refactor only `ForgeFocusModel` subclasses get it.

tests/runner/test_models.py
```diff
@@ -382,14 +397,16 @@ def test_all_models_jax(
...
 # original test names in test_all_models_torch and no need for collection-time deselection logic.
 
-# Build list of (test_entry, run_phase) pairs based on loader capabilities
+# Build list of (test_entry, run_phase) pairs based on loader capabilities.
+# Focus loaders own the prefill phase; regular ModelLoaders provide decode.
 _llm_test_params = []
 for entry in test_entries_torch:
     ModelLoader = entry.variant_info[1]
+    if _is_focus_loader(entry):
+        _llm_test_params.append((entry, RunPhase.LLM_PREFILL))
```
kmabeeTT, line 407: Do we have any ModelLoaders with load_inputs_prefill that aren't focus classed that would get silently dropped here?

kmabeeTT (commented):
Seems okay, one question about prefill, would be nice to mention in testing what models are covering this new path today and if they work.  Careful to only merge tt-forge-models once everything is approved and this is ready to be merged, then update this PR to point to officially released tt-forge-models in main rather than branch.

tests/runner/test_models.py
umalesTT: No, we have none.

kmabeeTT (approved):

PR #4396, opened 2026-04-24
Graceful shutdown of CallbackWorker in torch_plugin_tt

### LINKED ISSUE #4303: Tt-xla model demo test is failing
Resnet 50 demo model test is failing since 18.04. -> [failing job](https://github.com/tenstorrent/tt-xla/actions/runs/24642681050/job/72049960628).

This demo test is used as part of the forge release, so p0 (and is the model we use as an example in our [docs](https://github.com/tenstorrent/tt-forge#run-a-model) on tt-forge).

acicovicTT:
Torch XLA doesn't close the PJRT client object through the proper API call, which later causes race conditions during Python process teardown with the newly added CallbackWorker which still holds some Python references when teardown starts. This PR solves issue https://github.com/tenstorrent/tt-xla/issues/4303.

Some notes for reviewers:
- Shutdown was intentionally NOT registered in JAX plugin, JAX should handle everything correctly. Looking for feedback on whether we should call `shutdown()` in `onClientDestroy` which is called from JAX, or just let it be called during `CallbackWorker`'s destruction.
- The exported function was added to `dylib_entry_point.cc` which is not the obvious place for it, but adding it to its own translation unit like `dylib_exit.cc` would require the extraction of some macro definitions in `dylib_entry_point.cc.inc` to their own `.h` file, which seemed excessive to me, so I left it as it is.

### DIFF
--- a/pjrt_implementation/inc/api/event_instance.h
+++ b/pjrt_implementation/inc/api/event_instance.h
@@ -80,6 +80,14 @@ public:
   // lazily on first access.
   static CallbackWorker &getCallbackWorker();
 
+  // Shuts down the CallbackWorker: drains pending callbacks, joins the
+  // worker thread, and switches `enqueue` to synchronous execution on the
+  // caller's thread for any callbacks that arrive afterwards. Intended to
+  // be invoked from a Python `atexit` hook (via `tt_pjrt_shutdown`) so
+  // that pending and post-hook callbacks all run while Python state and
+  // the GIL are still valid.
+  static void destroyCallbackWorker();
+
   // See comment below for `m_indestructible`.
   void setIndestructible() { m_indestructible = true; }
   bool isIndestructible() const { return m_indestructible; }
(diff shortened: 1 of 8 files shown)


acicovicTT:
Release tests: https://github.com/tenstorrent/tt-xla/actions/runs/24987703821

pjrt_implementation/inc/api/event_instance.h
```diff
@@ -80,6 +80,11 @@ class EventInstance {
   // lazily on first access.
   static CallbackWorker &getCallbackWorker();
 
+  // Shuts down the CallbackWorker: drains pending callbacks, joins the
+  // worker thread, and switches enqueue() to synchronous execution on the
+  // caller's thread for any callbacks that arrive afterwards.
+  static void shutdownCallbackWorker();
```
mrakitaTT, line 86: Seems like a better place for these two would be in a `CallbackWorker` class

mrakitaTT (approved):

pjrt_implementation/inc/api/event_instance.h
acicovicTT: Yes, but the singleton was originally tied to this class, I suspect because it is logically owned by EventInstance, so I left it in the EventInstance as well

acicovicTT: Let's sync offline further, we can patch it up in a follow up PR, I want to unblock the release pipeline :)

PR #4419, opened 2026-04-27
Allow TT_RUNTIME_USING_DUALT3K to force fabric 2D init in non-distributed context

jameszianxuTT:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/43210

### Problem description
MeshDevice init for single BH galaxy does not work when it is cabled into the exabox cluster as part of a quad in the 4x32 config. Automatic topology detection fails under FABRIC_1D_RING, which is the default used for singledevice in tt-xla.

This is not expected and a metal bug. It seems to be part of a larger class of bugs where mesh device init in different fabric config contexts breaks on various topologies in a way correlated with their physical cabling. 

### What's changed
Move TT_RUNTIME_USING_DUALT3K variable to force fabric 2D init in distributed context to global context, now that this bug has been observed in singledevice context on single bh glx in exabox cluster. This is not necessary in general for bh glx, just the ones in exabox.

### Checklist
- [x] No CI coverage for this topology. Tested with `TT_RUNTIME_USING_DUALT3K=1 pytest -svv tests/torch/test_torch_xla_basic.py::test_fully_replicated_graph[True]` on exabox single BH galaxy, which fails without setting the env variable.

### DIFF
--- a/pjrt_implementation/src/api/client_instance.cc
+++ b/pjrt_implementation/src/api/client_instance.cc
@@ -478,16 +478,18 @@ tt_pjrt_status ClientInstance::compileMlirProgram(
 
 tt::runtime::MeshFabricConfig
 ClientInstance::computeFabricConfig(const std::vector<uint32_t> &mesh_shape) {
-  // Distributed uses FABRIC_1D for now.
+
+  // [Workaround] Override fabric config to FABRIC_2D due to device init bugs
+  // under default 1D ring + topology auto-detection required for dual t3k
+  // and single bh galaxy in exabox cluster
+  if (std::getenv("TT_RUNTIME_USING_DUALT3K") != nullptr &&
+      std::string(std::getenv("TT_RUNTIME_USING_DUALT3K")) != "0") {
+    return tt::runtime::MeshFabricConfig{tt::runtime::FabricConfig::FABRIC_2D,
+                                         {}};
+  }
+
   if (std::getenv("TT_RUNTIME_ENABLE_DISTRIBUTED") != nullptr &&
       std::string(std::getenv("TT_RUNTIME_ENABLE_DISTRIBUTED")) != "0") {
-
-    // [Workaround] Override fabric config to FABRIC_2D for dual t3k cluster
-    if (std::getenv("TT_RUNTIME_USING_DUALT3K") != nullptr &&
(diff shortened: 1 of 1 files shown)


jameszianxuTT:
Failure signature:
```
Graph specified in MGD could not fit in the discovered physical topology. Inter-mesh mapping failed after 2 attempt(s). Logical meshes being mapped: [0] (1 total). Physical meshes available: [0] (1 total). Failed mesh pair configurations tried: 1 out of 1 possible combinations. Inter-mesh validation mode: STRICT. Solver error: Mapping validation failed: 1 target node(s) are not mapped to any global node: 0. Failed mesh pairs from previous attempts: [(logical=0, physical=0)].. Either relax pinnings or modify the MGD. If this is unexpected, run ./build/test/tt_metal/tt_fabric/test_system_health to check connectivity.
```

sshonTT (approved):

pjrt_implementation/src/api/client_instance.cc
```diff
@@ -478,17 +478,19 @@ tt_pjrt_status ClientInstance::compileMlirProgram(
...
 tt::runtime::MeshFabricConfig
 ClientInstance::computeFabricConfig(const std::vector<uint32_t> &mesh_shape) {
-  // Distributed uses FABRIC_1D for now.
+
+  // [Workaround] Override fabric config to FABRIC_2D due to device init bugs
+  // under default 1D ring + topology auto-detection required for dual t3k
+  // and single bh galaxy in exabox cluster
+  if (std::getenv("TT_RUNTIME_USING_DUALT3K") != nullptr &&
+      std::string(std::getenv("TT_RUNTIME_USING_DUALT3K")) != "0") {
+    return tt::runtime::MeshFabricConfig{tt::runtime::FabricConfig::FABRIC_2D,
```
pilkicTT, line 487: might be good to log a warning that the fabric config is being overwritten...

pilkicTT (approved):

pjrt_implementation/src/api/client_instance.cc
jameszianxuTT: The only function of the env var `TT_RUNTIME_USING_DUALT3K` is to override the fabric config, I will probably need to rename it at some point (or wait for metal patch to solve the root issue). So I think a warning is a bit redundant, since user needs to manually set it TT_RUNTIME_USING_DUALT3K

PR #4424, opened 2026-04-28
Sgligorijevic/hf moe attn integration

sgligorijevicTT:
### Problem description
Our current MoE integration is not quite elegant.

### What's changed
Based on the existing `sparse_mlp.py`, I added an `moe_backend.py`. For context, huggingface has a concept of [expert backends](https://huggingface.co/docs/transformers/experts_interface) which exposes a uniform interface for inserting custom operations for MoE across all models. This PR adds the required code and two examples of usage.

Side benefit, we brought up Trinity-Nano and Olmoe with basically no effort :) .

Beware, code is basically all Claude generated, I hope to clean up any rough edges before merging.

Moving over all our tests to this code comes later, I already have a WIP branch for that.

### Checklist
- [ ] New/Existing tests provide coverage for changes - they don't, that will be a followup PR, for now it's only the two examples


AleksKnezevic:
@sshonTT can you please review?

sshonTT:
I believe this implementation offers a more formal and robust way to support MoE compared to the current version in sparse_mlp.py. Once we tidy up a few details, it should be ready to be mainlined.

@sgligorijevicTT, would you like to drive the remaining cleanup yourself, or would you like some help with the details?

PR #4429, opened 2026-04-28
Codegen multiple dirs on graph breaks instead of clobbering

### LINKED ISSUE #4383: EmitTTNN - fail gracefully on multiple graphs
Problem:
In situations where there are multiple graphs, emitting TTNN outputs a solution for each graph to the same dir, which overwrites previous graphs.

I get daily pings on this problem, so we should try and solve it ASAP.

Potential fix:
Output solutions to different dirs.
I do have a hotfix, but should consider a slightly better approach, e.g. something stateful that keeps track of the number of graphs produced within a single codegen call.

Hotfix:
```
git show 34d4aff -- pjrt_implementation/src/api/so_loaded_executable_instance.cc | git apply
```

cc @sdjordjevicTT @mrakitaTT @odjuricicTT

svuckovicTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/4383

### Problem description
In the codegen path, graph breaks would cause the compiler to output generated code into the same dir, i.e. if 2 graphs are generated (due to a graph break), second graph's generated code would overwrite the first one.

### What's changed
Keep track of number of graphs generated and write into `{export_path}/graph_{index}`.
E.g. if export path is `llama_codegenerated`, we'd now get:
```
llama_codegenerated/graph_0/...
llama_codegenerated/graph_1/...
...
```

Added a specific test for this.

### Checklist


svuckovicTT:
fyi @sdjordjevicTT @rpavlovicTT @odjuricicTT

sdjordjevicTT (approved):
Nice! :fire:

sgligorijevicTT (approved):
Thanks for fixing this!

Issue #4053, opened 2026-04-01
AOTAutograd exposes a runtime Segfault


sgligorijevicTT:
Repro steps:
```
git checkout sgligorijevic/aotautograd-on-by-default
cmake --build build
pytest -svv tests/torch/training/test_basic.py::test_qwen3_multichip_backward
```

Expected behavior:
Either passes completely, or correctly fails with bad PCC
Actual behavior:
SEGFAULT

I noticed this is a single chip graph without inputs that gets executed after a multichip graph that faults, during device reopening.

acolicTT:
I could't repro python segfaults in both debug and release builds (I rebased sgligorijevic/aotautograd-on-by-default on main yesterday). 
Instead, each time I have an error: ERR| Device count mismatch: 2 vs 1

This happens after the last graph execution, which has different mesh shape (1, 1) then previous ones (1, 2):

```
...
...      client_instance.cc:534      1| ClientInstance::getOrCreateMeshDevice - reshaping mesh device - [1, 2] -> [1, 1]
...          tensor_pool.cc:91       1| Moving tensors to host.
...      client_instance.cc:608      1| Closing parent mesh.
...      client_instance.cc:574      1| ClientInstance::openMeshDevice - setting fabric config: DISABLED
... loaded_executable_insta:266      1| LoadedExecutableInstance::PJRT_LoadedExecutable_GetExecutable
... loaded_executable_insta:286      1| LoadedExecutableInstance::PJRT_LoadedExecutable_AddressableDevices
...               stubs.inc:80    WARN| STUB: PJRT_Executable_GetCompiledMemoryStats
...       error_instance.cc:52       1| ErrorInstance::PJRT_Error_Message
...       error_instance.cc:61       1| ErrorInstance::PJRT_Error_GetCode
...       error_instance.cc:46       1| ErrorInstance::PJRT_Error_Destroy
...  executable_instance.cc:116      1| ExecutableInstance::PJRT_Executable_OptimizedProgram
...  executable_instance.cc:116      1| ExecutableInstance::PJRT_Executable_OptimizedProgram
...  executable_instance.cc:116      1| ExecutableInstance::PJRT_Executable_OptimizedProgram
...  executable_instance.cc:116      1| ExecutableInstance::PJRT_Executable_OptimizedProgram
...  executable_instance.cc:254      1| ExecutableInstance::PJRT_Executable_Serialize
[...]

PR #4439, opened 2026-04-28
Fix mesh shape when the graph has no inputs

pglusacTT:
### Ticket
Related to https://github.com/tenstorrent/tt-xla/issues/4053
Depends on https://github.com/tenstorrent/tt-mlir/pull/8167

### Problem description
When graph has no inputs we get a device mismatch since it opens a 1x1 mesh by default.

### What's changed
- Pass mesh shape to stablehlo pipeline
- Use all devices available

Note:
We might still want to handle 1x1 -> 1xN transfers?

### Checklist

### DIFF
--- a/pjrt_implementation/src/api/module_builder/module_builder.cc
+++ b/pjrt_implementation/src/api/module_builder/module_builder.cc
@@ -845,6 +845,13 @@ tt_pjrt_status ModuleBuilder::runCompilerStableHLOPipeline(
                                           mlir::PassManager::Nesting::Implicit);
   mlir::tt::stablehlo::StableHLOPipelineOptions stablehlo_pipeline_options;
   stablehlo_pipeline_options.resultPresharded = result_presharded;
+  // Forward the open parent mesh shape so AnalyzeMeshPass can use it as the
+  // fallback for modules with no shardings.
+  if (current_mesh_shape.has_value() && current_mesh_shape->size() == 2) {
+    stablehlo_pipeline_options.meshShape = {
+        static_cast<int64_t>((*current_mesh_shape)[0]),
+        static_cast<int64_t>((*current_mesh_shape)[1])};
+  }
   mlir::tt::stablehlo::createStableHLOPipeline(stablehlo_pipeline_pm,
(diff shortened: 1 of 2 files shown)


third_party/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 option(USE_CUSTOM_TT_MLIR_VERSION "Flag to use TT_MLIR_VERSION set by the user" OFF)
 
 if (NOT DEFINED TT_MLIR_VERSION OR NOT USE_CUSTOM_TT_MLIR_VERSION)
-    set(TT_MLIR_VERSION "f3ddbfb6b0eab2c2ec65fe45ec2347cc6ebedaca")
+    set(TT_MLIR_VERSION "pglusac/analyze-mesh-shape-fix")
```
pglusacTT: Will revert after it gets merged into main.

pjrt_implementation/src/api/module_builder/module_builder.cc
```diff
@@ -81,6 +81,18 @@ namespace tt::pjrt::module_builder {
 
 const std::string c_mlir_format_name = "mlir";
 
+static bool moduleHasAnyFuncArguments(mlir::OwningOpRef<mlir::ModuleOp> &m) {
+  bool found = false;
+  m.get().walk([&](mlir::func::FuncOp funcOp) {
+    if (funcOp.getNumArguments() > 0) {
```
sgligorijevicTT: How does this work when the entrypoint has no arguments but the module contains private functions as implementation detail that do have arguments? For example if Jax hoists out an RNG definition, or composites are used and a decomposition is emitted?
Elsewhere we assume that there is exactly one public function, perhaps you can assert that here and then count up the number of arguments of public functions only?

pglusacTT: Good point, done

sgligorijevicTT (approved):

sshonTT (approved):

acolicTT (approved):

jameszianxuTT (approved):

PR #4446, opened 2026-04-29
[vLLM plugin] Fix 2D-mesh device-sampler garbage output (and add coherence check)

### LINKED ISSUE #4440: [vLLM plugin] Incorrect/random output generated with TP generative test
Tensor parallel generative model tests are generating incorrect/random output for 2D mesh. This issue is caused by sampling on device. These tests generate correct results with either 1D mesh or 2D mesh with cpu sampling. 

Output for last few nightly runs for `Qwen/Qwen3-32B` with 2D mesh are below 
[CI job 27 Apr](https://github.com/tenstorrent/tt-xla/actions/runs/25026871133/job/73299881403#step:19:439)
```
prompt: I like taking walks in the, 
output:  restartgp erect workplaceSignpart Unblock_REGSamplerlekestado                                           (info Htmldom ///building neighborhoods somewhat '''iness(())
```
[CI job 26 Apr](https://github.com/tenstorrent/tt-xla/actions/runs/24970760261/job/73131748597#step:19:439)
```
prompt: I like taking walks in the, 
output:  dou调 arrangementyieldysisyieldSignatureited bitstern(\ THE Need(isset Handle Parameter arts post Gretaphael Handle Penguin surplus@SpringBootTestystem language
```
[CI job 25 Apr](https://github.com/tenstorrent/tt-xla/actions/runs/24944092006/job/73042387279#step:19:439)
```
prompt: I like taking walks in the, 
output:  restart twittererase workflow winnerfectiontreeimiento Regionpartizophren Son класс PureComponent '''class.PORT\Requests_LE volunte Bool Personen Concf typically Sony
```
[CI job 24 Apr](https://github.com/tenstorrent/tt-xla/actions/runs/24917761316/job/72973189022#step:19:439)
```
prompt: I like taking walks in the, 
output:  dou--){
hen(Main.PORTsession Unblock Unblock reslek artscoctract]=>SQL Unblock.cHaving_play availabilityestureRecognizer shares Unblock ''' Char_OPTION
```

kmabeeTT:
## Ticket

Closes #4440

## Problem description

- Tensor-parallel generative tests with 2D mesh + on-device sampling produced token-soup garbage (e.g. `'.presentientes winner actually! Bool volunte vertexlek language("");\n ReIGHlek...'`), while 1D mesh and 2D-mesh-with-CPU-sampling were both fine. Bisected to PR #3820 which gated the post-`compute_logits` replication-constraint to the 1D-mesh path only — the device sampler then operated on logits still sharded along the `model` axis.

- A second issue surfaced during the investigation: the bug had been landing as PASSED in CI because the tests asserted only that `llm.generate(...)` completed, not that its output was coherent. A small heuristic check is added so token-soup output now hard-fails.

## What's changed

- `integrations/vllm_plugin/vllm_tt/model_runner.py`: drop the `not self.use_2d_mesh` guard in `compute_logits` so the `sharding_constraint_tensor((None, None))` replication applies for both 1D and 2D mesh — restores pre-#3820 behavior. (Pre-commit isort also reordered two unrelated import lines in the same file.)
- `tests/integrations/vllm_plugin/generative/conftest.py`: add `assert_output_coherent` — flags empty output, >3% non-Latin chars, or <10% common-English-stopword ratio as garbage.
- `tests/integrations/vllm_plugin/generative/test_tensor_parallel_generation.py`: call `assert_output_coherent(output_text)` after `llm.generate(...)` in the three TP generation tests (`_n300`, `_llmbox_small`, `_llmbox_large`).

See follow-up comment for the bisect, root-cause walkthrough, and heuristic tuning.

## Checklist

- [x] Bug reproduces and clears with the patch on Qwen3-8B (smallest reproducer) and Qwen3-32B (the original failing test).
- [x] New `assert_output_coherent` validated against 8 known-good and 6 known-garbage outputs from local runs and CI logs (issue #4440).

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -2172,6 +2172,10 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
     @torch.compile(backend="tt", fullgraph=True, dynamic=False)
     def compute_logits(self, sample_hidden_states: torch.Tensor) -> torch.Tensor:
         logits = self.model.compute_logits(sample_hidden_states)
+        # Replicate logits for SPMD. Hooks can't reach ParallelLMHead
+        # (quant_method.apply bypasses __call__) and all_gather is a
+        # no-op (world_size=1). Must be inside the compiled graph —
+        # external sharding_constraint between compiled functions breaks.
         if self.enable_tensor_parallel:
             logits = sharding_constraint_tensor(logits, self.mesh, (None, None))
         return logits
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -2172,7 +2172,7 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
     @torch.compile(backend="tt", fullgraph=True, dynamic=False)
     def compute_logits(self, sample_hidden_states: torch.Tensor) -> torch.Tensor:
         logits = self.model.compute_logits(sample_hidden_states)
-        if self.enable_tensor_parallel and not self.use_2d_mesh:
+        if self.enable_tensor_parallel:
             logits = sharding_constraint_tensor(logits, self.mesh, (None, None))
         return logits
 
--- a/tests/integrations/vllm_plugin/generative/conftest.py
+++ b/tests/integrations/vllm_plugin/generative/conftest.py
@@ -2,7 +2,50 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 """Shared conftest for vLLM generative tests."""
+import re
+
 import psutil
+
+# Common English function words used by `assert_output_coherent` to detect
+# the 2D-mesh sampler garbage-output bug (issue #4440). Coherent natural-
+# language continuations contain several of these per ~30-token output;
+# token-soup garbage contains ~zero.
+_STOPWORDS = frozenset(
+    """
+    the a an and or but i you he she it we they is are was were be been
+    have has had do does did of to in on at for with by from as that this
+    my your her his their can will would should like go get make me so
+    not if when what how there here
+    """.split()
+)
+_WORD_RE = re.compile(r"[A-Za-z']+")
+_MIN_STOPWORD_RATIO = 0.10
+_MAX_NON_LATIN_RATIO = 0.03
+_MIN_WORDS = 5
+
+
+def assert_output_coherent(text: str) -> None:
+    """Heuristic assertion: text is English natural-language, not token soup.
+
+    Tuned for the English prompts used in the generative TP tests
+    (e.g. "I like taking walks in the"). Re-tune thresholds if a test
+    introduces non-English prompts or code-completion prompts.
+    """
+    s = text.strip()
+    assert s, f"empty output: {text!r}"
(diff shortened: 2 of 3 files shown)


kmabeeTT:
Full bisect, root-cause walkthrough, why-this-fix justification, and before/after model output are in the issue: [#4440 (comment)](https://github.com/tenstorrent/tt-xla/issues/4440#issuecomment-4340196591). Brief recap: PR #3820 gated off the post-`compute_logits` `sharding_constraint_tensor((None, None))` for 2D mesh, leaving logits sharded along the `model` axis when they entered the sampler graph; vocab-dim reductions in the sampler then operated per-shard and picked invalid tokens. This patch is a partial revert that re-enables that constraint for 2D as well.

A few extra notes for reviewers that aren't in the issue:

**Why 8B and not the existing CI tests caught it.** The 1D-vs-2D logit numerical drift is present at all model sizes (different reduction order between SPMD plans `(N, 1)` and `(R, C)`), but is sub-clinical at smaller scale — Qwen3-0.6B / 1.7B / 4B all generate coherent text on 2D + device sampling despite the same underlying issue. Qwen3-8B is the smallest size where the drift is large enough to flip enough sampler picks to produce garbage. The existing `_llmbox_small` test (Qwen3-0.6B) passes with or without the fix and didn't catch the regression.

**Why a heuristic coherence check rather than a stronger comparison.** The TP tests use temperature=0.8 + top-p=0.95, so even bit-identical logits produce different sampled tokens across runs (RNG path through the sampler). A strict text-equality check is therefore impossible without reseeding identically across configs, which the test harness doesn't expose. The heuristic (empty-string / non-Latin ratio / stopword ratio) is the cheapest reliable signal: 0 misclassifications across 8 known-good + 6 known-garbage outputs from local runs and the CI logs in #4440.

**Out-of-scope follow-up.** A weaker `assert_coherent` already lives in `tests/integrations/vllm_plugin/sampling/test_cpu_sampling.py` (ASCII ratio + min word count). Unifying both into a shared helper is worth doing, but I left it out of this PR to keep the scope tight.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2172,7 +2172,7 @@ def select_hidden_states(self, hidden_states, indices_do_sample):
     @torch.compile(backend="tt", fullgraph=True, dynamic=False)
     def compute_logits(self, sample_hidden_states: torch.Tensor) -> torch.Tensor:
         logits = self.model.compute_logits(sample_hidden_states)
-        if self.enable_tensor_parallel and not self.use_2d_mesh:
```
mmanzoorTT, line 2175: Please run the vLLM nightly tests with this change. I had tried this configuration in the past and few models were failing.

kmabeeTT: Darn, okay sounds good, kicked off here will watch https://github.com/tenstorrent/tt-xla/actions/runs/25087328990

Also in case this ends up working out good, I pushed a change just now to this PR to add back original 4 line comment above this change that got dropped during the 2d-mesh support change.

kmabeeTT: Hmm yeah okay needs some additional changes somewhere... n300 fails even in push, will try and take a closer look...

```
loc("custom-call.11"): error: 'sdy.sharding_constraint' op unknown mesh: @mesh
2026-04-29 03:17:43.073 ( 254.559s) [        307E7300]      module_builder.cc:839    ERR| Failed to run stablehlo pipeline
```

kmabeeTT:
replaced by a PR from Asif.

PR #4456, opened 2026-04-29
Simplify call-test.yml by moving conditions into test matrix config

vmilosevic:
## Summary

- **Fix silent bug**: `wheel_build` override for `require=release/alchemist` was immediately overwritten by an unconditional `echo` on the next line — the override never took effect
- **Move vllm detection to config**: replace 3×4-way `matrix.build.name` checks with a single `matrix.build['extra-wheel'] == 'vllm'` flag set on vllm entries in the JSON presets
- **Move forked detection to config**: replace a hardcoded list of 10 names in the workflow with a `"forked": true` field on the relevant matrix entries
- **Move forge-models detection to config**: replace fragile `dir`-string-sniffing (`*tests/runner/test_models.py*`) with an explicit `"forge-models": true` field on all `test_models.py` entries
- **Remove passthrough variable**: `artifact_run_id` was just aliasing `inputs.artifact_release_run_id` — all usages now reference the input directly

Adding a new test type that needs forking, vllm wheels, or forge-models treatment now only requires setting a flag in the JSON preset — the workflow no longer needs to be edited.

## Test plan
- [ ] Verify vllm jobs still download and install the vllm wheel
- [ ] Verify forge-models jobs still receive `--arch` flag and run system dep install
- [ ] Verify torch/forge-models jobs still run with `--forked`
- [ ] Verify `require=alchemist` jobs now correctly use `wheel_build=release`


nsumrakTT (commented):
with "require" and added "forge-models",  "forked", and "extra-wheel" maybe it would be a good idea to put document file inside test-matrix-presets dir?

.github/workflows/call-test.yml
```diff
@@ -320,7 +307,7 @@ jobs:
         echo "VERBOSITY=$VERBOSITY" >> $GITHUB_ENV
 
     - name: Install System Deps for Forge Models Tests
-      if: ${{ steps.strings.outputs.forge-models-test == 'true' }}
+      if: ${{ matrix.build['forge-models'] == true }}
```
vvukomanTT: Let's trying using a consolidated approach for dereferencing json object fields
i.e. `matrix.build['forge-models']` vs `matrix.build.forge-models`

vvukomanTT (approved):

nsumrakTT (approved):

Issue #4293, opened 2026-04-17
AI skill for automated triage of backward pass frontend failures


vladimirjovanovicTT:
## Problem

Many training tests are stuck at `FAILED_FE_COMPILATION` with a small set of recurring frontend errors. These are triaged manually today, which doesn't scale.

## Proposed solution

Build a Claude skill that triages each error pattern with a repeatable workflow. The general approach: run the model on CPU first to check if the error reproduces there (frontend issue) vs only on TT backend.

### Pattern-specific workflows

**`tt-forge-models doesn't implement unpack_forward_output for this model.`**
- Run model on CPU
- Inspect what the model returns
- Gather documentation for the model
- Suggest what outputs are needed for training loss (e.g. logits for LLMs)
- Reference examples in `training_utils.py` and `unpack_forward_output` overrides in model loaders

**`RuntimeError: mat1 and mat2 must have the same dtype, but got Float and BFloat16`**
**`RuntimeError: 'deformable_im2col' not implemented for 'BFloat16'`**
- Try running on CPU in float32 to check if they run at all — possibly just can't run in bfloat16

**`ValueError: You have to specify either decoder_input_ids or decoder_inputs_embeds.`**
**`ValueError: Expected more than 1 value per channel when training, got input size torch.Size([1, 256, 1, 1])`**
**`Model expects targets to be passed while in training mode`**
**`AssertionError: targets should not be none when in training mode`**
**`AttributeError: 'NoneType' object has no attribute 'max'`**
- We're not passing something to the model that we should. The skill should give a short description of what needs to be passed. After going through all these errors, aggregate results to see what's the best course of action to handle them in bulk.

### Reference files
- `third_party/tt_forge_models/training_utils.py`
- `third_party/tt_forge_models/base.py`
- `unpack_forward_output` overrides in model loaders
- `tests/runner/test_config/torch/test_config_training_single_device.yaml`

PR #4457, opened 2026-04-29
Add training triage skills for unpack and dtype FE compilation failures

agobeljicTT:
### Ticket
Addresses 2/3 of #4293. The third pattern (missing training inputs) will be handled in a followup queued to this PR.

### Problem description
Many training tests in `test_config_training_single_device.yaml` are stuck at `FAILED_FE_COMPILATION` with a small set of recurring frontend errors. These are triaged manually today, which doesn't scale. Issue #4293 proposed building a Claude skill that triages each error pattern with a repeatable workflow.

### What's changed
This PR introduces two of the three planned triage skills plus the shared tooling that drives them:

- **`tools/triage_fe_failures.py`** — script that scans the training YAML, groups `FAILED_FE_COMPILATION` entries by error pattern (`unpack` / `dtype` / `inputs`), and optionally runs each model on CPU to collect output classes / additional error context. Supports `--pattern`, `--run`, `--offset`, `--limit`, and `--output` flags so models can be triaged in batches without timing out.
- **`.claude/skills/training-triage-unpack/SKILL.md`** — workflow for `tt-forge-models doesn't implement unpack_forward_output for this model.` Walks through identifying the model's output class statically, registering it in `training_utils.py` via `_register_attr` (or overriding `unpack_forward_output` in the loader), and verifying the fix via pytest with downstream-error handling guidance.
- **`.claude/skills/training-triage-dtype/SKILL.md`** — workflow for the bfloat16 mismatch / unsupported-op errors. Re-runs the loader on CPU in float32 to determine whether the model is fundamentally broken or only fails under mixed precision, then directs the fix into the loader's mixed-precision path.
- **`tests/runner/test_config/torch/test_config_training_single_device.yaml`** — updated entries for models triaged with the new skills: failures that are no longer FE-compilation issues are reclassified to their real downstream stage (`FAILED_TTMLIR_COMPILATION`, `FAILED_RUNTIME`, PCC mismatches, LayerDrop nondeterminism, etc.), and models that now pass are flipped to `EXPECTED_PASSING`.

The `inputs` pattern (decoder inputs / training targets / `NoneType.max`) is intentionally out of scope here and will land in a followup PR queued to this one.

### Checklist
- [x] New/Existing tests provide coverage for changes (YAML statuses updated based on triage results)


agobeljicTT:
Superseded by #4459, which rolls in the third skill (training-triage-inputs) alongside the unpack and dtype skills from this PR. Closing as no longer needed.

Issue #4465, opened 2026-04-29
[Krea Realtime 14B] VAE decoder : RuntimeError: Value out of range (expected to be in range of [-1, 0], but got -2)


kamalrajkannan78:
### Describe the bug

- Decoder part fails with 

```
E   RuntimeError: Value out of range (expected to be in range of [-1, 0], but got -2)
E   
E   While executing %slice_1 : [num_users=1] = call_function[target=torch.ops.aten.slice.Tensor](args = (%mark_argument_attributes_102, 2, -2, 9223372036854775807), kwargs = {})
E   Original traceback:
E     File "/proj_sw/user_dev/kkannan/apr29_lb51/tt-xla/venv/lib/python3.12/site-packages/diffusers/models/autoencoders/autoencoder_kl_wan.py", line 876, in forward
E       cache_x = x[:, :, -CACHE_T:, :, :].clone()
E     File "/proj_sw/user_dev/kkannan/apr29_lb51/tt-xla/python_package/tt_torch/torch_overrides.py", line 22, in __torch_function__
E       return func(*args, **(kwargs or {}))
E   
E   Use tlparse to see full graph. (https://github.com/pytorch/tlparse?tab=readme-ov-file#tlparse-parse-structured-pt2-logs)
```

### Logs

- [apr29_krea_decoder.log](https://github.com/user-attachments/files/27216147/apr29_krea_decoder.log)

Issue #4464, opened 2026-04-29
[Krea Realtime 14B] Transformer : Complex dtype unsupported on TT (RoPE)


kamalrajkannan78:
### Describe the bug 

- Transformer part fails with 

```
Found an argument on non-XLA device: tensor([[ 1.0000+0.0000e+00j,  1.0000+0.0000e+00j,  1.0000+0.0000e+00j,
          ...,  1.0000+0.0000e+00j,  1.0000+0.0000e+00j,
          1.0000+0.0000e+00j],
        [ 0.5403+8.4147e-01j,  0.7913+6.1148e-01j,  0.9078+4.1948e-01j,
          ...,  1.0000+3.7276e-04j,  1.0000+2.4041e-04j,
          1.0000+1.5505e-04j],
        [-0.4161+9.0930e-01j,  0.2522+9.6768e-01j,  0.6481+7.6158e-01j,
          ...,  1.0000+7.4552e-04j,  1.0000+4.8082e-04j,
          1.0000+3.1010e-04j],
        ...,
        [-0.9998+1.7612e-02j,  0.8520-5.2354e-01j, -0.5420+8.4041e-01j,
          ...,  0.9284+3.7147e-01j,  0.9700+2.4300e-01j,
          0.9875+1.5765e-01j],
        [-0.5550-8.3182e-01j,  0.9943+1.0672e-01j, -0.8445+5.3555e-01j,
          ...,  0.9283+3.7181e-01j,  0.9700+2.4323e-01j,
          0.9875+1.5780e-01j],
        [ 0.4001-9.1649e-01j,  0.7215+6.9244e-01j, -0.9913+1.3190e-01j,
          ...,  0.9282+3.7216e-01j,  0.9699+2.4347e-01j,
          0.9874+1.5795e-01j]], dtype=torch.complex128). Passing a non-XLA tensor to TT compile was likely not intended. Force moving the argument to XLA.
```

### Logs

- [apr29_krea_transformer_sharded.log](https://github.com/user-attachments/files/27216104/apr29_krea_transformer_sharded.log)

PR #4298, opened 2026-04-18
[Wan 2.2] Inital tests for each part of the pipeline

mstojkovicTT:
### Ticket
fixes: https://github.com/tenstorrent/tt-xla/issues/4297

### What's changed

All component tests live in `tests/torch/models/wan2_2/`. Each test file exposes 4 pytest functions following the pattern `test_<component>_<resolution>[_sharded]`:

| File | Component | Tests |
|---|---|---|
| `test_umt5_text_encoder.py` | UMT5-XXL text encoder | `test_umt5_{480p,720p}[_sharded]` |
| `test_vae_encoder.py` | 3D Causal VAE encoder | `test_vae_encoder_{480p,720p}[_sharded]` |
| `test_vae_decoder.py` | 3D Causal VAE decoder | `test_vae_decoder_{480p,720p}[_sharded]` |
| `test_wan_dit.py` | WanDiT (5B transformer) | `test_wan_dit_{480p,720p}[_sharded]` |

**Running:**
```
pytest tests/torch/models/wan2_2/                                    # all 16 tests
pytest tests/torch/models/wan2_2/test_wan_dit.py::test_wan_dit_480p  # one test
pytest tests/torch/models/wan2_2/ -k sharded                         # only sharded variants
pytest tests/torch/models/wan2_2/ -k 720p                            # only 720p variants
```

**Shared infrastructure** (`tests/torch/models/wan2_2/shared.py`):
- `RESOLUTIONS`- dict with 480p and 720p shape configs
- `load_umt5` / `load_vae` / `load_dit`- real-weight loaders
- `setup_mesh`- adaptive 2D `(batch, model)` SPMD mesh (4 / 8 / 32 devices)
- `shard_umt5_weights` / `shard_vae_encoder_weights` / `shard_vae_decoder_weights` / `shard_dit_weights`

PR #4466, opened 2026-04-29
[Krea Realtime 14B] Add initial tests for each part of the pipeline

### LINKED ISSUE #4463: [Krea Realtime 14B] Bringup Initial tests for each part of the pipeline

kamalrajkannan78:
### Ticket

- fixes https://github.com/tenstorrent/tt-xla/issues/4463

### Problem description

- Add initial tests for each part of the pipeline
- Initial target device: Wormhole

### What's changed

 | File | Component | Tests |
|------|-----------|-------|
| `test_text_encoder.py` | UMT5-XXL text encoder | `test_text_encoder` (skip — OOM single device), `test_text_encoder_sharded` (pass — 2 chips ) |
| `test_transformer.py` | CausalWanModel (14B autoregressive DiT) | `test_transformer` (skip — OOM single device), `test_transformer_sharded` (xfail — 4 chips — [#4464](https://github.com/tenstorrent/tt-xla/issues/4464)) |
| `test_vae_decoder.py` | 3D Causal VAE decoder | `test_vae_decoder` (xfail — single chip — [#4465](https://github.com/tenstorrent/tt-xla/issues/4465)) |

Shared infrastructure (`tests/torch/models/krea_realtime/shared.py`):

- `HEIGHT / WIDTH / NUM_BLOCKS / MAX_SEQ_LEN` — default inference shape constants
- `KV_CACHE_NUM_FRAMES / FRAME_SEQ_LENGTH / SEQ_LENGTH / KV_CACHE_SIZE` — KV-cache config constants
- `load_text_encoder / load_transformer / load_vae` — real-weight loaders
- `krea_mesh` — adaptive 2D `(batch, model)` SPMD mesh (1 / 2 / 4 / 8 / 32 devices)
- `shard_text_encoder_specs / shard_transformer_specs` — per-component shard spec functions

Test-specific extras:

- `test_transformer.py` includes `fixed_sinusoidal_embedding_1d`, a runtime monkey-patch that replaces Krea's [CUDA-hardcoded time-embedding helper](https://huggingface.co/krea/krea-realtime-video/blob/main/transformer/model.py#L33) with a device-agnostic version, so the test runs without modifying the upstream package code.
- `test_transformer.py` includes `CausalWanWrapper`, a thin `nn.Module` adapter that exposes `forward(x, t, context)` and rebuilds zero-init KV / cross-attention caches per call (Krea's transformer expects 8 args including `list[dict]` caches; `run_graph_test` only accepts tensor positionals).

### Checklist
- [x] Verify the changes through local testing in WH

### Logs

- [apr29_krea_transformer_sharded.log](https://github.com/user-attachments/files/27215938/apr29_krea_transformer_sharded.log)
- [apr29_krea_decoder.log](https://github.com/user-attachments/files/27215940/apr29_krea_decoder.log)
- [apr29_krea_encoder_sharded.log](https://github.com/user-attachments/files/27215944/apr29_krea_encoder_shard_2_chips.log)

### DIFF
--- a/tests/torch/models/krea_realtime/shared.py
+++ b/tests/torch/models/krea_realtime/shared.py
@@ -1,0 +1,205 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""
+Shared helpers for Krea Realtime 14B component tests.
+
+Exposes:
+  - load_text_encoder / load_transformer / load_vae: HuggingFace model loaders
+  - krea_mesh: 2D ("batch", "model") SPMD mesh — adapts to device count
+  - shard_text_encoder_specs / shard_transformer_specs: per-component shard
+    spec functions returning dict[Tensor, partition_spec] for run_graph_test.
+"""
+
+import torch
+from diffusers import ModularPipeline
+from infra.utilities import Mesh
+from infra.utilities.torch_multichip_utils import get_mesh
+
+# ---------------------------------------------------------------------------
+# Model + dtype
+# ---------------------------------------------------------------------------
+
+REPO_ID = "krea/krea-realtime-video"
+DTYPE = torch.bfloat16
+
+
+# ---------------------------------------------------------------------------
+# Default inference shape
+# ---------------------------------------------------------------------------
+
+HEIGHT = 480
+WIDTH = 832
+NUM_BLOCKS = 1
+MAX_SEQ_LEN = 512  # text token length
+
+LATENT_H = HEIGHT // 8  # 60
+LATENT_W = WIDTH // 8  # 104
+NUM_FRAMES_PER_BLOCK = 3
+NUM_LATENT_FRAMES = NUM_BLOCKS * NUM_FRAMES_PER_BLOCK  # 3
+NUM_CHANNELS_LATENTS = 16
+TEXT_EMBED_DIM = 4096
+
+# KV-cache config
+KV_CACHE_NUM_FRAMES = 3
+FRAME_SEQ_LENGTH = 1560
+SEQ_LENGTH = 32760
+LOCAL_ATTN_SIZE = KV_CACHE_NUM_FRAMES + NUM_FRAMES_PER_BLOCK  # 6
+KV_CACHE_SIZE = LOCAL_ATTN_SIZE * FRAME_SEQ_LENGTH  # 9360
+
+
+# ---------------------------------------------------------------------------
+# Component loaders (real weights from HuggingFace)
+# ---------------------------------------------------------------------------
+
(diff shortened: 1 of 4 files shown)


saiarthiraguram (approved):

tests/torch/models/krea_realtime/shared.py
```diff
@@ -0,0 +1,205 @@
...
+LOCAL_ATTN_SIZE = KV_CACHE_NUM_FRAMES + NUM_FRAMES_PER_BLOCK  # 6
+KV_CACHE_SIZE = LOCAL_ATTN_SIZE * FRAME_SEQ_LENGTH  # 9360
+
+
+# ---------------------------------------------------------------------------
+# Component loaders (real weights from HuggingFace)
+# ---------------------------------------------------------------------------
+
+
+def _load_pipe():
```
ppadjinTT, line 56: i don't like using this way of loading components. for each component to be loaded, we have to load the whole pipeline and select one part. huggingface generally offers selective loading via subfolder keyword arg, e.g. for vae something like:

```
    vae = AutoencoderKL.from_pretrained(
            model_id,
            subfolder="vae",
            torch_dtype=dtype,
            device_map=device,
            trust_remote_code=True, # add this only if necessary
        )
```

Please edit all of these modules that use `_load_pipe` to instead each load only its respective component.

ppadjinTT (changes requested):
Why do we need explicit test files in the tt-xla repo? Is it possible to use tt-forge-models repo for this? If not, why not? It seems all of these tests are torch.nn.Modules themselves

kamalrajkannan78:
> Why do we need explicit test files in the tt-xla repo? Is it possible to use tt-forge-models repo for this? If not, why not? It seems all of these tests are torch.nn.Modules themselves

Yes, I agree with you. I referred to this PR: https://github.com/tenstorrent/tt-xla/pull/4298 to add initial tests for each part of the pipeline. I Will update it in suggested way.
If you prefer, I can move these tests to the tt-forge-models repository.

tests/torch/models/krea_realtime/shared.py
kamalrajkannan78: I referred to this PR: https://github.com/tenstorrent/tt-xla/pull/4298 to add initial tests for each part of the pipeline. I Will update it in suggested way.

ppadjinTT: Yes, unfortunately those were added for some speedrun effort, they will be moved to tt-forge-models as well. Thanks for understanding

kamalrajkannan78:
Below PR's will address the comments 

https://github.com/tenstorrent/tt-forge-models/pull/633
https://github.com/tenstorrent/tt-xla/pull/4513

PR #4490, opened 2026-04-30
[vLLM plugin] Add sharding for RMSNorm layer

### LINKED ISSUE #4489: [vLLM plugin] Add shardings for RMS layer

mmanzoorTT:
### Ticket
closes #4489 

### What's changed
Add shardings for RMSNorm layer

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
+++ b/integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
@@ -11,7 +11,6 @@ import torch.nn.functional as F
 import torch_xla.distributed.spmd as xs
 from torch.nn.parameter import Parameter
 from tt_torch.sharding import sharding_constraint_hook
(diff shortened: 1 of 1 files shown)


integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
```diff
@@ -264,6 +264,13 @@ def partition_vocab_parallel_embedding(
     return layer
```
AleksKnezevic, line 296: This should only apply on 2d, right?

mmanzoorTT: Currently, vLLM is using `(n x 1)` as `("batch", "model")` for 1D mesh; so the weights will be replicated for 1D mesh with this sharding. We will need to fix it once we switch to `(1 x n)` mesh shape.

PR #4503, opened 2026-05-04
Uplift PJRT C API header from v0.106 to v0.110

acicovicTT:
## Description

This PR uplifts the PJRT C API header from the upstream [OpenXLA repository](https://github.com/openxla/xla).

- **Version change:** `v0.106` -> `v0.110`
- **Source file:** https://github.com/openxla/xla/blob/main/xla/pjrt/c/pjrt_c_api.h

## Checklist

- [x] Review the changes in `pjrt_c_api.h` and ensure backward compatibility with the tt-xla implementation.
- [x] Review the changes in `stubs.inc` and discuss with the team whether a new feature should be flagged as useful for implementation.
- [x] Move the new signatures in `stubs.inc` to the most appropriate group in the same file.
- [x] Run PJRT unit tests (scheduled automatically upon PR creation).
- [x] Run the `model-test-extended.json` test matrix against this branch (needs to be run manually) -> https://github.com/tenstorrent/tt-xla/actions/runs/26396639240

### DIFF
--- a/third_party/pjrt_c_api/xla/pjrt/c/pjrt_c_api.h
+++ b/third_party/pjrt_c_api/xla/pjrt/c/pjrt_c_api.h
@@ -85,6 +85,7 @@ typedef enum {
   PJRT_Extension_Type_Collectives,
   PJRT_Extension_Type_MultiSlice,
   PJRT_Extension_Type_HostMemoryAllocator,
+  PJRT_Extension_Type_XlaTransform,
 } PJRT_Extension_Type;
 
 // PJRT_Extension_Base contains a type and a pointer to next
@@ -119,7 +120,7 @@ PJRT_DEFINE_STRUCT_TRAITS(PJRT_Extension_Base, next);
 // Changes include:
 // * Adding a new field to the PJRT_Api or argument structs
 // * Renaming a method or argument (doesn't affect ABI)
-#define PJRT_API_MINOR 106
+#define PJRT_API_MINOR 110
 
 // The plugin should set the major_version and minor_version of
 // PJRT_Api.pjrt_api_version to be the `PJRT_API_MAJOR` and `PJRT_API_MINOR` in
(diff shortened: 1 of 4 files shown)


third_party/pjrt_c_api/xla/pjrt/c/pjrt_c_api.h
```diff
@@ -2907,6 +2928,43 @@ PJRT_DEFINE_STRUCT_TRAITS(PJRT_TopologyDescription_Fingerprint_Args,
...
+  // Output shape (serialized ShapeProto)
+  const char *serialized_shape;                   // out
+  size_t serialized_shape_size;                   // out
+  void (*serialized_shape_deleter)(const char *); // out
+};
+PJRT_DEFINE_STRUCT_TRAITS(
+    PJRT_TopologyDescription_MakeCanonicalShapeForMemorySpace_Args,
+    serialized_shape_deleter);
+
+typedef PJRT_Error *PJRT_TopologyDescription_MakeCanonicalShapeForMemorySpace(
```
mrakitaTT, line 2952: Missed to add stub for this one

third_party/pjrt_c_api/xla/pjrt/c/pjrt_c_api.h
```diff
@@ -2000,8 +2017,12 @@ struct PJRT_ExecuteOptions {
   int *task_ids;
   int64_t *incarnation_ids;
   PJRT_MultiSlice_Config *multi_slice_config;
+  // If true, transpose the array from the device native format into major to
+  // minor format.
+  bool use_major_to_minor_data_layout_for_callbacks;
```
mrakitaTT, line 2022: We should add check to see if this is getting activated from the client, and throw error in case it is because we currently don't support it. Otherwise we might encounter some PCC issues that would be hard to debug and connect to this flag.

third_party/pjrt_c_api/xla/pjrt/c/pjrt_c_api.h
acicovicTT: There was a bug in the workflow that resulted in not catching new signatures that are split across multiple lines. I've fixed it as part of this PR

third_party/pjrt_c_api/xla/pjrt/c/pjrt_c_api.h
acicovicTT: Nice catch. I have added the check.

acicovicTT:
Temporarily disabled the workflow so that changes don't get overwritten.

mrakitaTT (approved):

vvukomanTT (approved):

PR #4511, opened 2026-05-04
[vLLM plugin] Cleanly shut down after a failed test to prevent device hang

### LINKED ISSUE #4510: [vLLM plugin] Assertion Failures Cause vLLM Tests to Hang
vLLM does not close properly/completely when we assert in a test. When such scenario occurs; the test framework leaves TT devices in a corrupted state, causing subsequent tests to hang indefinitely.

mmanzoorTT:
### Ticket
closes #4510 

### Problem description
Assertion Failures Cause vLLM Tests to Hang

### What's changed
Replace assert statements with pytest.fail() in vLLM plugin tests.

### Checklist

### DIFF
--- a/tests/integrations/vllm_plugin/conftest.py
+++ b/tests/integrations/vllm_plugin/conftest.py
@@ -1,0 +1,30 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Shared cleanup for vLLM-plugin tests.
+
+vLLM has no public LLM.shutdown(); engine teardown runs via
+weakref.finalize when the LLM becomes unreachable. On a failed test,
+pytest's report pins the test-frame locals (incl. `llm`) via the live
+traceback, so finalize never fires and the EngineCore subprocess
+holding the TT device outlives the test — hanging the next one. Shut
+them down explicitly on failure.
+"""
+import gc
+
+import pytest
+import vllm
+
+
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    outcome = yield
+    report = outcome.get_result()
+    if report.when != "call" or not report.failed:
+        return
+    for obj in gc.get_objects():
+        if isinstance(obj, vllm.LLM):
+            try:
+                obj.llm_engine.engine_core.shutdown()
+            except Exception:
+                pass
--- a/tests/integrations/vllm_plugin/generative/conftest.py
+++ b/tests/integrations/vllm_plugin/generative/conftest.py
@@ -5,6 +5,7 @@
 import re
 
 import psutil
+import pytest
 
 # Common English function words used by `assert_output_coherent` to detect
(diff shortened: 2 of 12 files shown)


AleksKnezevic:
@mmanzoorTT do asserts cause subsequent tests to hang? Or do we not recover cleanly from the assert and shut down the test? 

I don't like replacing with pytest.fail because our plugin should be callable from outside the scope of pytest in a production environment.

kmabeeTT:
For @AleksKnezevic comment, I mean these **are** pytests being modified and not plugin code, so don't think it's too real of a concern?

But, I don't love the change here and like to be able to use assert, here is an alternative I tested locally which solves problem when middle test asserts and allows subsequent tests to run fine and pass - can we consider it instead @mmanzoorTT ?

https://github.com/tenstorrent/tt-xla/commit/be7e84f137d8bf822128ce5ac048701754843486

Some details on what claude found about this problem, and above alternative fix:

## The problem

When a vLLM test fails with a bare `assert`, pytest grabs the failing exception and saves it for the failure report. That saved exception keeps a chain of references alive: traceback → test frame → local variables → the `llm` object → its EngineCore subprocess.

vLLM never gives you a `shutdown()` to call yourself. It relies on `weakref.finalize`, which only runs cleanup when the LLM is *unreachable*. But the report is still holding a reference, so the LLM stays reachable, finalize never fires, and the EngineCore subprocess (which owns the TT device) stays alive.

The next test starts, tries to claim the device — and hangs forever, because the previous test's worker is still sitting on it.

## Why `pytest.fail` works but `assert` doesn't

Same pin, different escape route. Somewhere in vLLM's cleanup path there's a `try: ... except Exception:` that catches `AssertionError` (which inherits from `Exception`) and silently skips the worker shutdown. `pytest.fail()` raises `Failed`, which inherits from `BaseException` — not `Exception` — so it slips past that handler and the cleanup runs normally.

## How the alternative fix bypasses it

Stop relying on the LLM becoming unreachable. After any failed test, walk every live Python object, find the `vllm.LLM` instances, and call `engine_core.shutdown()` on them directly. The subprocess dies, the device is freed, the next test runs cleanly. We don't care who's still holding references — we just kill the worker outright.

mmanzoorTT:
@AleksKnezevic @kmabeeTT `assert` terminates the test abruptly and the signal is not passed to vLLM process (or vLLM spawns new processes and signal is not passed to correct process). So, vLLM process is still running and holding the device. Subsequent test is waiting for the device causing CI to timeout. pytest.fail terminates process gracefully and pass terminate signal to vLLM ensuring completion of the vLLM process.

I have tried handling signal to process assert and shutdown gracefully to have a generic solution but we are not receiving the signal. The other option can be to handle `assert` in `conftest.py` for vLLM tests but it will also be specific to tests. 

Edit
I have cherry-picked Kyle's changes for `conftest.py`. This fixes the issue for the tests.

kmabeeTT (approved):
I think this is a good single clean workaround for now and we should ship it

AleksKnezevic (approved):

PR #4528, opened 2026-05-05
Allow overriding tt-mlir and tt-metal source dirs in third_party build

nsmithtt:
Add TTMLIR_SOURCE_DIR_OVERRIDE and TTMLIR_TTMETAL_SOURCE_DIR cache variables so cmake can use existing local checkouts instead of having ExternalProject_Add clone them. When the tt-mlir override is set, the canonical third_party/tt-mlir/src/tt-mlir path is symlinked to the override so scripts and CI workflows that hardcode that path (venv activate, install_ttmlir_requirements.sh, bisect_*.sh, tracy include path) keep working unchanged.

Primary motivation is making cross-project iteration easier -- especially for coding agents making coordinated changes across tt-xla, tt-mlir, and tt-metal. With the overrides, an agent can edit all three repos in place, point tt-xla at the local checkouts, and rebuild without each repo needing its own clone-and-rebuild cycle nested inside ExternalProject_Add.


nsmithtt:
An equivalent change being made to tt-mlir:
- https://github.com/tenstorrent/tt-mlir/pull/8211

sgligorijevicTT (approved):

PR #4535, opened 2026-05-05
Default batch_size=2 for training in test_all_models_torch

agobeljicTT:
### Ticket
N/A

### Problem description
`test_all_models_torch` parametrizes `run_mode` over `inference` and `training`, but training-mode runs use whatever batch size the underlying loader hardcodes (typically 1). There was no centralized way to bump the training batch size without editing each loader or each test's YAML entry.

Two issues blocked a clean default:
1. `TorchDynamicLoader.load_inputs` default path silently dropped `batch_size` — only the `LLM_DECODE`/`LLM_PREFILL` branches forwarded it. So even setting `batch_size: N` in YAML had no effect on regular (non-LLM) torch tests.
2. `DynamicTorchModelTester._get_input_activations` only used `batch_size` from `test_metadata`, with no run-mode-aware default.

This was noticed while triaging `ssd300_resnet50/pytorch-Base-single_device-training`:
```
In training mode an internal BatchNorm operates on a [1, 256, 1, 1]
tensor and rejects batch=1; bump to 2 only when the model is in
training mode so the inference test (EXPECTED_PASSING) keeps batch=1.
```

### What's changed
- `tests/runner/utils/dynamic_loader.py`: default branch of `TorchDynamicLoader.load_inputs` now forwards `batch_size` to the underlying loader if its signature accepts it (mirrors the existing pattern in the LLM branches).
- `tests/runner/testers/torch/dynamic_torch_model_tester.py`: when `run_mode == RunMode.TRAINING` and YAML didn't specify `batch_size`, default to 2.

YAML entries can still override via `batch_size: <n>`. Loaders without a `batch_size` kwarg are unaffected (`inspect`-based forwarding ignores them).

### Checklist

### DIFF
--- a/tests/runner/testers/torch/dynamic_torch_model_tester.py
+++ b/tests/runner/testers/torch/dynamic_torch_model_tester.py
@@ -133,6 +133,8 @@ class DynamicTorchModelTester(TorchModelTester):
             if self._test_metadata
             else None
         )
+        if batch_size is None and self._run_mode == RunMode.TRAINING:
+            batch_size = 2
 
         inputs = self.dynamic_loader.load_inputs(
             run_phase=self.run_phase,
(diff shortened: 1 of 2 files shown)


sgligorijevicTT (approved):

ndrakulicTT (approved):
LGTM

PR #4334, opened 2026-04-21
[vLLM] Replace sort-based sampling with multi-core topk for 2x non-greedy speedup

kmabeeTT:
## Ticket

#3940

## Problem

- Non-greedy (temperature > 0) vLLM sampling was slow: `apply_top_k_top_p` sorted the full probability distribution (`probs.sort()`) over the entire vocabulary (128k tokens for Llama). This compiled to single-core `SortDeviceOperation` at ~30ms/token, making non-greedy roughly 3x slower than greedy.

## What's changed

- `sampler.py`: replace `apply_top_k_top_p` (full-vocab sort) with `apply_top_k_top_p_fast` — splits vocab into power-of-2 chunks (≤ 32768), runs `torch.topk(k=32)` per chunk, and Gumbel-max samples from the ~128 resulting candidates. Compiles to multi-core `ttnn.topk` (~0.18ms/chunk) instead of single-core `SortDeviceOperation` (~30ms)
- `test_topk.py`: add `test_topk_vllm_sampling_shapes` covering power-of-2 chunk sizes (16384/32768/65536, k=32/64) with correct value-based comparator
-  Side effect: workarounds #4422 (`ttnn.sort` aborts under metal-trace capture on Wormhole) by removing full-vocab sort from the sampler graph.

## Checklist

- [x] Non-greedy output coherent on Llama 3.1-8B and 3.2-1B
- [x] Greedy throughput unchanged, Non-greedy 2x+ speedup vs sort baseline
- [x] Used this change for recent single chip + multichip (gemma4-31B-it) demos
- [x] vLLM Nightly at TOTT passing (except for known hang in n300-llmbox): https://github.com/tenstorrent/tt-xla/actions/runs/25337040803
- [x] tenstorrent/tt-mlir#7773 uplifted into tt-xla
- [x] Add composite op support to torch.gather via https://github.com/tenstorrent/tt-xla/pull/4373

Issue #3940, opened 2026-03-25
vLLM on-device sampling: performance optimization


kmabeeTT:
Tracking issue for optimizing on-device sampling performance in vLLM on Tenstorrent hardware. Individual issues will be opened as specific areas are investigated.

Non-greedy device sampling is currently ~2x slower than CPU sampling at batch=1:

| Configuration | OPT-125M tok/s | Llama-3.1-8B tok/s |
|---|---|---|
| Greedy device | 11.61 | 9.82 |
| Non-greedy device | 5.94 | 4.02 |
| Non-greedy CPU | 11.00 | 7.89 |

Investigation will start with IR analysis and Tracy profiling to identify the dominant ops in the non-greedy sampling graph.

Branch: `kmabee/vllm_perf_debug`

kmabeeTT:
A bit of an overdue update here (a few things being worked on in parallel)... Trace support for llama through VLLM not originally working, but working with some work via https://github.com/tenstorrent/tt-xla/issues/4220 that helps here a bit.

## Update — April 12, 2026

Branch: `kmabee/vllm_perf_apr12` based on `866d0abdd` (latest tt-xla from Apr 12)
Hardware: P150 Blackhole (qb2), Llama-3.1-8B-Instruct, seq_len=128, bfp_bf8

### Baseline (no fix, with metal trace fixes)

| Config | Batch=1 | Batch=32 |
|---|---|---|
| Greedy device | 22.2 tok/s | 14.2 tok/s |
| Non-greedy device | **5.1 tok/s** | **4.5 tok/s** |
| Non-greedy CPU | 19.9 tok/s | 6.9 tok/s |

Non-greedy device sampling is **4.4x slower** than greedy. CPU sampling is 4x faster than device for non-greedy at batch=1.

### Root cause (Tracy profiling)

The compiled non-greedy sampling graph adds ~60ms/token from:
- `SortDeviceOperation` (30ms, 110 cores) — full vocab sort for top-k/top-p
- `AccumulationDeviceOperation` (24ms, **1 core**) — cumsum for top-p
- `FillPadDeviceOperation` ×11 (14ms, **1 core**)

### Fix: replace sort with multi-core topk

[...]

Issue #4494, opened 2026-05-01
move topk chunking for sampling into tt-mlir


AleksKnezevic:
Sampling code currently chunks topk inputs into 32k slices so that they can be processed faster. This is very inefficient for multichip because it has logits all-gathered only to be re-sliced and processed sequentially. 

This logic should be moved into a ttnn WA so that logits can stay split and the all-gathering can happen on tiny top-k tensors. On gemma4 on qb2 this would save ~2ms. Galaxy savings will be much higher.

kmabeeTT:
Good idea, will pursue - just for reference, the topk chunking into 32k was used for demo, but didn't get merged to tt-xla yet (had some dependencies), it's PR is here and it gave 2.5x tok/sec for llama3.1-8b single chip model: https://github.com/tenstorrent/tt-xla/pull/4334

Edit May 13: Last of the previously worked on sampler improvements merged via #4536 and now is good time to evaluate this ticket.

alinakhanTT:
Change
Moves the top-k chunking into a ttnn workaround so logits stay split — the all-gather then happens on the tiny top-k tensors instead of the full-vocab logits. Each shard is padded to a power-of-2 and the per-shard k is capped so the multi-core top-k kernel engages.

PRs
 - [tt-mlir: #8601](https://github.com/tenstorrent/tt-mlir/pull/8601/)
 - [tt-xla: #5004](https://github.com/tenstorrent/tt-xla/pull/5004) 
 
Results
Measured on galaxy: the top-k op dropped from 14.7ms → 0.68ms, PCC ~1.0. Expected ~2ms savings on gemma4 on qb2, with galaxy savings substantially higher.

E2E impact
End-to-end throughput is roughly flat for now, at these batch/seq sizes the decode is host/dispatch-bound, so the device-side savings don't yet surface in tokens/sec (confirmed with before/after benchmarks on 1B and 70B). This lands as a correctness + device-efficiency improvement (vocab-sharded sampling, no full-vocab all-gather).

Next step
The tt.sampling multinomial kernel is now the largest remaining cost in the sampling step and is the right target for the next real sampling speedup.

PR #4536, opened 2026-05-05
[vLLM + Benchmark] 3 perf improvements: ttnn.sampling fused op, pad-batch-to-32, skip greedy on all_random

kmabeeTT:
## Ticket

Main: #4559 
Related: #3940 #4539 #4560 #4570

## Problem description

- vLLM non-greedy device-sampling on Llama has been the dominant per-step cost for vLLM demos. #4334 replaced single-core `ttnn.sort` with a multi-core chunked-topk pre-filter (5 → 21 tok/s on Llama-3.2-1B at b=2), but the remaining tail still ran a `softmax → Gumbel-max → gather` chain on the candidate set every step, the topk pre-filter ran at the user's batch size despite the kernel being ~14× faster at b=32, and `Sampler.sample()` did an unconditional argmax over the full vocab even when every row was non-greedy.

- This PR adds a fused `tt::sampling` kernel covering softmax + top-k + top-p + multinomial; pads the topk input batch to 32; and skips the dead `greedy_sample` + final `torch.where` when every row is sampling. With `enable_trace=True`, net on Llama-3.2-1B nongreedy-device: 19.94 → 42.62 tok/s (2.14×); on Llama-3.1-8B nongreedy-device: 12.19 → 17.16 tok/s (1.41×). Greedy and CPU-sampling paths unchanged within ±2% (and benefit from trace separately). See follow-up comment for the full perf table and TP validation results.

## What's changed

*   `python_package/tt_torch/custom_ops.py`: register `tt::sampling` as a StableHLO custom_call with a CPU multinomial reference for op-level testing. Replaces the prior softmax + Gumbel-max + gather chain with a single fused kernel call on the small candidate set (~128 tokens).
*   `integrations/vllm_plugin/vllm_tt/sampler.py`:
    *   Add `_ttnn_sampling_padded` helper: pads inputs to batch=32 (kernel requirement), invokes `tt::sampling`, slices the result back to `[batch]` and casts to int64.
    *   Rewrite `Sampler.sample()` to use the fused path; gate `greedy_sample` + the final `torch.where` on `if not sampling_metadata.all_random:` so torch.compile specializes the branch away on homogeneous-random batches. Mixed batches still take the slow path correctly.
    *   Rename `apply_top_k_top_p_fast` → `chunked_topk_candidates` and drop the now-dead `k`/`p` parameters + masking block (the fused kernel re-applies user top-k/top-p internally).
    *   Pad logits batch to 32 inside `chunked_topk_candidates` before the chunked topk; trim outputs back to `[:batch]`. Multi-core topk is ~14× faster at b=32 than at b=1 (1.08 ms vs 14.85 ms measured); dummy `-inf` rows cost nothing.
*   `tests/torch/ops/test_topk.py`: docstring + comment updates for the function rename.
*   `tests/benchmark/test_vllm_benchmarks.py`, `tests/benchmark/benchmarks/vllm_benchmark.py`: add Llama-3.2-1B and Llama-3.1-8B entries to `SINGLE_DEVICE_CONFIGS` at b=1 and b=32 (mirrors existing 3B shape). Sampling overrides are env-var driven so CI can re-run the same matrix with different configs without test fan-out (`TT_BENCHMARK_TEMPERATURE`, `TT_BENCHMARK_CPU_SAMPLING`). Default `enable_trace=True` (matches `DEFAULT_TRACE_ENABLED=True` from `test_llms.py`); flips the `cpu_sampling` default to `False` so benchmarks exercise the device sampler by default.
*   `tests/integrations/vllm_plugin/sampling/test_device_sampling.py`: new device-sampler correctness suite mirroring `test_cpu_sampling.py`. Five parameter knobs (greedy, temperature, top_p, top_k, repetition_penalty) parametrized across `single_device` / `n300` / `n300_llmbox` targets via `for_targets()` — 15 nightly cases covering the `chunked_topk_candidates → _ttnn_sampling_padded → tt::sampling` chain under metal trace at TP=1, TP=2, and TP=4+.
  * **Caveat workarounds and fail-fast guards** (see "Known caveats" in follow up comment):
      * `integrations/vllm_plugin/vllm_tt/platform.py`: `TTPlatform.validate_request()` rejects `SamplingParams(seed=...)` (#4539); `TTConfig.__post_init__` rejects `enable_trace=True + optimization_level>=1` at engine init (#4570).
      * `integrations/vllm_plugin/vllm_tt/sampler.py`: `chunked_topk_candidates` pads W with `-inf` so `Wt = W/32` is always a power of 2 and `>= 2` (#4560).
      * Test changes: new `test_non_greedy_unsupported_wt[wt1_1chunk/wt3_3chunks/wt6_6chunks]` (#4560 regression lock); `xfail(strict=True)` on `test_seed` + synthetic `test_seed_*` tests (#4539), `test_opt125m_trace_logprobs` and `test_llama3_3b_generation_trace` (#4570); new `test_llama3_3b_generation_trace_opt0` preserves 3B-trace nightly coverage at opt=0.

## Checklist
* [x] vLLM Nightly: https://github.com/tenstorrent/tt-xla/actions/runs/25479192132
* [x] Used these sampler changes locally for tt-deploy demos last week for single chip + gemma4-31b-it multichip for perf
* [x] Unit / integration tests added
* [x] Benchmark numbers verified locally (posted as follow-up comment)
* [x] TP validated on a 4-chip Blackhole (posted as follow-up comment)

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/sampler.py
+++ b/integrations/vllm_plugin/vllm_tt/sampler.py
@@ -404,4 +404,15 @@ def chunked_topk_candidates(
     # Concat: [batch, num_chunks * k_per_chunk]
     all_values = torch.cat(topk_values_list, dim=-1)
     all_indices = torch.cat(topk_indices_list, dim=-1)
+
+    # Pad W so that Wt = W/32 is a power of 2 AND Wt >= 2 to avoid the
+    # tt::sampling kernel hang (#4560). -inf values can't win the topk /
+    # multinomial draw, so output is unchanged.
+    cur_w = all_values.shape[-1]
+    target_w = max(64, _next_power_of_2(cur_w))
+    if cur_w < target_w:
+        pad = target_w - cur_w
+        all_values = torch.nn.functional.pad(all_values, (0, pad), value=float("-inf"))
+        all_indices = torch.nn.functional.pad(all_indices, (0, pad), value=0)
+
     return all_values[:batch], all_indices[:batch]
--- a/integrations/vllm_plugin/vllm_tt/sampler.py
+++ b/integrations/vllm_plugin/vllm_tt/sampler.py
@@ -31,6 +31,7 @@ def _get_topk_split_params(vocab_size: int) -> tuple[int, int]:
 
 
 _SAMPLING_EPS = 1e-5
+_TTNN_SAMPLING_BATCH_SIZE = 32  # ttnn.sampling kernel requires batch=32
 
 
 def count_tokens_ge(logprobs: torch.Tensor, threshold: torch.Tensor) -> torch.Tensor:
@@ -198,47 +199,18 @@ class Sampler(nn.Module):
 
         greedy_sampled = self.greedy_sample(logits)
 
-        # Apply temperature.
-        logits = self.apply_temperature(
-            logits, sampling_metadata.temperature, sampling_metadata.all_random
+        # Build the candidate set via chunked multi-core topk. The fused
+        # tt::sampling kernel applies user top-k, top-p, softmax, and
+        # multinomial downstream — no need to filter twice here.
+        filtered_logits, candidate_indices = chunked_topk_candidates(logits)
+        random_sampled = self._ttnn_sampling_padded(
+            filtered_logits, candidate_indices, sampling_metadata
         )
-
-        # Apply min_p.
-        if sampling_metadata.min_p is not None:
-            logits = self.apply_min_p(logits, sampling_metadata.min_p)
-
-        # Apply top_k and/or top_p using multi-core topk pre-filtering.
-        # This splits vocab into power-of-2 chunks (<= 32768) to trigger
-        # multi-core ttnn.topk (0.18ms per chunk vs 9ms single-core), then
-        # applies top-k/top-p on the small candidate set.
-        filtered_logits, candidate_indices = apply_top_k_top_p_fast(
-            logits,
-            sampling_metadata.top_k,
-            sampling_metadata.top_p,
-        )
-
-        # Random sample on reduced candidate set.
-        probs = filtered_logits.softmax(dim=-1, dtype=torch.float32)
-
-        # If seeded sampling, gather q_samples at candidate positions.
-        q_samples_reduced = None
-        if sampling_metadata.q_samples is not None:
-            q_samples_reduced = sampling_metadata.q_samples.gather(1, candidate_indices)
-
-        random_sampled_local = self.random_sample(
-            probs, sampling_metadata.generators, q_samples_reduced
-        )
-        # Map local candidate index back to global vocab index.
-        random_sampled = candidate_indices.gather(
-            1, random_sampled_local.unsqueeze(-1)
-        ).squeeze(-1)
-
-        sampled = torch.where(
+        return torch.where(
             sampling_metadata.temperature < _SAMPLING_EPS,
             greedy_sampled,
             random_sampled,
         )
-        return sampled
 
     def compute_logprobs(self, logits: torch.Tensor) -> torch.Tensor:
         return logits.log_softmax(dim=-1, dtype=torch.float32)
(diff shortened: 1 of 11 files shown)


kmabeeTT:
Extra follow up information useful for reference/review.

- I'm including benchmark test updates since I'd been using them locally and they were highly useful in testing these and prior sampler improverment PR changes beyond just existing llama3.2-3B model (especially beinng able to tweak sampler settings).. one less thing to maintain locally as sampler is continued to be improved.

## Detailed perf table

Pre-PR baseline below is with #4334 (chunked-topk only) at `optimization_level=0, enable_trace=False` (the only state pre-PR could run under without the trace+ttnn.sampling earlier blocker). The two "This PR" columns isolate the sampler-changes-only delta (trace=False) from the additional trace boost newly available now that the device-sampler+trace path is unblocked.

| Model | Path | Pre-PR (trace=False) | This PR (trace=False) | This PR (trace=True) |
|---|---|---:|---:|---:|
| 1B | greedy-device     | 61.12 | 61.34 | 77.91 |
| 1B | **nongreedy-device** | **19.94** | **37.70 (1.89×)** | **42.62 (2.14×)** |
| 1B | greedy-cpu        | 47.68 | 48.37 | 62.11 |
| 1B | nongreedy-cpu     | 46.02 | 46.10 | 59.70 |
| 8B | greedy-device     | 20.54 | 20.39 | 21.32 |
| 8B | **nongreedy-device** | **12.19** | **16.51 (1.35×)** | **17.16 (1.41×)** |
| 8B | greedy-cpu        | 18.86 | 18.90 | 19.22 |
| 8B | nongreedy-cpu     | 18.98 | 18.55 | 19.13 |

The trace=True column reflects the recommended production setting now that the trace + tt::sampling combination works (see "Note on metal trace" below). The trace boost is largest on small-model paths where the per-step launch overhead dominates (1B greedy/cpu paths gain 25–30%); on the 8B nongreedy bottleneck it's a smaller but still useful 4%.

In tracy at b=32 (sampler standalone, no forward), commit 3's `all_random` gate alone is **3.04 → 1.85 ms (1.65×)** — removes `ArgMaxDeviceOperation` (1.03 ms = 34% of sampler) and its downstream `Tilize`.

Tests run at `optimization_level=0` while the separate multicore-topk + opt=1 garbage bug is being fixed. Sampler logic is opt-agnostic.

## TP validation

Confirmed on a 4-chip Blackhole p150 system. With logits replicated via `compute_logits`'s `(None, None)` sharding constraint, the sampler is functionally single-chip math; remaining concern was SPMD partitioning of the new `tt::sampling` custom op + cross-device hardware-RNG state. Verified manually during PR development across greedy, non-greedy, and 2-prompt-batch non-greedy on Llama-3.2-3B at TP=4 — all produced coherent English output with no cross-device divergence and no engine-init failures.

`tests/integrations/vllm_plugin/sampling/test_device_sampling.py` provides the regression coverage going forward: five sampling knobs × {`single_device`, `n300`, `n300_llmbox`} = 15 nightly cases that exercise the device sampler at TP=1, TP=2, and TP=4+.

The pre-existing `test_tensor_parallel_generation_n300` (the `dual_chip` regression) also runs to completion under both `use_2d_mesh=True/False` shapes (124 s total).

## Skipped optimization (intentional)

A symmetric `all_greedy` early-return mirroring commit 3's gate could skip `chunked_topk_candidates` + `_ttnn_sampling_padded` when every row is greedy. Not added because `model_runner.sample_from_logits` already short-circuits to `torch.argmax` whenever the batch is all-greedy + has no penalties / logit_bias / bad_words / allowed_token_ids / min_tokens / generators — i.e., the common greedy case never enters `Sampler.sample` at all. Worth adding only if a future workload triggers `Sampler.sample` with `all_greedy=True` (greedy + some auxiliary feature like `allowed_token_ids`).

## Known caveats found during qualification

  Three sub-issues surfaced while qualifying this PR — all kernel- or compiler-side, all guarded with caller-side workarounds here. See umbrella #4559 for details.  Will follow up with tt-metal and tt-mlir issues next.

  - **#4539** — `tt::sampling` kernel uses a single seed across all 32 cores, breaking per-row seeded reproducibility. `TTPlatform.validate_request()` rejects `SamplingParams(seed=...)` with a `ValueError` pointing at `cpu_sampling=True`. `test_seed` and the synthetic `test_seed_*` tests are `xfail(strict=True)`.

  - **#4560** — `tt::sampling` hangs the device when the candidate tile count `Wt = W/32` is `1` (TinyLlama) or not a power of 2 (Qwen3-32B → Wt=5, gpt-oss-120b → Wt=7). `chunked_topk_candidates` pads W with `-inf` so `Wt` is always a power of 2 and at least 2. New `test_non_greedy_unsupported_wt` cases lock the workaround in.

  - **#4570** — `enable_trace=True + optimization_level>=1` triggers a tt-mlir compile-time bug: `TTNNGreedyMemoryLayoutPropagation` falls back to `system_memory` on the sampling op output, breaking trace capture. `TTConfig.__post_init__` rejects the combo at engine init and lists three workarounds (`optimization_level=0`, `cpu_sampling=True`, or
  `enable_trace=False`). `test_opt125m_trace_logprobs` and `test_llama3_3b_generation_trace` are `xfail(strict=True)`; new `test_llama3_3b_generation_trace_opt0` preserves 3B-trace nightly coverage at opt=0.

kmabeeTT:
Okay, there is high chance that I got everything in place here now after a few surprises along the way - going to mark ready for review :)

AleksKnezevic (approved):

mmanzoorTT (approved):

kmabeeTT:
Extra datpoint here about vllm benchmark suite changes to show that when switching to greedy sampling on device (switching `cpu_sampling=True` to be `cpu_sampling=False)` we don't regress but improve the existing 3B test:

  `pytest tests/benchmark/test_vllm_benchmarks.py` at pre-PR main (`38f0f7ef1`) vs PR tip (`fa134dcdd`):

  | test_id | pre-PR | tip | Δ |
  |---|---:|---:|---:|
  | `llama-3.2-1b` | — | 62.87 | new |
  | `llama-3.2-1b-batch32` | — | 40.76 | new |
  | `llama-3.2-3b` | 25.60 | 28.63 | 1.12× |
  | `llama-3.2-3b-batch32` | 16.90 | 18.69 | 1.11× |
  | `llama-3.1-8b` | — | 13.98 | new |
  | `llama-3.1-8b-batch32` | — | 9.88 | new |

Existing 3B coverage gains 11–12%; four new test variants added by the PR scaffolding (1B and 8B at b=1 and b=32). No regressions.

(8B added since it's more importnat workload, 1B added because it can be used easier/quicker and show sampler slowdown more drastically than bigger variants due to it's smaller device runtime).

kmabeeTT:
@odjuricicTT @bmalesevicTT @mvasiljevicTT @rpavlovicTT @vkovacevicTT may I please have review for benchmark suite additions, see my previous comment for some more context.

integrations/vllm_plugin/vllm_tt/sampler.py
```diff
@@ -325,25 +304,88 @@ def random_sample(
...
+        # Pad batch to 32 (kernel requirement).
+        if batch < _TTNN_SAMPLING_BATCH_SIZE:
+            pad_size = _TTNN_SAMPLING_BATCH_SIZE - batch
+            values = torch.nn.functional.pad(
+                values, (0, 0, 0, pad_size), value=float("-inf")
+            )
+            indices = torch.nn.functional.pad(indices, (0, 0, 0, pad_size))
+            k_tensor = torch.nn.functional.pad(k_tensor, (0, pad_size), value=1)
+            p_tensor = torch.nn.functional.pad(p_tensor, (0, pad_size), value=1.0)
+            temp_tensor = torch.nn.functional.pad(temp_tensor, (0, pad_size), value=1.0)
```
odjuricicTT, line 359: Can you open an issue for us to explore having this as some type of workaround on tt-mlir level instead of here?

odjuricicTT (approved):

integrations/vllm_plugin/vllm_tt/sampler.py
AleksKnezevic: https://github.com/tenstorrent/tt-xla/issues/4494

PR #4537, opened 2026-05-05
sparse_mlp: pad batch to tile-align tokens for bsz<32 support

sshonTT:
### Ticket
N/A

### Problem description
Downstream MoE ops in A2aSparseMLP.forward — moe_expert_token_remap, sparse_matmul, all_to_all_dispatch — assume the (batch * seq_len) token count is a multiple of TILE (32). Small-batch decode (e.g. bsz=8 seq_len=1 -> 8 tokens) violates this and either segfaults or fails op validation.

### What's changed
Pad batch dim by F.pad before dispatch so total tokens are tile-aligned. Padded entries get zero router scores/indices, run through dispatch / sparse_matmul / combine alongside real tokens, and are sliced off the final output via output[:batch_size] before return so callers see the original [batch_size, seq_len, hidden] shape unchanged.

### Checklist

### DIFF
--- a/python_package/tt_torch/sparse_mlp.py
+++ b/python_package/tt_torch/sparse_mlp.py
@@ -15,6 +15,7 @@ Usage:
     model = enable_sparse_mlp(model)  # Replace MLP layers with SparseMLP
 """
 
+import math
 from typing import Any, Dict, List, Optional, Type
 
 import torch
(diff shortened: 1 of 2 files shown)


python_package/tt_torch/sparse_mlp.py
```diff
@@ -526,6 +526,30 @@ def forward(self, hidden_states):
...
+        # otherwise — typical trigger is small-batch decode (e.g. bsz=8
+        # seq_len=1 → 8 tokens; pad batch by 4× → 32 tokens).
+        # The padded entries get zero router scores/indices, get processed
+        # alongside real tokens through dispatch/sparse_matmul/combine, and
+        # are sliced off the final output before return.
+        TILE = 32
+        total_tokens = batch_size * seq_len
+        pad_batch = 0
+        if total_tokens % TILE != 0:
+            pad_tokens = (-total_tokens) % TILE
```
jameszianxuTT: is it normal to write this pattern as negative mod? I think ceil is easier to understand, but a very minor nit

python_package/tt_torch/sparse_mlp.py
```diff
@@ -526,6 +526,30 @@ def forward(self, hidden_states):
...
+            assert pad_tokens % seq_len == 0, (
+                f"Cannot pad batch to tile-align tokens: pad_tokens={pad_tokens} "
+                f"not divisible by seq_len={seq_len}."
+            )
```
jameszianxuTT: I don't think think this calculation handles it exactly, but couldn't we always pad batch so that batch x seqlen is a multiple of tile size?

eg. batch = 5, seqlen =5 => 25

in worst case, we can pad batch to TILE x batch, but in practise it could probably result in some extreme DRAM waste

jameszianxuTT: Not really asking for a change here, just confused that this calculation for pad_batch does what it says it does

python_package/tt_torch/sparse_mlp.py
sshonTT: math.ceil reads bette. switching.

python_package/tt_torch/sparse_mlp.py
sshonTT: Good catch assert misses cases like seq_len=5. Switching to direct pad_batch calc:
```
batch_step = TILE // math.gcd(seq_len, TILE)
padded_batch = math.ceil(batch_size / batch_step) * batch_step
```
Hot path is seq_len=1 so it worked, but this handles all seq_len. Pushing fix.

AleksKnezevic (approved):

Issue #3927, opened 2026-03-25
Run GLM 4.7 and deepseek on BH Galaxy


AleksKnezevic:

btsoiTT:
Updates on progress for DeepSeek V3.1 / GLM-4.7 Galaxy end-to-end work

Branch [btsoi/deepseek_glm_galaxy_e2e_cleaned](https://github.com/tenstorrent/tt-xla/tree/btsoi/deepseek_glm_galaxy_e2e_cleaned) ships four tests - a prefill-only `*_full_sparse_moe` and an autoregressive `*_decode_static_cache` for each of DeepSeek V3.1 and GLM-4.7. All four share: batch 32, 32-token prompt, Galaxy (4,8) mesh.

## Prerequisites

1. Access to a Galaxy machine.
2. Confirm there is enough disk storage for the weights and swap
3. Set up swap. The full-layer DeepSeek V3.1 / GLM-4.7 runs peak well above system RAM during weight dequant + expert stacking, so a large swapfile is required to avoid OOM (1.5TB is probably enough, but more doesn't hurt):

   ```bash
   # To enable swap

   # only if the file system is btrfs
   sudo truncate -s 0 ./swapfile
   sudo chattr +C ./swapfile

   sudo fallocate -l 3TB ./swapfile
   sudo chmod 600 ./swapfile
   sudo mkswap ./swapfile
   sudo swapon ./swapfile


   # To disable swap

   sudo swapoff ./swapfile
   sudo rm ./swapfile
   ```

Note that this needs to be run outside of the container.

## Weight caches

To reduce runtime peak memory usage, a weight caching mechanism is implemented.  Both tests can load weights from a pre-built on-disk cache instead of re-running the heavy preprocessing each launch. There are two flavors:

- **Dequant cache** - DeepSeek weights ship as FP8 on HF but the model runs in BF16; dequantizing the full checkpoint takes a long time and the FP8 weights stick around which increases memory pressure, so we do it once and cache the BF16 result. GLM-4.7 is already BF16, so no dequant step is needed.
- **Post-sparse cache** - To run both models, we call `enable_sparse_mlp` to use `sparse_matmul`. The function stacks MoE experts as a single `[E, in, out]` tensor. The post- sparse cache caches this tensor to further reduce the memory consumption and mmap it to reduce disk write pressure during swapping.

[...]

PR #4538, opened 2026-05-05
Add 4 layer deepseek and glm tests with weight caching

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3927

### Problem description
We want to run deepseek v3 and glm 4.7 on galaxy end-to-end.

### What's changed
@btsoiTT added `test_deepseek_v3_1_decode_static_cache` and `test_glm4_7_decode_static_cache` tests that use a weight caching mechanism to reduce runtime peak memory usage. This also allows weights to be cached and re-used locally. More details about these changes can be found [in this comment](https://github.com/tenstorrent/tt-xla/issues/3927#issuecomment-4306868627).

A next step for these tests is to clean them up to use existing test infra where possible. It would be great to get these tests in nightly CI first to catch any breaking changes.

The full end-to-end tests (`n_layers=32`) require a bh galaxy with swap in order to run without hitting OOM. For now, the 4 layer versions of the tests have been added to `basic-test-nightly.yml` to run on the wh galaxy until a bh galaxy with swap is ready to use. 

[Edit]: this test makes changes to the modified_model.py for deepseek v3.2. Other tests using this modified_model.py have been updated to import the model from tt-forge-models. 

### Checklist

### DIFF
--- a/tests/torch/models/deepseek_v3_2_exp/build_weight_cache.py
+++ b/tests/torch/models/deepseek_v3_2_exp/build_weight_cache.py
@@ -1,0 +1,316 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Build a chunked BF16 weight cache for DeepSeek V3.1, one layer at a time.
+
+Processes each layer independently so peak memory is ~23 GB (one MoE layer)
+instead of ~1.37 TB (all 61 layers). No swap needed on a 512 GB machine.
+
+Usage:
+    python build_weight_cache.py                  # default: 61 layers
+    python build_weight_cache.py --n-layers 10
+    python build_weight_cache.py --n-layers 4 --repo deepseek-ai/DeepSeek-V3.1
+"""
+import argparse
+import json
+import os
+import re
+import sys
+import time
+
+import torch
+from huggingface_hub import hf_hub_download
+from safetensors import safe_open
+from safetensors.torch import load_file as safetensors_load_file
+from safetensors.torch import save_file as safetensors_save_file
+
+# Import the canonical helpers from the test file to stay in sync
+sys.path.insert(0, os.path.dirname(__file__))
+from test_deepseek_v3_2_exp import _rename_hf_key, _weight_dequant
+
+DEEPSEEK_V3_1_REPO = "deepseek-ai/DeepSeek-V3.1"
+
+
+def build_cache(repo_id, n_layers, n_dense_layers=1):
+    repo_slug = repo_id.replace("/", "--")
+    base = os.environ.get("HF_HOME", os.path.expanduser("~/.cache/huggingface"))
+    cache_dir = os.path.join(
+        base, "tt_xla_dequant_cache", f"{repo_slug}_{n_layers}layers"
+    )
+
+    if os.path.isdir(cache_dir) and any(
+        f.endswith(".safetensors") for f in os.listdir(cache_dir)
+    ):
+        print(f"Cache already exists at {cache_dir}, skipping.")
+        print(f"  Files: {sorted(os.listdir(cache_dir))}")
+        print("  Delete the directory to rebuild.")
+        return
+
+    t_total_start = time.perf_counter()
+
+    # Load index
+    index_path = hf_hub_download(repo_id, "model.safetensors.index.json")
+    with open(index_path) as f:
+        index = json.load(f)
+    weight_map = index["weight_map"]
+
+    # Build per-layer mapping: which ckpt keys (and their scales) belong to
+    # which layer/group.
+    # Groups: "shared" (embed, norm, head) and layer_0000..layer_NNNN
+    groups = {}  # group_name -> {ckpt_key: model_key}
+    scale_keys = {}  # ckpt_weight_key -> ckpt_scale_key
+    all_key_to_shard = {}  # ckpt_key -> shard_file (weights + scales)
+
+    for ckpt_key, shard_file in weight_map.items():
+        layer_m = re.match(r"model\.layers\.(\d+)\.", ckpt_key)
+        if layer_m and int(layer_m.group(1)) >= n_layers:
+            continue
+
+        if "weight_scale_inv" in ckpt_key:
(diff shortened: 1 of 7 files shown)


gengelageTT:
These newly added tests are passing here and took about 46 min: https://github.com/tenstorrent/tt-xla/actions/runs/25384221440/job/74455772384

vvukomanTT (approved):

nsumrakTT (approved):

gengelageTT:
Re-running after rebase and changes here: https://github.com/tenstorrent/tt-xla/actions/runs/25440566263

tests/torch/models/deepseek_v3_2_exp/modified_model.py
```diff
@@ -1,6 +1,10 @@
 # SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
 #
 # SPDX-License-Identifier: Apache-2.0
+
+# This version of the modified model is only for tests/torch/models/deepseek_v3_2_exp/test_deepseek_v3_1.py.
```
AleksKnezevic, line 5: Can you move this to tt-forge-models? Does not need to be in this PR.

tests/torch/models/deepseek_v3_2_exp/modified_model.py
```diff
@@ -972,7 +1200,9 @@ def forward(self, x: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
         Returns:
             Tuple[torch.Tensor, torch.Tensor]: Routing weights and selected expert indices.
         """
-        scores = linear(x.float(), self.weight.float())
+        # TODO(tt-xla#XXXX): fp32 linear hits an accuracy issue in the XLA
```
AleksKnezevic: please open issue.

tests/torch/models/glm4_moe/build_weight_cache_glm.py
```diff
@@ -0,0 +1,262 @@
...
+
+import torch
+from huggingface_hub import hf_hub_download
+from safetensors import safe_open
+from safetensors.torch import save_file as safetensors_save_file
+
+GLM_REPO = "zai-org/GLM-4.7"
+
+
+def _safe_open_hf(path):
```
AleksKnezevic, line 40: Can we extract some of the common weight cache functions into a utility file?

tests/torch/models/glm4_moe/test_glm4_7.py
```diff
@@ -0,0 +1,762 @@
...
+
+        # Drop source tensors for this layer to keep peak memory in check.
+        for j in range(n_experts):
+            for name in ("gate_proj", "up_proj", "down_proj"):
+                expert_tensors.pop((layer_idx, j, name), None)
+
+    return state_dict
+
+
+def _build_and_load_model_for_cpu(config, repo_id, mesh_shape):
```
AleksKnezevic, line 262: what's this for?

AleksKnezevic (approved):

tests/torch/models/deepseek_v3_2_exp/build_weight_cache.py
```diff
@@ -0,0 +1,406 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Build a chunked BF16 weight cache for DeepSeek V3.1, one layer at a time.
```
jameszianxuTT, line 5: This weight cache idea is a great dev tool, and nontrivial to be useful in CI

The main issues being:
- the cache would be regenerated between runs, because it's saved to an ephemeral location in the container
- even if the cache was persistent (e.g. like ttnn tensor cache), we would need to come up with some proper keying and eviction logic, and infra to share the caches to make it useful to accelerate CI

This is not an argument to block the PR, I think it's fine to merge with the knowledge that weight cache is regenerated per CI run. And it has the benefit of lowering peak RSS by buffering the dequantized result to an ephemeral location on disk, which may ease swap thrashing as well... so overall a useful change. (and similar in spirit to @sshonTT's split-layer compile idea, though it only touches the stack at a much earlier and smaller point)

jameszianxuTT: I've put up some additional thoughts/experiments related to weight caching in #4564. I think this is also something to be productized, but requires more work to spec out.

tests/torch/models/glm4_moe/test_glm4_7.py
gengelageTT: My understanding is that it on device we use a stacked and transformed version of the weights for the sparse moe. However the cpu path in sparse_mlp.py expects the original weights for golden computation. I think Brian set this test up so that when not run on cpu, the weights aren't loaded (model contains meta tensors) until later and at that point we expect them to be in this stacked format which is precomputed and stored in `post_sparse_cache`. That's the `_build_and_load_model_post_sparse` path. In `_build_and_load_model_for_cpu` the weights are instead loaded earlier in the original format so the model can run on cpu.

tests/torch/models/deepseek_v3_2_exp/modified_model.py
gengelageTT: I should be able to remove this version of the modified model. Brian and I made similar changes but in different places. When I clean these tests up I'll try to get this deepseek test running with the modified_model already in tt-forge-models. Or make adjustments in tt-forge-models where needed.

tests/torch/models/deepseek_v3_2_exp/modified_model.py
gengelageTT: Thanks for catching this, I didn't see it. I tested this locally and it doesn't cause a failure for deepseek-v3.1 so I reverted this change but I'll keep note of it for when I test deepseek-v3.2-exp and create an issue then if it's still a problem.

tests/torch/models/glm4_moe/build_weight_cache_glm.py
gengelageTT: Yes, that's the next thing I'll work on for these tests, alongside using existing test infra where possible

gengelageTT:
Tests are still passing after rebase here: https://github.com/tenstorrent/tt-xla/actions/runs/25572644562/job/75073863919

PR #4545, opened 2026-05-06
Add `finding-missed-fusions` skill

ppadjinTT:
## Overview

A skill that looks at the model IRs (either through raw log file or isolated .mlir files) and finds missed fusions. I classified missed fusions in two categories:
1. direct missed fusions: there is an existing TTNN op we could fused into
2. theoretical missed fusions: if torch/cuda/triton/... has similar fused op and we don't. This category is less important now but will be very useful when we get tt-lang support through the compiler 

## Example

I attached a `fusion_todo.yml` that this skill generated for the Wan 5B DiT. In this example notable things it detected:
1. missing ttnn.rms_norm fusion
2. pattern where we explicitly calculate the `scale` parameter of SDPA with multiply -> permute -> typecast sequence instead of just passing different scale to the SDPA op, saving us a few ops

fyi @djstefanovicTT @sdjordjevicTT 
[fusion_todo.yml](https://github.com/user-attachments/files/27434313/fusion_todo.yml)

### DIFF
--- a/.claude/skills/finding-missed-fusions/SKILL.md
+++ b/.claude/skills/finding-missed-fusions/SKILL.md
@@ -1,0 +1,161 @@
+---
+name: finding-missed-fusions
+description: Use when auditing a TTNN model's IR for missed op fusion opportunities — both direct TTNN fusions (a fused ttnn op already exists) and theoretical fusions (the pattern is a single kernel in torch/triton/cuda)
+---
+
+# Finding Missed Fusions
+
+## Overview
+
+Op fusion is critical for TTNN model performance. This skill audits a model's TTNN IR for **two classes** of missed fusion:
+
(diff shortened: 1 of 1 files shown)


sdjordjevicTT (approved):
Nice!

AleksKnezevic (approved):
LGTM. As a future improvement you can suggest to offer to add [filecheck](https://github.com/tenstorrent/tt-xla/blob/b604b62e981b2d7c92cc57f40b918b5ff033b7bc/tests/torch/graphs/fusion_patterns/test_repvgg_conv_sum.py#L15) tests for fusion patterns.

ppadjinTT:
nice, i didn't know you can do filecheck through tt-xla pytest

PR #4549, opened 2026-05-06
Add triage skill for unpack_forward_output FE failures

agobeljicTT:
> [!NOTE]
> This PR complements [tenstorrent/tt-forge-models#637](https://github.com/tenstorrent/tt-forge-models/pull/637), which contains the model-side fixes (registry handlers + per-loader overrides) produced by running this skill. The YAML status updates here reflect the new failure stages observed once those fixes are in.

### Ticket
Refs: [#4293](https://github.com/tenstorrent/tt-xla/issues/4293) (kept open — this is only the first pattern of FE failures being triaged).

### Problem description
A large block of training tests was stuck at \`FAILED_FE_COMPILATION\` with reason:

> \`tt-forge-models doesn't implement unpack_forward_output for this model.\`

Triaging these by hand is mechanical: run the model on CPU, inspect the forward output structure, then either register the output class in \`training_utils.py\` or write a per-loader \`unpack_forward_output\` override. Repeated dozens of times across yolov6, yolov9, retinanet, fpn, swin, dpr, segformer, mgp_str, yolos, opt, albert, bert, bge_m3, distilbert, etc.

### What's changed
- **New Claude Code skill \`.claude/skills/triage-unpack-forward-output/SKILL.md\`** that handles exactly one failure pattern (the one above), invoked with a YAML key or \`model_info.name\`. The skill:
  1. Resolves the test name to a YAML entry + loader path. The YAML \`reason:\` is read for context but is not authoritative — Step 3 (CPU triage) is the real gate, since reasons can be stale.
  2. Runs the model on CPU via the bundled \`scripts/triage_forward_output.py\` and prints the forward output structure, gathers model docs, and decides between a registry entry vs a per-loader override (rule of thumb: registry for HuggingFace \`ModelOutput\` subclasses with a single loss-relevant attr; override for bare \`list\`/\`tuple\`/\`dict\` outputs, structural transforms, or when a registry entry exists but is wrong — i.e. its registered attribute is itself a tuple/list/dict).
  3. Writes the fix in tt-forge-models, then verifies with \`pytest --force-run --runxfail --junitxml=<path>\` to bypass the current YAML status and capture the real outcome. The YAML is updated once, after observing the real outcome.
  4. When the registry path is chosen, re-checks every other model that returns the same output class — registry is global by class name, so the fix affects all of them and the YAML must reflect that.
- **YAML updates** in \`tests/runner/test_config/torch/test_config_training_single_device.yaml\` reflecting the new failure stages observed after the companion tt-forge-models PR is in: most entries advance from \`FAILED_FE_COMPILATION\` to \`INCORRECT_RESULT\` (PCC drops, recorded as \`EXPECTED_PASSING\` + \`assert_pcc: false\`), \`FAILED_TTMLIR_COMPILATION\` (e.g. \`ElementsAttr::getValues UNREACHABLE\` for retinanet/yolov9/fpn ResNet50), or \`FAILED_RUNTIME\` (segformer, swinv2, mgp_str CB page sizing). A few unrelated environment errors are also surfaced now that FE compilation no longer hides them (\`'_has_torch_function' already has a docstring\`, \`is_torch_fx_available\`, MapTR boolean-tensor ambiguity).

### Checklist
- [x] New/Existing tests provide coverage for changes — the affected training tests are the ones whose YAML statuses are updated; each was re-run as part of the skill's verification step.


tests/runner/test_config/torch/test_config_training_single_device.yaml
```diff
@@ -295,27 +295,27 @@ test_config:
...
     bringup_status: FAILED_FE_COMPILATION
-    reason: "tt-forge-models doesn't implement unpack_forward_output for this model."
+    reason: "RuntimeError: function '_has_torch_function' already has a docstring"
   yolov6/pytorch-M-single_device-training:
     status: NOT_SUPPORTED_SKIP
     bringup_status: FAILED_FE_COMPILATION
-    reason: "tt-forge-models doesn't implement unpack_forward_output for this model."
+    reason: "RuntimeError: function '_has_torch_function' already has a docstring"
   distilbert/question_answering/pytorch-Base_Cased_Distilled_Squad-single_device-training:
     status: NOT_SUPPORTED_SKIP
```
ndrakulicTT: Now that we're past FAILED_FE_COMPILATION, let’s move the status to KNOWN_XFAIL so these models run in the nightly/weekly suites. For PCC failures, we can use EXPECTED_PASSING with assert_pcc = False.

tests/runner/test_config/torch/test_config_inference_single_device.yaml
```diff
@@ -1253,7 +1254,6 @@ test_config:
   yolov9/pytorch-T-single_device-inference:
     required_pcc: 0.94
     status: EXPECTED_PASSING
-    markers: [push, nightly]
```
ndrakulicTT, line 1252: This was not supposed to be removed?

tests/runner/test_config/torch/test_config_training_single_device.yaml
```diff
@@ -295,27 +295,27 @@ test_config:
...
-    bringup_status: FAILED_FE_COMPILATION
-    reason: "tt-forge-models doesn't implement unpack_forward_output for this model."
+    bringup_status: INCORRECT_RESULT
```
ndrakulicTT: as there is both static and runtime bringup_status can you check which one is being uploded to superset, similar to reason

tests/runner/test_config/torch/test_config_training_single_device.yaml
```diff
@@ -346,175 +346,175 @@ test_config:
...
-    reason: "tt-forge-models doesn't implement unpack_forward_output for this model."
+    bringup_status: FAILED_RUNTIME
+    reason: "ValueError: Error code: 13 from _xla_sync_multi during bev_pool"
   maptr/pytorch-Tiny_Fusion_24e-single_device-training:
-    status: NOT_SUPPORTED_SKIP
-    bringup_status: FAILED_FE_COMPILATION
-    reason: "tt-forge-models doesn't implement unpack_forward_output for this model."
+    status: KNOWN_FAILURE_XFAIL
+    markers: [notimeout]
+    reason: "Test exceeds default pytest timeout during torch.compile compilation"
```
ndrakulicTT, line 499: For my curiosity, what is this timeout?

tests/runner/test_config/torch/test_config_training_single_device.yaml
```diff
@@ -346,175 +346,175 @@ test_config:
...
     status: NOT_SUPPORTED_SKIP
     bringup_status: FAILED_FE_COMPILATION
-    reason: "tt-forge-models doesn't implement unpack_forward_output for this model."
+    reason: "RuntimeError: Boolean value of Tensor with more than one value is ambiguous"
   maptr/pytorch-Tiny_R50_24e_Bevpool-single_device-training:
     status: NOT_SUPPORTED_SKIP
-    bringup_status: FAILED_FE_COMPILATION
-    reason: "tt-forge-models doesn't implement unpack_forward_output for this model."
+    bringup_status: FAILED_RUNTIME
+    reason: "ValueError: Error code: 13 from _xla_sync_multi during bev_pool"
```
ndrakulicTT: This can probably be a separate skill of reading error code 13 stdout/stderr, as this is in most case not the main error. And would probably generally help the claude code in tt-xla repo. We can discuss this offline.

.claude/skills/triage-unpack-forward-output/SKILL.md
```diff
@@ -0,0 +1,239 @@
+---
+name: triage-unpack-forward-output
+description: Triage and auto-fix one tt-forge-models training test stuck at FAILED_FE_COMPILATION with reason "tt-forge-models doesn't implement unpack_forward_output for this model." Runs the model on CPU, inspects forward outputs, gathers model docs, then either registers a handler in training_utils.py (preferred for HuggingFace ModelOutput classes) or writes a per-loader unpack_forward_output override (for bare list/tuple/dict outputs or structural transforms). Re-runs the affected training and inference tests and updates the YAML.
```
ndrakulicTT: Shouldn't this description be shorter, to not pollute the context.

.claude/skills/triage-unpack-forward-output/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+allowed-tools: Bash Read Edit Write Grep Glob WebFetch
```
ndrakulicTT: Isn't bash a bit to much to allow, by default can we scope down it a bit for example Bash (source venv/activate && pytest *) or something like that?

.claude/skills/triage-unpack-forward-output/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+- Existing per-loader override examples to mimic when the override is the right tool:
+  - `third_party/tt_forge_models/yolov9/pytorch/loader.py:171-194` — list-of-tensors via `extract_tensors_recursive`
+  - `third_party/tt_forge_models/centernet/pytorch/loader.py:190-209` — list-of-dicts
+  - `third_party/tt_forge_models/clip/pytorch/loader.py:209-220` — tuple
+- **Implication of choosing the registry path:** the registry entry is global by class name, so it fixes every variant of the current model **and** every other model in the codebase that returns the same output class. You must therefore re-run *all* affected training tests, not just the one you were invoked with, and update each of their YAML entries.
+- Test-runner glue: `tests/runner/test_models.py` parametrizes `test_all_models_torch` with `(test_entry, parallelism, run_mode)`. Pytest test IDs combine in stack order: `<test_entry_id>-<parallelism_id>-<run_mode_id>`. The `<test_entry_id>` is `<model_path>-<variant>` (see `tests/runner/utils/dynamic_loader.py:319`).
```
ndrakulicTT: Side note, running test probably can be skill for itself, or even better we should have part of documentation to explain the test structure and how to run specific test. Anyways not the scope of the current work.

.claude/skills/triage-unpack-forward-output/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+- Test-runner gates on YAML status: when an entry is `NOT_SUPPORTED_SKIP`, pytest reports `SKIPPED` and your fix never runs. Before invoking pytest for verification, temporarily flip the affected entries to `status: EXPECTED_PASSING` (drop `bringup_status` and `reason`), run the tests, then write the final YAML based on the observed outcome.
```
ndrakulicTT: We should probably somehow enable running `NOT_SUPPORTED_SKIP` models by some cli argument but not the scope of this PR

.claude/skills/triage-unpack-forward-output/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Step 2 — Gate on the failure pattern
+
+1. Read the YAML entry. Its `reason:` must contain the substring `unpack_forward_output`. If not, abort: `This skill only handles unpack_forward_output failures. The failing reason here is: <reason>.`
+2. `grep -n "def unpack_forward_output(" <loader_path>`. If the override **already exists**, skip to Step 6 — the YAML entry may simply be stale. Note this in the final report.
+3. `grep -n "<OutputClassName>" third_party/tt_forge_models/training_utils.py` once you know the output class name (after Step 3). If the class is **already registered**, also skip to Step 6 — same staleness case.
+
+### Step 3 — CPU triage script
+
+Write `/tmp/triage_<model_dir>_<variant>.py`:
```
ndrakulicTT: This is okay for now, but would probably be good to have this as part of the test function that you can just call somewhere in the runner. Anyway not currently the scope of the PR. I will create issue for this.

.claude/skills/triage-unpack-forward-output/SKILL.md
```diff
@@ -0,0 +1,239 @@
...
+
+```python
+import sys, torch
+sys.path.insert(0, "/localdev/<user>/tt-xla")
+from third_party.tt_forge_models.<model_dir>.pytorch.loader import ModelLoader, ModelVariant
+
+loader = ModelLoader(variant=ModelVariant.<VARIANT>)  # omit variant arg if loader has no ModelVariant
+model = loader.load_model().eval()
+inputs = loader.load_inputs()
```
ndrakulicTT: we set the `batch_size=2` so maybe we should also have it here

ndrakulicTT (commented):
I read until the Step 5 will continue a bit later.

tests/runner/test_config/torch/test_config_training_single_device.yaml
agobeljicTT: wanna put that in the skill or just manually updat?

tests/runner/test_config/torch/test_config_inference_single_device.yaml
agobeljicTT: I agree

tests/runner/test_config/torch/test_config_training_single_device.yaml
agobeljicTT: I merged the PR for that, now the runtime will be the first one

tests/runner/test_config/torch/test_config_training_single_device.yaml
agobeljicTT: i think there is a default timeout and this marker means run without timeouts

tests/runner/test_config/torch/test_config_training_single_device.yaml
agobeljicTT: It was included that it needs to read the entire error code 13. I dont know why it didnt read the whole thing... this is probably an edge case, I've seen it does in fact read for some models

.claude/skills/triage-unpack-forward-output/SKILL.md
agobeljicTT: I have no idea xd

.claude/skills/triage-unpack-forward-output/SKILL.md
agobeljicTT: I agree, I wanted to remove the allowed-tools altogether...

.claude/skills/triage-unpack-forward-output/SKILL.md
agobeljicTT: i think that there is a force run test somewhere

.claude/skills/triage-unpack-forward-output/SKILL.md
agobeljicTT: wdymean?

.claude/skills/triage-unpack-forward-output/SKILL.md
agobeljicTT: done

.claude/skills/triage-unpack-forward-output/SKILL.md
```diff
@@ -0,0 +1,255 @@
...
+
+#### Step 5b — Per-loader override
+
+Add `def unpack_forward_output(self, forward_output)` to that model's `ModelLoader` class. Two non-negotiable rules:
+
+1. **Return only what the loss depends on** (universal rule above).
+2. **Docstring with these three sections.** The override must carry them — they are the audit trail for why this override exists when a registry entry would otherwise have sufficed:
+   - **Forward output structure** — full type/shape signature of what the model returns. Example: `list[Tensor] of length 3, shapes [(B, 256, 80, 80), (B, 256, 40, 40), (B, 256, 20, 20)]`.
+   - **What is selected and why** — which subset is returned and which loss it is the gradient source of. Example: `returning the three detection-head tensors concatenated; these are the only outputs consumed by YOLOLoss; auxiliary feature maps are dropped`.
+   - **Why a registry entry was not sufficient** — one line. Example: `default registry cannot key on bare list[Tensor]`, or `output class is shared with model X which uses a different loss target`.
```
ndrakulicTT: Can we remove this it just polutes the docsting, and in most cases is possible to deduce from the first two points

.claude/skills/triage-unpack-forward-output/SKILL.md
```diff
@@ -0,0 +1,255 @@
...
+
+Implementation guidance:
+
+- For nested list/tuple/dict structures: import `extract_tensors_recursive` from `third_party.tt_forge_models.tools.utils`, walk the loss-relevant subtree, and `torch.cat(tensors, dim=0)` (or stack) — see `yolov9/pytorch/loader.py:171-194`.
+- For a plain tuple where only one element is the loss target: index it.
+- Keep the implementation minimal. No extra abstractions, no inline comments beyond the required docstring.
+
+After writing (either path), run `pre-commit run --files <changed_files>` if pre-commit is installed; otherwise just confirm `python -c "from third_party.tt_forge_models.<model_dir>.pytorch.loader import ModelLoader"` succeeds.
```
ndrakulicTT: Pre-commit should always be installed, we should just tell it to activate env if its not installed

.claude/skills/triage-unpack-forward-output/SKILL.md
```diff
@@ -0,0 +1,255 @@
...
+
+### Step 6 — Verify with pytest
+
+**Before running pytest, temporarily flip every affected training entry to `status: EXPECTED_PASSING`** (drop `bringup_status` and `reason`). Otherwise the test runner skips them at collection (`tests/runner/test_models.py:118` gates on `NOT_SUPPORTED_SKIP`) and you observe `SKIPPED` instead of the real outcome. The set of "affected entries" depends on the path you took in Step 5:
```
ndrakulicTT: Lets update this as #4572 is merged

.claude/skills/triage-unpack-forward-output/SKILL.md
```diff
@@ -0,0 +1,255 @@
...
+
+Then run pytest. **No `less`. No `tail`. No piping into anything that hides exit codes.** Dump straight to a file and Read it back. Run the affected training tests in a single pytest invocation, plus the inference test for the originally-requested entry:
+
+```bash
+timeout 600 pytest -svv \
+  "tests/runner/test_models.py::test_all_models_torch[<entry1>-single_device-training]" \
+  "tests/runner/test_models.py::test_all_models_torch[<entry2>-single_device-training]" \
+  &> /tmp/verify_<scope>_training.log
+
+timeout 300 pytest -svv \
```
ndrakulicTT: Lets try maybe with `--junixml=report.xml` it can then parse the xml and look up the runtime reason why it failes, still needs the fallback to generate it from the stdout/stderr. With this option `-s` is not needed

.claude/skills/triage-unpack-forward-output/SKILL.md
```diff
@@ -0,0 +1,255 @@
...
+### Step 7 — Classify the training outcome
+
+- **Pass:** outcome = `EXPECTED_PASSING`.
+- **Fail:** classify the new error and pick the matching `bringup_status`:
+
+  | Symptom                                                              | `bringup_status`             |
+  | -------------------------------------------------------------------- | ---------------------------- |
+  | Crash in TTNN runtime / device assert / runtime stack trace          | `FAILED_RUNTIME`             |
+  | Compiler error after the frontend (TTIR / StableHLO / TTNN compile)  | `FAILED_TTMLIR_COMPILATION`  |
+  | Numerics divergence (PCC / atol failure, comparison output)          | `INCORRECT_RESULT`           |
```
ndrakulicTT: Lets add that `INCORRECT_RESULT` should be EXPECTED_PASSING with assert_false on PCC

agobeljicTT: done

.claude/skills/triage-unpack-forward-output/SKILL.md
```diff
@@ -0,0 +1,244 @@
...
+  <key>:
+    status: KNOWN_FAILURE_XFAIL
+    bringup_status: <FAILED_FE_COMPILATION | FAILED_TTMLIR_COMPILATION | FAILED_RUNTIME>
+    reason: "<one-line excerpt of the real error, ~120 chars>"
+  ```
+  Do not add a `markers:` field unless one was already present for an unrelated reason. Never leave a triaged entry at `NOT_SUPPORTED_SKIP` — the runner skips those (`tests/runner/test_models.py:118`) and the failure is invisible in nightly reports.
+
+The inference log is **informational only**. Do not modify the inference YAML entry. Flag any inference regression to the user in the final report so they can investigate separately.
+
+### Step 9 — Final report
```
mmilosevicTT: Some quotation is missing

mmilosevicTT: I think some clean-up is required, this will keep logs for no reason.

.claude/skills/triage-unpack-forward-output/SKILL.md
```diff
@@ -0,0 +1,244 @@
...
+  2. **Per-loader override** — add `def unpack_forward_output(self, forward_output)` to that model's `ModelLoader`. Required when the output has no stable class name to key on (bare `list`/`tuple`/`dict`), when extracting the loss-relevant tensors needs a structural transform (concat, stack, slice), or when two models share an output class but legitimately need different loss targets.
+- Existing per-loader override examples to mimic when the override is the right tool:
+  - `third_party/tt_forge_models/yolov9/pytorch/loader.py:171-194` — list-of-tensors via `extract_tensors_recursive`
+  - `third_party/tt_forge_models/centernet/pytorch/loader.py:190-209` — list-of-dicts
+  - `third_party/tt_forge_models/clip/pytorch/loader.py:209-220` — tuple
```
ndrakulicTT: nit: it would probably be better to specify something like third_party/tt_forge_models/clip/pytorch/loader.py::ModelLoader::unpack_forward_output, as lines will get out of sync very fast.

.claude/skills/triage-unpack-forward-output/SKILL.md
```diff
@@ -0,0 +1,244 @@
...
+  ```
+  Drop `bringup_status` and `reason`.
+- **Any other failure** (`FAILED_TTMLIR_COMPILATION`, `FAILED_RUNTIME`, or even still `FAILED_FE_COMPILATION` for an unrelated env/import reason like `_has_torch_function` docstring, `is_torch_fx_available`, `Boolean value of Tensor … ambiguous`, `ModuleNotFoundError`): mark as `KNOWN_FAILURE_XFAIL` so nightly/weekly invokes the test and reports XFAIL with the documented reason instead of silently skipping:
+  ```yaml
+  <key>:
+    status: KNOWN_FAILURE_XFAIL
+    bringup_status: <FAILED_FE_COMPILATION | FAILED_TTMLIR_COMPILATION | FAILED_RUNTIME>
+    reason: "<one-line excerpt of the real error, ~120 chars>"
+  ```
+  Do not add a `markers:` field unless one was already present for an unrelated reason. Never leave a triaged entry at `NOT_SUPPORTED_SKIP` — the runner skips those (`tests/runner/test_models.py:118`) and the failure is invisible in nightly reports.
```
ndrakulicTT: Maybe we can leave NOT_SUPPORTED_SKIP if we get FAILED_FE_COMPILATION, as those errors will not be fixed without explicit fixing from our side.

ndrakulicTT (approved):
LGTM, few small nits, dont forget to drop the custom forge-models commit

.claude/skills/triage-unpack-forward-output/SKILL.md
sgligorijevicTT: So for the practice in xla was to leave everything that doesn't bring down the entire process as xfailed

.claude/skills/triage-unpack-forward-output/SKILL.md
agobeljicTT: where

agobeljicTT: it's all in /tmp, perhaps a note that a cleanup is necessary.
i needed to manually check some logs afterwards, so i am against auto-cleanup

AleksKnezevic (approved):

PR #4552, opened 2026-05-06
Clamp OOB negative slices via FX pass in tt_torch (#4465)

### LINKED ISSUE #4465: [Krea Realtime 14B] VAE decoder : RuntimeError: Value out of range (expected to be in range of [-1, 0], but got -2)

kamalrajkannan78:
### Ticket

- fixes https://github.com/tenstorrent/tt-xla/issues/4465

### Problem description

- `x[..., -N:, ...]` fails on the tt backend when `N > dim_size` (e.g. `-CACHE_T:` on a size-1 dim in the krea VAE decoder).
- On CPU, `Tensor.__getitem__` silently clamps the out-of-range negative start. On XLA tensors it lazy-lowers straight to the strict `aten.slice.Tensor`, which rejects `start < -size` and raises `RuntimeError: Value out of range`.
- Same code path the diffusers `AutoencoderKLWan` decoder relies on, so this blocks krea VAE compilation outright.

### What's changed

- For each `aten.slice.Tensor` node in the FX graph whose `start < -dim_size`, rewrite the literal to `-dim_size`. Produces the same output CPU eager `__getitem__` would (it clamps silently) and lets the strict `aten.slice` overload accept the arg.
- Implemented as `clamp_neg_slice_starts` in `python_package/tt_torch/backend/passes.py`; called from `_call_experimental_compile` (`backend.py`) after `legalize_graph`, so it runs before the torch_xla bridge's FX collector validates the graph.
- Added `tests/torch/ops/test_slice.py` (repro for the model's issue) to catch regressions in nightly until the model passes e2e.

### Checklist

### Logs

- [may8_slice_before_fix.log](https://github.com/user-attachments/files/27516019/may8_slice_bf.log)
- [may8_vae_decoder_before_fix.log](https://github.com/user-attachments/files/27516022/may8_vae_decoder_bf.log)

### DIFF
--- a/python_package/tt_torch/backend/backend.py
+++ b/python_package/tt_torch/backend/backend.py
@@ -29,6 +29,7 @@ from .passes import (
     bypass_assert_tensor_metadata,
     bypass_dtype_promotion_and_redundant_cast,
     bypass_redundant_getitem,
+    clamp_neg_slice_starts,
     handle_composite_ops,
     insert_argument_type_markers,
     rewrite_adaptive_avgpool_to_mean,
@@ -190,6 +191,8 @@ class XLAExecutor:
             # Calling legalize_graph rebuilds the graph in topological order(from usage information), and fixes up the prev/next pointers in the process - which fixes our issue.
             # All this is a problem because DynamoBridge Partitioner can get confused by wrong next nodes and partition the graph in a way which fails to execute.
             legalize_graph(program.graph_module)
+
+            clamp_neg_slice_starts(program.graph_module)
 
             # Collect the params and constants from the exported program.
             self.params_and_consts = self._build_params_and_consts(program)
--- a/python_package/tt_torch/backend/passes.py
+++ b/python_package/tt_torch/backend/passes.py
@@ -308,6 +308,39 @@ def bypass_redundant_getitem(gm):
     return gm
 
 
+def clamp_neg_slice_starts(gm):
+    """
+    Clamp out-of-range negative `start` values on aten.slice.Tensor nodes.
+
+    Eager `x[..., -N:, ...]` clamps when N > dim size, but the strict
+    aten.slice.Tensor overload rejects start < -size. Normalize the literal in
+    the FX graph so downstream lowering sees a legal index.
+    """
+    changed = False
+    for node in gm.graph.nodes:
+        if (
+            node.op != "call_function"
+            or node.target is not torch.ops.aten.slice.Tensor
+            or len(node.args) <= 2
+        ):
+            continue
+        self_node, dim, start = node.args[0], node.args[1], node.args[2]
+        if not isinstance(start, int) or not isinstance(dim, int):
+            continue
+        val = self_node.meta.get("val")
+        if val is None:
+            continue
+        size = val.shape[dim]
+        if not isinstance(size, int):
+            continue
+        if start < -size:
+            node.args = (self_node, dim, -size, *node.args[3:])
+            changed = True
+    if changed:
+        gm.recompile()
+    return gm
+
+
 def run_shape_prop(gm, example_inputs):
     """
     Propagates shape and dtype information through the graph.
--- a/tests/torch/models/krea_realtime/test_vae_decoder.py
+++ b/tests/torch/models/krea_realtime/test_vae_decoder.py
@@ -16,9 +16,9 @@ from third_party.tt_forge_models.krea_realtime_video.pytorch import (
 )
 
 
-@pytest.mark.xfail(
-    reason="VAE temporal slice fails on TT (out-of-range slice on size-1 dim) — https://github.com/tenstorrent/tt-xla/issues/4465"
-)
+# @pytest.mark.xfail(
+#     reason="Out of Memory: Not enough space to allocate 4907335680 B DRAM buffer across 12 banks, where each bank needs to store 408944640 B, but bank size is 1071821792 B - https://github.com/tenstorrent/tt-xla/issues/4551"
+# )
 def test_vae_decoder():
     xr.set_device_type("TT")
     torch.manual_seed(42)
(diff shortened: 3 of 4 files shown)


saiarthiraguram (approved):

ppadjinTT (commented):
This override is a bit tricky, we're hitting similar problems with this on Wan model. Let's put this PR to a hold until we figure out the best way to solve these types of problems (e.g. maybe it is just better for fix to be contained in the test script)

kamalrajkannan78:
> This override is a bit tricky, we're hitting similar problems with this on Wan model. Let's put this PR to a hold until we figure out the best way to solve these types of problems (e.g. maybe it is just better for fix to be contained in the test script)

Following up on your concern about the always-on override in shared tt_torch code — would a test-scoped TorchFunctionMode work? Same clamp logic, but only active inside `with _NegSliceClamp():` for this one test. No change to `tt_torch/torch_overrides.py`, no need to re-implement the diffusers forwards, and Wan can adopt the same pattern (or we can lift it into a shared test util once both PRs settle) if needed.

```python
from torch.overrides import TorchFunctionMode

class _NegSliceClamp(TorchFunctionMode):
    """XLA __getitem__ rejects start < -size; CPU eager clamps. Match eager."""

    def __torch_function__(self, func, types, args=(), kwargs=None):
        if getattr(func, "__name__", "") == "__getitem__" and len(args) >= 2:
            self_t, key = args[0], args[1]
            if isinstance(key, tuple) and Ellipsis not in key:
                shape = self_t.shape
                changed, new_key, dim = False, [], 0
                for k in key:
                    if k is None:
                        new_key.append(k)
                        continue
                    if isinstance(k, slice) and isinstance(k.start, int) and k.start < -shape[dim]:
                        k = slice(-shape[dim], k.stop, k.step)
                        changed = True
                    new_key.append(k)
                    dim += 1
                if changed:
                    args = (self_t, tuple(new_key))
        return func(*args, **(kwargs or {}))


def test_vae_decoder():
    xr.set_device_type("TT")
    torch.manual_seed(42)

    loader = ModelLoader(ModelVariant.VAE)
    model = loader.load_model(dtype_override=torch.bfloat16)
    inputs = loader.load_inputs(dtype_override=torch.bfloat16)

    with _NegSliceClamp():
        run_graph_test(
            model,
            inputs,
            framework=Framework.TORCH,
        )

```

Does this fit the direction you're aiming for, or were you thinking of something different?

I've tried this fix on sanity and model & it works! Here are the logs, scripts with fix for your reference

- [may8_slice_before_fix.log](https://github.com/user-attachments/files/27516019/may8_slice_bf.log)
- [may8_slice_after_fix.log](https://github.com/user-attachments/files/27516016/may8_slice_af.log)
- [may8_vae_decoder_before_fixf.log](https://github.com/user-attachments/files/27516022/may8_vae_decoder_bf.log)
- [may8_vae_decoder_after_fix.log](https://github.com/user-attachments/files/27516020/may8_vae_decoder_af.log)
- [test_slice.py](https://github.com/user-attachments/files/27516023/test_slice.py)
- [test_vae_decoder.py](https://github.com/user-attachments/files/27516024/test_vae_decoder.py)

ppadjinTT:
This seems better yes, let's go with this

ppadjinTT:
After offline chat, we agreed this is a good solution for now, thanks @kamalrajkannan78 !

ppadjinTT (approved):

tests/infra/__init__.py
```diff
@@ -29,6 +29,7 @@
     MLACache,
     MLAStaticLayer,
     Model,
+    NegSliceClamp,
```
jameszianxuTT: I think we should really avoid these patching contexts (and in general, model-specific override behaviour that changes compiler mechanics) whenever it's not absolutely necessary. Otherwise, a user will see that we have 1000 patching wrappers and they will need to figure out which N of the wrappers are necessary to use to get their model to compile. 

Is it not possible to put an override somewhere else such that this can apply automatically? For example a torchfx decomposition, existing torch_overrides.py, or even a patch to MLIR to clamp OOB negative slice?

jameszianxuTT (changes requested):

kamalrajkannan78:
cc: @jameszianxuTT @ppadjinTT

Synced offline with @jameszianxuTT.

Direction agreed on: following @ppadjinTT's note about TorchFunctionMode overrides causing graph breaks that are hard to catch in our current infra, the fix should live as a torch FX pass — e.g., in `python_package/tt_torch/backend/passes.py`. The pass walks the FX graph for whatever __getitem__ + slice lowers to and rewrites it there.

Status: I've tried this approach (current PR has the changes). It fixes the sanity test but not the krea VAE model. I'll take a closer look at the model later and come back with a cleaner fix if possible.

[may9_slice_sanity_after_fix.log](https://github.com/user-attachments/files/27555684/may9_slice_sanity_after_fix.log), [may9_vae_decoder_after_fix.log](https://github.com/user-attachments/files/27555686/may9_vae_decoder_after_fix.log)

Issue #3905, opened 2026-03-24
Add kimi-k2, deepseek v3.2 exp and glm 4.5 to benchmark tests


gengelageTT:
Add perf benchmark tests for kimi-k2, deepseek v3.2 exp and glm 4.5. Tests don't need to run the full model but need to include an MoE layer.

gengelageTT:
Sharding for the [kimi-k2 benchmark test added here](https://github.com/tenstorrent/tt-xla/pull/4247) should be updated to use (after the MoE sharding fix to EP 32):
https://github.com/tenstorrent/tt-xla/blob/main/tests/benchmark/test_llms.py#L1660

gengelageTT:
Each model has a 2-4 layer benchmark now running in CI. Next steps for these benchmarks are being tracked by other issues.

PR #4557, opened 2026-05-06
Add deepseek-v3.2-exp benchmark

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3905

### Problem description
We want to test 2 layers of deepseek-v3.2-exp on galaxy through the benchmark tester.

### What's changed
Changes to benchmark tests/infra:
- Added test_deepseek_v3_2_exp_tp_galaxy_2_layers (with trace disabled). The required pcc is set to `0.75` since that's what's seen for decode but the prefill pcc is `0.92`
- Added support for indexer cache (`k_cache`) initialization when `use_indexer_cache` is set
- Added `_init_indexer_cache()` in benchmark/llm_utils/decode_utils 

Changes to tests/torch:
- removed local copies of the original model and kernel. Keeping the local modified_model.py until the e2e deepseek test is set up to use the same modifications made for the version in tt-forge-models
- the tests in test_deepseek_v3_2_exp.py are updated to import the model from tt-forge-models in the e2e deepseek test pr

### Checklist


gengelageTT:
Benchmark ran here: https://github.com/tenstorrent/tt-xla/actions/runs/25380393604/job/74431843739

Will re-run on latest before merging.

nsumrakTT (approved):

mvasiljevicTT:
> Benchmark ran here: https://github.com/tenstorrent/tt-xla/actions/runs/25380393604/job/74431843739
> 
> Will re-run on latest before merging.

Please add an issue that will track bad outputs for all users on this model, note that they are also all different

gengelageTT:
@mvasiljevicTT This test is only 2 layers of the model with random weights initialized. As a result the outputs are expected to be incoherent - I can add a comment about this, thanks for bringing it up. There's a separate issue in progress to bring up this model e2e with weights loaded [here](https://github.com/tenstorrent/tt-xla/issues/3927). Eventually this test will be replaced by a test of the full model with pre-trained weights that should have coherent outputs.

vvukomanTT (approved):

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -157,6 +163,18 @@ def construct_inputs(
     input_ids = inputs["input_ids"] if isinstance(inputs, dict) else inputs.input_ids
     cache_position: torch.Tensor = torch.arange(0, input_ids.shape[1])
 
+    # Initialize indexer cache before model.to(device) so it's not compiled as a separate XLA
+    # graph. Models using this cache are expected to handle stale values since we cannot reset
+    # the indexer key cache once the model is transferred to device.
+    if use_indexer_cache:
```
mvasiljevicTT: If I understand this correctly, once indexer is initialized it will be used for the whole run? The `use_indexer_cache` flag is a bit confusing to me as it is relying on the user to set it to true only for the first `construct_inputs` call? If someone misinterpreted it and used it in every call would it break or just silently work wrong?

Could `init_indexer_cache` call with this same `if` be moved directly to `benchmark_llm_torch_xla` before the first call of `construct_inputs`?

gengelageTT: Yes it should only be set true for the first `construct_inputs` call and then the model is expected to ensure the indexer cache doesn't use stale values from previous runs. It will break compilation if set true after the model is moved to device.

I can initialize the indexer outside of `construct_inputs` if that's preferred :) 

However, setting `use_indexer_cache=True` only affects the first call of construct_inputs, since it's only passed into the first call. The remaining calls see no `use_indexer_cache` passed in and would use the default of `False`. In order for a user to see a test break for this reason they would have to make changes in llm_benchmark.py. 

Additionally, I'm not sure if this will be used for other models, this model is modified to have an `early_intialization` function in its Indexer. If a user tries to use this flag they would likely see a failure early on due to this function not existing in the model they're running. If/when we encounter more models that need this functionality then the `init_indexer_cache` will likely have to be updated to work more generally - perhaps in a way that requires it to be initialized more than once. I'm not sure yet what this would look like since deepseek-v3.2-exp isn't a transformers model.

mvasiljevicTT: Thanks, moving it outside would be nice

vkovacevicTT:
From the offline discussion:

We want to avoid adding models to the benchmark that don't work initially.
Since Deepseek and [Kimi](https://github.com/tenstorrent/tt-xla/pull/4566) are our top priority models, I agree that they should be added to perf benchmark as soon as possible, but until they work correctly (pass PCC check) I would suggest one of two options:
1. Add them to experimental benchmark. They will run just like other priority models in the regular benchmark, we just wouldn't track them and file P0 issues until PCC error is fixed and they are green. We can then transfer them to the regular benchmark group.
2. Lower PCC thresh and add them to regular benchmarks. I would avoid this option ([Kimi](https://github.com/tenstorrent/tt-xla/pull/4566) has even worse PCC than Deepseek).

In any case, PCC issue for both models should be addressed ASAP.
Until then, from the benchmark side, we won't treat their regressions as P0.

What do you guys think?
cc @odjuricicTT @AleksKnezevic

AleksKnezevic:
@vkovacevicTT, as these are priority models, to me it makes most sense to parallelize these efforts and go with 2 since it protects against further regressions by ensuring that they treated as P0s (i.e. things that completely break compile). 

@gengelageTT will, of course, keep working on fixing PCC, but why risk losing the model fully until that is done?

odjuricicTT (approved):
@gengelageTT I'm fine with merging this with a lowered PCC threshold and a with a rerun ton main to make sure the test still passes in CI.

gengelageTT:
Both deepseek-v3.2-exp and kimi-k2 are seeing inconsistent PCC between tests run on the same commit/build. Created [an issue here](https://github.com/tenstorrent/tt-xla/issues/4632) and removed the pcc requirement (`-1.0`) so that we can test the model continues to compile and run in nightly CI until this issue is resolved.

Running with latest changes here: https://github.com/tenstorrent/tt-xla/actions/runs/25702752921

AleksKnezevic (approved):

Issue #4562, opened 2026-05-06
[vLLM] Flaky output-coherence assertion in test_tensor_parallel_generation


kmabeeTT:
## Problem

`assert_output_coherent` in `tests/integrations/vllm_plugin/generative/conftest.py` flags any character with `ord(c) > 127` as non-Latin and fails when the ratio exceeds 3%. This catches legitimate output that is not the bug it was designed to detect (#4440 sampler token-soup):

| Test | Output | Ratio | Issue |
|---|---|---|---|
| Llama-3.2-3B | `' rain. I think it's because I used'` | 0.030 | Smart quote (U+2019) |
| Qwen3-0.6B | `' park. I have a dog... my可愛猫. I like to play...'` | 0.034 | Coherent English with 3 CJK chars |
| Qwen3-0.6B | `' park. I always like to do what? A. Go to the watering vatＢ. Go toขาย (Shop) C.CH'` | 0.050 | Fullwidth Latin + Thai |
| Qwen3-0.6B | `' park, so今天小编就整理了相关的内容...'` | 0.833 | Full Chinese continuation |

The check is misnamed — `ord(c) > 127` is a non-ASCII test, not non-Latin. It flags standard English typography (smart quotes, em dashes, ellipses) and legitimate multilingual sampling on Qwen3.

## Mitigation (this PR)

- Drop the non-Latin ratio assertion. The stopword-ratio check (the actual #4440 token-soup detector per its docstring) stays.
- Prefix Qwen3 prompt with `"Continue in English:"` to reduce multilingual drift.

## Follow-up to consider

Whether a tighter token-soup detector is worth adding back (e.g. lower `_MIN_WORDS` so the stopword check fires on shorter outputs, or a different heuristic). Not urgent — stopword check covers the original bug.

kmabeeTT:
https://github.com/tenstorrent/tt-xla/pull/4563 closes this

PR #4563, opened 2026-05-06
[vLLM] Fix flaky output coherence assertion in TP generation test

kmabeeTT:
## Ticket
#4562

## Problem
`assert_output_coherent` flagged legitimate output as failures: smart quotes (U+2019) on Llama-3.2-3B, and Unicode punctuation / multilingual sampling drift on Qwen3. The `ord(c) > 127` check is a non-ASCII test, not a non-Latin test, so it caught standard English typography.

## What's changed
- `tests/integrations/vllm_plugin/generative/conftest.py`: drop the non-Latin character ratio assertion. Stopword-ratio check (the actual #4440 token-soup detector) stays.
- `tests/integrations/vllm_plugin/generative/test_tensor_parallel_generation.py`: prefix the Qwen3-0.6B test prompt with `"Continue in English:"` to reduce multilingual drift at the source.

## Checklist
- [x] Affected tests run locally, results in comment

### DIFF
--- a/tests/integrations/vllm_plugin/generative/conftest.py
+++ b/tests/integrations/vllm_plugin/generative/conftest.py
@@ -20,24 +20,18 @@ _STOPWORDS = frozenset(
 )
 _WORD_RE = re.compile(r"[A-Za-z']+")
 _MIN_STOPWORD_RATIO = 0.10
-_MAX_NON_LATIN_RATIO = 0.03
 _MIN_WORDS = 5
 
 
 def assert_output_coherent(text: str) -> None:
-    """Heuristic assertion: text is English natural-language, not token soup.
+    """Heuristic assertion: text is natural-language, not token soup.
 
-    Tuned for the English prompts used in the generative TP tests
-    (e.g. "I like taking walks in the"). Re-tune thresholds if a test
-    introduces non-English prompts or code-completion prompts.
+    Uses English stopword ratio as the token-soup detector — coherent
+    continuations contain several stopwords per ~30-token output, while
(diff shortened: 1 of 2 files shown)


mmanzoorTT (approved):

kmabeeTT:
**Local stress-test findings** (50 generations per prompt, single chip, no TP, same `temperature=0.8, top_p=0.95, max_tokens=32`):

|  | OLD assertion (with non-Latin check) | NEW assertion (stopword only) |
|---|---|---|
| ORIG prompt (`"I like taking walks in the"`) | **15/50 failed (30%)** | 0/50 failed |
| NEW prompt (`"Continue in English: ..."`) | 7/50 failed (14%) | 0/50 failed |

Multilingual drift in the raw output:
- Outputs containing any non-Latin char: 26/50 (ORIG) → 13/50 (NEW) — roughly halved
- Outputs that are pure Latin: 24/50 (ORIG) → 37/50 (NEW)

Three takeaways:
1. The old assertion was very flaky (~30% per run). CI's apparent rarity was just a low sample count.
2. The prompt prefix halves drift but doesn't eliminate it — Qwen3-0.6B still emits CJK/Cyrillic/Thai chars in ~26% of outputs even when explicitly asked to continue in English.
3. The stopword-only check caught zero false positives across all 100 outputs — it's robust against drift and would still fire on the #4440 token-soup pattern (which has ~0 stopwords).

Confirms the failures aren't TP-related — single-chip reproduces them at the same rates.

AleksKnezevic (approved):

Issue #4592, opened 2026-05-08
Weight streaming feature enable


sshonTT:
****Background****

Running large MoE models (DeepSeek-V4 / DeepSeek-V4-Pro and similar) through the our system exposes a tension between two existing execution modes:

Mode 1 — whole-model compile. Load all weights into host RAM, run torch.compile once over the full model, and execute the resulting single graph for prefill + decode. This is the production path: one compiled artifact, no per-step retrace, best runtime throughput.

Mode 2 — per-layer execute. Load and run one layer at a time at inference time. Slower at runtime (every step re-traces and re-compiles each layer), but useful as a tool: it lets us run a model on a smaller device than the production target (helpful during hardware bring-up), and it cleanly separates intra-layer PCC bugs from inter-layer ones (when a per-layer reference matches but a chained whole-model output diverges, the bug is in cross-layer plumbing — sharding, dispatch/combine, KV state — rather than inside the block).

Neither mode covers large MoE on its own:

 - Mode 1 OOMs. Pro has 60+ layers × 384 routed experts × 3 weights → ~1.5 TB after bf16 dequant, ~360 GB even in fp4 packed form. Hosts in the 512–566 GB range can't hold that. On top of the model itself, the PJRT plugin keeps host-side at::Tensor staging buffers for every parameter it ships until ensure_layout migrates them to device, pushing peak host RAM past available memory.
 - Swap doesn't generalize as a fallback for Mode 1: swap sizes vary per host (CI / dev / lab), making runs non-reproducible; weight stacking / dequant access patterns thrash on swap and don't finish in practical time; and device DRAM is independently bounded (~4 GB per bank in bh glx) — host swap doesn't help that side at all.
Mode 2's per-step retrace is too expensive at runtime to be the production execution path.

****What we need****
A hybrid weight streaming feature that combines the two:

Host RAM bounded to roughly one layer's staging at any time — so the model runs on hosts that cannot fit the full weight set in memory (Mode 2's loading discipline).
Inference runs through a single whole-model compiled graph — preserving Mode 1's runtime efficiency, no per-step retrace.
This issue tracks the work to land that hybrid path.

sshonTT:
### Required components
**PJRT plugin (pjrt_plugin_tt)**
[...]

PR #4593, opened 2026-05-08
[pjrt] expose const-eval-inputs-to-system-memory pipeline option

sshonTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/4592)

### Problem description
We are not able to use existing ttnn option `enable_const_eval_inputs_to_system_memory`

### What's changed
Adds a PJRT compile option `enable_const_eval_inputs_to_system_memory` that forwards to the TTIRToTTNNCommonPipelineOptions field of the same name.

By default the compiler annotates const-eval function inputs as system memory in the executable's expected layout, which keeps their host-side copies alive for the lifetime of the model. For workloads that do not need the host copies after the first execute (e.g. when host RAM headroom is tight), exposing this option lets the caller opt out so the runtime migrates the inputs to device on first use and the host-side copy is released.

Default mirrors the compiler default (true), so existing behavior is unchanged.

### Checklist

### DIFF
--- a/pjrt_implementation/inc/api/compile_options.h
+++ b/pjrt_implementation/inc/api/compile_options.h
@@ -92,6 +92,22 @@ struct CompileOptions {
   // precision for all operations, which can improve accuracy for some models.
   bool enable_const_eval_on_cpu = true;
 
+  // Forces const-eval function inputs to be annotated as system memory in
+  // the executable's expected layout.
+  //
+  // ON  (default): consteval inputs stay on host RAM, cached outputs on
+  //                device DRAM. The host-side staging is held for the
+  //                BufferInstance's full lifetime (ensure_layout sees the
+  //                source already on host and skips migration).
+  //
+  // OFF:           consteval inputs migrate to device DRAM on first use,
+  //                so the host-side staging can be released between
+  //                stages. Useful for staged-load workloads (e.g. layer-
+  //                by-layer ship) where host data must be freed between
+  //                stages, at the cost of holding both inputs and cached
+  //                outputs on device DRAM.
+  bool enable_const_eval_inputs_to_system_memory = true;
+
   // Enables transpose + matmul and transpose + linear ops fusion.
   // This controls fusing of transpose + matmul and transpose + linear ops.
   // When disabled, transpose is kept as a separate op which can be constevaled,
--- a/pjrt_implementation/src/api/compile_options.cc
+++ b/pjrt_implementation/src/api/compile_options.cc
@@ -51,6 +51,10 @@ CompileOptions CompileOptions::parse(
   options.enable_const_eval_on_cpu =
       internal::parseBoolOption(compile_options, "enable_const_eval_on_cpu")
           .value_or(options.enable_const_eval_on_cpu);
+  options.enable_const_eval_inputs_to_system_memory =
+      internal::parseBoolOption(compile_options,
+                                "enable_const_eval_inputs_to_system_memory")
+          .value_or(options.enable_const_eval_inputs_to_system_memory);
   options.experimental_enable_permute_matmul_fusion =
       internal::parseBoolOption(compile_options,
                                 "experimental_enable_permute_matmul_fusion")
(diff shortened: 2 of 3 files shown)


pjrt_implementation/inc/api/compile_options.h
```diff
@@ -92,6 +92,21 @@ struct CompileOptions {
   // precision for all operations, which can improve accuracy for some models.
   bool enable_const_eval_on_cpu = true;
 
+  // Forces const-eval function inputs to be annotated as system memory in
+  // the executable's expected layout.
+  //
+  // When ON (default, mirroring the tt-mlir compiler default), the
+  // TTNNConstEvalInputsToSystemMemory pass marks inputs consumed by const-
```
jameszianxuTT: Inputs consumed "only" by const-eval functions?

pjrt_implementation/inc/api/compile_options.h
```diff
@@ -92,6 +92,21 @@ struct CompileOptions {
...
+  // eval functions as host-resident. The runtime then sees them as already
+  // in the right place and never migrates them to device, so the host-side
+  // copy stays alive for the entire model run.
+  //
+  // Setting this to false leaves those inputs with their natural device-
+  // memory annotation, so the runtime migrates them on first use and the
+  // host-side copy is released via RAII. Useful when host RAM headroom is
+  // tight and the host copy is not needed after the first execute.
```
jameszianxuTT: Is there a downside here? Given that we are now unloading tensors to device DRAM. 

I think this can lead to potentially double allocation inside device memory, since from
(1) PJRT we hold onto the source tensor (input to consteval) through and 
(2) mlir runtime consteval tensor cache holds onto the post-consteval tensor (output to consteval) - with lifetime attached/subordinate to (1)

I thought that was the problem that offloading consteval inputs to system memory was intended to solve.

If my understanding is right (please correct if I'm wrong), then this leads to both consteval inputs and outputs landing on device DRAM. That would occupy extra device DRAM and lower headroom; making this feature useful for situations where we have a lot of device DRAM and not much host memory, but where the model easily fits in device DRAM?

jameszianxuTT: I am not opposed to allowing user to flip this default or expose as option through tt-xla, I just want to make sure I understand the consequences and that it's clear from the comments

pjrt_implementation/inc/api/compile_options.h
sshonTT: `only` is correct, I'll update the comment. The pass [docstring](https://github.com/tenstorrent/tt-mlir/blob/fecd1901cb0264e8de2ff7c39b6cd37a34336fd2/include/ttmlir/Dialect/TTNN/Transforms/Passes.td#L496-L497) confirms it. Will update to `consumed only by const-eval functions`

pjrt_implementation/inc/api/compile_options.h
sshonTT: There is a clear trade-off when this is disabled. Pushing inputs to the device means the Device DRAM must hold both consteval inputs and outputs simultaneously for that path. host RAM is freed in exchange. Therefore, this option is primarily useful when the device has sufficient headroom but Host RAM is the bottleneck.

The motivation behind this isn't just host ram tight. By default, host-side staging for multi-device buffers persists for the entire lifetime of the BufferInstance. Since the hasLayout check in ensure_layout returns true when the pass is active, tensor reassignment and thus host memory release never occurs. This causes host memory to accumulate across stages during layer-by-layer shipping, preventing us from staying within host RAM limits even if a single layer fits comfortably.

jameszianxuTT: Thanks for the explanation. I think your change to comment helps to clarify the tradeoff, my concern was just that the tradeoff of extra DRAM requirement of holding DRAM input and output wasn't clear in comments but it is now!

jameszianxuTT (approved):

hshahTT (approved):

Issue #4493, opened 2026-05-01
Can we drop tensor borrowed memory from torch_xla after buffer transfer to device?


jameszianxuTT:
We currently call done with host buffer event at buffer instance destruction (tied to pytorch tensor object destruction -> buffer instance destructor). This is so we can borrow memory from torchxla, which itself has moderately intelligent constructs for sharing shard memory thanks to @mstojkovicTT and @acolicTT. 

When we borrow from torchxla, we are trusting that torchxla will optimally represent tensor shards in host memory and use it as the source of truth (and source of literal data). That has worked well so far.

However, in our current execution path with the changes to create PJRTTensor abstraction, we no longer create a mirrored host/device allocation per bufferInstance, and I think we drop the reference to the CPU tensor (borrowing from torchxla data) once we construct a multidevice host tensor and push it to device. I believe we have no way to recover this pointer as well, making this data effectively dead.

I believe that historically, we designed bufferInstance implementation to keep a persistent allocation on host _and_ device so that if we ever lost the device copy, e.g. due to mesh device reshape, that we would have a restore point and could just re-layout the host copy to device. However, we (1) drop the host copy and (2) pull tensors back to host on mesh reshape; so this is not necessary any more. 
 
As such, I think when we drop the host copy of the PJRTTensor, it should be safe to fully drop the allocation on torch_xla side by calling done with host buffer. Maybe this should even be registered as destructor of host runtime tensor to do this with RAII...

Somewhat related - #4347 - "Transfer" seems a bit ambiguous now that I think of it, does this refer to the time that ptxla is transferring the data to us, the callee, and so the "transfer" time is defined by the caller? or to the time where the callee (us) says the transfer to device is done, which may happen in async. 

Relevant to: https://tenstorrent.slack.com/archives/C0AV6L239SQ/p1777658232990129?thread_ts=1777656200.055009&cid=C0AV6L239SQ. Thanks to @sshonTT for making this interesting observation.

jameszianxuTT:
> However, in our current execution path with the changes to create PJRTTensor abstraction, we no longer create a mirrored host/device allocation per bufferInstance, and I think we drop the reference to the CPU tensor (borrowing from torchxla data) once we construct a multidevice host tensor and push it to device. I believe we have no way to recover this pointer as well, making this data effectively dead.

[...]

PR #4594, opened 2026-05-08
[pjrt] release host source after layout migration

sshonTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/4592)

### Problem description
Under host buffer semantics where the plugin keeps the source alive until transfer completes (kImmutableUntilTransferCompletes), the on-done callback registered at copyFromHost time would otherwise only fire in deleteData(), i.e. when the BufferInstance is being destroyed. For workloads that issue many host->device transfers sharing one BufferInstance lifetime, that delays the framework-side(torch-xla) release until model teardown.

### What's changed
Adds BufferInstance::fireDoneWithHostBufferEvent() and calls it from PjrtTensor::ensure_layout right after toLayout migrates the runtime tensor to the requested (device) layout.
Firing the event right after toLayout completes lets the framework drop its source reference promptly. The helper is idempotent (no-op when the event was already fired or never set), so it is safe under both host buffer semantics.

### Checklist

### DIFF
--- a/pjrt_implementation/src/api/tensor.cc
+++ b/pjrt_implementation/src/api/tensor.cc
@@ -85,6 +85,16 @@ void PjrtTensor::ensure_layout(const tt::runtime::Device &device,
   const bool retain = tt::runtime::getTensorRetain(m_runtime_tensor);
   m_runtime_tensor =
       tt::runtime::toLayout(m_runtime_tensor, device, layout, retain);
+
+  // Notify each shard so the host-buffer-done event fires now and the
+  // framework releases its source reference, instead of holding it for
+  // the BufferInstance's full lifetime. No-op when the event was already
+  // fired or never set.
+  for (BufferInstance *shard : m_shards) {
+    if (shard != nullptr) {
+      shard->fireDoneWithHostBufferEvent();
+    }
+  }
 }
 
 // Moves pjrt tensor to host.
--- a/pjrt_implementation/inc/api/buffer_instance.h
+++ b/pjrt_implementation/inc/api/buffer_instance.h
@@ -113,6 +113,21 @@ public:
   // Deletes the buffer data.
   void deleteData();
 
+  // Marks the host-buffer-done event ready, releasing the framework-side
+  // reference to the source host buffer.
+  //
+  // For host buffer semantics that require the plugin to keep the source
+  // alive until transfer completes (e.g. kImmutableUntilTransferCompletes),
+  // the on-done callback is registered at copyFromHost time but only fires
+  // when the BufferInstance is destroyed. Calling this method earlier --
+  // e.g. once the runtime has migrated the data to device and no longer
+  // needs the host source -- lets the framework release that source
+  // promptly instead of holding it for the BufferInstance's lifetime.
+  //
+  // Safe to call multiple times: no-op when the event is null
+  // (semantics that fire eagerly inside copyFromHost) or already ready.
+  void fireDoneWithHostBufferEvent();
+
   // This method should asynchronously copy data into device buffer from the
   // given host buffer. Currently our runtime expects all input buffers to be on
   // host and to be copied to device during execution, because it needs to read
--- a/pjrt_implementation/src/api/buffer_instance.cc
+++ b/pjrt_implementation/src/api/buffer_instance.cc
@@ -171,6 +171,35 @@ void BufferInstance::deleteData() {
   }
 }
 
+void BufferInstance::fireDoneWithHostBufferEvent() {
+  std::lock_guard<std::mutex> deleted_lock(m_data_deleted_mutex);
+  if (m_done_with_host_buffer_event == nullptr) {
+    return;
+  }
+  if (m_done_with_host_buffer_event->isReady()) {
+    m_done_with_host_buffer_event = nullptr;
+    return;
+  }
+  TT_FATAL(m_done_with_host_buffer_event->isIndestructible(),
+           "Expected done_with_host_buffer_event to be indestructible");
+
+  // Same cleanup-on-callback hack used by deleteData(): the callback
+  // deletes the EventInstance after the framework callback chain runs
+  // on a separate thread (see comment in deleteData).
+  m_done_with_host_buffer_event->onReady(
+      [](PJRT_Error *error, void *user_arg) {
+        delete reinterpret_cast<EventInstance *>(user_arg);
+      },
+      m_done_with_host_buffer_event);
+
+  EventInstance::markAsReadyAndCallback(m_done_with_host_buffer_event,
+                                        tt_pjrt_status::kSuccess);
+
+  // Null out so a later deleteData() does not try to fire it again
+  // (which would assert on isReady()).
+  m_done_with_host_buffer_event = nullptr;
+}
+
 // Constructing buffer instance for the first time.
 void BufferInstance::copyFromHost(
     const void *host_buffer, PJRT_Buffer_Type data_type,


pjrt_implementation/src/api/tensor.cc
```diff
@@ -85,6 +85,19 @@ void PjrtTensor::ensure_layout(const tt::runtime::Device &device,
...
+  // buffer is no longer needed by this tensor. Notify each shard so that
+  // any pending host-buffer-done event fires now and the framework can
+  // release its source reference, instead of holding it for the
+  // BufferInstance's full lifetime. No-op for shards whose event was
+  // already fired (or never set, e.g. semantics that release host data
+  // eagerly inside copyFromHost).
+  for (BufferInstance *shard : m_shards) {
+    if (shard != nullptr) {
+      shard->fireDoneWithHostBufferEvent();
+    }
```
jameszianxuTT, line 96: Maybe I misunderstood the discussion on #4493, but how does this work for consteval inputs? 

Won't it allow torch_xla to deallocate the borrowed tensor memory backing consteval host input tensors before consteval can use it?

jameszianxuTT: Unless we have https://github.com/tenstorrent/tt-xla/pull/4593 set to disable the consteval system memory offload?

sshonTT: Good catch,  let me walk through both cases.
The [early `hasLayout` check](https://github.com/tenstorrent/tt-xla/pull/4594/changes#diff-05e97220ef3a9724deb4cca31d2159f17b550e78c56203cdb0f31cc369449ca9L82-L83) is what makes this safe in both modes:

 - Consteval inputs path (system-memory offload ON):
 The pass annotates consteval inputs as SystemMemory in the executable's expected layout. ensure_layout sees the source already matches (host vs host), hasLayout returns true, the function early-returns. toLayout never runs, the fire loop never runs, the borrowed at::Tensor stays alive — consteval pulls it to device via its own internal to_device op when LoadCachedOp fires, and the framework only releases the host data when the BufferInstance is torn down (today's behavior).
 --> No change in this PR for the default consteval path.

 - 4593 disabled path (system-memory offload OFF):
 Consteval inputs keep their natural device-memory annotation, so the expected layout is device. ensure_layout sees a layout mismatch, runs toLayout, the host-resident borrowed tensor is migrated to device, and the resulting device tensor is what the consteval function (and cache) operate on. By the time we fire the on-done event, the device copy already exists and the borrowed host pointer is no longer needed by anyone, including consteval, since it now reads from the device tensor.
 --> Safe to fire here, that's the whole point of the option.

jameszianxuTT: Oh, sorry, I missed that part about has_layout doing an early return :)

jameszianxuTT:
Also closes #4493

pjrt_implementation/src/api/tensor.cc
jameszianxuTT: Maybe someone else on runtime (@pilkicTT ?) can comment a bit more here, but I wonder if we can be a little more defensive in this implementation. 

From my understanding - Framework tensor lifetime is non-RAII, we need to explicitly signal when to deallocate borrowed memory. Here, we manually signal because we know that to_layout will drop the runtime tensor which borrows framework memory, so now the framework tensor data ptr is dangling with no users -> so we feel it is safe to signal deallocation. But I think it may be safer to detect that the framework tensor data ptr is dangling and signal automatically. 

Some options for safety:
- Check that the new tensor layout on m_runtime_tensor is actually on device, gate this fireDoneWithHostBufferEvent with some boolean `canSafelyDropFrameworkHandle` for a better guarantee
- Some optional, custom destructor callback for CreateBorrowedHostTensor that we can attach `fireDoneWithHostBufferEvent` payload to? -- would require more runtime changes, but maybe a better long term solution? I can look at complexity of adding a `tt::runtime::Tensor attachOnDestructionHook(void fn ...)`

Sorry if my comments are too pedantic. I see that from the tests, this does work and I don't want to block what is a useful feature. But any incremental safety improvements especially if they are cheap, seems prudent

jameszianxuTT: Seems like mlir TTNN runtime tensor already has OnDestroyTensorCallback, but not exposed to public API

sshonTT: Not pedantic at all, these are excellent safety considerations.

Regarding option (b), I actually prototyped a similar RAII approach on a side branch (exposing a public registerOnDestroyCallback and firing the event from the wrapper destructor), but it triggered a segfault in memcpy_to_device.

The issue is that toLayout(host->device) enqueues the copy on the device Command Queue and returns before the actual transfer is finished. With a naive RAII implementation, the old host wrapper is destroyed immediately at the m_runtime_tensor reassignment site, triggering the 'done' event prematurely. The framework then deallocates the host buffer while the device is still reading from it.

For (b) to be truly safe, we need a way to defer the 'done' signal until the queued copy actually completes, without blocking the host thread. Since ttnn::MeshEvent currently only supports blocking synchronization or polling, a clean non-blocking implementation would likely require an EventOnReady-style primitive at the metal layer.

In the meantime, I agree that option (a) so pushed the change. Fire is now gated behind guard.

jameszianxuTT: > The issue is that toLayout(host->device) enqueues the copy on the device Command Queue and returns before the actual transfer is finished. With a naive RAII implementation, the old host wrapper is destroyed immediately at the m_runtime_tensor reassignment site, triggering the 'done' event prematurely. The framework then deallocates the host buffer while the device is still reading from it.

Thanks for proving that out and providing your findings ... that is really annoying. 

I guess we could double buffer (memcpy to a safe-transfer area prior to launching async transfer) in toLayout transiently when moving to device to deal with this issue, though that probably comes with its own set of problems. I am pretty stumped here then, as I really want to limit the added complexity for a small safety guarantee. 

I am okay with this as it is after adding (a)  [did you push?], but I think more feedback from other runtime stakeholders who know more about C++ than me would be good 😄

jameszianxuTT (approved):
Approving, though I think feedback from someone else with PJRT core / runtime knowledge is helpful

pjrt_implementation/src/api/tensor.cc
pilkicTT: i am not sure if i am missing something, but in this change we are firing the event immediatelly after `toLayout()`, so this issue can happen here as well:

>The issue is that toLayout(host->device) enqueues the copy on the device Command Queue and returns before the actual transfer is finished. With a naive RAII implementation, the old host wrapper is destroyed immediately at the m_runtime_tensor reassignment site, triggering the 'done' event prematurely. The framework then deallocates the host buffer while the device is still reading from it.

am i missing something?

pjrt_implementation/src/api/tensor.cc
```diff
@@ -85,6 +85,22 @@ void PjrtTensor::ensure_layout(const tt::runtime::Device &device,
   const bool retain = tt::runtime::getTensorRetain(m_runtime_tensor);
   m_runtime_tensor =
       tt::runtime::toLayout(m_runtime_tensor, device, layout, retain);
+
+  // Only signal release when toLayout actually produced the requested
+  // layout; otherwise keep the source alive until BufferInstance teardown.
+  if (!tt::runtime::hasLayout(m_runtime_tensor, layout)) {
```
pilkicTT: i don't see how this if condition can ever be satisfied? after `toLayout` the expectation is that the returned tensor is of the same layout as requested.

sshonTT: Good catch. We don't need this condition. I'll remove this one.

edit)
This is very naive (or bad) version of the guard, but now that it seems no longer necessary, I will remove anyway for this reason.

pjrt_implementation/src/api/tensor.cc
sshonTT: @pilkicTT 
That's a valid concern. I've dig into the tt-metal side to confirm.

Slight correction on the mechanism: toLayout doesn't actually return before the transfer is finished. 
tt-metal's per-shard staging is synchronous - [DeviceCommand::add_data](https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/impl/dispatch/device_command.cpp#L850-L855) does a memcpy from the user buffer into device, and enqueue_write_shards waits on the dispatch thread pool before returning from [here](https://github.com/tenstorrent/tt-metal/blob/15df477236fe9400a9a6fa6c8d1a73faaf42898e/tt_metal/distributed/mesh_command_queue_base.cpp#L266). 
By the time toLayout returns, all per-shard memcpys are done and the device dispatch only reads from sysmem (cmd_region), not the original user buffer.

So firing at the m_runtime_tensor reassignment site (right after toLayout returns) is actually safe, the user buffer is no longer referenced by anyone at that point.

Where RAII still broke for me that the host wrapper can be destroyed not just at the reassignment site, but also at intermediate copy endpoints inside toLayout / convertTensorLayout (e.g. a wrapper copy going out of scope mid-call, before metal's add_data has run for that shard). Those endpoints are what fired the on_done event too early, not the post-toLayout reassignment.

Sorry for the ambigous and misleading.

@jameszianxuTT 
I think we are no longer necessary guard logic because of this.

pilkicTT: ok, let's go with this approach in that case.

pilkicTT (approved):

PR #4574, opened 2026-05-07
Add triage skill for bfloat16 dtype-mismatch FE failures

agobeljicTT:
> [!NOTE]
> This PR complements [tenstorrent/tt-forge-models#641](https://github.com/tenstorrent/tt-forge-models/pull/641), which contains the model-side fixes (loader `dtype_override` propagation + one missing unpack registry entry) produced by running this skill. The YAML status updates here reflect the new failure stages observed once those fixes are in.

### Ticket
Refs: [#4293](https://github.com/tenstorrent/tt-xla/issues/4293) (kept open — this is the second pattern of FE failures being triaged).

### Problem description
A cluster of training tests was stuck at `FAILED_FE_COMPILATION` with a `BFloat16` `RuntimeError`. Two distinct flavors hide behind that wording:

- **Cross-dtype operands** — e.g. `mat1 and mat2 must have the same dtype, but got Float and BFloat16`, `Index put requires the source and destination dtypes match, got Float for the destination and BFloat16 for the source.`, `expected scalar type Float but found BFloat16`. Caused by a tensor that the loader constructs without honoring `dtype_override`. The runner forces `dtype_override=torch.bfloat16` for both training and inference (`tests/runner/utils/dynamic_loader.py:362-419`); any `torch.zeros(...)` / `torch.ones(...)` / `torch.tensor(...)` without an explicit `dtype=` stays float32 and triggers the mismatch downstream. Loader-fixable.
- **Op-not-implemented for BFloat16** — e.g. `'deformable_im2col' not implemented for 'BFloat16'`. The PyTorch op simply has no bfloat16 kernel. Not loader-fixable; xfail with the verbatim error.

Triaging these by hand is mechanical and looks the same every time: reproduce on plain CPU, classify the flavor, propagate `dtype_override` into the offending constructor (cross-dtype) or skip straight to xfail (op-not-implemented), then re-run the affected variants and update the YAML to reflect the post-fix outcome.

### What's changed
- **New Claude Code skill `.claude/skills/triage-dtype-bfloat16/SKILL.md`** that handles exactly this failure pattern, invoked with a YAML key or `model_info.name`. The skill:
  1. Resolves the test name to a YAML entry + loader path; aborts if the failure reason is not the bfloat16 pattern.
  2. Classifies the flavor (cross-dtype operands vs op-not-implemented).
[...]

PR #4609, opened 2026-05-11
Add triage skill for missing-input FE failures

pglusacTT:
> [!NOTE]
> Stacked on [#4574](https://github.com/tenstorrent/tt-xla/pull/4574) (pattern 2 — bfloat16 dtype). Complements [tenstorrent/tt-forge-models#642](https://github.com/tenstorrent/tt-forge-models/pull/642), which contains the loader fixes (`decoder_input_ids` / `targets` / `batch_size=2`) produced by running this skill. The YAML status updates here reflect the new failure stages observed once those fixes are in.

### Ticket
Refs: [#4293](https://github.com/tenstorrent/tt-xla/issues/4293) (kept open — this is the third pattern of FE failures being triaged).

### Problem description
A cluster of training tests was stuck at `FAILED_FE_COMPILATION` with one of these reasons:

- `ValueError: You have to specify either decoder_input_ids or decoder_inputs_embeds.`
- `Model expects targets to be passed while in training mode`
- `AssertionError: targets should not be none when in training mode`
- `AttributeError: 'NoneType' object has no attribute 'max'`
- `ValueError: Expected more than 1 value per channel when training, got input size torch.Size([1, 256, 1, 1])`

All five are the same root cause: in `model.train()` mode the model requires an input that the loader is not passing. The runner (`tests/runner/utils/dynamic_loader.py:362-419`) calls `load_inputs` once before `model.train()` is set (`tests/infra/testers/single_chip/model/model_tester.py:68,82`), so the loader cannot branch on run mode — any added input must be safe for inference too. Three distinct sub-flavors hide behind the wording:

- **`decoder_inputs`** — HF encoder-decoder LMs (Whisper, T5, BART, MarianMT, …) called without `decoder_input_ids`. Often caused by a `load_inputs` that returns a positional `list` where the decoder slot lands in the `attention_mask` position; the model still raises. Loader-fixable by switching to a `dict` return that names `decoder_input_ids`.
- **`missing_targets`** — torchvision detection models (`SSD`, `SSDLite`, `RetinaNet`, …) that require `targets=[{"boxes": …, "labels": …}, …]` in `train()` mode. Loader-fixable by returning `{"images": …, "targets": …}`; `eval()` ignores `targets`, so inference is unaffected.
- **`bn_single_value`** — `BatchNorm{1,2,3}d` rejects inputs where every non-channel dim is 1 (e.g. `[1, 256, 1, 1]` after a deeplabv3 ASPP global pool on `batch=1`). Sometimes loader-fixable by setting `batch_size=2`; not fixable if a feature map collapses to 1 spatially regardless of batch.

Triaging these by hand is mechanical and looks the same every time: reproduce on plain CPU, classify the sub-flavor, edit the loader, re-run the affected variants, update the YAML.

### What's changed
- **New Claude Code skill `.claude/skills/triage-inputs/SKILL.md`** that handles exactly this failure pattern, invoked with a YAML key or `model_info.name`. The skill:
  1. Resolves the test name to a YAML entry + loader path; aborts if the failure reason is not one of the five strings above.
  2. Classifies the sub-flavor (`decoder_inputs` / `missing_targets` / `bn_single_value`).
  3. Reproduces on plain CPU with `model.train()` set (no TT involvement) to confirm the same error.
  4. For `decoder_inputs` and `missing_targets`: edits `load_inputs` in tt-forge-models to return a dict that names the missing key, then re-runs CPU to confirm the fix. For `bn_single_value`: tries `batch_size=2` only if the inference YAML entry tolerates it; otherwise goes straight to xfail.
  5. Updates **every** training entry sharing the affected loader (loader fixes are loader-wide), per the observed post-fix outcome — `EXPECTED_PASSING` if now passing, `KNOWN_FAILURE_XFAIL` with the new error if it advances to a different failure, `NOT_SUPPORTED_SKIP` retained when the new failure is a process-crashing tt-mlir bug.
  6. Never edits the inference YAML or `tests/runner/utils/dynamic_loader.py`.
- **YAML updates** in `tests/runner/test_config/torch/test_config_training_single_device.yaml` reflecting the new stages observed after the companion tt-forge-models PR is in:
  - `whisper/pytorch-{Tiny,Base,Small,Medium,Large,Large_v3_Turbo}`: loader switched to a dict return that includes `decoder_input_ids`; tests now advance past FE and xfail on `TT_THROW: Failed to generate binaries for eltwise_sfpu (trisc1 build failed)`. `NOT_SUPPORTED_SKIP` → `KNOWN_FAILURE_XFAIL`.
  - `yolox/pytorch-{Nano,Tiny,S,M,L,Darknet}`: loader now returns `{"x": …, "targets": …}` with a synthetic single-label tensor; tests now xfail on `Unhandled FakeTensor Device Propagation for aten.add.Tensor` (xla:0 vs cpu). `NOT_SUPPORTED_SKIP` → `KNOWN_FAILURE_XFAIL`.
  - `ssdlite320_mobilenetv3/pytorch-Ssdlite320_Mobilenet_v3_Large`: loader returns `{"images": …, "targets": …}` with synthetic targets; now hits an OOM at runtime. `NOT_SUPPORTED_SKIP` → `KNOWN_FAILURE_XFAIL` with the OOM reason.
  - `ssd300_vgg16/pytorch-Default`, `ssd300_resnet50/pytorch-Base`: targets fix lands; both now surface a tt-mlir `ElementsAttr::getValues UNREACHABLE on dense<4294967295> : tensor<ui32>` crash. Status kept at `NOT_SUPPORTED_SKIP` because the failure crashes the process before pytest can xfail it.
  - `mobilenetv2/pytorch-Deeplabv3_Mobilenet_v2_1.0_513`: `batch_size=2` unblocks BN; the test now compiles but times out. Status kept at `NOT_SUPPORTED_SKIP`, reason updated.
- **Tweaks to `.claude/skills/triage-dtype-bfloat16/SKILL.md`** carried in this branch: simplified the Step 7 verification flow to a single \"flip → run → write\" pattern matching `triage-unpack-forward-output`, instead of the previous `--force-run --runxfail` + JUnit-XML dance. These belong on top of #4574 logically; they're surfaced here only because both skills were iterated on together.

### Checklist
- [x] New/Existing tests provide coverage for changes — the affected training tests are the ones whose YAML statuses are updated; each was re-run as part of the skill's verification step.


pglusacTT:
Not opened by me... @agobeljicTT

PR #4610, opened 2026-05-11
Don't download weights in GPT OSS training test

### LINKED ISSUE #4595: Don't download GPT-OSS-20B weights on CIv2 - test timeouts observed
Re: https://github.com/tenstorrent/tt-xla/pull/4181/changes

I've seen several failures recently showing that this test sees a timeout, likely in downloading full model weights https://github.com/tenstorrent/tt-xla/actions/runs/25509187014/job/74869874335

Let's use fake weights if possible, or if really necessary, only download a subset of the weights needed for the limited test. 

I don't think Civ2 model weight caching infra is there yet so we don't have that path right now. 

Also we should cap this test t/o to much less than 4 hours... seems like successful run should take a few minutes.

pglusacTT:
### Ticket
Closes https://github.com/tenstorrent/tt-xla/issues/4595

### Problem description
Model weights get loaded since weight caching doesn't work in CIv2. 

### What's changed
Randomly init weights.

### Checklist

### DIFF
--- a/tests/torch/training/test_moe.py
+++ b/tests/torch/training/test_moe.py
@@ -9,6 +9,7 @@ import torch_xla.distributed.spmd as xs
 import torch_xla.runtime as xr
 from infra.utilities.torch_multichip_utils import enable_spmd
 from torch_xla.distributed.spmd import Mesh
+from transformers import AutoModelForCausalLM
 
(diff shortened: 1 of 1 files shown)


jameszianxuTT (approved):

jameszianxuTT:
Thanks!

PR #4617, opened 2026-05-11
Implement deferred transfer of host tensor data to multihost workers

### LINKED ISSUE #4011: Multihost CreateOwnedHostTensor causes host memory waste scaling by device count on workers
The optimizations we have for singlehost sharing of tensor memory do not work across a host boundary, so a graph like this causes host memory usage scaling over 

(replicated x+=1 over 64 devices, x is 1GB tensor `16384x32768xbf16`) induces a 1GB host allocation per shard!)

```
#loc1 = loc("p0.3")
module @SyncTensorsGraph.11 attributes {mhlo.cross_program_prefetches = [], mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
  sdy.mesh @mesh = <["x"=1, "y"=1]> loc(#loc)
  func.func @main(%arg0: tensor<16384x32768xbf16> {mhlo.sharding = "{replicated}", ttcore.argument_type = #ttcore.argument_type<input>, ttcore.local_shape = #ttcore<local_shape local_shape = tensor<16384x32768xbf16>>, ttcore.shard_status = #ttcore.shard_status<presharded>, ttir.name = "args_0"} loc("p0.3")) -> (tensor<16384x32768xbf16> {ttcore.local_shape = #ttcore<local_shape local_shape = tensor<16384x32768xbf16>>, ttcore.shard_status = #ttcore.shard_status<presharded>}) {
    %cst = stablehlo.constant dense<1.000000e+00> : tensor<bf16> loc(#loc)
    %0 = stablehlo.broadcast_in_dim %cst, dims = [] : (tensor<bf16>) -> tensor<16384x32768xbf16> loc(#loc)
    %1 = stablehlo.reshape %arg0 : (tensor<16384x32768xbf16>) -> tensor<1x16384x32768xbf16> loc(#loc2)
    %2 = stablehlo.reshape %1 : (tensor<1x16384x32768xbf16>) -> tensor<16384x32768xbf16> loc(#loc3)
    %3 = stablehlo.add %2, %0 : tensor<16384x32768xbf16> loc(#loc4)
    return %3 : tensor<16384x32768xbf16> loc(#loc)
  } loc(#loc)
} loc(#loc)
#loc = loc(unknown)
#loc2 = loc("reshape.4")
#loc3 = loc("reshape.6")
#loc4 = loc("add.9")
```

Resolution - some form of host buffer caching/weight sharing in multihost worker context

jameszianxuTT:
### Ticket
Closes #4011 

### Problem description
Multihost workers currently lack a mechanism for buffer data aliasing. While the host controller has a shared ownership model (borrowing allocations from caller) workers materialize every shard and replica as an independent allocation leading to extreme host memory waste scaling with device count.

The root issue: at the current copyFromHost callsite, torch_xla doesn't provide enough information to identify which BufferInstances share the same source data.

### What's changed
1. Introduce HostTensorShell: A struct that captures copyFromHost arguments (host buffer pointer, shape, strides, dtype) without materializing a runtime tensor. Acts as deferred metadata until execution time, as a member of PJRTTensor
2. Deduplicate at execution time: In prepareInputTensor, group BufferInstances by their host buffer pointer address.
For each group:
- First buffer gets an owned tensor (actual copy from host)
- Subsequent buffers get createUnsafeBorrowedHostTensor aliasing the first, avoiding redundant copies
3. Multihost-only codepath: This behavior is gated on getCurrentHostRuntime() == Distributed. Single-host runtime is unchanged.

Discussion and comparison of alternative approaches was made in [this RFC.](https://docs.google.com/document/d/1n-_KaQo9yVHLazT8HPJaHyIbDufTe5OqroLEnAME_rc/edit?tab=t.0)

### Checklist
- [x] Existing multihost tests passing https://github.com/tenstorrent/tt-xla/actions/runs/25684369837

### DIFF
--- a/pjrt_implementation/src/api/buffer_instance.cc
+++ b/pjrt_implementation/src/api/buffer_instance.cc
@@ -12,6 +12,7 @@
 
 // c++ standard library includes
 #include <cstring>
+#include <cstddef>
 #include <memory>
 #include <numeric>
 #include <optional>
@@ -125,6 +126,27 @@ size_t BufferInstance::logicalTensorSize() const {
                          });
 }
 
+bool BufferInstance::aliasesSameBorrowedHostBase(
+    const BufferInstance &other) const {
+  const void *a = m_borrowed_host_base_ptr;
+  const void *b = other.m_borrowed_host_base_ptr;
+  return a != nullptr && a == b;
+}
+
+bool BufferInstance::borrowedHostByteRangesOverlap(const BufferInstance &a,
+                                                   const BufferInstance &b) {
+  const void *pa = a.m_borrowed_host_base_ptr;
+  const void *pb = b.m_borrowed_host_base_ptr;
+  if (pa == nullptr || pb == nullptr) {
+    return false;
+  }
+  const auto *ba = static_cast<const std::byte *>(pa);
+  const auto *bb = static_cast<const std::byte *>(pb);
+  const size_t sa = a.logicalTensorSize();
+  const size_t sb = b.logicalTensorSize();
+  return ba < bb + sb && bb < ba + sa;
+}
+
 std::string BufferInstance::toShapeStr() const {
   return tt::pjrt::utils::to_string(m_dimensions);
 }
@@ -214,6 +236,7 @@ void BufferInstance::copyFromHost(
       data_type_utils::getPJRTBufferTypeString(data_type));
 
   m_pjrt_tensor.reset();
+  m_borrowed_host_base_ptr = nullptr;
 
   ::tt::target::DataType runtime_data_type =
       tt::pjrt::data_type_utils::convertPJRTToRuntimeDataType(m_data_type);
@@ -269,6 +292,8 @@ void BufferInstance::copyFromHost(
         const_cast<void *>(host_buffer), shape, strides, element_size,
         runtime_data_type);
 
+    m_borrowed_host_base_ptr = host_buffer;
+
     // Memory is aliased, we need to hold on to host buffer until this buffer is
     // deleted.
     m_done_with_host_buffer_event = done_with_host_buffer_event.get();
@@ -291,6 +316,8 @@ void BufferInstance::copyFromHost(
 void BufferInstance::copyFromBuffer(BufferInstance *src_buffer) {
   DLOG_F(LOG_DEBUG, "BufferInstance::copyFromBuffer");
   TT_FATAL(src_buffer->getPjrtTensor(), "Source buffer has no data.");
+
+  m_borrowed_host_base_ptr = nullptr;
 
   ::tt::target::DataType runtime_data_type =
       tt::pjrt::data_type_utils::convertPJRTToRuntimeDataType(
--- a/pjrt_implementation/src/api/buffer_instance.cc
+++ b/pjrt_implementation/src/api/buffer_instance.cc
@@ -259,7 +259,7 @@ void BufferInstance::copyFromHost(
         host_buffer, shape, strides, element_size, runtime_data_type};
     m_done_with_host_buffer_event = done_with_host_buffer_event.get();
     m_done_with_host_buffer_event->setIndestructible();
-    PjrtTensor::from_host_tensor_shell({this}, std::move(*host_tensor_shell));
+    PjrtTensor::from_host_tensor_shell({this}, std::move(host_tensor_shell));
   } else {
     if (cannot_borrow_host_buffer) {
       runtime_tensor = tt::runtime::createOwnedHostTensor(
--- a/pjrt_implementation/src/api/buffer_instance.cc
+++ b/pjrt_implementation/src/api/buffer_instance.cc
@@ -230,21 +230,17 @@ void BufferInstance::copyFromHost(
 
   tt::runtime::Tensor runtime_tensor;
 
-  // In distributed runtime, for ImmutableOnlyDuringCall semantics, or for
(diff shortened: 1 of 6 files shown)


third_party/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 option(USE_CUSTOM_TT_MLIR_VERSION "Flag to use TT_MLIR_VERSION set by the user" OFF)
 
 if (NOT DEFINED TT_MLIR_VERSION OR NOT USE_CUSTOM_TT_MLIR_VERSION)
-    set(TT_MLIR_VERSION "bb1deb417cef0e2c60147072cf7b6926a49ccca7")
+    set(TT_MLIR_VERSION "jzx/deferred_sharding")
```
jameszianxuTT: wait for uplift of tt-mlir

pjrt_implementation/src/api/buffer_instance.cc
```diff
@@ -214,31 +218,45 @@ void BufferInstance::copyFromHost(
...
-          tt::runtime::HostRuntime::Distributed ||
-      host_buffer_semantics ==
+
+  bool is_distributed = ::tt::runtime::getCurrentHostRuntime() ==
+        tt::runtime::HostRuntime::Distributed;
+  
+  bool can_borrow_host_buffer = host_buffer_semantics ==
           PJRT_HostBufferSemantics_kImmutableOnlyDuringCall ||
-      !::tt::runtime::utils::isSupportedDataType(runtime_data_type)) {
+      ::tt::runtime::utils::isSupportedDataType(runtime_data_type);
```
jameszianxuTT, line 252: I've rewritten the logic here to be a little clearer:

in distributed case - we make a shell tensor
in singlehost case - we check if we can borrow - if no - create owned host tensor. If yes - create borrowed host tensor.

jameszianxuTT:
image

Some [test graphs from the RFC](https://docs.google.com/document/d/1n-_KaQo9yVHLazT8HPJaHyIbDufTe5OqroLEnAME_rc/edit?tab=t.0). This is better, but, quoting @pilkicTT, there is still room for improvement

jameszianxuTT:
Update from EOD may 11 - I need another hour or two to review changes, fixed a couple bugs here in the singlehost path... then I will open this up for review tmrw. MLIR changes have been merged but may not be uplifted for some time; I believe test regressions here are likely due to mlir changes

jameszianxuTT:
Note - at commit e5676072ecf7545f8d09d22445502615405cc8cd - we are passing everything except for known failures in tt-mlir uplift.

jameszianxuTT:
This may conflict with #4594, which should be reviewed first.

jameszianxuTT:
Also, as evidence that this actually fixes the original problem - kimi k2 instruct single sparse MOE layer used to really blow up host machine RAM - after this change RSS and runtime are _much_ improved. 

image

pjrt_implementation/src/api/buffer_instance.cc
pilkicTT: can we assert in the distributed case that we can borrow the host buffer? or add handling for case where we need to create an owned host buffer (std::memcpy)?

pjrt_implementation/src/api/flatbuffer_loaded_executable_instance.cc
```diff
@@ -78,6 +85,53 @@ FlatbufferLoadedExecutableInstance::prepareInputTensor(
...
+  std::unordered_map<const void *, std::vector<BufferInstance *>>
+      borrowed_host_base_ptr_to_buffers;
+
+  for (BufferInstance *buf : arg_buffers) {
+    // Buffers without a shell have already been materialized
+    const auto &shell = buf->getPjrtTensor()->host_tensor_shell();
+    if (shell.has_value()) {
+      borrowed_host_base_ptr_to_buffers[shell->host_buffer].push_back(buf);
+    }
+  }
```
pilkicTT, line 100: can we assert that all shells containing the same ptr are indeed equal (shape, data type, etc.)?

pjrt_implementation/src/api/flatbuffer_loaded_executable_instance.cc
```diff
@@ -78,6 +85,53 @@ FlatbufferLoadedExecutableInstance::prepareInputTensor(
   return tensor.runtime_tensor();
 }
 
+void FlatbufferLoadedExecutableInstance::materializeShellTensors(
```
pilkicTT, line 88: this materializes all shells for a particular tensor - can we have same shells which are used by multiple tensors?

jameszianxuTT: I don't think so. The shell captures torchxla's allocation of a tensor shard, which are generated on each pytorch tensor uniquely without any mechanism to share shards between them.

pjrt_implementation/src/api/buffer_instance.cc
jameszianxuTT: Sorry, can you elaborate on this comment? We cannot borrow the host buffer on multihost in any case, so I am ignoring the value of `cannot_borrow_host_buffer` for distributed case.

jameszianxuTT: Added a safety assert against PJRT_HostBufferSemantics_kImmutableOnlyDuringCall; implicitly denying Jax generality for multihost

pilkicTT (approved):

PR #4627, opened 2026-05-11
Fix mesh-shape inference for no-input graphs on multichip hosts

### LINKED ISSUE #4624: Single chip tests can no longer be run in multichip environments
Due to https://github.com/tenstorrent/tt-mlir/pull/8167, which fixes some separate issue with #4053 - We might need to expose some more control knobs from compiler side to infer what should happen when no mesh is specified:

1. Run without a mesh, fully replicated  
2. Run on existing previously opened mesh if available

PR changed behaviour from 1->2, which made multidevice tests that were broken before work, but now breaks singledevice tests in multidevice environment. 

Ideally, we can figure out some solution that allows runtime to automatically infer an unproblematic mesh shape rather than requiring user to select. 

see: https://tenstorrent.slack.com/archives/C08H2NCQGHJ/p1778523635588749?thread_ts=1778523371.928969&cid=C08H2NCQGHJ

pglusacTT:
### Ticket
Closes https://github.com/tenstorrent/tt-xla/issues/4624

### Problem description
Single-chip tests need to be able to run on multichip machines.
This can be achieved by checking the `num_partitions` field [passed by torch-xla](https://github.com/tenstorrent/pytorch-xla/blob/5c82b1034208844b584e6088b2e043838935e875/torch_xla/csrc/runtime/pjrt_computation_client.cpp#L584).

 `num_partitions` is the right discriminator because torch-xla only sets it to `device_count()` on the SPMD branch, the non-SPMD branch keeps `num_partitions=1` and instead sets `num_replicas=device_count()` (each replica is dispatched as its own 1-device `ExecuteSharded` call), so using the product `num_partitions x num_replicas` would mis-classify single-device compiles as multichip.

### What's changed
  - Read `num_partitions` / `num_replicas` from `ExecutableBuildOptionsProto`.
  - Use `target_num_devices` into `runCompilerStableHLOPipeline`.
  - In the no-input branch of `runCompilerStableHLOPipeline`, only pass the parent mesh shape into `AnalyzeMesh` when `target_num_devices > 1`. Single-chip compiles fall back to AnalyzeMesh's default `{1,1}` and the existing reshape path reshapes the parent mesh downstream.
  - Added `tests/torch/graphs/test_mesh_shapes.py` covering 7 cases: single/multi chip x with/without args, sharded vs replicated inputs, and sequential single-device compiles.
  
###  Known Limitation!
Changing mode mid execution is faulty. Tests with different modes (single and multichip) fail when they are ran subsequently. Will need further discussion.

### Checklist

### DIFF
--- a/pjrt_implementation/inc/api/compile_options_parser.h
+++ b/pjrt_implementation/inc/api/compile_options_parser.h
@@ -31,6 +31,14 @@ namespace tt::pjrt {
 //
 // The protobuf layout is defined in:
 // https://github.com/openxla/xla/blob/main/xla/pjrt/proto/compile_options.proto
+struct ExecutableDeviceShape {
+  std::optional<int64_t> num_partitions;
+  std::optional<int64_t> num_replicas;
+
+  // Number of devices the executable will distribute its computation across.
+  int64_t targetNumDevices() const { return num_partitions.value_or(1); }
+};
+
 class CompileOptionsParser {
 public:
   // Parses compile options protobuf and extracts both custom compile options
@@ -38,7 +46,8 @@ public:
   static tt_pjrt_status parseCompileOptions(
       const char *compile_options_data, size_t compile_options_size,
       std::unordered_map<std::string, std::string> &out_compile_options,
-      std::optional<std::vector<int64_t>> &out_replica_device_ids);
+      std::optional<std::vector<int64_t>> &out_replica_device_ids,
+      ExecutableDeviceShape &out_device_shape);
 
   // Parses compile options protobuf data into UnknownFieldSet.
   // Returns true if parsing succeeded, false otherwise.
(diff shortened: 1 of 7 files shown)


AleksKnezevic:
Do you want to add an option for the user to be able to force a mesh size using something similar to [this](https://github.com/tenstorrent/tt-xla/blob/cda3a8132073f37122fd118a7991cb526e8fb628/python_package/torch_plugin_tt/__init__.py#L50)?

acicovicTT:
Assuming the issue occurs for JAX models as well, does the solution work for JAX graphs?

pjrt_implementation/inc/api/compile_options_parser.h
```diff
@@ -31,14 +31,23 @@ namespace tt::pjrt {
 //
 // The protobuf layout is defined in:
 // https://github.com/openxla/xla/blob/main/xla/pjrt/proto/compile_options.proto
+struct ExecutableDeviceShape {
```
acicovicTT, line 34: It seems like the comment above is referring to `CompileOptionsParser`?

PR #4635, opened 2026-05-12
Refactoring the sliding attention using generics and flags.

devisettymahidhar608:
### Ticket

### Problem description
Move sliding-window attention overrides out of tt-forge-models loaders into generic tt-xla test infrastructure

### What's changed

overrides_sliding_attention is an opt-in boolean YAML flag (default false) that tells the test runner to apply tt-xla's TT-friendly sliding-window logic — swap HF's StaticSlidingWindowLayer for the compile-safe TTStaticSlidingWindowLayer in the input cache, and monkey-patch the model's create_sliding_window_causal_mask with the compile-safe version. Set it to true for sliding-window LLMs (olmo3, gpt_oss, ministral, …) that need a torch.compile-friendly attention path on TT hardware.

Note: No code dependencies in the tt-forge-models repository.

```
YAML:  overrides_sliding_attention: true   (per test row)
  │
  ▼ config_loader.py validates against ALLOWED_FIELDS
  ▼ conftest.py builds ModelTestConfig and attaches to item._test_meta
  ▼ test_metadata fixture returns it
  │
  ▼ test_models.py forwards test_metadata to DynamicTorchModelTester
  │
  ▼ DynamicTorchModelTester.__init__:
       │
       ├─ super().__init__() (loads model, builds inputs)
       │    └─ _get_input_activations:
       │         ├─ inputs = loader.load_inputs()         # tokenizer output only
       │         └─ if flag: inputs = self._build_sliding_window_input_args(inputs)
       │                ├─ build HF StaticCache
       │                ├─ early_initialization
       │                ├─ override_cache_sliding_window_layers (TT cache layer swap)
       │                ├─ pad attention_mask to max_cache_len
       │                └─ return input_args dict
       │
       └─ Post-super hooks:
            └─ if flag: self._apply_sliding_window_mask_override(self._model)
                   └─ override_sliding_window_causal_mask(model)
                          ├─ resolve transformers.models.<type>.modeling_<type>
                          └─ patch create_sliding_window_causal_mask
  │
  ▼ tester.test() → torch.compile traces forward with TT-friendly mask
                    and TT-friendly StaticCache layers
```
logs:
[gpt_ooss.log](https://github.com/user-attachments/files/27631214/gpt_ooss.log)
[olmo3_1125_32b.log](https://github.com/user-attachments/files/27631218/olmo3_1125_32b.log)

### Checklist

### DIFF
--- a/examples/pytorch/gpt_oss_20b.py
+++ b/examples/pytorch/gpt_oss_20b.py
@@ -27,7 +27,7 @@ from transformers.modeling_outputs import CausalLMOutputWithPast
 from tt_torch.sharding import sharding_constraint_hook
 from tt_torch.transformers_overrides import (
     override_cache_sliding_window_layers,
-    override_gpt_oss_sliding_window_causal_mask,
+    override_sliding_window_causal_mask,
 )
 
 DEFAULT_PROMPTS = ["Explain quantum mechanics."]
@@ -152,7 +152,7 @@ def setup_model_and_tokenizer(
         attn_implementation="eager",
     )
     # Override the gpt_oss model to use the TT-friendly sliding window causal mask
-    override_gpt_oss_sliding_window_causal_mask()
+    override_sliding_window_causal_mask(model)
     model = model.eval()
 
     tokenizer: PreTrainedTokenizer = AutoTokenizer.from_pretrained(model_name)
--- a/python_package/tt_torch/transformers_overrides.py
+++ b/python_package/tt_torch/transformers_overrides.py
@@ -8,43 +8,20 @@ from transformers.cache_utils import StaticCache, StaticLayer, StaticSlidingWind
 from transformers.masking_utils import create_sliding_window_causal_mask
 
 
-def override_gpt_oss_sliding_window_causal_mask():
+def override_sliding_window_causal_mask(model):
+    """Monkey-patch the create_sliding_window_causal_mask function of the
+    transformers modeling module that backs ``model`` with the TT-friendly
+    compile-safe version.
+
+    Call this AFTER ``from_pretrained`` and BEFORE ``torch.compile`` so that
+    dynamo traces through the patched function.
     """
-    Override gpt_oss's modeling so that its
-    create_sliding_window_causal_mask points to the TT-friendly version.
+    import importlib
 
-    Call this before torch.compile so that dynamo traces through the
-    patched function.
-    """
-    import transformers.models.gpt_oss.modeling_gpt_oss as gpt_oss_mod
-
-    gpt_oss_mod.create_sliding_window_causal_mask = tt_create_sliding_window_causal_mask
-
-
-def override_olmo3_sliding_window_causal_mask():
-    """
-    Override olmo3's modeling so that its
-    create_sliding_window_causal_mask points to the TT-friendly version.
-    """
-    import transformers.models.olmo3.modeling_olmo3 as olmo3_mod
-
-    olmo3_mod.create_sliding_window_causal_mask = tt_create_sliding_window_causal_mask
-
-
-def override_ministral_sliding_window_causal_mask():
-    """
-    Override ministral's modeling so that its
-    create_sliding_window_causal_mask points to the TT-friendly version.
-
-    Note: ministral (mistralai/Ministral-*) uses a separate module
-    transformers.models.ministral.modeling_ministral, distinct from
-    transformers.models.mistral.modeling_mistral.
-    """
-    import transformers.models.ministral.modeling_ministral as ministral_mod
-
-    ministral_mod.create_sliding_window_causal_mask = (
-        tt_create_sliding_window_causal_mask
-    )
+    model_type = model.config.model_type
+    module_name = f"transformers.models.{model_type}.modeling_{model_type}"
+    mod = importlib.import_module(module_name)
+    mod.create_sliding_window_causal_mask = tt_create_sliding_window_causal_mask
 
 
 def override_cache_sliding_window_layers(
--- a/tests/runner/test_config/torch/test_config_inference_single_device.yaml
+++ b/tests/runner/test_config/torch/test_config_inference_single_device.yaml
@@ -1684,6 +1684,7 @@ test_config:
     arch_overrides:
       p150:
         status: EXPECTED_PASSING
+        overrides_sliding_attention: true
 
       n150:
         status: NOT_SUPPORTED_SKIP
@@ -2557,6 +2558,7 @@ test_config:
         reason: "Out of Memory: Not enough space to allocate 52428800 B L1 buffer across 64 banks, where each bank needs to store 819200 B, but bank size is only 1331936 B"
       p150:
         status: EXPECTED_PASSING
+        overrides_sliding_attention: true
 
   olm_ocr/image_text_generation/pytorch-olmOCR-7B-0225-preview-single_device-inference:
     supported_archs: ["p150"]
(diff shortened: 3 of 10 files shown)


devisettymahidhar608:
FYI, olmo3/causal_lm/pytorch-3_1025_7b and mistral/pytorch-Ministral_8B_Instruct are both failing with a device mismatch error on a single device, whereas both models pass in tensor parallel mode.

logs:
[mistral_8b.log](https://github.com/user-attachments/files/27633144/mistral_8b.log)
[olmo_7b.log](https://github.com/user-attachments/files/27633146/olmo_7b.log)

sonalibaskaran2499 (approved):

tests/runner/test_config/torch/test_config_inference_single_device.yaml
```diff
@@ -2557,6 +2558,7 @@ test_config:
         reason: "Out of Memory: Not enough space to allocate 52428800 B L1 buffer across 64 banks, where each bank needs to store 819200 B, but bank size is only 1331936 B"
       p150:
         status: EXPECTED_PASSING
+        overrides_sliding_attention: true
```
AleksKnezevic, line 2561: I think this should be a flag in tt-forge-models. @kmabeeTT thoughts?

examples/pytorch/gpt_oss_20b.py
```diff
@@ -152,7 +152,7 @@ def setup_model_and_tokenizer(
         attn_implementation="eager",
     )
     # Override the gpt_oss model to use the TT-friendly sliding window causal mask
-    override_gpt_oss_sliding_window_causal_mask()
+    override_sliding_window_causal_mask(model)
```
AleksKnezevic, line 155: we can avoid the complexity of importlib and just do something like

`model.create_sliding_window_causal_mask = tt_create_sliding_window_causal_mask`

tests/runner/testers/torch/dynamic_torch_model_tester.py
```diff
@@ -298,3 +318,112 @@ def combined_shard_spec_fn(model, _fn=shard_spec_fn, _names=mesh_names):
                 return get_moe_shard_specs(model, _fn, _names)
 
             self._workload.shard_spec_fn = combined_shard_spec_fn
+
+    def _build_sliding_window_input_args(self, inputs):
```
AleksKnezevic, line 322: shouldn't building inputs be in tt-forge-models?

AleksKnezevic (changes requested):

tests/runner/testers/torch/dynamic_torch_model_tester.py
devisettymahidhar608: @AleksKnezevic , I hope that editing the inputs in TT-XLA is the more optimized approach. If we start making changes in TT-Forge-Models, we would need to update all the loaders as well, which is not an optimized way to handle it.

Advantages of keeping this function in TT-XLA:

- We can enable overrides_sliding_attention=True for any model and run it directly without needing to modify the loaders again.

- Currently, we only have two models (olmo3 and mistral) that require loader changes in TT-Forge-Models. However, in the future, we may have many more models with sliding attention support. Updating loaders individually is not the right approach, as it breaks uniformity across TT-Forge-Models. By maintaining a single build_input function in TT-XLA, changes can be applied consistently across all models at the same time.

Thank You

tests/runner/test_config/torch/test_config_inference_single_device.yaml
devisettymahidhar608: Disadvantages of keeping this flag in TT-Forge-Models:

We would need to apply the same changes across every model variant, which is not an optimized or scalable approach.
loader.py in TT-Forge-Models would lose its uniform structure over time.

So, I was thinking we should implement this in a more optimized way in TT-XLA, such that any future model can support these changes with minimal code modifications by simply enabling overrides_sliding_attention=True.

Thank You

examples/pytorch/gpt_oss_20b.py
devisettymahidhar608: Let me check this once, Aleks.

Thank You

kmabeeTT: Agreed, was expecting trimmed version of original code from override_gpt_oss_sliding_window_causal_mask() could just be used here.

tests/runner/test_config/torch/test_config_inference_single_device.yaml
kmabeeTT: Agreed with Aleks, yeah I was expecting flag to be optional new flag in just the few models that need it in tt-forge-models rather than here at test config time. Here it is uglier and needs to be maintained per arch, and per variant, when it should just come with the model loader.py itself,  and used by the infra in tt-xla.  Can you try moving it to tt-forge-models?

python_package/tt_torch/transformers_overrides.py
```diff
@@ -8,43 +8,20 @@
...
-    """
-    import transformers.models.ministral.modeling_ministral as ministral_mod
-
-    ministral_mod.create_sliding_window_causal_mask = (
-        tt_create_sliding_window_causal_mask
-    )
+    model_type = model.config.model_type
+    module_name = f"transformers.models.{model_type}.modeling_{model_type}"
+    mod = importlib.import_module(module_name)
+    mod.create_sliding_window_causal_mask = tt_create_sliding_window_causal_mask
```
kmabeeTT, line 24: Not sure how safe this is for *all* models, was originally thinking/suggesting in [proposal](https://github.com/tenstorrent/tt-xla/pull/4287#discussion_r3199055366) a scan of the entire model to replace instances of create_sliding_window_causal_mask with our overriden version and it would be less sensitive to module names for the import and patching than above.

tests/runner/testers/torch/dynamic_torch_model_tester.py
```diff
@@ -298,3 +318,112 @@ def combined_shard_spec_fn(model, _fn=shard_spec_fn, _names=mesh_names):
...
+        return {
+            "input_ids": input_ids,
+            "past_key_values": static_cache,
+            "cache_position": cache_position,
+            "use_cache": True,
+            "attention_mask": full_attention_mask,
+        }
+
+    @staticmethod
+    def _apply_sliding_window_mask_override(model):
```
kmabeeTT, line 399: With idea to scan model and replace instances, this could maybe be entirely removed and we just call override_sliding_window_causal_mask() for the models that ask for it. If any new exception checking needed can add directly into override_sliding_window_causal_mask() for one less layer of functions to jump through.

tests/runner/testers/torch/dynamic_torch_model_tester.py
kmabeeTT: Agreed with Aleks, I think tt-forge-models location (like https://github.com/tenstorrent/tt-forge-models/pull/604 added) is the right approach, it is ideally common/shared function that is available as a util and simply used by model loaders who have (as also suggested above in other comment) the new flag `override_sliding_attention` set.

kmabeeTT (changes requested):
Some feedback to more closely align with original proposal, if we make them it will greatly reduce the size of this PR to basically just a new function, and some cleanup of surrounding code (usage of old functions).  You'd want to make the tt-forge-models changes, test them with tt-xla, get them merged, integrated to tt-xla and then tt-xla updated.

python_package/tt_torch/transformers_overrides.py
AleksKnezevic: I'd honestly leave this as 
`mod.create_sliding_window_causal_mask = tt_create_sliding_window_causal_mask`
is until we hit an issue, it's not for all models, just the ones we want to override the causal mask creation function for. If it's not where we expect it, maybe it's not the exact same API, so more changes might be needed.

devisettymahidhar608:
Addressed in this [PR](https://github.com/tenstorrent/tt-xla/pull/4975). Closing the pr.

PR #4636, opened 2026-05-12
Enable optimization level configuration for tests in test infra

acicovicTT:
Adds a new valid config file field and a CLI-level override used to configure optimization level in tests.

Still a work in progress for configuration files: validation of changes needed + comments noting what are the corresponding perf benchmarks need to be removed + temporary test matrix needs to be removed.


sgligorijevicTT (approved):

.github/workflows/test-matrix-presets/model-test-optimization-level.json
```diff
@@ -0,0 +1,160 @@
+[
+  {"runs-on": "n150", "name": "run_forge_models", "dir": "./tests/runner/test_models.py::test_all_models_torch", "test-mark": "n150 and inference and single_device", "contains": "bert/sentence_embedding_generation/pytorch-emrecan/bert-base-turkish-cased-mean-nli-stsb-tr-single_device-inference", "forge-models": true, "forked": true},
```
nsumrakTT: I do not understand why you took approach to state each model? With this approach:
- One has to edit file multiple times although we already have model config yaml's
- Each of these tests runs on separate machine. Whole run will need 160 instances and 30-60% of machine time will be wasted on machine setup/teardown.
I guess these models are one with optimization level set. Since you implemented --optimization-level in conftest.py it might be a good approach to add optimization level to matrix parser (parameter of this json) so multiple tests could be run per machine and it could utilize parallelism feature as in other matrix jsons.

vvukomanTT: +1 on the points:
- it seems like a lot of duplicated, unnecessary yaml code
- consolidate jobs into fewer jobs to not waste machine time

acicovicTT: I mentioned this in PR description, this is a temporary matrix that I need to remove. I'm using it as a one-off check to see what will fail with higher optimization levels.

Issue #3094, opened 2026-02-03
[vLLM plugin] Non deterministic failure in CI for Qwen3-Embedding-0.6B


mmanzoorTT:
`Qwen3-Embedding-0.6B` model is failing randomly in CI for `batch_size=2` due to seg fault. 
[CI failure] (https://github.com/tenstorrent/tt-xla/actions/runs/21544645182/job/62084523198#step:17:704)

PR #4673, opened 2026-05-13
[vLLM plugin] Unify pooling runner input layout to [num_reqs, tokens]

### LINKED ISSUE #4672: [vLLM plugin] Remove batch_size dual-mode input batching for pooling models
Pooling_runner uses two paths for preparing inputs. 
batch_size=1 currently fuses all requests into [1, total_tokens]
batch_size>1 uses per-request batching [batch_size, tokens_per_req]

Unify these two paths and always use per-request batched layout [num_reqs, padded_tokens_per_req]. This will simplify pooling_runner by removing batch_size-controlled dual-mode logic:

mmanzoorTT:
### Ticket
closes #4672 

### Problem description
Pooling runner uses two different input layouts (fused vs. batched).

### What's changed
Always use per-request batched layout [num_reqs, padded_tokens]
- Remove batch_size conditionals in _prepare_inputs
- Unify token padding to max_model_len
- Update _dummy_run and _precompile_backbone signatures
- Add config warning if batch_size provided for pooling

### Checklist


integrations/vllm_plugin/vllm_tt/pooling_runner.py
```diff
@@ -1711,12 +1604,12 @@ def profile_run(
         # Trigger compilation for general shape.
         torch._dynamo.config.dynamic_shapes = False
         self._dummy_run(
-            num_tokens, self.num_reqs_max_model_len, self.max_num_blocks_per_req
+            self.num_reqs_max_model_len, num_tokens, self.max_num_blocks_per_req
```
kmabeeTT: Hmm is this a bug fix too? Was taking 4 args before but we only passed 3... do we have some tests covering this? I'd think they would have been failing.

integrations/vllm_plugin/vllm_tt/pooling_runner.py
```diff
@@ -891,136 +875,63 @@ def _prepare_inputs(self, scheduler_output: "SchedulerOutput", start_index: int)
...
+        # Create zero tensor of shape [num_reqs x max_token_len] for position tensor.
+        arange = np.zeros((num_reqs, max_num_scheduled_tokens_all_reqs), dtype=np.int32)
+
+        # Get batched arange; padded with zero.
+        # E.g., [2, 5, 3] -> [[0, 1, 0, 0, 0],
+        #                     [0, 1, 2, 3, 4],
+        #                     [0, 1, 2, 0, 0]]
+        # For each scheduled token, what is its position in corresponding req.
+        for i, n in enumerate(num_scheduled_tokens_per_req):
+            arange[i, :n] = np.arange(n, dtype=np.int32)
```
kmabeeTT, line 894: Seems like the old batch_size == 1 path was adding num_computed_tokens_cpu to arange for positions, and the new path just uses bare arange. Is that fine because pooling never resumes mid-sequence (no chunked prefill)? Just want to confirm we're not silently dropping the offset for some cases..

integrations/vllm_plugin/vllm_tt/pooling_runner.py
```diff
@@ -241,24 +241,24 @@ def __init__(
...
-            self.batch_size > 0
-        ), f"Positive batch_size allowed, received: batch_size: {self.batch_size}"
-        assert not (
-            self.batch_size > 1
-            and self.batch_size != self.scheduler_config.max_num_seqs
-        ), f"batch_size ({self.batch_size}) must equal max_num_seqs ({self.scheduler_config.max_num_seqs}) when batch_size > 1"
+        # InputBatch needs to work with sampling tensors greater than padding
+        # to avoid dynamic shapes. Also, avoid suboptimal alignment.
+        self.max_num_reqs = max(self.scheduler_config.max_num_seqs, MIN_NUM_SEQS)
+        if "batch_size" in (vllm_config.additional_config or {}):
```
kmabeeTT, line 248: Since we're ignoring batch_size here, might be worth cleaning up two related things in platform.py while you're at it:

https://github.com/tenstorrent/tt-xla/blob/main/integrations/vllm_plugin/vllm_tt/platform.py#L53-L60

  1. TTConfig.batch_size (line 53) — nothing reads it after this PR, so it's effectively dead config. Either remove it or add a deprecation note.
  2. The enable_data_parallel comment (lines 56–59) still says "requires batch_size > 1", which isn't accurate anymore — DP gates on max_num_reqs > 1 now.

kmabeeTT (commented):
Few comments and little worried about test coverage - test_batched_inference currently skips Qwen3-0.6B at batch_size=2 (#3094), so the only decoder-only multi-request signal on push comes from the default batch_size=1, max_num_seqs=2 config. Have you spot-checked PCC on a decoder-only multi-request case outside CI with this PR applied?

integrations/vllm_plugin/vllm_tt/pooling_runner.py
mmanzoorTT: We were actually consuming 2 arguments in `_dummy_run`. Fixed it. 2 arguments passed and 2 consumed.

integrations/vllm_plugin/vllm_tt/pooling_runner.py
mmanzoorTT: The reason for adding num_computed_tokens_cpu to arange for positions is that we flatten/fused multiple requests into single input sequence and then pad it. We had two paths; 1) flatten input sequence for batch_size==1 as [1 x combined_flattened_tokens] and 2) batched inputs for batch_size > 1 as [batch_size x num_tokens]. 
This PR unify both paths and always use batched inputs as [batch_size, num_tokens] regardless of batch_size.

integrations/vllm_plugin/vllm_tt/pooling_runner.py
mmanzoorTT: Thanks for pointing it out. updated `platform.py` as suggested.

mmanzoorTT:
> Few comments and little worried about test coverage - test_batched_inference currently skips Qwen3-0.6B at batch_size=2 (#3094), so the only decoder-only multi-request signal on push comes from the default batch_size=1, max_num_seqs=2 config. Have you spot-checked PCC on a decoder-only multi-request case outside CI with this PR applied?

I had ran the [nightly CI](https://github.com/tenstorrent/tt-xla/actions/runs/25807286681) which tests Qwen3-0.6B with batch_size=4. I'll improve test coverage for larger batch_size=[8, 16, 32] in a subsequent PR.

kmabeeTT (approved):
thx for followup

PR #4676, opened 2026-05-13
Add relative L2 error similarity metric to op-tests and benchmarks

dgolubovicTT:
## Summary

  Adds `rel_l2 = ||device - golden||_2 / ||golden||_2` as a new similarity metric, alongside the existing PCC / atol / allclose checks.
  
  **Motivation — every metric we use today has a known failure mode:**
  
  - **PCC** is alignment/shape-only and scale-blind: scaling a tensor by some factor still yields PCC = 1.0.
  - **max atol** is dominated by a single outlier element, hiding the overall quality of the tensor.
  - **max rtol** blows up wherever the golden tensor is near zero (e.g. LLM logits routinely hit `|y| ~ 1e-7`, sending rtol to `~1e5` even when the result is numerically fine).

  `rel_l2` is the standard numerical-linear-algebra / HPC validation metric:
  
  - **Scale-sensitive** (catches drift that PCC misses).
  - **Stable near zero globally** (denominator is `||y||_2`, not per-element `|y_i|`).
  - **Captures distributed degradation** rather than one bad element.
  - **Smooth, robust, well-conditioned.**

  All norms are computed in float64 to avoid bf16/fp32 underflow.
  
  ### What changes

  - **Test infra evaluators** (`tests/infra/evaluators/`): new `RelL2Config` (disabled by default, threshold `1.0`), `rel_l2` field on the comparison result, and Torch + JAX implementations. Integrated into the existing comparison pipeline alongside PCC / atol / allclose.
  - **Benchmark utilities** (`tests/benchmark/`): standalone `compute_rel_l2` helper, and rel_l2 is printed next to PCC at each verification point in the LLM benchmark. Observability-only — no new asserts.
  - Rename `pcc_mask` to `real_token_mask` since the mask is now consumed by both PCC and rel_l2 (and any future sequence-aware metric). Also renames the corresponding helper in `tests/runner/testers/torch/dynamic_torch_model_tester.py`.

  ### Defaults
  
  `RelL2Config` defaults to `enabled=False` with a permissive threshold of `1.0`. Observability-only: no existing test changes pass/fail status. Tightening thresholds per-model is a follow-up.
  
  ## Tests
`pytest -v tests/infra_tests/single_chip/test_result_assertion.py` passes (pre-existing + new rel_l2 tests).

### DIFF
--- a/tests/benchmark/benchmarks/llm_benchmark.py
+++ b/tests/benchmark/benchmarks/llm_benchmark.py
@@ -36,6 +36,7 @@ from tt_torch.weight_dtype import apply_weight_dtype_overrides
 from utils import (
     build_xla_export_name,
     compute_pcc,
+    compute_rel_l2,
     create_benchmark_result,
     get_benchmark_metadata,
     get_xla_device_arch,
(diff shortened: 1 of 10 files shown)


vkovinicTT (approved):

mrakitaTT (approved):
Took a quick look, LGTM

vkovacevicTT (approved):
Benchmark changes look good!

mstojkovicTT (approved):

PR #4712, opened 2026-05-14
[TT-Lang] Jackzhang/tt lang integration

jackzhangTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/4813#event-25777791477)

### Problem description
This is the first POC integration of tt-lang into the tt-xla compiler pipeline.

### What's changed
In the current flow, the user's tt-lang @operation kernel is captured at decoration time and stored in a Python kernel registry. The kernel is then emitted as an opaque stablehlo.custom_call in the graph and processed through the StableHLO → TTIR → TTNN pipeline without the kernel body, the body lives only in the Python registry, only metadata (kernel_id, arg_roles, etc.) are kept in the IR.

We defer invoking the tt-lang compiler until the TTNN stage, because tt-lang needs the post-Shardy, shard-local shapes and the final TTNN layout encoding on each operand. This is handled by two tt-mlir passes appended to the TTIR→TTNN pipeline (not by the plugin walking the module):

--ttnn-resolve-tt-lang-kernels walks each ttnn.tt_lang_op, invokes the tt-lang compiler via pybind11 (tt_torch.tt_lang.resolve_operation) with the final operand types and grid, and attaches the returned artifact as the kernel_artifact attribute (JSON produced by the Python side).
--ttnn-lower-tt-lang-to-generic parses that kernel_artifact and rewrites the ttnn.tt_lang_op into an equivalent ttnn.generic op (building a #ttnn.program with the kernels' C++ source inline and per-operand TensorAccessor markers).

By the time TTNN→flatbuffer emission runs, the op is already a plain ttnn.generic; the flatbuffer emitter needs no tt-lang-specific handling and does not read kernel_artifact. TensorAccessor compile-time args are filled in by the tt-mlir runtime at execute time.

### Tests added

**Python unit tests** (`tests/torch/ops/test_tt_lang_kernel.py`, 27 tests):
- `@tt_torch.tt_lang_operation` decorator: required `arg_roles`, rejection of non-XLA tensors, rejection of zero "out" operands.
- Kernel registry: registration and lookup by `kernel_id`.
- `torch.ops.tt.tt_lang_op` dispatch: XLA-only enforcement, `_validate_tt_lang_op_out_indices` rejects zero outs / duplicate indices / out-of-range indices.
- Device-less compile path: `_StubTtnnTensor` / `_StubTtnnDevice` plumbing and the L1-budget probe bypass.
- Artifact serialization: `_ARTIFACT_FORMAT_VERSION` round-trip, JSON schema validation.

**Hardware e2e tests** (`tests/torch/ops/test_tt_lang_kernel_e2e.py`, 2 tests × 2 shape params = 4 invocations, n150):

- `test_tt_lang_eltwise_add_e2e` — Bare-kernel path. A tt-lang tilewise elementwise-add (`examples/eltwise_add.py`-style: explicit reader/compute/writer threads, 2-block CB pipeline, `grid="auto"`) is wrapped with `@tt_torch.tt_lang_operation`, called from a `torch.nn.Module` on the XLA device, and exercised end-to-end:
  `@tt_torch.tt_lang_operation` → `stablehlo.custom_call @tt.tt_lang_op` → TTIR → TTNN → embedded-Python bridge resolves the kernel → `kernel_artifact` populated → flatbuffer `GenericOp` → execute on silicon → bf16 PCC against torch golden.

- `test_tt_lang_add_stitched_with_torch_add_e2e` — **Graph-stitching path.** Same kernel, but wrapped between two regular `torch.add` ops inside a single XLA program:

### DIFF
--- a/.github/workflows/call-test.yml
+++ b/.github/workflows/call-test.yml
@@ -191,7 +191,16 @@ jobs:
 
         echo "=== Installing wheel ==="
         source venv/activate
-        uv pip install wheels/*.whl
+        if [[ "${{ matrix.build.extra-wheel }}" == "vllm" ]]; then
+          # tt-lang's wheel metadata caps numpy<2, but tt-lang works with numpy 2.x.
+          # vllm (via opencv-python-headless>=4.13) requires numpy>=2, so pjrt_plugin_tt
+          # and vllm_tt cannot co-resolve. Override the stale cap; every other
+          # dependency still resolves normally.
+          echo 'numpy>=2' > wheels/numpy-override.txt
+          uv pip install --override wheels/numpy-override.txt wheels/*.whl
+        else
+          uv pip install wheels/*.whl
+        fi
 
     - name: Install custom torch_xla from workflow run
       if: ${{ inputs.torchxla_build_run_id != '' }}
--- a/docs/source/index.md
+++ b/docs/source/index.md
@@ -32,6 +32,12 @@ model_auto_discovery_tests
 :maxdepth: 2
 
 fusing_and_composite_ops
+```
+
+```{toctree}
+:caption: TT-Lang Integration
+:maxdepth: 2
+
 tt_lang_integration
 ```
 
--- a/docs/source/index.md
+++ b/docs/source/index.md
@@ -32,6 +32,7 @@ model_auto_discovery_tests
 :maxdepth: 2
 
 fusing_and_composite_ops
+tt_lang_integration
 ```
 
 ```{toctree}
--- a/docs/source/tt_lang_integration.md
+++ b/docs/source/tt_lang_integration.md
@@ -14,7 +14,7 @@ The torch custom op it emits is registered in
 import tt_torch
 import ttl  # tt-lang
 
-@tt_torch.kernel(kernel_id="moe.routed_mlp.v1", arg_roles=("in", "in", "out"))
+@tt_torch.tt_lang_operation(operation_id="moe.routed_mlp.v1", arg_roles=("in", "in", "out"))
 @ttl.operation(grid=(8, 8))
 def routed_mlp(lhs, rhs, out):
     ...
@@ -26,7 +26,7 @@ routed_mlp(lhs, rhs, out)
 The decorator does three things at call time:
 
 1. Validates inputs and copies argument-role metadata into a process-global
-   registry keyed by `kernel_id`.
+   registry keyed by `operation_id`.
 2. Emits `stablehlo.custom_call @tt.tt_lang_op` whose `frontend_attributes`
    carry the metadata the plugin will need at compile/execute time.
 3. Copies the functional custom-call results back into the user's
@@ -39,7 +39,7 @@ Shardy, TTIR legalization, and TTNN lowering all see it as a normal op.
 
 ```text
 PyTorch model
-   |  @tt_torch.kernel(kernel_id=...)  -- registers callable, emits custom op
+   |  @tt_torch.tt_lang_operation(operation_id=...)  -- registers callable, emits custom op
    v
 stablehlo.custom_call @tt.tt_lang_op
    { kernel_id, arg_roles, version_tag, shard_spec }
@@ -56,7 +56,7 @@ TTIR -> TTNN                    (ttnn.tt_lang_op, kernel_artifact = empty)
    v  ModuleBuilder::resolveTtLangKernels (compile time, post-TTNN)
 tt-mlir's --ttnn-resolve-tt-lang-kernels pass walks the module for
 `ttnn.tt_lang_op`s and calls
-    tt_torch.tt_lang.resolve_kernel(kernel_id, version_tag, shapes, dtypes, ...)
+    tt_torch.tt_lang.resolve_operation(operation_id, version_tag, shapes, dtypes, ...)
 through the embedded Python interpreter (host's libpython, GIL acquired).
    |
    v
@@ -75,10 +75,10 @@ already-bound kernel -- no Python on the hot path.
 | Piece | Location | Status |
 |---|---|---|
 | `torch.ops.tt.tt_lang_op` (variadic custom op + fake + autograd-raises) | `python_package/tt_torch/custom_ops.py` | done |
-| `@tt_torch.kernel` decorator | `python_package/tt_torch/tt_lang.py` | done |
-| `kernel_id -> KernelEntry` registry | `python_package/tt_torch/tt_lang.py` | done |
-| `resolve_kernel(...)` entry point | `python_package/tt_torch/tt_lang.py` | **stub** -- raises `NotImplementedError`; signature is stable for the plugin to call against |
-| Embedded-Python resolver (`pybind11`, calls `tt_torch.tt_lang.resolve_kernel`) | tt-mlir `--ttnn-resolve-tt-lang-kernels` pass (`lib/Dialect/TTNN/Transforms/TTNNResolveTtLangKernels*.cpp`) | done |
+| `@tt_torch.tt_lang_operation` decorator | `python_package/tt_torch/tt_lang.py` | done |
+| `operation_id -> OperationEntry` registry | `python_package/tt_torch/tt_lang.py` | done |
+| `resolve_operation(...)` entry point | `python_package/tt_torch/tt_lang.py` | **stub** -- raises `NotImplementedError`; signature is stable for the plugin to call against |
+| Embedded-Python resolver (`pybind11`, calls `tt_torch.tt_lang.resolve_operation`) | tt-mlir `--ttnn-resolve-tt-lang-kernels` pass (`lib/Dialect/TTNN/Transforms/TTNNResolveTtLangKernels*.cpp`) | done |
 | `ModuleBuilder::resolveTtLangKernels(...)` compile-time hook (runs the tt-mlir pass via a `PassManager`) | `pjrt_implementation/src/api/module_builder/module_builder.cc` | done -- no-op until tt-mlir emits `ttnn.tt_lang_op` |
 
 The pybind11 / libpython dependency lives entirely in tt-mlir's
@@ -103,7 +103,7 @@ Its operands are the input tensors followed by the pre-allocated outputs
 
 | attribute | meaning |
 |---|---|
-| `kernel_id` | Stable identifier the plugin will pass back to `resolve_kernel`. |
+| `kernel_id` | Stable identifier the plugin will pass back to `resolve_operation`. |
 | `arg_roles` | Comma-separated per-operand role, constrained to `in* out+` (all inputs first, then outputs), e.g. `"in,in,out"`. |
 | `version_tag` | Hash of kernel source; the plugin sends it back so the resolver can refuse stale matches. |
 | `shard_spec` | Optional opaque sharding hint (empty string when unused). |
@@ -152,10 +152,10 @@ and the runtime backend handoff. It drives tt-mlir's
    `kernel_artifact` is still empty (already-baked ops are skipped, so the
    pass is composable with future ahead-of-time artifact paths).
 2. Acquires the GIL on the host Python (the one that loaded the plugin)
-   and imports `tt_torch.tt_lang.resolve_kernel` once per pass.
+   and imports `tt_torch.tt_lang.resolve_operation` once per pass.
 3. For each op, reads the four named attributes, builds per-operand
    `(shape, dtype, layout)` triples from the op's operand types, and
-   calls `resolve_kernel(...)` through pybind11.
+   calls `resolve_operation(...)` through pybind11.
 4. Attaches the returned `bytes` (the JSON artifact) back onto the op
    as the `kernel_artifact` `StringAttr`.
 
@@ -170,7 +170,7 @@ What this still needs:
   emitter described below.
 * Layout hand-off in the resolver pass (T6). **Done** -- the pass
   now stringifies each operand's `ttnn.ttnn_layout` encoding and passes
-  it through to `resolve_kernel`; the Python side records it under
+  it through to `resolve_operation`; the Python side records it under
   `operand_metadata.layouts` in the artifact JSON so the flatbuffer
   emitter can consume it later.
 
@@ -198,14 +198,14 @@ descriptive error pointing at the missing wheel.
 
 ### 4. tt-lang: compile driver (done)
 
-`resolve_kernel` drives tt-lang's existing compile path with mock torch
+`resolve_operation` drives tt-lang's existing compile path with mock torch
 tensors and the `TTLANG_COMPILE_ONLY=1` env var, monkey-patching
 `ttl.ttl_api._compile_kernel` to capture the resulting
 `CompiledTTNNKernel`. The captured bundle is serialized into a JSON byte
 blob (kernel C++ sources, thread types, configs/CB/core-ranges, tensor
 indices, plus `operand_metadata` recording the shapes/dtypes/layouts
 the kernel was compiled against). See
-`python_package/tt_torch/tt_lang.py::_serialize_compiled_kernel` and the
+`python_package/tt_torch/tt_lang.py::_serialize_compiled_operation` and the
 versioned `_ARTIFACT_FORMAT_VERSION` constant; the tt-mlir flatbuffer
 emitter (T4) decodes the same schema.
 
@@ -432,7 +432,7 @@ coverage.
    required -- the same path every other DPS op (matmul, etc.) takes.
 
    The `in* out+` ordering is enforced in three places: the
-   `@tt_torch.kernel` decorator (`_normalize_arg_roles`), the TTIR
+   `@tt_torch.tt_lang_operation` decorator (`_normalize_arg_roles`), the TTIR
    verifier, and the TTNN verifier. The flatbuffer emitter
    (`TTNNToFlatbuffer.cpp`) relies on it too: operand declaration order
    already matches the runtime's `io_tensors` order (ins first, then
@@ -449,14 +449,14 @@ Today there is no cache. The PJRT plugin already caches whole compiled
 executables; tt-lang should cache its own compilations. Only if profiling
 shows our resolve callback firing multiple times for the same arguments
 should we add caching here -- and then it can be a single
-`functools.lru_cache` on `resolve_kernel`, not a separate subsystem.
+`functools.lru_cache` on `resolve_operation`, not a separate subsystem.
 
 ### 6. Autograd (if needed)
 
 `torch.ops.tt.tt_lang_op` currently raises from its autograd hook to
 prevent silent zero-gradient bugs. When backward support is required,
-the canonical pattern is to author a separate `@tt_torch.kernel` (with
-its own `kernel_id`) for the backward and have the autograd hook look it
+the canonical pattern is to author a separate `@tt_torch.tt_lang_operation` (with
+its own `operation_id`) for the backward and have the autograd hook look it
 up and invoke it.
 
 ## Why this shape
@@ -466,11 +466,12 @@ A few invariants kept the design honest:
 * The custom op must stay inside the XLA graph end to end so that Shardy,
   layout, and ttnn.generic placement all compose with it. Anything that
   routes around the graph (e.g. a host callback) defeats the integration.
-* The Python side owns only what it must own: the user-facing API, a
-  `kernel_id` registry, and a single resolve entry point. Compile
+* The Python side owns only what it must own: the user-facing API, an
+  `operation_id` registry, and a single resolve entry point. Compile
   caching, ABI mirrors, and backend selection are deferred until there is
   real code to integrate them with.
-* The `kernel_id` is the only stable identifier crossing the C / Python
+* The operation id (the `kernel_id` MLIR/wire attribute) is the only stable
+  identifier crossing the C / Python
   boundary; everything else (shapes, dtypes, mesh) is normal PJRT runtime
   data the plugin already understands.
 
@@ -479,8 +480,8 @@ A few invariants kept the design honest:
 ```
 python_package/tt_torch/
 ├── custom_ops.py           # tt::tt_lang_op definition + fake + autograd
-└── tt_lang.py              # decorator + registry + resolve_kernel
-                            # + _serialize_compiled_kernel JSON producer
+└── tt_lang.py              # decorator + registry + resolve_operation
+                            # + _serialize_compiled_operation JSON producer
                             # (_ARTIFACT_FORMAT_VERSION is the schema gate)
 
 pjrt_implementation/inc/api/module_builder/
@@ -519,7 +520,7 @@ lib/Dialect/TTNN/Transforms/TTNNResolveTtLangKernels.cpp
                             # --ttnn-resolve-tt-lang-kernels pass: IR walk +
                             # mesh-shape parsing (-fno-rtti, no pybind11)
 lib/Dialect/TTNN/Transforms/TTNNResolveTtLangKernelsPython.cpp
-                            # pybind11 call into tt_torch.tt_lang.resolve_kernel
+                            # pybind11 call into tt_torch.tt_lang.resolve_operation
                             # (-frtti -fexceptions)
 
 test/ttmlir/Conversion/StableHLOToTTIR/tt_lang_op.mlir
--- a/docs/source/tt_lang_integration.md
+++ b/docs/source/tt_lang_integration.md
@@ -1,0 +1,530 @@
+# tt-lang integration in tt-xla
+
+This document describes how user-authored
+[tt-lang](https://github.com/tenstorrent/tt-lang) kernels get from PyTorch
+source code down to running on Tenstorrent hardware through tt-xla.
+
+The Python surface lives in a single file: `python_package/tt_torch/tt_lang.py`.
+The torch custom op it emits is registered in
+`python_package/tt_torch/custom_ops.py` alongside the rest of `torch.ops.tt.*`.
+
+## Author-facing API
+
+```python
+import tt_torch
+import ttl  # tt-lang
+
+@tt_torch.kernel(kernel_id="moe.routed_mlp.v1", arg_roles=("in", "in", "out"))
+@ttl.operation(grid=(8, 8))
+def routed_mlp(lhs, rhs, out):
+    ...
+
+# On XLA tensors:
+routed_mlp(lhs, rhs, out)
+```
+
+The decorator does three things at call time:
+
+1. Validates inputs and copies argument-role metadata into a process-global
+   registry keyed by `kernel_id`.
+2. Emits `stablehlo.custom_call @tt.tt_lang_op` whose `frontend_attributes`
+   carry the metadata the plugin will need at compile/execute time.
+3. Copies the functional custom-call results back into the user's
+   pre-allocated `"out"` operands so the mutation-style API is honored.
+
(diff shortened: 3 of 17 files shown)


docs/source/tt_lang_integration.md
```diff
@@ -0,0 +1,520 @@
...
+```python
+import tt_torch
+import ttl  # tt-lang
+
+@tt_torch.kernel(kernel_id="moe.routed_mlp.v1", arg_roles=("in", "in", "out"))
+@ttl.operation(grid=(8, 8))
+def routed_mlp(lhs, rhs, out):
+    ...
+
+# On XLA tensors:
```
AleksKnezevic, line 22: on pytorch tensors?

docs/source/tt_lang_integration.md
```diff
@@ -0,0 +1,520 @@
...
+The torch custom op it emits is registered in
+`python_package/tt_torch/custom_ops.py` alongside the rest of `torch.ops.tt.*`.
+
+## Author-facing API
+
+```python
+import tt_torch
+import ttl  # tt-lang
+
+@tt_torch.kernel(kernel_id="moe.routed_mlp.v1", arg_roles=("in", "in", "out"))
```
AleksKnezevic: we still need to add shard propagation, correct? This should/can be in a follow-up PR.

pjrt_implementation/src/api/module_builder/tt_lang_bridge.cc
```diff
@@ -0,0 +1,219 @@
...
+// off-RTTI everywhere else.
+//
+// The plugin is always loaded by a running Python process (JAX or
+// torch.compile invokes us), so we attach to the existing interpreter via
+// pybind11's GIL acquisition rather than starting a fresh one. If no
+// interpreter is present (e.g. the plugin is dlopen'd from a C++-only
+// binary), `import "tt_torch.tt_lang"` will fail and we report a
+// descriptive error.
+
+#include "api/module_builder/tt_lang_bridge.h"
```
AleksKnezevic: Can this file exist within tt-mlir? That way we could compile tt-lang ops even if they weren't placed in by the FE, but lowered from some graph pattern.

python_package/tt_torch/tt_lang.py
```diff
@@ -0,0 +1,1151 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""
```
AleksKnezevic: As above, can this line in tt-mlir for FE independant tt-lang integration? Follow-up PR is fine.

docs/source/tt_lang_integration.md
jackzhangTT: Yes'll add it in a follow up PR

pjrt_implementation/src/api/module_builder/tt_lang_bridge.cc
jackzhangTT: I will look into this in a follow up PR

tests/torch/ops/test_tt_lang_kernel_e2e.py
```diff
@@ -0,0 +1,374 @@
...
+                if row < row_tiles:
+                    r0, r1 = row * GRANULARITY, (row + 1) * GRANULARITY
+                    for local_col in range(cols_per_node):
+                        col = node_col * cols_per_node + local_col
+                        if col < col_tiles:
+                            with out_dfb.wait() as out_blk:
+                                tx = ttl.copy(out_blk, out[r0:r1, col : col + 1])
+                                tx.wait()
+
+    @tt_lang_kernel(
```
zoecarver: Nit: we call the tt-lang kernel a `ttl.operation` now, so maybe this should be called an operation as well. Maybe we should also make it clear that this is a forge wrapper/interface. Maybe `@ttl.export_operation`?

tests/torch/ops/test_tt_lang_kernel_e2e.py
```diff
@@ -0,0 +1,374 @@
...
+def _make_eltwise_add_kernel(kernel_id: str):
+    """Build the tt-lang kernel + ``@tt_torch.kernel`` wrapper.
+
+    The kernel mirrors `tt-lang/examples/eltwise_add.py`: a 2-tile-block
+    pipelined elementwise add with explicit reader/compute/writer
+    threads and CB-based block staging. Block count = 2 gives the
+    runtime a minimal pipeline depth to exercise.
+    """
+
+    @ttl.operation(grid="auto")
```
zoecarver, line 109: Maybe another test with a different grid size?

tests/torch/ops/test_tt_lang_kernel_e2e.py
```diff
@@ -0,0 +1,374 @@
...
+                if row < row_tiles:
+                    for local_col in range(cols_per_node):
+                        col = node_col * cols_per_node + local_col
+                        if col < col_tiles:
+                            with (
+                                a_dfb.wait() as a_blk,
+                                b_dfb.wait() as b_blk,
+                                out_dfb.reserve() as out_blk,
+                            ):
+                                out_blk.store(a_blk + b_blk)
```
zoecarver, line 142: Maybe another operation that does a matmul? Also maybe one that uses pipes as that will test that semaphores are piped correctly.

third_party/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 option(USE_CUSTOM_TT_MLIR_VERSION "Flag to use TT_MLIR_VERSION set by the user" OFF)
 
 if (NOT DEFINED TT_MLIR_VERSION OR NOT USE_CUSTOM_TT_MLIR_VERSION)
-    set(TT_MLIR_VERSION "464a5f341908d5a3f790e697be5ffa6fd3fb6f32")
+    set(TT_MLIR_VERSION "jackzhang/tt_lang_integration")
```
AleksKnezevic: reminder to remove this before merging.

tests/torch/ops/test_tt_lang_kernel_e2e.py
```diff
@@ -0,0 +1,374 @@
...
+        abc = add_kernel(ab, c_xla, out_buf)
+        result_xla = abc + d_xla
+        # One mark_step -- everything must compile into a single
+        # executable. If torch_xla split the graph at the custom_call
+        # boundary, we'd observe extra compiles in the logs.
+        xm.mark_step()
+        result = result_xla.to("cpu")
+    except RuntimeError as e:
+        msg = str(e)
+        if "undefined symbol" in msg or "_ZN2tt8tt_metal" in msg:
```
AleksKnezevic: Why is this needed?

tests/torch/ops/test_tt_lang_kernel_e2e.py
```diff
@@ -0,0 +1,374 @@
...
+import tt_torch  # noqa: F401  -- registers torch.ops.tt.*
+from tt_torch import tt_lang as tt_lang_mod
+from tt_torch.tt_lang import kernel as tt_lang_kernel
+
+# ---------------------------------------------------------------------------
+# Hardware / version gates
+# ---------------------------------------------------------------------------
+
+
+def _have_tt_device() -> bool:
```
AleksKnezevic: most of this repo assumes you have a device, so it's probably ok to do the same here.

tests/torch/ops/test_tt_lang_kernel_e2e.py
```diff
@@ -0,0 +1,374 @@
...
+#     VENDORED_METAL=<tt-xla>/third_party/tt-mlir/src/tt-mlir/
+#                    third_party/tt-metal/src/tt-metal
+#     PJRT_DEVICE=TT TT_METAL_HOME=$VENDORED_METAL \
+#       PYTHONPATH=$VENDORED_METAL/ttnn:$PYTHONPATH pytest ...
+#
+# If a vendored ttnn isn't on the path, ttnn won't import here and we
+# skip below.
+try:
+    import ttnn  # type: ignore  # noqa: F401
+except ImportError as e:
```
AleksKnezevic: pythonpath is set in the activate script

python_package/tt_torch/tt_lang.py
```diff
@@ -0,0 +1,1004 @@
...
+    captured: List[Any] = []
+    original_compile = _ttl_api._compile_kernel
+
+    @functools.wraps(original_compile)
+    def _capturing_compile(*args, **kwargs):
+        compiled = original_compile(*args, **kwargs)
+        captured.append(compiled)
+        return compiled
+
+    # DEMO HACK: tt-lang inspects real ttnn objects in a couple of
```
AleksKnezevic, line 566: is this still needed?

pjrt_implementation/src/api/module_builder/module_builder.cc
```diff
@@ -1043,6 +1059,46 @@ tt_pjrt_status ModuleBuilder::convertFromTTIRToTTNN(
...
+    std::string mesh_csv;
+    for (size_t i = 0; i < mesh_shape.size(); ++i) {
+      if (i != 0) {
+        mesh_csv.push_back(',');
+      }
+      mesh_csv += std::to_string(mesh_shape[i]);
+    }
+    options.meshShape = std::move(mesh_csv);
+  }
+  resolve_pm.addPass(mlir::tt::ttnn::createTTNNResolveTtLangKernels(options));
```
AleksKnezevic: as discussed, please add this to a pipeline.

tests/torch/ops/test_tt_lang_kernel_e2e.py
jackzhangTT: I've renamed the kernels to operation instead so its consistent with the tt-lang naming. However, I'm unable to change the name of the attribute's package unless it's part of tt-lang or add a monkey patch and attach it to the ttl module at runtime. For now I've kept it as `tt_torch.tt_lang_operation` since it's part of the tt_torch package and it clearly indicates that it's a tt-lang operation.

tests/torch/ops/test_tt_lang_kernel_e2e.py
jackzhangTT: this was from a previous bug I found when the tt-xla bundle and tt-lang wheel referenced mismatched tt-metal versions. Its probably too specific of a check

AleksKnezevic: yeah, I think we should remove it.

python_package/tt_torch/tt_lang.py
jackzhangTT: yea this is still needed the tt-lang compiler expects "real" tt-nn tensors but we create fake ones to circumvent this requirement and monkey patch some of its functions

tests/torch/ops/test_tt_lang_kernel_e2e.py
jackzhangTT: that sounds like a good idea. I'll add them in a follow up PR

AleksKnezevic (approved):

sshonTT (approved):

nsumrakTT (approved):

.github/workflows/call-test.yml
```diff
@@ -191,7 +191,16 @@ jobs:
...
         echo "=== Installing wheel ==="
         source venv/activate
-        uv pip install wheels/*.whl
+        if [[ "${{ matrix.build.extra-wheel }}" == "vllm" ]]; then
+          # tt-lang's wheel metadata caps numpy<2, but tt-lang works with numpy 2.x.
+          # vllm (via opencv-python-headless>=4.13) requires numpy>=2, so pjrt_plugin_tt
+          # and vllm_tt cannot co-resolve. Override the stale cap; every other
+          # dependency still resolves normally.
+          echo 'numpy>=2' > wheels/numpy-override.txt
+          uv pip install --override wheels/numpy-override.txt wheels/*.whl
```
nsumrakTT, line 200: This is a patch. If it is not properly registered it will stay.
It seems to me you should open issue in tt-lang, refer it here, so when it is resolved we can remove the patch.

python_package/requirements.txt
```diff
@@ -20,3 +20,6 @@ click==8.3.1
 pandas==3.0.0
 # tracy's serve_wasm.py imports websockets
 websockets==16.0
+
+# tt-lang (light build, 05/29)
+tt_lang@https://pypi.eng.aws.tenstorrent.com/tt-lang/tt_lang-1.1.1.dev20260529%2Blight-cp312-cp312-manylinux_2_35_x86_64.whl
```
nsumrakTT, line 32: Is there a reason to pin exact version here?
Would we have to uplift this version in the future?

jackzhangTT: Since both tt-lang and tt-mlir (which we refer to in third_party) both point to metal this is the version of tt-lang who's code that it generates is compatible with the version of tt-metal that tt-mlir points to. Yes we would have to uplift this version in the future

.github/workflows/call-test.yml
jackzhangTT: I agree I think it would be best if we could sync the numpy versions between tt-lang and tt-xla there's been some discussion but nothing concrete yet.

python_package/requirements.txt
hshahTT: Could you add a comment here stating this? So we remember to keep this in sync with the tt-metal version that this repo's tt-mlir commit uses

hshahTT: and also in the venv/requirements-dev.txt file

.github/workflows/call-test.yml
vvukomanTT: If this is just a matter of changing the cap for the numpy version in tt-lang (which it already works with), let's please do that and avoid this version overriding.

jackzhangTT: The change for tt-lang to drop the numpy dependency for their light builds (the builds that do not package tt-metal) will not go in until `1.1.4` which from my understanding also generates code for a newer version of tt-metal than the one that tt-mlir points to. So until tt-mlir uplifts metal until the same point as tt-lang we will need this.

hshahTT (approved):

venv/requirements-dev.txt
```diff
@@ -61,3 +62,13 @@ torchcodec
...
+
+# tt-lang (light build, 05/29).
+# SYNC WITH tt-metal: tt-lang generates kernel code (TTNN/tt-metal C++) that
+# must match the tt-metal API/ABI it runs against. tt-mlir (pinned in
+# third_party) points to a specific tt-metal, and tt-xla executes the
+# tt-lang-generated kernels through that tt-metal. This pin is therefore the
+# tt-lang version whose generated code is compatible with the tt-metal that
+# our current tt-mlir points to. When tt-mlir (and thus its tt-metal) is
+# uplifted, this tt-lang pin must be uplifted in sync.
+tt_lang==1.1.1.dev20260529+light
```
acicovicTT, line 74: tt-mlir uplifts are automated in the `.github/workflows/schedule-uplift.yml`, so this bump must be automated as well as part of the uplift PR created by that workflow

docs/source/index.md
```diff
@@ -32,6 +32,7 @@ model_auto_discovery_tests
 :maxdepth: 2
 
 fusing_and_composite_ops
+tt_lang_integration
```
acicovicTT, line 41: I would make this even more visible by having this be a top-level item "TT-Lang Integration".

Also, if you didn't already, you should check out the look-and-feel of your new docs page(s) locally before merging, here are the instructions: https://github.com/tenstorrent/tt-xla/tree/main/docs

python_package/requirements.txt
```diff
@@ -20,3 +20,13 @@ click==8.3.1
...
+
+# tt-lang (light build, 05/29).
+# SYNC WITH tt-metal: tt-lang generates kernel code (TTNN/tt-metal C++) that
+# must match the tt-metal API/ABI it runs against. tt-mlir (pinned in
+# third_party) points to a specific tt-metal, and tt-xla executes the
+# tt-lang-generated kernels through that tt-metal. This pin is therefore the
+# tt-lang version whose generated code is compatible with the tt-metal that
+# our current tt-mlir points to. When tt-mlir (and thus its tt-metal) is
+# uplifted, this tt-lang pin must be uplifted in sync.
+tt_lang@https://pypi.eng.aws.tenstorrent.com/tt-lang/tt_lang-1.1.1.dev20260529%2Blight-cp312-cp312-manylinux_2_35_x86_64.whl
```
acicovicTT, line 32: Same comment as I left in `venv/requirements-dev.txt`.

venv/requirements-dev.txt
jackzhangTT: With the integration of tt-lang there's a much more complicated process for uplifting both tt-lang and tt-mlir since they both have dependencies on tt-metal. For example, if an error occurs from uplifting tt-metal the problem could be from tt-xla, tt-mlir, or tt-lang or some combination of the three. Furthermore, just the tt-mlir uplift could break tests in tt-lang thus I think for now keeping the uplifting process manual is okay for now while we assess how the three projects integrate with each other.

acicovicTT: Who will own the uplift process? I feel like tt-xla team is streched very thin at the moment.

IMO it's fine to keep it manual for now as long as we ensure that broken tt-lang tests will not block the tt-mlir uplift. Assuming that the root cause for broken tt-lang tests will mostly be version incompatibility introduced by tt-mlir uplift, we should not gate the tt-mlir uplift on these tests.

jackzhangTT: yes I agree for now we should not gate the tt-mlir uplift on these tests and I can be responsible for the manual uplift in the mean time

acicovicTT (approved):

python_package/tt_torch/custom_ops.py
```diff
@@ -1739,6 +1739,134 @@ def sampling_fake(
...
+        if not 0 <= idx < num_tensors:
+            raise ValueError(
+                f"out index {idx} is out of range for {num_tensors} operands."
+            )
+        if idx in seen:
+            raise ValueError(f"out index {idx} appears more than once.")
+        seen.add(idx)
+
+
+@torch.library.custom_op("tt::tt_lang_op", mutates_args=[], device_types=["xla"])
```
jameszianxuTT, line 1767: Is it guaranteed that the ttlang kernel will not mutate the input arguments inplace?

python_package/tt_torch/tt_lang.py
```diff
@@ -0,0 +1,1012 @@
...
+from tt_torch.custom_ops import tt_lang_op_dispatch
+
+__all__ = [
+    "OperationEntry",
+    "TTLangError",
+    "tt_lang_operation",
+    "get_registered_operation",
+    "iter_registered_operations",
+    "resolve_operation",
+]
```
jameszianxuTT: If this is not an intentional way to expose the public interface to the ttlang feature to the tt-xla caller I suggest to remove/not advertise this, you already expose it in tt_torch.__init__.py

python_package/tt_torch/tt_lang.py
```diff
@@ -0,0 +1,1012 @@
...
+def _derive_version_tag(fn: Callable) -> str:
+    """Stable short hash of the operation source. Best-effort.
+
+    Falls back to the qualified name if the source isn't introspectable
+    (e.g. C-backed callables); identical source produces identical tags.
+    """
+    try:
+        material = inspect.getsource(fn)
+    except (OSError, TypeError):
+        material = getattr(fn, "__qualname__", fn.__class__.__name__)
```
jameszianxuTT: Is it better to throw here? isn't this tt-lang integration only supposed to support python callables as user input? why would a c-backed callable end up here?

python_package/tt_torch/tt_lang.py
```diff
@@ -0,0 +1,1012 @@
...
+
+
+def _torch_dtype_from_mlir_string(name: str) -> torch.dtype:
+    """Map an MLIR-style element-type string to a ``torch.dtype``.
+
+    Unknown / non-element types fall back to ``torch.float32`` so the
+    compile can still proceed (tt-lang doesn't read dtype off the stub
+    tensor on the compile-only path).
+    """
+    return _MLIR_DTYPE_TO_TORCH.get(name.strip().lower(), torch.float32)
```
jameszianxuTT: nit: I feel like you should warn/throw if the datatype mismatches instead of silently falling back to float32?

would signal a torch/mlir incompatibility

python_package/tt_torch/tt_lang.py
```diff
@@ -0,0 +1,1012 @@
...
+# ``ttnn.Tensor`` arguments and runs ``isinstance(arg, ttnn.Tensor)``
+# checks. Opening a real ttnn device in the plugin process races against
+# PJRT's own ``init_command_queue_device_with_topology`` (both register
+# kernel binaries in their own ``Program`` caches), so we feed tt-lang
+# duck-typed stand-ins and monkey-patch ``is_ttnn_tensor`` for the
+# duration of the compile. See ``docs/source/tt_lang_integration.md``
+# ("Device-less compile path") for the full discussion; the proper fix
+# is a ``_compile_ttnn_kernel_from_spec`` entry point upstream in
+# tt-lang, after which these stubs and the monkey-patch can be deleted.
+# Grep for ``DEMO HACK`` to find every site.
```
jameszianxuTT, line 387: TBD - look at mock device for this?

python_package/tt_torch/tt_lang.py
```diff
@@ -0,0 +1,1012 @@
...
+
+    def __init__(
+        self,
+        *,
+        grid_cols: int = 8,
+        grid_rows: int = 8,
+        arch: str = "wormhole_b0",
+    ) -> None:
+        self._grid = _StubGridSize(grid_cols, grid_rows)
+        self.arch = arch
```
jameszianxuTT, line 419: discussed offline - should remove default values here so it fails loudly if these are not overwritten

python_package/tt_torch/tt_lang.py
```diff
@@ -0,0 +1,1012 @@
...
+            f"layout={self.layout!r}, mem={self._memory_config.buffer_type!r})"
+        )
+
+
+def _memory_space_from_layout(layout_str: str) -> str:
+    """Map a printed ``#ttnn.ttnn_layout<...>`` substring to L1/DRAM.
+
+    Returns the bare string the stub's ``buffer_type`` should contain.
+    The default is ``"DRAM"`` because that's what tt-lang itself
+    defaults to when no explicit memory_config is plumbed through.
```
jameszianxuTT, line 487: When making such assumptions about tt-lang internal impl, I think it is good to expose a unit test so we can know if these assumptions are invalidated

python_package/tt_torch/tt_lang.py
```diff
@@ -0,0 +1,1012 @@
...
+    # @ttl.operation calls in unusual operations); we take the last one,
+    # which is the outermost-finishing compile.
+    return captured[-1]
+
+
+# Mapping from a tt-metal/ttnn DataType (or torch dtype) to the ttcore
+# DataType enum name that the tt-mlir flatbuffer emitter expects. Both
+# producers (this file) and consumer (TTNNToFlatbuffer.cpp's
+# `parseDataType`) must stay in sync with
+# `third_party/tt-mlir/.../Target/Common/types.fbs::enum DataType`.
```
jameszianxuTT, line 660: How will this stay in sync? is there a way to read this off a header or expose some interface from ttmlir to view the enum rather than duplicating this lookup table

python_package/tt_torch/tt_lang.py
```diff
@@ -0,0 +1,1012 @@
...
+def _tile_bytes_from_dtype_name(name: str) -> int:
+    """Tile byte size for a ttcore DataType name.
+
+    Mirrors ``tt-lang/dtype_utils::tile_bytes_from_dtype`` but indexed by
+    flatbuffer enum name so we don't have to import ttnn just to compute
+    sizes. Each TT-Metal tile is 32x32 elements; the byte cost depends
+    on element width and format-specific exponent overhead.
+    """
+    return {
+        "BFloat16": 32 * 32 * 2,
```
jameszianxuTT, line 732: 32*32 as a named const or something? Also defaulting to 32x32x2 seems really risky, and throwing in this case seems better

python_package/tt_torch/tt_lang.py
```diff
@@ -0,0 +1,1012 @@
...
+        # PipeNet sync-semaphore count. Keep the artifact key as
+        # "num_pipe_nets" (the tt-mlir flatbuffer parser reads that key) and
+        # fall back across ttl versions.
+        "num_pipe_nets": int(
+            getattr(
+                compiled,
+                "num_pipe_sync_semaphores",
+                getattr(compiled, "num_pipe_nets", 0),
+            )
+        ),
```
jameszianxuTT: Is this how we want to maintain backwards compatibility (do we care)? it seems like this schema and the mlir flatbuffer parser will accumulate branches and exceptions for every ttlang interface change until it becomes untenable. I think that is fine if we have a coherent plan for backwards compatibility across the stack at some point, but I don't think that is currently the case. 

i.e. this backwards compatibility mechanism extends to your schema format if tt-lang / tt-mlir become out of sync, so your schema is always parsable into a ttnn generic op. However, won't other things also break if tt-lang and tt-mlir become out of sync, eg. metal interfaces? 

not totally sure if I understand the intent of this, please feel free to clarify

python_package/tt_torch/tt_lang.py
```diff
@@ -0,0 +1,1012 @@
...
+                return bool(_orig(t)) if _orig is not None else False
+
+            _patch(module_name, "is_ttnn_tensor", _accept_stub)
+
+        _patch("ttl.ttl_api", "get_min_remaining_l1_for_device", lambda _device: 0)
+
+    prev_env = os.environ.get("TTLANG_COMPILE_ONLY")
+    with _DRIVE_LOCK:
+        _ttl_api._compile_kernel = _capturing_compile  # type: ignore[attr-defined]
+        os.environ["TTLANG_COMPILE_ONLY"] = "1"
```
jameszianxuTT, line 623: Not gating, but can we ask tt-lang to give us a proper interface for compile only, like an API for that instead of this env var state machine that the caller needs to manage

venv/requirements-dev.txt
jameszianxuTT, line 74: We may want to package this dep into tt-xla wheel too (eg. setup.py). 

I think the reason this works in CI is because we run our venv/activate script before CI tests, but that will not be the case for a user downloading the wheel. 

Please make sure that wheel build works on its own (eg. download the wheel, install in a blank venv, download your test file and run it) Adding a tt-lang standalone e2e example to examples/pytorch would also be good!

jameszianxuTT (changes requested):
Some nits. Blocking because I think there is some risk of breaking the tt-xla wheel without packaging tt_lang in setup.py, please confirm that the standalone wheel does work (for tt-lang tests and other random examples like mnist.py) before merging this.

This is a testing gap in our CI that real users exercise and is not usually a source of errors, except when adding project dependencies like is done in this PR, so I just want to make sure

python_package/tt_torch/custom_ops.py
jackzhangTT: sorry "unknown" is probably better. tt-lang kernels will almost always mutate an input args but it's unknown at this point which arg it'll mutate

jameszianxuTT (approved):
Removing my block, we tested manual wheel install and it doesn't break anything. looks like tt-lang was inserted into wheel dependency

thanks for all the discussion and answering my questions

Issue #4632, opened 2026-05-11
Inconsistent PCC for kimi-k2 and deepseek-v3.2-exp benchmarks


gengelageTT:
Both kimi-k2 and deepseek-v3.2-exp see different PCC results between various runs off the same commit. See logs for [kimi-k2 here](https://github.com/tenstorrent/tt-xla/actions/runs/25676680580) and [deepseek here](https://github.com/tenstorrent/tt-xla/actions/runs/25679930838). The same can be replicated locally.

Locally running `test_kimi_k2_layer_sparse_moe`shows different topk results between runs, despite the ttnn graphs being equivalent. The weights are consistent between runs so the random weights doesn't seem to be the issue.

Run 1:
```
[router topk diagnostic] captured 1 router call(s):
  call#0: device=xla:0 shape=(2048, 8) sum=467764 hash=7fd447c5fedfc2a2 first10=[312, 169, 57, 10, 313, 194, 248, 163, 312, 165]
```

Run 2:
```
[router topk diagnostic] captured 1 router call(s):
  call#0: device=xla:0 shape=(2048, 8) sum=436141 hash=52c2e9b7e6fb4bb6 first10=[169, 57, 10, 313, 194, 248, 163, 134, 165, 380]
```

Run 3:
```
[router topk diagnostic] captured 1 router call(s):
  call#0: device=xla:0 shape=(2048, 8) sum=436141 hash=52c2e9b7e6fb4bb6 first10=[169, 57, 10, 313, 194, 248, 163, 134, 165, 380]
```

gengelageTT:
**Update**
Locally was able to fix inconsistent PCC for kimi using fix in `gengelage/fix-kimi-bias-init` in tt-forge-models. Both deepseek and kimi still see inconsistent PCC when run multiple times in CI. My current suspicion is that `torch.manual_seed` in the `run_around_tests()` pytest fixture is not being properly applied to the benchmark tests. This would explain why this behaviour is only seen for these models (to my knowledge), since they load random weights.

gengelageTT:
Both tests are now passing with expected PCC in nightly CI [here](https://github.com/tenstorrent/tt-xla/actions/runs/26133759592/job/76864367697#step:15:509) and [here](https://github.com/tenstorrent/tt-xla/actions/runs/26133759592/job/76864367674#step:15:320).

PR #4713, opened 2026-05-14
Apply torch.manual_seed fixture to benchmark tests

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/4632

### Problem description
Both kimi-k2 and deepseek-v3.2-exp see different PCC results between various runs off the same commit. See logs for [kimi-k2 here](https://github.com/tenstorrent/tt-xla/actions/runs/25676680580) and [deepseek here](https://github.com/tenstorrent/tt-xla/actions/runs/25679930838). Also can be seen in nightly CI runs.

This is because the benchmark workflows run tests with `--confcutdir=tests/benchmark`. This means the following pytest fixture from tests/conftest.py isn't used by benchmark tests running in CI:
```
@pytest.fixture(autouse=True)
def run_around_tests():
    torch.manual_seed(0)
    yield
    torch._dynamo.reset()
    _release_dynamo_bridge_tensors()
```

For models with random weights (or anything randomly initialized) this means that the test will produce different results each time it's run. This isn't the case locally since tests aren't run with `--confcutdir=tests/benchmark` set. The reason kimi-k2 saw variations in PCC locally was since it had an uninitialized weight.

### What's changed
Import `run_around_tests` in tests/benchmark/conftest.py so that it's used by benchmark tests running in CI. This ensures any randomly initialized weights/tensors/etc. will be consistent between runs. 

Set `required_pcc` to default `0.94` for kimi-k2 and `0.92` for deepseek-v3.2-exp.

### Checklist

### DIFF
--- a/tests/benchmark/conftest.py
+++ b/tests/benchmark/conftest.py
@@ -3,6 +3,8 @@
 # SPDX-License-Identifier: Apache-2.0
 import pytest
 
+from tests.conftest import run_around_tests  # noqa: F401 — autouse fixture
+
 
 def make_validator_positive_int(option_name):
     """Create a positive integer validator with the option name in error messages."""
--- a/tests/benchmark/conftest.py
+++ b/tests/benchmark/conftest.py
@@ -2,8 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 import pytest
-
-from tests.conftest import run_around_tests  # noqa: F401 — autouse fixture
+import torch
 
 
 def make_validator_positive_int(option_name):
@@ -144,3 +143,9 @@ def decode_only(request):
 @pytest.fixture
 def check_fusions(request):
     return request.config.getoption("--check-fusions")
+
+
+@pytest.fixture(autouse=True)
+def seed_torch():
+    """Seed torch before every benchmark test for reproducible PCC runs."""
+    torch.manual_seed(0)
--- a/tests/benchmark/test_llms.py
+++ b/tests/benchmark/test_llms.py
@@ -2072,7 +2072,6 @@ def test_kimi_k2_tp_galaxy_2_layers(
         arch="wormhole_galaxy",
         optimization_level=0,
         trace_enabled=False,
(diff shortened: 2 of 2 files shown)


gengelageTT:
Both tests see the same results both times when re-run in the workflow here: https://github.com/tenstorrent/tt-xla/actions/runs/25872872055

The calculated PCCs also match what is seen when running these tests locally:
kimi-k2
```
Prefill PCC verification passed with PCC=0.946076
First decode PCC verification passed with PCC=0.998662
```
deepseek-v3.2-exp
```
Prefill PCC verification passed with PCC=0.927263
First decode PCC verification passed with PCC=0.921720
```

gengelageTT:
Kimi-k2 also requires a fix [in pr here](https://github.com/tenstorrent/tt-forge-models/pull/657) in order to see consistent results between runs (this fix is included in the CI runs linked above). This pr shouldn't be merged until that fix has been uplifted into tt-xla.

tests/benchmark/conftest.py
```diff
@@ -3,6 +3,8 @@
 # SPDX-License-Identifier: Apache-2.0
 import pytest
 
+from tests.conftest import run_around_tests  # noqa: F401 — autouse fixture
```
odjuricicTT: I wonder if everything that this function does is needed, or is it just enough to set torch.manual_seed(0).

Can you try to reimplement only the part that is needed to fix our issue instead of importing the whole thing?

I'd like to keep benchmark tests separate from the other testing infra, so no imports, if possible.

gengelageTT: Made this change and it worked locally with `--confcutdir=tests/benchmark` set. Waiting for [run here](https://github.com/tenstorrent/tt-xla/actions/runs/25926252604) to pass with same PCC to confirm it works in CI.

gengelageTT:
Re-running tests here after rebase which includes fixes fix for kimi in tt-forge-models and changes to decode pcc calculations: https://github.com/tenstorrent/tt-xla/actions/runs/25929114099

[Edit]: tests are passing

vkovacevicTT (approved):

Issue #4204, opened 2026-04-10
Assert that LLM benchmarks generate exactly 2 graphs


odjuricicTT:
## Problem

LLM benchmarks (prefill + decode) should produce exactly **2 graphs** (one prefill, one decode). Graph breaks silently introduce extra graphs that cause host-device synchronization overhead in the decode loop and break trace capture, leading to significant performance regressions.

For example, the Llama 3.1 70B Galaxy benchmark recently had a graph break producing a 3rd tiny graph (position index slice/reshape) that was eliminated by a tt-mlir uplift — resulting in a ~28% throughput improvement (4.39 → 5.64 samples/sec).

## Proposed Change

Add an assertion in the LLM benchmark harness that validates the number of compiled graphs equals the expected count (2 for standard LLMs with prefill + decode). This would catch graph break regressions immediately rather than relying on perf regression checks to surface them after the fact.

## Context

- Nightly run with the regression: https://github.com/tenstorrent/tt-xla/actions/runs/24165410414
- Nightly run after fix: https://github.com/tenstorrent/tt-xla/actions/runs/24219888811/job/70708848882
- The perf report already tracks `ttnn_num_graphs` — this just needs to be asserted against an expected value.

PR #4724, opened 2026-05-15
Add rotary embedding fusion validation to LLM benchmarks

sdjordjevicTT:
### Ticket
N/A

### Summary
- Introduce OpCount and OpCountBetween dataclasses in fusion_check.py for structured, per-layer op count validation in compiled TTNN IR
- Add ttnn.rotary_embedding fusion checks to all LLM benchmark tests (4 per layer for most models, 3-4 range for gpt-oss models, skipped for phi1.5/phi2)
- Fusion stats are always printed; --check-fusions flag makes mismatches fail the test
- Rename PCC/TOPK graph exports with _pcc suffix to distinguish them from perf graphs

### Checklist

### DIFF
--- a/tests/benchmark/fusion_check.py
+++ b/tests/benchmark/fusion_check.py
@@ -5,9 +5,9 @@
 """
 Fusion-pattern verification for benchmark tests.
 
-Accepts a list of expected op strings (e.g. ["ttnn.rms_norm", "ttnn.scaled_dot_product_attention"]),
-globs the ttnn IR files produced by the current run, and searches for each op
-by simple string match. A pattern must appear in every graph file.
+Accepts a list of OpCount objects that specify expected per-layer op counts,
+globs the ttnn IR files produced by the current run, and verifies exact
+occurrence counts (per_layer_count * num_layers) in each graph file.
 
 The current run is identified by `export_model_name`, which embeds a unique
 `_run<hex>_` token, so files from previous runs are naturally excluded.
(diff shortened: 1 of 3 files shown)


tests/benchmark/fusion_check.py
```diff
@@ -5,62 +5,149 @@
...
+            if not passed:
+                failures.append(msg.strip())
+        else:
+            expected_count = check.count * num_layers
+            passed = actual_count == expected_count
+            status = "PASS" if passed else "FAIL"
+            msg = (
+                f"  {status}: '{check.op}' count={actual_count}, "
+                f"expected {expected_count} "
+                f"({check.count} * {num_layers} layers)"
```
vkovacevicTT, line 146: `actual_count` counts all occurrences of a given op in all TTNN IRs.

So what does `expected_count` we specify mean? 
E.g. 
```
expected_ops=[
            OpCount("ttnn.rotary_embedding", 4),
        ],
```
Does `4` here mean we expect 2 ops per layer in prefill + 2 ops per layer in decode graph?

Same for `min_expected` and `max_expected`.

sdjordjevicTT: Yep, that is correct:
2 prefill, 2 decode -> 4

I agree that it needs to be much more explicit. It isn't very obvious now.

sdjordjevicTT: I want to perform this math so that we can validate cases when running models with fewer layers. For example, it would be helpful if validation worked effectively when we run a benchmark with just a single layer.

vkovacevicTT: Got it :)

For clarity, I think it is better to have this number represent per-graph, per-layer count.
So instead of `4` in OpCount, we have `2` and just multiply it with the number of graphs.

sdjordjevicTT: Yes, I agree, make sense totally! Will update in next iteration.

sdjordjevicTT: I was thinking about this, but this math won't fit generally, like for SDPA. For SDPA, we have one fused op for prefill, and the other one for decode, hence the multiply by 2 in this case won't work. :(

vkovacevicTT: If I understand correctly you are referring to `"ttnn.scaled_dot_product_attention"` in prefill and `"ttnn.scaled_dot_product_attention_decode"` in decode.

I agree, but then we need more fine-grained specification e.g.

```
expected_ops=[
    OpCount("ttnn.scaled_dot_product_attention", prefill_count=1, decode_count=0)
    OpCount("ttnn.scaled_dot_product_attention_decode", prefill_count=0, decode_count=1)
    OpCount("ttnn.rotary_embedding", prefill_count=2, decode_count=2)
]
```
since we have decode specific ops like `"ttnn.nlp_concat_heads_decode"`, `"ttnn.nlp_create_qkv_heads_decode"` etc.

We don't necessarily need it in this PR, but let's point out in docstrings/comments what the number in OpCount is representing (sum of counts per layer across all TTNN IRs).

sdjordjevicTT: Yeah, but the problem that I see is that we currently can't solve with a fine-grained count, is that we don't know which graph is prefill and which one is decode, right?

vkovacevicTT: We do know, our dumped IRs look like this:
```
ttnn_falcon3_7b_bs32_isl128_run2955_g0_1776647312003.mlir
ttnn_falcon3_7b_bs32_isl128_run2955_g1_1776647475918.mlir
ttnn_falcon3_7b_bs32_isl128_run2955_g2_1776647694432.mlir
ttnn_falcon3_7b_bs32_isl128_run2955_g3_1776647775132.mlir
```
so we can distinguish g0 as prefill and g1 as decode (or by timestamp at the end of filename).

g2 and g3 are prefill and decode of PCC/accuracy run, although your [change](https://github.com/tenstorrent/tt-xla/pull/4724/changes#diff-6e5a8960831b592893a9a8714226e0e42cddbcb01315ea8f8882d82925f337b4R586-R588) adds "_pcc" to filename and makes it more obvious.

sdjordjevicTT: But in some general case, can there be more than 2 graphs?

vkovacevicTT: Benchmark LLMs should produce exactly two graphs: https://github.com/tenstorrent/tt-xla/issues/4204
If there are more graphs, we should treat it as a graph break regression.

We could assert it in general case as suggested in https://github.com/tenstorrent/tt-xla/issues/4204 or handle it in fusion check logic.

Let's sync offline :)

Issue #2227, opened 2025-11-18
Find a way to clear program cache from user side due to cross-test leaks


jameszianxuTT:
**Background**
We do not have a way to clean-reset device state from user side (eg. python). We do not even have a way to get a handle to the MeshDevice to user python. 

Our devices are less stable than TPUs and our dependencies require caller-side manual resource management (eg. program cache does not have automatic LRU eviction, nor does consteval cache). So, we will need some way for the user to reset device state cleanly. 

This is the sequence of events leading to program cache leaks:
1. Program cache lifetime is bound to device lifetime. 
2. We don't close device between tests, only at process closure. We have no good options for test process isolation. We should not close device between tests to avoid hiding device state persistence bugs.
3. Program cache therefore leaks between tests that reuse the same mesh device
4. Metal exposes a [clear_program_cache member function on a MeshDevice](https://github.com/tenstorrent/tt-metal/blob/023b335eed1adcc536434c34a51f30d52cc42dca/tt_metal/distributed/mesh_device.cpp#L647)
5. PJRT is the only place we have a handle to MeshDevice
6. PJRT does not know when a pytest is over so we have nowhere to hook a call to parentMesh -> clear_program_cache()*MLIR runtime does not / cannot easily track open devices, so a direct pybind to MLIR runtime would require us to pass in the device from Python context. Which we do not have, since it lives in PJRT.

This would be more complex to fix than consteval leaks, but I have some thoughts below:

**Some possible solutions:**
1. Expose clear_program_cache() binding in ttmlir ttnn runtime. Point (3) above is not 100% true, and we can exploit the fact that my post-pytest torchxla clear_compilation_cache function reliably triggers LoadedExecutable destruction at pytest exit. Then, we can hook into LoadedExecutable destruction to clear program cache or do any other post-test cleanup. (e.g. "external RAII")
2. We can make a proper workaround binding into PJRT to call clear_program_cache_of_parent_mesh()  which would hook into the ClientInstance. 
3. We could add some tracking to MLIR runtime so we can track and query devices. This would probably involve some singleton OpenDeviceList  object that gets appended to, and a function like ClearOpenDeviceProgramCaches that could be used as a pure pybind from a python context. Then we would make bindings to the MLIR runtime and expose them in tt-xla.
[...]

PR #4734, opened 2026-05-15
[PJRT] stop clearing program cache on executable destroy

### LINKED ISSUE #4051: [PJRT] stop clearing program cache on executable destroy
Clearing of the program cache when destroying executable instance was introduced to help alieviate OOM issues on device - this was because the conv-like ops were using persisted config tensors (stored in `L1_SMALL`) for their execution. The only way to clean up these tensors is to clear the program cache. Recently, we've switched to storing these small tensors in DRAM - hence it is likely that we don't need to clear the program cache anymore.

mstojkovicTT:
### Ticket
Fixes: #4051

### Problem description
Described in issue.

### What's changed
- Program cache is no longer cleared on `LoadedExecutableInstance` destroy

### Checklist
- [x] Run locally llms to see `L1_SMALL` usage
- [x] Run nightly
- [x] Run perf benchmark

### DIFF
--- a/pjrt_implementation/inc/api/loaded_executable_instance.h
+++ b/pjrt_implementation/inc/api/loaded_executable_instance.h
@@ -47,8 +47,7 @@ public:
   static void bindApi(PJRT_Api *api);
 
   // Virtual destructor for proper cleanup of derived classes
-  // Clears program cache on instance destroy.
-  virtual ~LoadedExecutableInstance();
+  virtual ~LoadedExecutableInstance() = default;
 
   // Casts this loaded executable instance to PJRT_LoadedExecutable pointer.
   operator PJRT_LoadedExecutable *() {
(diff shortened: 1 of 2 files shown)


mstojkovicTT:
As per https://github.com/tenstorrent/tt-xla/issues/2227 discussion, i tested models that were problematic with program_cache ON (conv models) like `DLA` and `VGG` with `TT_RUNTIME_MEMORY_LOG_LEVEL=operation`: 
- [dla.log](https://github.com/user-attachments/files/27815563/dla.log) 
- [vgg11.log](https://github.com/user-attachments/files/27815564/vgg11.log)

As per logs, we can see that `L1_SMALL` is no longer used, and that should be enough for us to turn this feature back on

pilkicTT (approved):
let's just monitor the nightly when this gets merged, to see if there are any unexpected issues with this.

PR #4737, opened 2026-05-17
[fix] tt_torch dynamo: patch wrap_values NameError in NNModuleVariable.call_method

devisettymahidhar608:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/4738

### Problem description
On torch==2.10.0, Dynamo's `NNModuleVariable.call_method` has a bug where the `wrap_values` helper returns `ListIteratorVariable(named_children, ...)` instead of its local `result` list, raising `NameError: cannot access free
variable 'named_children' ...` whenever a traced `forward` calls`.parameters()`, `.children()`, `.modules()` or `.buffers()`. The bug is only reachable when `inline_inbuilt_nn_modules = False`, which `torch_plugin_tt` sets globally, so every such model is broken on tt-xla today (perceiver/Language_Perceiver).Fixed upstream in pytorch/pytorch#174399 (post-2.10), so this PR installs an
equivalent runtime patch until we bump torch.

### What's changed
- Add `apply_dynamo_nn_module_wrap_values_patch()` in `python_package/tt_torch/utils.py` and call it from
  `apply_dynamo_compatibility_patches()` (already runs at `import tt_torch`).
- The patch sources `NNModuleVariable.call_method` via `inspect.getsource`, rewrites the single buggy `named_children` reference inside `wrap_values` to `result` (`count=1`; the legitimate use in the `name == "named_children"`  branch is left untouched), and `exec`s it in the original module's globals.
- Idempotent and self-disables when the upstream source no longer contains the buggy pattern, so the helper becomes a no-op (and can be deleted) once we move to a torch release containing pytorch/pytorch#174399.

Before fix:
```
test_all_models_torch[perceiver/pytorch-Language_Perceiver-single_device-inference]                                                        language    generality  bfloat16       n150         FAILED_FE_COMPILATION      PASSING       N/A                    0.99       PCC_EN   single_device    14.253    N/A           
test_all_models_torch[perceiver/pytorch-Language_Perceiver-single_device-inference]                                                        language    generality  bfloat16       p150         FAILED_FE_COMPILATION      PASSING       N/A                    0.99       PCC_EN   single_device    10.429    N/A           

```
After fix:(models are passing)
https://github.com/tenstorrent/tt-xla/actions/runs/25987835900
https://github.com/tenstorrent/tt-xla/actions/runs/25987856765

### Checklist

### DIFF
--- a/python_package/tt_torch/utils.py
+++ b/python_package/tt_torch/utils.py
@@ -31,6 +31,139 @@ def apply_dynamo_compatibility_patches():
     """Apply TorchDynamo compatibility patches for TT devices."""
     torch._C._accelerator_getDeviceIndex = cpu_device_index
     torch._C._accelerator_getStream = cpu_stream
+    apply_dynamo_nn_module_wrap_values_patch()
+
+
+# TODO: https://github.com/tenstorrent/tt-xla/issues/4806
+_DYNAMO_WRAP_VALUES_BUG_TORCH_VERSION = "2.10.0"
+
+
+def _wrap_values_patch_warn(msg: str) -> None:
+    """
+    Emit a warning about the ``wrap_values`` Dynamo patch.
+
+    Prefers the project's ``ttxla_tools`` logger (matches the convention in
+    the rest of ``tt_torch``), but lazily imported so this module stays
+    importable even if ``ttxla_tools`` is not on the path yet. Falls back to
+    the stdlib ``warnings`` module so the message is never silently dropped.
+    """
+    full_msg = f"[tt_torch] dynamo wrap_values patch: {msg}"
+    try:
+        from ttxla_tools.logging import logger
+
+        logger.warning(full_msg)
+    except Exception:
+        import warnings
+
+        warnings.warn(full_msg, RuntimeWarning, stacklevel=3)
+
+
+def apply_dynamo_nn_module_wrap_values_patch() -> None:
+    """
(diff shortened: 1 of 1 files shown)


sonalibaskaran2499 (approved):

kmabeeTT (commented):
Thanks for finding this failure and digging into it - fix seems fine to me, has some good protections in place and not much else we can do until we consume newer pytorch 2.11.0 with the fix.  On the other hand, this is low priority model, so leaving it fail until we consume new pytorch is another idea unless this error is present in nightly/red models too (is it?)

Let's discuss in meeting tomorrow before merge just to double check we're on same page. If we go forward here, should file a separate issue as reminder to remove workaround and mention it as # TODO comment above this temporary code for completeness I'd say.

devisettymahidhar608:
@kmabeeTT 
FYI — this issue has also been observed in the nightly (red) models, including qwen2_5_vl and qwen_3_5_vl, as well as a few other models.

devisettymahidhar608:
Hi @kmabeeTT , I’ve added a TODO comment along with the issue ticket reference. Could you please review it again?
Thank You

kmabeeTT (approved):

AleksKnezevic (changes requested):
We should have pytorch uplifted to 2.11 soon, which means this fix won't be needed, correct? If that's the case, let's skip.

Issue #4742, opened 2026-05-18
Add Deepseek v4 to CI


hshahTT:
Need to add the following modified DSv4 model tests to tt-xla CI to catch regressions:
On LLMBox:
- Attention tests (Prefill and Decode)
- Flash MoE (hash and non-hash layers)
- Block Prefill and Decode (single layer including MoE and Attention)
- Transformer Prefill and Decode (1, 2, and 3 layers)

On BH Galaxy:
- Full E2E coherence test

PR #4743, opened 2026-05-18
Add LLMBox Deepseek v4 tests

hshahTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/4742

### Problem description
Need to add DSv4 tests to catch regressions.

### What's changed
Added the following LLMBox tests:
- Attention tests (Prefill and Decode)
- Flash MoE (hash and non-hash layers)
- Block Prefill and Decode (single layer including MoE and Attention)
- Transformer Prefill and Decode (1, 2, and 3 layers)

Note: it takes around 35 min to run all these tests.

### Checklist


tests/torch/models/deepseek_v4/test_deepseek_v4_tp.py
```diff
@@ -0,0 +1,399 @@
...
+        shard_specs[attn.indexer.wq_b.weight] = ("model", None)
+        shard_specs[attn.indexer.weights_proj.weight] = ("model", None)
+        shard_specs[attn.indexer.compressor.kv_cache] = ("batch", None, None)
+        shard_specs[attn.indexer.compressor.kv_state] = ("batch", None, None)
+        shard_specs[attn.indexer.compressor.score_state] = ("batch", None, None)
+
+    ffn = layer.ffn
+    if hasattr(ffn, "mlp") and hasattr(ffn.mlp, "experts"):
+        experts = ffn.mlp.experts
+        compound = ("batch", "model") if "batch" is not None else "model"
```
sshonTT: Isn't `if "batch" is not None` is always true?

tests/torch/models/deepseek_v4/test_deepseek_v4_tp.py
```diff
@@ -0,0 +1,399 @@
...
+        # - start_pos==127 with num_layers==4 hits `(start_pos + 1) % ratio == 0`
+        #   for both layer 2 (ratio=4) and layer 3 (ratio=128) — exercises the
+        #   compressor's "compress step" branch (kv_state roll, kv_cache write).
+        (1, 4, utils.PCC_99),
+        (2, 4, utils.PCC_99),
+        (3, 128, utils.PCC_99),
+        (3, 127, utils.PCC_99),
+        (3, 4, utils.PCC_99),
+        (3, 128, utils.PCC_99),
+        (3, 127, utils.PCC_99),
```
sshonTT, line 357: seems duplicated tests (3, 128) and (3, 127)

tests/torch/models/deepseek_v4/test_deepseek_v4_tp.py
```diff
@@ -0,0 +1,399 @@
...
+def transformer_shard_spec(model, args, kwargs):
+    shard_specs: dict = {}
+    for layer in model.layers:
+        _layer_shard_spec(layer, shard_specs)
+    shard_specs[args[0]] = ("batch", None)  # inputs_id: [b, s]
+    return shard_specs
+
+
+@pytest.mark.nightly
+@pytest.mark.parametrize("model_name", ["deepseek-ai/DeepSeek-V4-Flash"])
```
sshonTT, line 294: We probably need a `@pytest.mark.llmbox`

tests/torch/models/deepseek_v4/test_deepseek_v4_moe.py
```diff
@@ -0,0 +1,146 @@
...
+    block = model.Block(HASH_LAYER_ID, args)
+    block.ffn.load_state_dict(
+        weight_loader.load_moe_state_dict(model_name, layer_id=HASH_LAYER_ID)
+    )
+
+    mesh = utils.make_2d_mesh()
+    enable_sparse_mlp(block, mesh=mesh.mesh_shape, cluster_axis=0, config=args)
+    runner = _HashMoERunner(block.ffn)
+
+    ffn = block.ffn
```
sshonTT: Could you remove this one?

tests/torch/models/deepseek_v4/test_deepseek_v4_tp.py
```diff
@@ -0,0 +1,399 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from pathlib import Path
```
sshonTT: Could you clean up the unused imports? Path, np, ComparisonConfig, etc

tests/torch/models/deepseek_v4/weight_loader.py
```diff
@@ -0,0 +1,349 @@
...
+    Returns keys like: gate.weight, gate.bias, experts.{E}.w{1,2,3}.weight,
+    shared_experts.w{1,2,3}.weight — all in bf16 (except gate.bias which
+    remains in its checkpoint dtype, typically fp32).
+    """
+    base = f"layers.{layer_id}.ffn."
+    raw = _load_raw_subset(model_name, [base])
+    return _dequant_paired(raw, base)
+
+
+def load_expert_state_dict(
```
sshonTT: Seems this one is dead code. Could you remove it?

tests/torch/models/deepseek_v4/test_deepseek_v4_tp.py
kmabeeTT: Yeah, needed so it would hit this entry https://github.com/tenstorrent/tt-xla/blob/main/.github/workflows/test-matrix-presets/basic-test-nightly.json#L10C1-L11C1:

```
  { "runs-on": "n300-llmbox",        "name": "run_torch_multi_chip",   "dir": "./tests/torch",                            "test-mark": "nightly and llmbox",                      "shared-runners": "false" },
```

tests/torch/models/deepseek_v4/test_deepseek_v4_moe.py
```diff
@@ -0,0 +1,140 @@
...
+@pytest.mark.nightly
+@pytest.mark.llmbox
+@pytest.mark.parametrize("model_name", ["deepseek-ai/DeepSeek-V4-Flash"])
+@pytest.mark.parametrize("batch_size", [32])
+@pytest.mark.parametrize("seq_len", [1, 32])
+def test_flash_moe_hash(model_name, batch_size, seq_len):
+    enable_spmd()
+    xr.set_device_type("TT")
+
+    args = weight_loader.load_config_args(model_name, False)
```
sshonTT, line 97: nit) not a blocker for this PR, but this is quite different config from real_args() that is used in _tp.py .

sshonTT (approved):

tests/torch/models/deepseek_v4/realistic_inputs.py
```diff
@@ -0,0 +1,443 @@
...
+`hidden_states` corresponds to the input that goes directly into
+`Block(layer_id=LAYER_ID).ffn` — i.e. the value right before the MoE gate
+matmul. To keep the CPU prefix tractable, `LAYER_ID` defaults to
+`args.n_hash_layers` (= 3), the first score-routed layer; the prefix run
+covers `embed -> hc_expand -> layers[0..2] (hash MoE) -> layers[LAYER_ID]
+(hc_pre + attn_norm + attn + hc_post + hc_pre + ffn_norm)`.
+
+First-run cost: ~12-15 GB HF download + several minutes of CPU forward.
+The result is cached as a single `.pt` checked into the repo so subsequent
+runs (and other contributors) just `torch.load` a ~16 MB file.
```
kmabeeTT: Hmm "checked into the repo", is this accurate? I don't see a .pt file checked into tt-xla repo here in this PR and that's probably a good thing because we were always advised against using git-lfs for large files even 16MB in tree.

We should probably consider/try moving _CACHE_DIR under HF_HOME to avoid the 16MB file being regnerated every Nightly CI run. HF_HOME is already persistent (/mnt/dockercache/huggingface in our CI), so the dequant + prefix forward would amortize the same way the weight downloads already do.

tests/torch/models/deepseek_v4/utils.py
```diff
@@ -0,0 +1,148 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
kmabeeTT: nit; 2026

tests/torch/models/deepseek_v4/test_deepseek_v4_tp.py
```diff
@@ -0,0 +1,389 @@
...
+    )
+
+
+@pytest.mark.nightly
+@pytest.mark.llmbox
+@pytest.mark.parametrize("model_name", ["deepseek-ai/DeepSeek-V4-Flash"])
+@pytest.mark.parametrize("bsz", [1, 4, 32])
+@pytest.mark.parametrize("prefill_seq_len", [32, 64, 128])
+@pytest.mark.parametrize("is_compression_layer", [False, True])
+def test_attention_decode(model_name, bsz, prefill_seq_len, is_compression_layer):
```
kmabeeTT, line 125: `prime_decode_kv_buffers` is called in `test_transformer_decode` but not in `test_attention_decode`/`test_block_decode`. Won't the kv_cache be all zeros there, so decode reads zero context? If so, PCC will pass trivially — is that the intended coverage, or should we prime here too?

tests/torch/models/deepseek_v4/realistic_inputs.py
```diff
@@ -0,0 +1,443 @@
...
+        h, block.hc_attn_fn, block.hc_attn_scale, block.hc_attn_base
+    )
+    h = block.attn_norm(h)
+    h = block.attn(h, 0)
+    h = block.hc_post(h, residual, post, comb)
+    residual = h
+    h, post, comb = block.hc_pre(
+        h, block.hc_ffn_fn, block.hc_ffn_scale, block.hc_ffn_base
+    )
+    h = block.ffn_norm(h)
```
kmabeeTT, line 378: Hmm... could drift over time and produce stale activations, no? Worth adding an assertion that the captured hidden_states.shape matches what a real Block(...).ffn_norm output would be, or refactoring to invoke the real forward with a hook that captures the post-ffn_norm value if possible.

kmabeeTT (approved):
Few small things, approved in the meantime. 50 minutes for this single model in nightly is a noticeable amount, worried about job timeouts, would be nice if it could be trimmed without losing too much coverage.

tests/torch/models/deepseek_v4/realistic_inputs.py
hshahTT: Yeah "checked in" is the wrong phrase. It just generates a .pt file under the `tests/torch/models/deepseek_v4` directory the first time any test asking for realistic inputs is ran. I can move it to under HF_HOME though.

tests/torch/models/deepseek_v4/test_deepseek_v4_tp.py
hshahTT: I added it to test_block_decode but we don't need it in test_attention_decode since the primary intention of that test is just to test the code path when seq_len=1

hshahTT:
Addressed comments and removed some parameterizations to reduce the total time from 50min to around 35 min. We can probably remove more if there's job timeout issues.

PR #4750, opened 2026-05-18
[WIP] Remove identity variant of `mesh_shard` op

dmilinkovicTT:
### Ticket
N/A

### Problem description
Aligns the PJRT plugin with the tt-mlir [PR](https://github.com/tenstorrent/tt-mlir/pull/8380) that removes the `Identity` variant of the `mesh_shard`.

### What's changed
- **`module_builder.{h,cc}`**: replace the per-frontend `collectInput/OutputShardingsGSPMD/Shardy` + `createShardingsFromGSPMD/Shardy` machinery with single calls to the new `mlir::tt::sharding_utils::collectArgMeshShardings` / `collectResultMeshShardings` helpers, in order to deduplicate the `MeshSharding` instantiation logic between tt-mlir and tt-xla.
- **`loaded_executable_instance.cc`**: drop the now-dead `MeshShardType::Identity` branches in `getOutputShape` (only `Replicate` short-circuits return `output_shape` as-is) and `fillStrategyMapFromSharding` (no longer maps any case to the `"identity"` strategy).
- **`executable_image.cc`**: drop the per-output shape sanity validation in `FlatbufferExecutableImage`'s constructor. It compared MLIR's pre-conversion (global, complex-preserving) shape to the flatbuffer's per-device, complex-split shape, which diverge by exactly the SHLO→TTIR conversion's intentional rewrites - the check is unsound at that layer. The count checks above still catch real flatbuffer/MLIR drift.

### Checklist
- [x] Existing tests provide coverage

### DIFF
--- a/pjrt_implementation/src/api/executable_image.cc
+++ b/pjrt_implementation/src/api/executable_image.cc
@@ -228,48 +228,6 @@ FlatbufferExecutableImage::FlatbufferExecutableImage(
            "getNumOutputs()={}, output_specs.size()={}",
            this->getNumOutputs(), output_specs.size());
 
-  int output_dims_so_far = 0;
-  for (size_t output_index = 0; output_index < getNumOutputs();
-       ++output_index) {
-    // Complex types are lowered by the TT-MLIR compiler to float tensors with
-    // a trailing dimension of 2 (interleaved real/imag pairs), so the
-    // flatbuffer shape will have one extra trailing dimension compared to the
-    // MLIR module shape.
-    std::vector<std::uint32_t> expected_shape =
-        getOutputDimensions()[output_index];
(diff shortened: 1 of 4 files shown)


pjrt_implementation/src/api/executable_image.cc
```diff
@@ -228,48 +228,6 @@ FlatbufferExecutableImage::FlatbufferExecutableImage(
            "getNumOutputs()={}, output_specs.size()={}",
            this->getNumOutputs(), output_specs.size());
 
-  int output_dims_so_far = 0;
```
acicovicTT: Can you provide an explanation for this removal in the PR description, since it's not obvious how it is connected to the removal of identity mesh_shard op?

PR #4789, opened 2026-05-19
Reduce vLLM decode graphs from 5 to 2

### LINKED ISSUE #4468: reduce vllm decode graphs
Currently there are 5 graphs for each vllm decode step:

1. prepare inputs
2. main model
3. select hidden states
4. compute logits
5. sample

This should be reduced to 1/2 graphs.

alinakhanTT:
## Ticket
closes #4468

## Problem description
Each vLLM decode step dispatched 5 separate compiled graphs (select hidden states, compute logits, structured decode, sample, plus backbone), adding unnecessary dispatch overhead on every generated token.

## What's changed
Each decode step now dispatches **2 compiled graphs** (backbone + fused postprocess), down from 5.

- Combine `select_hidden_states`, `compute_logits`, `structured_decode`, and `sample_from_logits` into a single `decode_postprocess` compiled function
- Restrict `decode_postprocess` precompilation to `num_tokens=1` only — decode always processes exactly 1 token per request, so larger bucket sizes compile unnecessary variants. For prefill batches, the last token is selected eagerly outside the compiled function so `decode_postprocess` always receives `(max_num_reqs, 1, hsize)`
- Restore separate `select_hidden_states`, `compute_logits`, `structured_decode` precompiles for the `cpu_sampling=True` path which still needs them directly
- Return both `selected_token_ids` and `logits` from `decode_postprocess` so the fused path can satisfy `gather_logprobs` (which needs raw logits) without recomputation
- Precompile `decode_postprocess` with non-trivial dummy values for `require_struct_decoding` and `grammar_bitmask` so XLA's optimizer doesn't constant-fold the structured-decode masking branch away based on zero-valued CPU warmup buffers (which previously caused regex-constrained outputs to ignore the grammar bitmask)

Total precompiled variants for opt-125m: **6** (3 backbone + 3 postprocess), down from ~10 on main. Per-step dispatch is unchanged at 2 graphs. The cache count is 6 rather than 5 because the structured-decode masking now genuinely participates in the graph; the earlier 5 came from XLA eliding the masking path against zero-valued precompile dummies.

Verified breakdown for opt-125m:
- backbone num_tokens=1: 2 graphs
- backbone num_tokens=128: 1 graph
- postprocess all_greedy=False: 2 graphs (decode_postprocess + sampler)
- postprocess all_greedy=True: 1 graph (decode_postprocess argmax path)

## Checklist
- [x] `opt-125m-fused-measure` benchmark passes with 6 total compiled graphs
- [x] `VLLM_XLA_CHECK_RECOMPILATION=1` confirms no recompilation during inference
- [x] `test_logprobs[*]` passes locally (single_device) and in CI (n300)
- [x] `test_structured_outputs_regex[single_device]` passes locally with correctly constrained output
- [x] passing all nightly vllm tests - https://github.com/tenstorrent/tt-xla/actions/runs/26673774736


AleksKnezevic:
Awesome work! Two questions:

1. How does benchmark perf compare before/after your changes? 
2. Is it possible to merge the two graphs such that there is only one decode graph?

alinakhanTT:
Perf benchmark passed: https://github.com/tenstorrent/tt-xla/actions/runs/26177614614

AleksKnezevic:
I don't think we currently assert on performance. Can you verify whether there were any drops in perf?

alinakhanTT:
Thank you!! Here is the manual comparison

1. Benchmark perf before/after
  Compared run before changes vs run after:
   TP models (LLMBox) : 8–24% throughput improvement:
    | Model           | Before    | After     | Δ    |
    |-----------------|-----------|-----------|------|
    | qwen3-8b-tp     | 10.41 sps | 12.93 sps | +24% |
    | falcon3-10b-tp  | 8.07 sps  | 9.73 sps  | +21% |
    | qwen3-14b-tp    | 9.51 sps  | 10.68 sps | +12% |
    | falcon3-7b-tp   | 11.87 sps | 13.19 sps | +11% |
    | ministral-8b-tp | 12.33 sps | 13.36 sps | +8%  |
  
TP models benefit most because each graph dispatch on multi-chip hardware involves inter-chip synchronization - reducing 5 dispatches to 2 also cuts sync points from 5 to 2.
 
Single-chip (n150): mostly stable, a few improvements (opt-125m +16%, qwen2.5-0.5b +16%). Two regressions - qwen3-1.7b (-20%) and qwen3-0.6b (-9%) both very low absolute throughput (<1  sps), will verify with a follow-up benchmark run

2. Can we get to 1 decode graph?
Yes, but after looking into it I think the gain would be minimal. The 2 dispatches per decode step are the backbone (which dominates decode time - all attention + FFN compute) and the fused decode_postprocess (this PR). Merging them requires moving sampling metadata construction before the backbone and deferring prompt logprobs. But since the backbone dominates, eliminating the 2nd dispatch wouldnt save much in practice. Happy to explore 2->1 as a follow-up if useful

AleksKnezevic:
Thanks for the analysis. I think it's probably worth exploring. What I have found is that with trace mode enabled each graph requires its inputs to be on host before trace replay can start. This means that we have to move hidden state back to host between the backbone and sampling. I agree that it's probably not a huge overhead, but due to the way generative workloads run (you can't start the next token until you've finished the current one) any overhead in the hot path should be optimized. 

I agree that it can be done in a subsequent PR.

alinakhanTT:
That makes a lot of sense about the trace mode behavior, I hadn't considered the transfer cost between graphs on every token. Should I go ahead and open a tracking issue for it, or wait until it gets assigned?

AleksKnezevic:
Please do and work with @mmanzoorTT on prioritizing it.

alinakhanTT:
Will do!!

alinakhanTT:
Updated passing performance benchmark: https://github.com/tenstorrent/tt-xla/actions/runs/26406115859

alinakhanTT:
## Benchmark: before vs after

Baseline: main pre-PR. After: [run 26406115859](https://github.com/tenstorrent/tt-xla/actions/runs/26406115859).

### TP models (LLMBox, 8-chip): +8–24% throughput improvement

| Model | Before (sps) | After (sps) | Δ |
|---|---:|---:|---:|
| qwen3-8b-tp | 10.41 | 12.93 | +24% |
| falcon3-10b-tp | 8.07 | 9.73 | +21% |
| qwen3-14b-tp | 9.51 | 10.68 | +12% |
| falcon3-7b-tp | 11.87 | 13.19 | +11% |
| ministral-8b-tp | 12.33 | 13.36 | +8% |

TP models benefit most because each graph dispatch on multi-chip hardware involves inter-chip synchronization, reducing 5 dispatches to 2 also cuts sync points from 5 to 2.

### Single-chip (n150): mostly stable

A few improvements: opt-125m +16%, qwen2.5-0.5b +16%. Two regressions: qwen3-1.7b (-20%) and qwen3-0.6b (-9%) - both very low absolute throughput (<1 sps). Will verify with a follow-up benchmark run and update this comment.

nightly-reports/benchmark_compare_26177614614_vs_26176774426.md
```diff
@@ -0,0 +1,132 @@
+# vLLM Benchmark Comparison: 5→2 Decode Graph Reduction
+
+**PR branch run:** [26177614614](https://github.com/tenstorrent/tt-xla/actions/runs/26177614614) (`akhan/vllm-reduce-decode-graphs-clean`, SHA `6a956c97`)
```
mvasiljevicTT: Do we need this file on main?

test_graph_counts.py
```diff
@@ -0,0 +1,108 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Graph count validation for PR #4789 (reduce vLLM decode graphs 5→2).
```
mvasiljevicTT: Is this something that will be used outside this PRs testing? Please move it to `tests/benchmark/scripts` if you think it will be useful to have it on main, or remove if it was temporary file

test_tp_quick.py
```diff
@@ -0,0 +1,87 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Quick TP validation using opt-125m (no HuggingFace token needed).
+
+Tests both use_2d_mesh=True and use_2d_mesh=False to exercise the
```
mvasiljevicTT: Same question as for the other test script, this one seems to be very specific.

alinakhanTT: removed all 3 files, as they are not required on main!

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1962,6 +2080,7 @@ def profile_run(
     ) -> None:
         if self.tt_config.decode_only:
             return
+        return  # TODO(#4468): profile_run adds no value currently; remove once confirmed
```
kmabeeTT: Just curious, was there an issue here without this return you were seeing?  My change recently merged that is now on your branch makes this change not needed so you could drop it, I already commented out calling of `profile_run()` via https://github.com/tenstorrent/tt-xla/commit/76e75e52781f6f8dce0957690a9c07b3b1d4234b

tests/benchmark/test_vllm_benchmarks.py
```diff
@@ -150,6 +150,11 @@ def _tp_config(
         _tp_config("mistralai/Mistral-Small-24B-Instruct-2501", 1),
         id="mistral-small-24b-instruct-2501-tp",
     ),
+    # temp: measure fused decode_postprocess graph count (cpu_sampling=False path)
```
kmabeeTT: Is this test intended to be merged? I noticed the word "temp" here, wasn't sure :)

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2222,6 +2341,49 @@ def sample_from_logits(
             out_tokens = self.sampler(logits, sampling_metadata).sampled_token_ids
         return out_tokens
 
+    @torch.compile(backend="tt", fullgraph=True, dynamic=False)
+    def decode_postprocess(
```
kmabeeTT, line 2525: Before this PR, both `cpu_sampling=True` and `cpu_sampling=False` shared a single call to `select_hidden_states` / `compute_logits` / `structured_decode`; only the final sampler call branched. After this PR, `decode_postprocess` inlines copies of all four bodies, while the `cpu_sampling=True` path keeps calling the originals — so e.g. a future fix to `apply_grammar_bitmask` only updates one path unless someone remembers to mirror it.

  Two possible options to consider if you didn't already:
  1. Try calling the helpers directly from inside `decode_postprocess` — with `fullgraph=True`, Dynamo should inline them into the same graph. If that works, the duplication goes away.
  2. If nested `@torch.compile` / SPMD constraints prevent that, add a comment at the top of `decode_postprocess` flagging that its body must stay in sync with the four originals.

kmabeeTT (approved):
Nice - looks great in general and perf improvements are nice, few small comments .

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Dropped it! Didn't realise your commit had already stopped calling profile_run()

tests/benchmark/test_vllm_benchmarks.py
alinakhanTT: It is intended to be merged, removed the temp prefix

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Went with option B ,added a NOTE to the docstring flagging that this method mirrors select_hidden_states, compute_logits, structured_decode/apply_grammar_bitmask, and sample_from_logits, and that fixes need to be mirrored!!

vvukomanTT (approved):

PR #4350, opened 2026-04-21
[d2m] tt-xla integration

dlokeTT:
### Summary
Adds TTMetal flatbuffer ("d2m") backend support to the PJRT plugin. The TTNN path is unchanged — TTMetal is opt-in via a new compile option and a matching client-create option.

### What's changed
- **New compile backend `ttmetal_flatbuffer`** — `BackendRuntime::TTMetalFlatbuffer` added; `module_builder` grows a TTIR → TTMetal pipeline (`convertFromTTIRToTTMetal`, `createTTMetalFlatbufferBinary`, `buildModuleForTTMetalRuntime`).
- **New PJRT client-create option `backend` (`"ttnn"` / `"ttmetal"`)** — `client_instance.cc` calls `tt::runtime::setCurrentDeviceRuntime(...)` before any mesh device / buffers are created, since tt-runtime tensors are tagged with their runtime and can't cross runtimes

### Checklist
- [ ] New/Existing tests provide coverage for changes

PR #4811, opened 2026-05-20
Add streaming inference for DeepSeek-V4-Flash

sshonTT:
### Ticket
#4592 

### Problem description
DeepSeek-V4-Flash full model can't be loaded into host RAM in one shot, so we need a layer-by-layer streaming path: load one block, ship to device, free host storage, repeat. The existing full-model loader OOMs the host before inference starts.

### What's changed
Each layer is loaded from HF, sparse-MLP-rewritten, shipped, then released before the next layer starts. KV-cache state lives in persistent device buffers spliced into each block so cross-step state survives the streaming load.

streaming.core.run_streaming is model-agnostic; the per-model knobs (skeleton, sharding, forward signatures, weight loader, MoE swap hook) sit behind streaming.adapters.base.ModelAdapter. DeepSeek-V4-Flash adapter + HF weight loader live alongside.

See streaming/README.md for usage and env-var overrides.

### Limitations

- **Batch size ≤ 8**: larger batches (16, 32) hit a numerical divergence in the MoE routed-combine path at the last sequence position, producing incoherent / multilingual leakage in the decoded output. Root cause is under investigation.
- **Blackhole galaxy only** : For bhglx at exabox env (or other possible bhglx) you need a patch for addressing mesh issue. Please see the attached patch file.
[hard_code_exabox.patch](https://github.com/user-attachments/files/28072078/hard_code_exabox.patch)

### Checklist

### DIFF
--- a/tests/torch/models/deepseek_v4/test_deepseek_v4_e2e_streaming.py
+++ b/tests/torch/models/deepseek_v4/test_deepseek_v4_e2e_streaming.py
@@ -1,0 +1,595 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Streaming-inference end-to-end test for DeepSeek-V4-Flash on Tenstorrent.
+
+The full 43-layer MoE model is too large to hold in host RAM all at once, so
+we stream it: build a weight-less skeleton, then for each transformer block
+load its weights from HuggingFace, ship them (sharded) to device, and release
+the host copy before moving to the next block. Peak host RAM stays bounded to
+roughly one block's worth of staging.
+
+The flow, top to bottom:
+
+    build skeleton (no weights)
+    ship top-level params (embed / norm / head)
+    pre-allocate persistent KV buffers on device
+    for block in model.layers:
+        load HF weights → swap MoE → ship sharded → dummy-flush
+    re-zero KV state
+    model = torch.compile(model, backend="tt")
+    prefill + decode loop
+
+This is a run-to-completion smoke test: it generates tokens and prints the
+decoded continuations, but does not check PCC against a CPU reference.
+
+    pytest -svv tests/torch/models/deepseek_v4/test_deepseek_v4_e2e_streaming.py
+"""
+from __future__ import annotations
+
+import gc
+import logging
+import sys
+import time
+import warnings
+from typing import Dict, List, Tuple
+
+import numpy as np
+import pytest
+import torch
+import torch.nn as nn
+import torch_xla
+import torch_xla.core.xla_model as xm
+import torch_xla.distributed.spmd as xs
+import torch_xla.runtime as xr
+from infra.utilities.torch_multichip_utils import enable_spmd
+from torch_xla.distributed.spmd import Mesh
+from tt_torch.sharding import sharding_constraint_hook
+from tt_torch.sparse_mlp import enable_sparse_mlp
+from ttxla_tools.logging import logger
+
+from third_party.tt_forge_models.deepseek_v4.modified_model import (
+    model_decode_opt as mdo,
+)
+
+from . import weight_loader
+
+# ---- run configuration (intentionally fixed for this example) ----
+MODEL_NAME = "deepseek-ai/DeepSeek-V4-Flash"
+BATCH_SIZE = 8
+MAX_NEW_TOKENS = 16
+PROMPT_LEN = 128
+
+# First BATCH_SIZE of these are tokenized and run.
+PROMPTS = [
+    "How are you today?",
+    "What is the capital of France?",
+    "Explain machine learning briefly.",
+    "Who painted the Mona Lisa?",
+    "What is two plus two?",
+    "Tell me a fun fact about space.",
+    "What is photosynthesis?",
+    "How does a transformer model work?",
+]
+
+
+# ---------------------------------------------------------------------------
+# Host-memory diagnostics
+# ---------------------------------------------------------------------------
+def _log_mem(tag: str) -> None:
+    """Log host RSS so the bounded per-layer footprint is visible. Big weight
+    tensors are mmap-backed, so freeing them (del + gc) returns the pages to
+    the OS and RSS drops without an explicit malloc_trim."""
+    import os
+
+    import psutil
+
+    rss = psutil.Process(os.getpid()).memory_info().rss / 1e9
+    sys_used = psutil.virtual_memory().used / 1e9
+    logger.info(f"[mem {tag:24s}] rss={rss:6.2f} sys={sys_used:6.2f} GB")
+
+
+# ---------------------------------------------------------------------------
+# Mesh + sharded upload helpers
+# ---------------------------------------------------------------------------
+def _make_mesh() -> Tuple[Mesh, Tuple[int, int]]:
+    """2D device mesh: axis 0 = batch (data) parallel, axis 1 = model/tensor."""
+    n = xr.global_runtime_device_count()
+    if n == 32:
+        mesh_shape = (4, 8)
+    elif n == 8:
+        mesh_shape = (2, 4)
+    else:
+        mesh_shape = (1, n)
+    logger.info(f"[mesh] num_devices={n} mesh_shape={mesh_shape}")
+    return Mesh(np.arange(n), mesh_shape, ("_axis_0", "_axis_1")), mesh_shape
+
+
+def _upload(cpu_tensor: torch.Tensor, mesh, partition_spec, device) -> torch.Tensor:
+    """Move a CPU tensor to the XLA device and annotate its shard spec.
+    `partition_spec=None` leaves it replicated across all devices."""
+    xla_t = cpu_tensor.to(device)
+    if partition_spec is None:
+        return xla_t
+    if len(partition_spec) != cpu_tensor.dim():
+        raise ValueError(
+            f"partition_spec {partition_spec!r} rank != tensor shape "
+            f"{tuple(cpu_tensor.shape)}"
+        )
+    xs.mark_sharding(xla_t, mesh, partition_spec)
+    return xla_t
+
+
+def _ship_module(module: nn.Module, spec_by_id: Dict[int, Tuple], mesh, device) -> None:
+    """Replace every Parameter and Buffer in `module` with a device-resident,
+    sharded copy. Drops the source CPU tensors so the caller can gc them.
+    Tensors absent from `spec_by_id` upload replicated."""
+    for sub in module.modules():
+        for name, p in list(sub._parameters.items()):
+            if p is None or p.device.type != "cpu":
+                continue
+            xla_t = _upload(p.data.detach(), mesh, spec_by_id.get(id(p)), device)
+            sub._parameters[name] = nn.Parameter(xla_t, requires_grad=False)
+        for name, b in list(sub._buffers.items()):
+            if b is None or b.device.type != "cpu":
+                continue
+            xla_t = _upload(b.detach(), mesh, spec_by_id.get(id(b)), device)
+            sub._buffers[name] = xla_t
+
+
+# ---------------------------------------------------------------------------
+# Buffer plumbing (persistent KV state lives on device across the run)
+# ---------------------------------------------------------------------------
+def _buffer_paths(block) -> List[Tuple[nn.Module, str, str]]:
+    """(submodule, buffer_name, dotted_path) for every registered buffer."""
+    out = []
+    for sub_path, sub in block.named_modules():
+        for name, buf in list(sub._buffers.items()):
+            if buf is None:
+                continue
+            full = f"{sub_path}.{name}" if sub_path else name
+            out.append((sub, name, full))
+    return out
+
+
+def _ship_persistent_buffers(block, mesh, device) -> Dict[str, torch.Tensor]:
+    """Ship every CPU buffer in `block` to device (batch-sharded on axis 0 for
+    rank>=3, else replicated). Returns {path: device_tensor} for re-splicing."""
+    out: Dict[str, torch.Tensor] = {}
+    for sub, name, full in _buffer_paths(block):
+        b = sub._buffers[name]
+        if b is None or b.device.type != "cpu":
+            if b is not None:
+                out[full] = b
+            continue
+        spec = (
+            ("_axis_0",) + (None,) * (b.dim() - 1)
+            if b.dim() >= 3
+            else (None,) * b.dim()
+        )
+        xla_t = _upload(b.detach(), mesh, spec, device)
+        sub._buffers[name] = xla_t
+        out[full] = xla_t
+    torch_xla.sync(wait=True)
+    return out
+
+
+def _splice_persistent_buffers(block, bufs: Dict[str, torch.Tensor]) -> None:
+    """Point each block buffer at the persistent device tensor of the same path."""
+    for sub, name, full in _buffer_paths(block):
+        if full in bufs:
+            sub._buffers[name] = bufs[full]
+
+
+# DS V4 buffers the per-block dummy forward mutates and that are sensitive to
+# their initial value. `score_state` is consumed by softmax(dim=1) and must
+# start at -inf so unwritten slots get zero probability mass; the rest start
+# at zero. Window-attn `kv_cache` is included because the dummy flush dirties
+# it too.
+_MUTABLE_BUFFERS = {"kv_cache", "kv_state", "score_state"}
+
+
+def _mutable_init(name: str, shape, dtype) -> torch.Tensor:
+    if name == "score_state":
+        return torch.full(shape, float("-inf"), dtype=dtype)
+    return torch.zeros(shape, dtype=dtype)
+
+
+def _rezero_kv(block, bufs: Dict[str, torch.Tensor], mesh, device) -> None:
+    """Re-ship clean init values for the mutable KV buffers a dummy flush
+    dirtied, updating both the block and the persistent-buffer map."""
+    for sub, name, full in _buffer_paths(block):
+        if name not in _MUTABLE_BUFFERS:
+            continue
+        b = sub._buffers[name]
+        if b is None:
+            continue
+        spec = (
+            ("_axis_0",) + (None,) * (b.dim() - 1)
+            if b.dim() >= 3
+            else (None,) * b.dim()
+        )
+        cpu_init = _mutable_init(name, b.shape, b.dtype)
+        xla_t = _upload(cpu_init, mesh, spec, device)
+        sub._buffers[name] = xla_t
+        bufs[full] = xla_t
+        del cpu_init
+
+
+# ---------------------------------------------------------------------------
+# Model construction + per-model sharding specs (DeepSeek-V4-Flash)
+# ---------------------------------------------------------------------------
+def _build_skeleton() -> nn.Module:
+    """Empty (weight-less) Transformer in bf16 sized for this run. Uses the
+    full layer count from the model config."""
+    args = weight_loader.load_config_args(MODEL_NAME, force_bf16=True)
+    args.n_mtp_layers = 0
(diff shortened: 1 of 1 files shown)


AleksKnezevic:
Why do we need the patch? Can we either:

1. Provide defaults for num_devices such that llmbox goes to 1x8 and glx goes to 8x4
2. Allow the user to open up the correct mesh size from the get-go via `client_create_options` as is used [here](https://github.com/tenstorrent/tt-xla/pull/4350) to select backed

streaming/streaming_loader.py
```diff
@@ -0,0 +1,171 @@
...
+        raise ValueError(
+            f"partition_spec {partition_spec!r} not applicable to tensor "
+            f"with shape {tuple(cpu_tensor.shape)}; pass partition_spec=None "
+            "if replication is intentional."
+        )
+    return xm.send_cpu_data_to_device(
+        [cpu_tensor],
+        device,
+        input_sharding=spec,
+    )[0]
```
jameszianxuTT: I am a bit scared to overwrite the standard transfer-to-device path in torchxla without fully understanding the implications of abandoning the backing `XLATensor::tensor_data` field - is there precedent somewhere that this works and doesn't violate other invariants later on, or introduce other bad state? (Other than the fact that it passes tests)

jameszianxuTT: There are two paths right now to make it to BufferFromHostBuffer with long callstacks, where we either:
- use mark_sharding to eagerly push a tensor to xla device with a given shard spec
- wait for sync to implicitly push a tensor to xla device with replicated spec

And both involving CreateTensorsData. 

This is a third path now, seemingly bypassing (some of?) the global side effects involved in the other two paths

streaming/README.md
```diff
@@ -0,0 +1,54 @@
+# Streaming Inference
+
+Layer-streaming inference for large MoE models that don't fit on host as a
+whole. Each layer is loaded from HuggingFace, sparse-MLP-rewritten, and
+shipped to device one at a time so peak host RAM stays bounded.
```
jameszianxuTT: Moving consteval inputs to device is required for streaming to work in its current form right (to avoid lingering host copies for inputs at each layer)? If so, we should mention that the device DRAM footprint will be increased, and streaming feature will induce device OOM if there is not much device DRAM headroom

streaming/streaming_loader.py
sshonTT: Good question, I tightened the comment block in `streaming_loader.py` to make this clearer. We are not bypassing the standard transfer path: `_upload_with_sharding` calls `xm.send_cpu_data_to_device(..., input_sharding=spec)`, which is the same public API torch_xla uses internally for the sync-time push you described. 
The only difference vs relying on lazy sync is that we pass an explicit shard spec so per-shard transfer happens at call time instead of being deferred to first use as replicated.

The reason we need this third path is that both standard paths upload the tensor as replicated first (`mark_sharding` annotates an already- on-XLA tensor; lazy sync uploads as replicated). For DS V4 Flash that would (a) spike device DRAM to ~num_chips × weight_size during the upload window, and (b) leave behind a `tensor_data` CPU shadow that pins host RAM per layer, both of which break the bound streaming is trying to enforce. The `send_cpu_data_to_device(input_sharding=spec)` call pre-shards on host and uploads handle-only, which is what lets streaming actually free both host and device memory between layers.

streaming/README.md
sshonTT: Yes, `enable_const_eval_inputs_to_system_memory=False` is required for streaming to actually bound host RAM (otherwise const-eval would repopulate the host shadow we just freed). The trade-off is that
const-eval inputs stay in device DRAM for the lifetime of the program. I added a "Device DRAM trade-off" section to `streaming/README.md` calling out the implication and that this can OOM on device if DRAM headroom is tight.

Also for bigger model like deepseek pro, I already have plan to dealloc logic that frees input so we can reduce dram footprint once we get it.

AleksKnezevic: They're not there for the lifetime of the program, right, just until the layer is run. This is how we were able to actually run the 1.7T pro model. Otherwise it would not have fit

sshonTT:
> Why do we need the patch? Can we either:
> 
> 1. Provide defaults for num_devices such that llmbox goes to 1x8 and glx goes to 8x4
> 2. Allow the user to open up the correct mesh size from the get-go via `client_create_options` as is used [here](https://github.com/tenstorrent/tt-xla/pull/4350) to select backed

I think we can go with option 2. The thing is, this issue is specifically observed on BH GLX in the exabox environment, whereas other machines like `g08blx04` aren't showing this behavior. We can discuss how to handle those edge cases in a separate issue.

AleksKnezevic:
> > Why do we need the patch? Can we either:
> > 
> > 1. Provide defaults for num_devices such that llmbox goes to 1x8 and glx goes to 8x4
> > 2. Allow the user to open up the correct mesh size from the get-go via `client_create_options` as is used [here](https://github.com/tenstorrent/tt-xla/pull/4350) to select backed
> 
> I think we can go with option 2. The thing is, this issue is specifically observed on BH GLX in the exabox environment, whereas other machines like `g08blx04` aren't showing this behavior. We can discuss how to handle those edge cases in a separate issue.

Ah, got it. Yeah, if we expect this to pass in CI as-is, then in can be in a follow-up.

streaming/streaming_loader.py
sshonTT: I've tried standard version, and I have no issue on it. Swapped `_upload_with_sharding` over to `.to(device)` + `mark_sharding`. host RSS profile (per-layer load → flush) actually came out a touch
similar (even lower) to the explicit `send_cpu_data_to_device(input_sharding=…)` path, and outputs match. Removing the third-path concern. Thanks for review

jameszianxuTT: Thanks for testing out existing to device path. I haven't inspected memory lifetime of XLA tensor data closely so I don't really know what is accurate w/r/t memory overhead `xm.send_cpu_data_to_device` vs to device + mark sharding.

 I am more comfortable going down the standard to device + mark sharding path, so I think it is good that we can keep it.

jameszianxuTT: I've spent a while looking at this on my multihost setup and trying to understand by debugging torch-xla internals to see when the allocations get freed

> I've tried standard version, and I have no issue on it. Swapped _upload_with_sharding over to .to(device) + mark_sharding. host RSS profile (per-layer load → flush) actually came out a touch
similar (even lower) to the explicit send_cpu_data_to_device(input_sharding=…) path, and outputs match.

I think there may be some truth in the leak of tensor host memory by torch-xla in a way we cannot easily deallocate. I see a dupe allocation after moving tensor.to XLA device that I cannot get to deallocate except by process exit, even though done with host buffer callbacks are firing and atensource destructors are being called... investigating.

streaming/weight_loaders/deepseek_v4_flash.py
```diff
@@ -0,0 +1,508 @@
...
+    if repo_id:
+        REPO_ID = repo_id
+    # CPU affinity: parent passes worker's core IDs via DSV4_WORKER_CORES
+    # (comma-separated). Apply via sched_setaffinity so both this Python
+    # process and any threads/OMP children stay inside the partition.
+    cores_str = os.environ.get("DSV4_WORKER_CORES", "")
+    if cores_str:
+        cores = {int(c) for c in cores_str.split(",") if c}
+        os.sched_setaffinity(0, cores)
+    # Match torch's BLAS pool size to the partition.
```
jameszianxuTT: What is going on here? What is being paralleled over these worker cores, why are we overriding torch num threads and all these other global side effects? 

I also don't see these prefetch worker functions called anywhere?

streaming/core.py
```diff
@@ -0,0 +1,613 @@
...
+        bsz=bsz,
+        prompt_len=config.prompt_len,
+        max_new_tokens=config.max_new_tokens,
+    )
+    layers = adapter.get_layers(model)
+    n_layers = len(layers)
+    _log("baseline")
+
+    # freqs_cis aliasing (model-side optimization; adapter may rely on this
+    # to keep compile cache hits across same-cr layers).
```
jameszianxuTT: Is this actually being used? Having model specific behavioral overrides in streaming/core seems like a confusing antipattern; and why not make this change to the modified model (or is it a change only needed in context of streaming, as the comment above might suggest)

jameszianxuTT: It seems like this could belong in some model adapters callback table if we really need model awareness in streaming

streaming/weight_loaders/deepseek_v4_flash.py
sshonTT: Good catch. leftovers from an earlier prefetch experiment the runner no longer uses.

streaming/core.py
sshonTT: Agreed. the model specific logic no lives in the flash adapter.

sshonTT:
@jameszianxuTT, as we [discussed](https://www.google.com/search?q=https://github.com/tenstorrent/tt-xla/issues/4592%23issuecomment-4577204486), are we good to land this? If there are still things left to review, I'd be happy to take care of them!

jameszianxuTT:
@sshonTT Let me do a final pass. This PR is quite complex and large, so thanks for your patience and explanations so far.

streaming/config.py
```diff
@@ -0,0 +1,57 @@
...
+    # whole_graph: streaming load → one `torch.compile(model)` → prefill + decode
+    # layer_eager: streaming load → per-layer compile + sequential execute
+    mode: Mode = "whole_graph"
+
+    # ---- weight dtype overrides ----
+    # Adapter interprets these values (e.g. "bfp_bf8", "bfp_bf4") into
+    # model-specific {path: dtype} via `adapter.weight_dtype_overrides(...)`.
+    # "bf16" / "" / "none" → no override for that weight class.
+    expert_dtype: str = "bf16"  # MoE expert weights
+    attn_dtype: str = "bf16"  # Attention weights
```
jameszianxuTT: This means default no-dtype-override right?

streaming/core.py
```diff
@@ -0,0 +1,599 @@
...
+def _block_relative_overrides(weight_overrides: Dict[str, str]) -> Dict[str, str]:
+    """Strip `layers.*.` prefix so overrides apply to a single Block."""
+    relative: Dict[str, str] = {}
+    layer_prefix = "layers.*."
+    for pattern, dtype in weight_overrides.items():
+        if pattern.startswith(layer_prefix):
+            relative[pattern[len(layer_prefix) :]] = dtype
+        elif pattern == "default":
+            relative["default"] = dtype
+    return relative
```
jameszianxuTT: Why do we only apply weight overrides to a single block?

streaming/core.py
```diff
@@ -0,0 +1,599 @@
...
+        prompts = _DEFAULT_PROMPTS
+    prompts_used = [prompts[i % len(prompts)] for i in range(config.batch_size)]
+
+    prompt_ids = adapter.tokenize_prompts(
+        prompts_used,
+        config.batch_size,
+        config.prompt_len,
+    )
+    prompt_ids_tt = prompt_ids.to(device)
+    xs.mark_sharding(prompt_ids_tt, mesh, ("_axis_0", None))
```
jameszianxuTT: Side effect of this function is to implicitly shard prompts on batch? What about batch 1?

streaming/core.py
```diff
@@ -0,0 +1,599 @@
...
+
+
+def _run_inference_whole_graph(
+    adapter: ModelAdapter,
+    model,
+    mesh,
+    device,
+    config: StreamingConfig,
+) -> StreamingResult:
+    """Whole-model `torch.compile` once, then prefill + decode."""
```
jameszianxuTT: >Whole-model `torch.compile` once

This means whole model in prefill configuration right?

streaming/core.py
```diff
@@ -0,0 +1,599 @@
...
+    generated: List[List[int]] = [[] for _ in range(bsz)]
+
+    sp_tt = torch.tensor(config.prompt_len, dtype=torch.long).to(device)
+    _log("pre-prefill")
+    t0 = time.time()
+    prefill_logits = adapter.call_model(compiled, prompt_ids_tt, sp_tt)
+    torch_xla.sync(wait=True)
+    logger.info(f"[prefill] compile+exec {time.time() - t0:.1f}s")
+    _log("post-prefill")
+    prefill_logits_cpu = prefill_logits.detach().to("cpu")
```
jameszianxuTT: Why detach this tensor from device before moving it to CPU? is it so argmax isn't traced into graph execution?

streaming/core.py
jameszianxuTT: To clarify - this is fine, but if we want to do this, please assert somewhere that this is a requirement from current streaming implementation. 

Here and in other places, I am scared to run into implied constraints inside streaming that caller does not know about, and user having to figure that out

streaming/config.py
sshonTT: right, this reduces device dram pressure as we not yet doesn't drop const-eval input in device.

streaming/core.py
sshonTT: It's intended to run layer by layer first for shipping weight data. Also there's some debugging feature and this is part of it.

streaming/core.py
sshonTT: Both prefill and decode yeah.

streaming/core.py
sshonTT: Ah good catch. This was intended to debug pcc, will fix it.

streaming/streaming_loader.py
```diff
@@ -0,0 +1,137 @@
...
+    module: nn.Module,
+    spec_by_id: Dict[int, Tuple],
+    mesh,
+    device,
+    *,
+    verbose: bool = False,
+    tag: str = "",
+) -> None:
+    """Replace every Parameter and Buffer in `module` with an
+    XLA-resident, handle-only copy uploaded via `_xla_tensors_from_aten`.
```
jameszianxuTT: I think this comment is now out of date? `_xla_tensors_from_aten` is removed per previous discussion I think

streaming/streaming_loader.py
```diff
@@ -0,0 +1,137 @@
...
+            xla_t = _upload_with_sharding(
+                b.detach(),
+                mesh,
+                partition_spec,
+                device,
+            )
+            sub._buffers[name] = xla_t
+            n_buffers += 1
+    if verbose:
+        logger.info(
```
jameszianxuTT: Small nit but we are not supposed to directly import loguru logger and use it, there is new logger in ttxla_tools

jameszianxuTT: Actually nvm, it seems that streaming reconfigures logger with `_configure_logging`... I think we can just leave this in as tech debt for now than trying to integrate it with existing logger, it's a really small nit

streaming/core.py
sshonTT: Fair point. Will add a guard for it.

streaming/streaming_loader.py
```diff
@@ -0,0 +1,137 @@
...
+
+import torch
+import torch_xla.distributed.spmd as xs
+from loguru import logger
+from torch import nn
+
+
+def _strip_cpu_golden_refs(block) -> None:
+    """Break the CPU-golden references kept by `enable_sparse_mlp` so the
+    original per-expert weights can be GC'd before we ship to device.
```
jameszianxuTT: I believe the CPU golden references were kept so that CPU golden evaluation could take place. Assuming we are on a host memory limited system, by running this, are we implicitly giving up on running CPU golden evaluation on the same host that's doing streaming, and if the user tries, let them hit whatever runtime error as a result of mutating the model?

I am fine with this; I think we should shift to a model of storing CPU reference tensors anyways; but I just want to understand if this has the side effect of making the normal CPU-run PCC check impossible in streaming mode

jameszianxuTT: As a side-idea, I wonder if the original experts meant to be used for CPU run can be replaced by some python object which throws with a more detailed error ("streaming mode deallocates CPU tensors so CPU golden evaluation cannot be run with these tensors") when accessed; or if that makes sense at all

sshonTT: Good question. The refs being stripped are enable_sparse_mlp's internal CPU-golden path (the sparse MoE falling back to _original_mlp/original_experts to verify its own rewrite). The streaming end-to-end CPU PCC check isn't built from those. It's done layer-by-layer. Each block's CPU forward runs while its weights are resident on host (right before ship), accumulating the reference hidden state, which is then compared against the device prefill/decode.

jameszianxuTT (approved):
IMO this PR threads the needle through torch-xla computation cache, torch-xla XLA data node cache, consteval cache and PJRTTensor cache in a pretty amazing way to cap host memory. 

I am a bit worried about generality and other hidden implementation-defined constraints since this is quite a large change. However I think we are merging this as an opt-in experimental application level side path as our best current option for host-memory-bound inference through forge. If nothing else, it is a good jumping off point for considering the future of large model inference through forge, as our current solution of allocating tons of swap is pretty untenable.

Thank you for explanations and discussion on this PR, as well as the side improvements to runtime brought up as prerequisite.

streaming/streaming_loader.py
sshonTT: Oh no, There was a pcc comparing logic that matches my explanation, but I decided to leave it out during the refactoring. 
I forgot this and I will add it back in the next commit.

streaming/core.py
```diff
@@ -0,0 +1,599 @@
...
+        )
+        try:
+            _ = run_block_flush(block, *dummy_args)
+            torch_xla.sync(wait=True)
+            xm.wait_device_ops()
+        finally:
+            hook.remove()
+
+        # 6. Re-init mutable KV buffers.
+        _reinit_mutable_kv_buffers(
```
jameszianxuTT: Sorry, another thing - why is this necessary? I thought underlying ttnn APIs were fine writing to a dirty cache; was it experimentally found that operating on a dirty cache caused issues?

jameszianxuTT: More precisely I think the fill/update cache ops are fine with a dirty cache. I guess we can have buffers that are mutated in other ways that may be sensitive to cache initialization

sshonTT: You're right that the plain attention kv_cache is fine with a dirty cache. fill/update_cache overwrite the slots they touch. The re-init isn't for that path.

The reason it's needed, the per-layer dummy flush runs the full block forward at real prefill shape on zero inputs, purely to migrate the plugin's host staging buffers onto device (the host-RAM bound streaming relies on). That forward writes garbage into every mutable state buffer, and DSV4's compressor has buffers that are mutated in non-fill-cache ways and are sensitive to init, exactly the case you flagged:

 - score_state is consumed by softmax(dim=1) and must start at -inf so unwritten slots get zero probability mass; a dirty finite value from the flush makes them equiprobable masked-fill candidates and silently corrupts routing.
 - kv_state / compressor kv_cache are accumulated/read assuming a clean (zero) start.
 
So the re-init is the cleanup for the dummy flush's side effect, scoped to exactly the buffers the adapter declares as mutable+init-sensitive (mutable_buffer_names + mutable_buffer_init_value, where score_state -> -inf). The plain kv_cache is included only because the flush dirties it too. if a model has only fill-cache-safe buffers, its adapter can return an empty mutable_buffer_names and skip this entirely.

jameszianxuTT: Thanks Sungjoon, it is a pretty annoying constraint that all mutable buffers cannot be resistant to dirty initialization, but I guess it comes from model implementation rather than anything streaming specific. And I doubt we can detect whether a buffer is sensitive to dirty init right now since it depends on the semantics of how it's used and how the underlying ops handle it... so this seems like the only good solution at the moment

Appreciate the explanation.

streaming/streaming_loader.py
sshonTT: This is simple fix so make it use ttlxa_tools

streaming/core.py
sshonTT: Fixed it to argmax goes to device.

AleksKnezevic (approved):

Issue #4785, opened 2026-05-19
Assess benchmark LLM output quality


vkovacevicTT:
Currently in benchmarks, PCC of prefill and first decode logits for User 0 is the main metric used to monitor and assert LLM output quality.

Some LLMs pass this check, but produce invalid output for some or all users (total garbage, repetitive tokens, empty tokens, ...)

Until we enable more extensive PCC checks (KV cache https://github.com/tenstorrent/tt-xla/issues/4783, full decode length, all users) and other metrics such as TopK, we should find the cause of current LLM output quality degradation.

[Claude report with outputs](https://github.com/user-attachments/files/28021313/llm-outputs-19-05.zip)

vkovacevicTT:
Run with potential fix for recent degradation: https://github.com/tenstorrent/tt-xla/actions/runs/26099205589/job/76745471261

Note: Some older issues are still present e.g. [llama_3_2_1b](https://github.com/tenstorrent/tt-xla/actions/runs/26099205589/job/76745471261) every 4th user's output is garbage.

mvasiljevicTT:
Partial fix tracked in [this issue](https://github.com/tenstorrent/tt-mlir/issues/8479)

investigating remaining wrong outputs

mvasiljevicTT:
the rest seems to be same as [this issue](https://github.com/tenstorrent/tt-mlir/issues/8068), it was there since we started printing all users so we can't bisect it.

mvasiljevicTT:
### Per-user prefill divergence in Llama: root cause + fix

#### What was wrong
Bisecting the 2-layer Llama-3.2-1B prefill with a per-user-consistency probe (same inputs across batch lanes → divergent outputs) pointed to `ttnn.experimental.rotary_embedding`. Repro at the device level:
- `ttnn::tilize_with_val_padding(input, ..., PadValue(0.0f))` is the canonical zero-pad-then-tile call inside the rotary wrapper.
- For already-`TILE` inputs `tilize_with_val_padding` short-circuits and silently drops the requested pad value (`if (layout == TILE) return input_tensor;`), so the implicit tile padding keeps whatever bytes were in L1 from prior ops.
- The rotary compute kernel processes the full padded `seq_len`, so it multiplies that L1 garbage into the output's padded rows — non-deterministic per-batch-lane per-run.

#### Fix
[...]

PR #4818, opened 2026-05-21
[Benchmark] Add kv_cache PCC check

### LINKED ISSUE #4783: Enable KV cache PCC check for benchmark LLMs
Currently in benchmarks PCC is calculated for prefill and first decode logits (for User 0).

Next step should be enabling KV cache PCC check.

vkovacevicTT:
### Ticket
Closes #4783 

### Problem description
Currently we measure PCC only for first user's prefill and first decode logits.
As a next step in extending PCC check, we want to add PCC check for KV cache.

### What's changed
- Added KV cache PCC check, for same number of steps as output logits PCC check.
- Modified `generate_and_benchmark` teacher-forcing logic so it can teacher-force a smaller window of ground truth tokens, and then continue autoregressivelly to full length of decode steps. This removes the need to split prefill and decode calls introduced in https://github.com/tenstorrent/tt-xla/commit/392e200b9ff69edcea01568330ed1ac09ca8b016

### Notes
We expect many models to fail because of this issue: #4785 

### Checklist
- [ ] [n150 LLMs](https://github.com/tenstorrent/tt-xla/actions/runs/26224737758)
- [ ] [LLMBOX LLMs](https://github.com/tenstorrent/tt-xla/actions/runs/26224756956)
- [ ] [Galaxy LLMs](https://github.com/tenstorrent/tt-xla/actions/runs/26224777626)
- [ ] [n150 LLMs on Marina's fix](https://github.com/tenstorrent/tt-xla/actions/runs/26505643781)
- [ ] [falcon3_10b_tp on Marina's fix](https://github.com/tenstorrent/tt-xla/actions/runs/26505570056)

### DIFF
--- a/tests/benchmark/benchmarks/llm_benchmark.py
+++ b/tests/benchmark/benchmarks/llm_benchmark.py
@@ -406,18 +406,26 @@ def benchmark_llm_torch_xla(
     cpu_output_logits: list = []
     cpu_output_tokens: Optional[torch.Tensor] = None
     if not accuracy_testing:
-        cpu_wrapper = LLMSamplingWrapper(model, read_logits_fn, return_logits=True)
+        cpu_wrapper = LLMSamplingWrapper(
+            model, read_logits_fn, return_logits=True, return_kv_cache=True
+        )
         cpu_wrapper.eval()
 
         # Iter 0: prefill. After this, input_args holds the post-prefill decode
         # state (input_ids=next_token_0, cache_position=[prompt_len]).
-        cpu_prefill_logits, cpu_prefill_tokens, _ = generate_and_benchmark(
+        (
+            cpu_prefill_logits,
+            cpu_prefill_tokens,
+            cpu_prefill_kv,
+            _,
+        ) = generate_and_benchmark(
             cpu_wrapper,
             input_args,
             torch.device("cpu"),
             1,
             verbose=False,
             collect_logits=True,
+            collect_kv_cache=True,
         )
 
         # Snapshot first-decode inputs before CPU iter 1 advances them.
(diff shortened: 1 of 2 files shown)


tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -417,18 +425,30 @@ def benchmark_llm_torch_xla(
...
-            cpu_wrapper,
-            input_args,
-            torch.device("cpu"),
-            1,
-            verbose=False,
-            collect_logits=True,
-        )
+        # Run NUM_PCC_DECODE_STEPS decode steps.
+        cpu_decode_logits: list = []
+        if NUM_PCC_DECODE_STEPS > 0:
```
odjuricicTT: This is always true? Is there a need for this if?

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -417,18 +425,30 @@ def benchmark_llm_torch_xla(
...
+                NUM_PCC_DECODE_STEPS,
+                verbose=False,
+                collect_logits=True,
+            )
 
         cpu_output_logits = cpu_prefill_logits + cpu_decode_logits
 
+        # CPU output tokens, used to teacher-force the device PCC run
+        # for the first len(cpu_output_logits) steps.
+        cpu_output_tokens = torch.tensor(
```
odjuricicTT, line 457: Can we take the tokens directly from the wrapper instead of doing argmax here again?

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -791,35 +781,67 @@ def benchmark_llm_torch_xla(
...
-            decode_pcc_value >= required_pcc
-        ), f"First decode PCC failed. PCC={decode_pcc_value:.6f}, Required={required_pcc}"
-        print(
-            "First decode PCC verification passed with PCC={:.6f}".format(
-                decode_pcc_value
+
+        # Pre-transfer device KV cache to CPU once.
+        cpu_pkv = decode_only_cache
+        device_layers_cpu = [
+            (layer.keys.to("cpu"), layer.values.to("cpu"))
```
odjuricicTT: Have you tried obtaining the kv cache by adding it to model output? It already is part of the output of the ttnn ir graph so it should not change the behaviour. Tho we need to test whether it impacts perf or memory usage.

It would be a more general approach instead of reading it from model layer data.

odjuricicTT (commented):
What is the number of models failing? You should open an issue to track all models that fail this check so we have an overview.

tests/benchmark/benchmarks/llm_benchmark.py
vkovacevicTT: Sure.

tests/benchmark/benchmarks/llm_benchmark.py
vkovacevicTT: Yes, we could extract what we need from output.

Issue #3763, opened 2026-03-17
Move models to tt-forge-models


AleksKnezevic:
Please move kimi k2 and deepseek models to tt-forge-models

gengelageTT:
Kimi-k2 modified model has been moved to tt-forge-models now.  [tests/torch/models/kimi_k2/original_modeling_deepseek.py](https://github.com/tenstorrent/tt-xla/blob/main/tests/torch/models/kimi_k2/original_modeling_deepseek.py) should be moved to tt-forge-models as well.

PR #4830, opened 2026-05-21
Update kimi-k2 mla cache test to use tt-forge-models

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3763

### Problem description
Moved the original model for kimi-k2 to tt-forge-models and can now remove the copy in tt-xla. Need to update import for 
the kimi-k2 mla cache test. This test currently isn't running in CI (seems accidental since marked with nightly).

### What's changed
Removed kimi_k2/original_modelling_deepseek.py and updated import for kimi-k2 tests. 

Marked `test_kimi_k2_mla_cache` as `single_device` so that it will run in CI. I'll request a cpu-only `"runs-on"` option so eventually this test won't be running on n150/p150s unnecessarily. 

Updated `torch.equal` to `torch.all_close` to avoid failure seen when results see negligible differences.

### Checklist


gengelageTT:
Testing in CI here: https://github.com/tenstorrent/tt-xla/actions/runs/26232048463

jameszianxuTT (approved):
Thanks for the port to tt-forge-models! You might want to mention this as part of the PR title :)

PR #4833, opened 2026-05-21
Reduce vLLM decode graphs from 2 to 1

### LINKED ISSUE #4824: Reduce vLLM decode graphs from 2 to 1
## Summary

Follow-up to #4789 (which reduced decode graphs from 5 to 2). This issue tracks merging the backbone and `decode_postprocess` into a single compiled graph per decode step.

## Background

Currently each decode step dispatches 2 graphs:
1. **Backbone** : all attention + FFN compute
2. **decode_postprocess** : select hidden states, compute logits, structured decode, sample

With trace mode enabled, each graph dispatch requires inputs to be on host before replay can start, meaning the hidden state must make a D2H + H2D round-trip between the two dispatches on every token. Since generative workloads run sequentially (can't start the next token until the current one is done), this overhead sits on the critical path.

## What needs to happen

- Move sampling metadata construction before the backbone
- Defer prompt logprobs handling
- Merge backbone + decode_postprocess into a single compiled function

## Notes

- Raised during review of #4789 by @AleksKnezevic
- Coordinate prioritization with @mmanzoorTT

alinakhanTT:
## Ticket
closes #4824

## Problem description
After #4789 reduced decode graphs from 5 to 2, each decode step still dispatches the backbone and `decode_postprocess` as two separate compiled programs. With trace mode enabled, each dispatch requires inputs to be on host before replay, meaning the hidden state must make a D2H + H2D round-trip between the two dispatches on every generated token. Since decode is sequential (can't start the next token until the current one is done), this overhead sits on the critical path.

## What's changed

**Task 1: Move sampling metadata and grammar prep before the backbone dispatch**
`sampling_metadata` and grammar tensors were previously constructed in Python after the backbone ran, creating unavoidable Python work between the backbone dispatch and `decode_postprocess`. Both only read from CPU state (`input_batch`, config) and do not depend on `hidden_states`, so they can be prepared before the backbone. This eliminates Python between the two dispatches, which is a prerequisite for fusion.

**Task 2: Precompile backbone + decode_postprocess as a single XLA program**
Added `_precompile_decode_combined` which traces backbone (num_tokens=1) and `decode_postprocess` together in one continuous lazy-execution window. Three ordering constraints are required for the fusion to work:
- `_precompile_backbone` skips `num_tokens=1` - if backbone(n=1) is cached standalone first, XLA splits the combined lazy graph at the cached boundary and dispatches 2 programs instead of 1
- `_precompile_decode_combined` runs before `_precompile_decode_postprocess` - same reason: if `decode_postprocess` is cached standalone first, XLA reuses it separately
- `wait_device_ops()` is called inside the greedy/non-greedy loop - without this, both variants merge into one double-sized lazy graph that never matches the single backbone+dp graph used at inference

During inference, backbone and decode_postprocess accumulate in the same lazy graph (no sync between them after Task 1), XLA matches the combined cached program and dispatches as one chip operation.

**Note: first token latency**
The first generated token still compiles 1 new graph at inference time. The dummy tensors used during precompile (`torch.zeros`) don't carry the same XLA SPMD provenance as real input tensors from `_prepare_inputs`, causing a cache miss on the first token. The combined backbone+dp program is recompiled once with real inputs and cached - from the second token onwards, decode dispatches as 1 graph with no recompilation.

**Task 3: Remove mid-graph sync for prompt logprobs**
The prompt logprobs hidden state extraction previously did a `.cpu()` on `hidden_states` between backbone and `decode_postprocess`, forcing a mid-graph sync that would break fusion. Following the suggestion to keep a single code path: a reference to the full backbone output is saved before any reshape, and the `.cpu()` extraction happens after `decode_postprocess` completes instead of before it.

## Checklist
- [ ] `VLLM_XLA_CHECK_RECOMPILATION=1` confirms no recompilation during inference
- [ ] Decode step confirmed dispatching as 1 graph (device-level verification)
- [x] Perf benchmark vs #4789 baseline

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -1482,6 +1482,47 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
             ) = self._prepare_model_call_tensors(
                 input_ids, self.position_ids, inputs_embeds
             )
+
+            # Build sampling_metadata and grammar inputs before the backbone so
+            # there is no Python work between the backbone dispatch and the
+            # decode_postprocess dispatch.  Both only read from CPU state
+            # (input_batch, config) and do not depend on hidden_states.
+            sampling_device = (
+                torch.device("cpu") if self.tt_config.cpu_sampling else self.device
+            )
+            logger.warning_once(
+                "Sampling on %s (cpu_sampling=%s)",
+                sampling_device,
+                self.tt_config.cpu_sampling,
+            )
+            sampling_metadata = XLASupportedSamplingMetadata.from_input_batch(
+                self.input_batch,
+                self.max_num_reqs,
+                sampling_device,
+                vocab_size=self.vocab_size,
+            )
+            if not self.tt_config.cpu_sampling:
+                # Prepare grammar tensors for the fused path.  The function only
+                # needs the device and num_reqs, not actual logit values, so we
+                # pass a shape proxy instead of waiting for hidden_states.
+                if grammar_output is not None:
+                    _shape_proxy = torch.empty(
+                        self.max_num_reqs, self.vocab_size, device=self.device
+                    )
+                    require_struct_decoding, grammar_bitmask_padded, bitmasks = (
+                        self.prepare_structured_decoding_input(
+                            _shape_proxy, grammar_output
+                        )
+                    )
+                else:
+                    self.require_structured_out_cpu.zero_()
+                    require_struct_decoding = self.require_structured_out_cpu[
+                        : self.max_num_reqs
+                    ].to(self.device)
+                    grammar_bitmask_padded = self.grammar_bitmask_cpu[
+                        : self.max_num_reqs
+                    ].to(self.device)
+                    bitmasks = self.structured_decode_bitmasks.to(self.device)
             torch_xla.sync(wait=False)
             # Run the decoder
             with (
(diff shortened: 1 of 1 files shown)


alinakhanTT:
Perf benchmark passed: https://github.com/tenstorrent/tt-xla/actions/runs/26406101582

alinakhanTT:
## Benchmark: decode graph reduction progression

[Run 26406101582](https://github.com/tenstorrent/tt-xla/actions/runs/26406101582) vs baseline (main, pre-#4789).

### TP models (LLMBox, 8-chip)

| Model | Baseline | After #4789 (5->2 graphs) | After #4833 (2->1 graph) | Δ vs baseline |
|---|---:|---:|---:|---:|
| qwen3-8b-tp | 10.41 sps | 12.93 sps | 13.56 sps | +30% |
| falcon3-10b-tp | 8.07 sps | 9.73 sps | 10.11 sps | +25% |
| ministral-8b-tp | 12.33 sps | 13.36 sps | 13.39 sps | +9% |
| falcon3-7b-tp | 11.87 sps | 13.19 sps | 13.13 sps | +11% |
| qwen3-14b-tp | 9.51 sps | 10.68 sps | 8.98 sps | -6% |

`qwen3-14b-tp` regresses vs #4789, will investigate in a follow-up run.

### Single-chip (n150): mostly stable vs #4789

kmabeeTT:
Can you incorporate feedback from https://github.com/tenstorrent/tt-xla/pull/4789 into this one first, then I'll give it more thorough review?

alinakhanTT:
feedback from https://github.com/tenstorrent/tt-xla/pull/4789 has been incorporated!!

PR #4857, opened 2026-05-22
 Fix JAX optimization_level not reaching PJRT plugin

### LINKED ISSUE #5072: Codegen doesn't seem to use the optimization_level flag
### Describe the bug

Exporting code from JAX doesn't seem to generate optimized TTNN code in the IR folder in the export path:
```
jitted_f = jax.jit(
    f,
    compiler_options={
        "optimization_level":"2",
        "backend": "codegen_py",
        "export_path": "exported"
    }
)
```

This seems to be due to [module_builder](https://github.com/tenstorrent/tt-xla/blob/a7e22a9faad5304db0c62ffb1812f0ea0e74b12f/pjrt_implementation/src/api/module_builder/module_builder.cc#L1476) dropping the optimization_level flag from compile_options

```
import jax.numpy as jnp
import jax

A = jnp.array([1,2,3,4,5,6,7,8,9])
def f(x):
        return A*x

jitted_f = jax.jit(
    f,
    compiler_options={
        "optimization_level":"2",
        "backend": "codegen_py",
        "export_path": "exported"
    }
)

jitted_f(jnp.zeros(A.shape))
```
Running this code with optimization 2 and 0 produces identical code, unlike what ttmlir-opt produces

### Expected behavior

optimization_level should be used in codegen's output

- Ubuntu 24.04
- Python 3.12
- Jax 0.7.1
- TT-XLA commit 9782561e7b53ff80347ca1e21e5ede2481fce234

aorlovicTT:
### Ticket
Fixes #5072 

### Problem description
- Fix a naming collision where JAX intercepts optimization_level from compiler_options before it reaches the TT PJRT plugin
- JAX treats optimization_level as a built-in XLA ExecutableBuildOptions field, ([code](https://github.com/jax-ml/jax/blob/main/jax/_src/compiler.py#L207)) it pops it from env_option_overrides (protobuf field 7) and places it on ExecutableBuildOptionsProto (protobuf field 3). Since tt-xla only read field 7, the value was silently lost and always defaulted to 0.

### What's changed
- The fix extracts optimization_level from field 3 as a fallback. Values from field 7 (e.g. from torch_xla.set_custom_compile_options()) still take precedence via overwrite.

### Checklist

### DIFF
--- a/pjrt_implementation/src/api/compile_options_parser.cc
+++ b/pjrt_implementation/src/api/compile_options_parser.cc
@@ -4,6 +4,9 @@
 
 #include "api/compile_options_parser.h"
 
+// PJRT implementation headers
+#include "api/compile_options.h"
+
 // third-party includes
 #include <google/protobuf/io/coded_stream.h>
 #include <google/protobuf/io/zero_copy_stream_impl_lite.h>
@@ -209,7 +212,7 @@ tt_pjrt_status CompileOptionsParser::extractCustomProtobufFields(
           if (exec_field.number() == kOptimizationLevelFieldNumber &&
               exec_field.type() ==
                   google::protobuf::UnknownField::TYPE_VARINT) {
-            out_compile_options.emplace("optimization_level",
+            out_compile_options.emplace(CompileOptions::optimization_level_key,
                                         std::to_string(exec_field.varint()));
           }
         }
--- a/pjrt_implementation/src/api/compile_options_parser.cc
+++ b/pjrt_implementation/src/api/compile_options_parser.cc
@@ -191,13 +191,32 @@ tt_pjrt_status CompileOptionsParser::extractCustomProtobufFields(
   // options map entry.
   constexpr int kMapValueFieldNumber = 2;
 
+  // ExecutableBuildOptionsProto is field #3 in CompileOptionsProto.
+  constexpr int kExecutableBuildOptionsFieldNumber = 3;
+
+  // Optimization level is field #24 inside ExecutableBuildOptionsProto.
+  constexpr int kOptimizationLevelFieldNumber = 24;
+
   for (int i = 0; i < unknown_fields.field_count(); ++i) {
     const google::protobuf::UnknownField &field = unknown_fields.field(i);
 
-    // Currently, we only support custom compiler options serialized in the
-    // `kCustomCompilerOptionsFieldNumber` field. In case we encounter
-    // options being serialized into some other field we will need to update
-    // this to support them.
+    if (field.number() == kExecutableBuildOptionsFieldNumber) {
+      google::protobuf::UnknownFieldSet exec_build_fields;
+      if (parseNestedProtobufField(field, exec_build_fields)) {
+        for (int j = 0; j < exec_build_fields.field_count(); ++j) {
+          const google::protobuf::UnknownField &exec_field =
+              exec_build_fields.field(j);
+          if (exec_field.number() == kOptimizationLevelFieldNumber &&
+              exec_field.type() ==
+                  google::protobuf::UnknownField::TYPE_VARINT) {
+            out_compile_options.emplace("optimization_level",
+                                        std::to_string(exec_field.varint()));
+          }
+        }
+      }
+      continue;
+    }
+
     if (field.number() != kCustomCompilerOptionsFieldNumber ||
         field.type() != google::protobuf::UnknownField::TYPE_LENGTH_DELIMITED) {
       continue;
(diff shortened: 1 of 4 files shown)


pjrt_implementation/src/api/unit_tests/compile_options_parser_tests.cc
```diff
@@ -0,0 +1,153 @@
...
+
+  google::protobuf::UnknownFieldSet fields;
+  ASSERT_TRUE(parseToFieldSet(data, fields));
+
+  std::unordered_map<std::string, std::string> compile_options;
+  tt_pjrt_status status = CompileOptionsParser::extractCustomProtobufFields(
+      fields, compile_options);
+  ASSERT_TRUE(tt_pjrt_status_is_ok(status));
+  ASSERT_NE(compile_options.find("optimization_level"), compile_options.end());
+  EXPECT_EQ(compile_options["optimization_level"], "1");
```
acicovicTT: We don't need two of the exactly same tests

pjrt_implementation/src/api/unit_tests/compile_options_parser_tests.cc
```diff
@@ -0,0 +1,153 @@
...
+
+  std::unordered_map<std::string, std::string> compile_options;
+  tt_pjrt_status status = CompileOptionsParser::extractCustomProtobufFields(
+      fields, compile_options);
+  ASSERT_TRUE(tt_pjrt_status_is_ok(status));
+  ASSERT_NE(compile_options.find("optimization_level"), compile_options.end());
+  EXPECT_EQ(compile_options["optimization_level"], "0");
+}
+
+TEST(CompileOptionsParserUnitTests,
```
mrakitaTT: I don't see much point to having three identical unit tests which differ just in the integer value that is being stored/read. I would only leave one test, which tests for example lvl 2 (just so it's not zero).

pjrt_implementation/src/api/unit_tests/compile_options_parser_tests.cc
```diff
@@ -0,0 +1,153 @@
...
+  google::protobuf::UnknownFieldSet fields;
+  ASSERT_TRUE(parseToFieldSet(data, fields));
+
+  std::unordered_map<std::string, std::string> compile_options;
+  tt_pjrt_status status = CompileOptionsParser::extractCustomProtobufFields(
+      fields, compile_options);
+  ASSERT_TRUE(tt_pjrt_status_is_ok(status));
+  ASSERT_NE(compile_options.find("optimization_level"), compile_options.end());
+  EXPECT_EQ(compile_options["optimization_level"], "2");
+}
```
acicovicTT, line 96: This seems to be the same as previous two, remove it pls

pjrt_implementation/src/api/unit_tests/compile_options_parser_tests.cc
mrakitaTT, line 1: Please also open us an issue to add some python JAX end to end test which will test if optimization level from jax is being properly read, in case OpenXLA changes the proto structure these unit tests won't catch that.

pjrt_implementation/src/api/unit_tests/compile_options_parser_tests.cc
```diff
@@ -0,0 +1,153 @@
...
+  tt_pjrt_status status = CompileOptionsParser::extractCustomProtobufFields(
+      fields, compile_options);
+  ASSERT_TRUE(tt_pjrt_status_is_ok(status));
+  EXPECT_EQ(compile_options.find("optimization_level"), compile_options.end());
+}
+
+TEST(CompileOptionsParserUnitTests,
+     extractCustomProtobufFields_unrelatedFieldsIgnored) {
+
+  // Field #5 should be ignored entirely.
```
acicovicTT: It's unclear why. pls add a comment explaining these two constants and why should they be ignored

pjrt_implementation/src/api/compile_options_parser.cc
```diff
@@ -191,13 +191,32 @@ tt_pjrt_status CompileOptionsParser::extractCustomProtobufFields(
...
+    if (field.number() == kExecutableBuildOptionsFieldNumber) {
+      google::protobuf::UnknownFieldSet exec_build_fields;
+      if (parseNestedProtobufField(field, exec_build_fields)) {
+        for (int j = 0; j < exec_build_fields.field_count(); ++j) {
+          const google::protobuf::UnknownField &exec_field =
+              exec_build_fields.field(j);
+          if (exec_field.number() == kOptimizationLevelFieldNumber &&
+              exec_field.type() ==
+                  google::protobuf::UnknownField::TYPE_VARINT) {
+            out_compile_options.emplace("optimization_level",
```
mrakitaTT: `"optimization_level"` string is being used in `pjrt_implementation/src/api/compile_options.cc` too, and in your unit test in bunch of places, so if we decide to change it one day we'll have to change it in multiple places. Lets define it as a constant inside `CompileOptions` struct in `pjrt_implementation/inc/api/compile_options.h` so it can be reused in all these places.

mrakitaTT (approved):

PR #4881, opened 2026-05-22
Pad attention heads

AleksKnezevic:
Added ability to pad attention heads when num_heads is not divisible by TP size

### Problem description
You can't shard a model if num-heads is not divisible by TP size

### What's changed
Added optional padding enabled by pad_attention_heads which gets activated automatically if num kv or q heads is not divisible by TP size. 

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -250,6 +250,8 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
             if self.use_2d_mesh and 1 in mesh_shape:
                 self.use_2d_mesh = False
 
+            self._maybe_pad_attention_heads(vllm_config, mesh_shape[0])
+
         self.enforce_eager = model_config.enforce_eager
 
         self.num_xla_graphs = 0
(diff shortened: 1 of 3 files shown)


AleksKnezevic:
@kmabeeTT this should allow you to run gemma4 31B on llmbox.

kmabeeTT:
This needs more work for gemma4-31b on n300-llmbox. I tried cherry picking it to branch `kmabee/aknezevic/gemma4-2d-mesh` (which allows model to run with 2D mesh) and runnning with 1D mesh and with the new flag to enable padding, but got same error. Claude found some issues..

```
● Diagnostic answer: no, it doesn't compile with 1D mesh + pad_attention_heads. Same Could not apply propagated tensor shardings to tensor dimensions error as the original Gemma-4 + 8-chip attempt. Two reasons:

  Why Aleks's padding feature doesn't fix Gemma-4 on 1D 8-chip

  1. It only inspects num_attention_heads / num_key_value_heads. For Gemma-4-31B those are 32 / 16 (the sliding-attention layers) — both already divide cleanly into the 8-chip "batch" axis. The padding decision computes pad_q=32, pad_kv=16 (i.e. no change), returns early before printing its Padding attention heads log line, and the model proceeds unpadded.
  2. It never looks at num_global_key_value_heads=4 (the full-attention layers, every 6th layer). That's the actual divisibility offender on 8-chip — 4 % 8 != 0 — and the MLIR sharding-propagation pass still fails on the global-attention reshape exactly as before my fixes.

  I also had to fix one issue in the cherry-pick along the way: _maybe_pad_attention_heads reads hf_config.num_attention_heads directly, but Gemma-4 is multimodal and nests those attrs under hf_config.text_config. Added a text_config fallback at model_runner.py:606-615 (uncommitted), but as expected that just gets past the AttributeError and into the original MLIR failure.
  
    Options to move forward

  1. Extend _maybe_pad_attention_heads to also pad num_global_key_value_heads when present. Conceptually clean — same pad-up-to-multiple logic, just applied to the second KV head count too. Requires matching surgery in _surgery_pad_attention_modules to know which layers are global vs sliding.
  <snip>
  ```
  
  Edit: Pushed the cherry pick and a WIP commit that reflects above to branch for safe keeping.

AleksKnezevic:
Thanks @kmabeeTT. I can extend it once gemma4 is merged.

Issue #4596, opened 2026-05-08
Mainline all changes used for gemma4 tt-deploy demo


kmabeeTT:
For tt-deploy we had some last minute changes for supporting gemma4 31B demo and were working off branches of tt-xla and tt-mlir.  This ticket is track what is remaining to get merged to main (some already done).

tt-xla branch was `origin/kmabee/aknezevic/gemma4` it contained:

```
2026-05-01 (107e769d6) by Kyle Mabee (Author kmabee@tenstorrent.com) : Uplift tt-mlir for Sungjoon's SDPA fix for max_model_len>1024 with Gemma4-31B-it
2026-04-30 (da1cecccf) by Kyle Mabee (Author kmabee@tenstorrent.com) : Remove redundant enable_trace
2026-04-30 (7dc0468ce) by Kyle Mabee (Author aknezevic@tenstorrent.com) : Hack for speedier batch=1 sampling (7.6 tok/sec non-greedy is now 10.6 tok/sec for Gemma4-31B-it at max_model_len=512)
2026-04-30 (56ce12b1b) by Kyle Mabee (Author kmabee@tenstorrent.com) : Add a debug print to show TTConfig options for extra proof from tt-inference-server runs
2026-04-30 (ef6f1e6be) by Kyle Mabee (Author kmabee@tenstorrent.com) : Uplift to tenstorrent/tt-mlir:rpavlovic-squashed-onto-kmabee-gemma4 one commit
2026-04-30 (51cd2da4a) by Kyle Mabee (Author kmabee@tenstorrent.com) : Non-Greedy device sampler improvements (3x perf improvement for cpu_sampling=False temp!=0)
2026-04-30 (83a78123d) by Kyle Mabee (Author hshah@tenstorrent.com) : add composite gather op
2026-04-30 (5135654e6) by Kyle Mabee (Author kmabee@tenstorrent.com) : Uplift tt-mlir for non-greedy sampler perf improvements (was d188ac454 from aknezevic/gemma4)
2026-04-30 (5c0944110) by Aleks Knezevic (Author aknezevic@tenstorrent.com) : pre-fetch per_layet_attn_metadata
2026-04-29 (339622143) by Aleks Knezevic (Author aknezevic@tenstorrent.com) : VALIDATE, re-use device buffers across iterations
2026-04-29 (6d80d882d) by Aleks Knezevic (Author aknezevic@tenstorrent.com) : Expand SDPA override, fix sharding
[...]

PR #4591, opened 2026-05-08
[vLLM plugin] Use [1, num_devices] shape for 1D mesh

mmanzoorTT:
### Ticket
closes #4590 

### Problem description
vLLM is using [num_devices, 1] as mesh shape for 1D mesh.

### What's changed
Use [1, num_devices] as mesh shape for 1D mesh

### Checklist
- [X] New/Existing tests provide coverage for changes

PR #4889, opened 2026-05-24
Add Gemma4 e4b, 31b support for vLLM

sshonTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/4596)

### Problem description
There is a working branch that supports Gemma4 (origin/kmabee/aknezevic/gemma4), need to be mainlined

### What's changed
- Enables Gemma-4 text-only inference (E4B single-device, 31B BHQB TP=4): cross-layer KV sharing, sliding-window SDPA, mesh-aware KV cache axis selection.
- Detects nested `ParallelLMHead` for Gemma-4 multimodal config so the sharded `compute_logits` path fires.
- Adds `flat_model_io` TTConfig option for HF forwards needing a flat token stream (Gemma-4 PLE).

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/attention.py
+++ b/integrations/vllm_plugin/vllm_tt/attention.py
@@ -284,7 +284,13 @@ class TTAttentionBackendImpl(AttentionImpl):
 
         # kv_cache is [k_cache, v_cache] after init, but a scalar placeholder
         # during profiling — isinstance distinguishes the two cases.
-        if isinstance(kv_cache, (list, tuple)) and kv_cache[0].numel() > 0:
+        # Skip cache update when sharing KV cache with another layer — the
+        # target layer already wrote to the shared cache.
+        if (
+            isinstance(kv_cache, (list, tuple))
+            and kv_cache[0].numel() > 0
+            and self.kv_sharing_target_layer_name is None
+        ):
             self._handle_paged_attention(inputs, kv_cache, attn_metadata)
 
         # Compute attention based on mode:
@@ -293,7 +299,9 @@ class TTAttentionBackendImpl(AttentionImpl):
         # - is_prefill=False: Paged decode attention (generative models only)
         if inputs.is_prefill:
             assert self.sinks is None, "Attention sink is unsupported in SDPA prefill"
-            output = self._compute_full_attention(inputs, attn_metadata)
+            # Pass kv_cache so shared-KV layers can gather K/V from the target
+            # layer's paged cache (see ``_compute_full_attention`` docstring).
+            output = self._compute_full_attention(inputs, kv_cache, attn_metadata)
         else:
             output = self._compute_decode_attention(inputs, kv_cache, attn_metadata)
 
@@ -502,7 +510,10 @@ class TTAttentionBackendImpl(AttentionImpl):
         kv_cache[1].copy_(v_cache)
 
     def _compute_full_attention(
-        self, inputs, attn_metadata: TTMetadata
+        self,
+        inputs,
+        kv_cache,
+        attn_metadata: TTMetadata,
     ) -> torch.Tensor:
         """Compute full attention using scaled dot-product attention (non-paged).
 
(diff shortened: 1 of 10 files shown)


third_party/CMakeLists.txt
sshonTT: Will remove when uplift is in.

integrations/vllm_plugin/vllm_tt/platform.py
```diff
@@ -68,6 +69,10 @@ class TTConfig:
     # Target dtype for weight conversion (e.g. "bfp_bf8", "bfp_bf4"). Empty disables.
     experimental_weight_dtype: str = ""
 
+    # Toggle the tt-mlir permute+matmul fusion optimization. Mirrors the PJRT
+    # compile option; defaults to True to match the PJRT default.
+    experimental_enable_permute_matmul_fusion: bool = True
```
AleksKnezevic, line 73: Why do we need this?

integrations/vllm_plugin/vllm_tt/pooling_runner.py
```diff
@@ -1221,8 +1313,8 @@ def execute_model(
 
             # Mark inputs for data parallel sharding.
             if self.enable_data_parallel:
-                xs.mark_sharding(input_ids, self.mesh, ("batch", None))
-                xs.mark_sharding(self.position_ids, self.mesh, ("batch", None))
+                xs.mark_sharding(input_ids, self.mesh, ("model", None))
```
sshonTT, line 1224: Some of these changes are introduced https://github.com/tenstorrent/tt-xla/pull/4591, once it gets merged, will remove this.

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
```diff
@@ -28,6 +28,43 @@
...
+        axis_sizes = mesh.shape()
+
+    safe_spec = []
+    for i, axis in enumerate(partition_spec):
+        if axis is None or i >= tensor.ndim:
+            safe_spec.append(None)
+            continue
+        mesh_axis_size = axis_sizes.get(axis)
+        if mesh_axis_size is None or tensor.shape[i] % mesh_axis_size != 0:
+            logger.debug(
```
AleksKnezevic: should this be a warning so it's always displayed?

integrations/vllm_plugin/vllm_tt/vllm_utils.py
```diff
@@ -36,7 +36,7 @@ def determine_mesh_shape(num_devices: int, use_2d_mesh: bool) -> tuple[int, int]
             return mesh_shape
     else:
         # For 1D mesh, all devices are in one dimension.
-        mesh_shape = (num_devices, 1)
+        mesh_shape = (1, num_devices)
```
AleksKnezevic, line 39: @mmanzoorTT FYI

python_package/tt_torch/sharding.py
```diff
@@ -19,6 +19,42 @@
 _MESH_IDX_PREFIX = "mesh_idx_"
 
 
+def _normalize_partition_spec_for_rank(partition_spec, tensor_rank: int) -> tuple:
```
AleksKnezevic, line 22: this seems dangerous. Would it be possible to shard on the wrong dim using this leading to superfluous CCLs?

third_party/CMakeLists.txt
AleksKnezevic: is this intended?

integrations/vllm_plugin/vllm_tt/platform.py
sshonTT: This is used for demo which is not in this commit.

python_package/tt_torch/sharding.py
sshonTT: Fair point. Narrowed the helper to only resize fully-replicated  (all-None) specs. Any spec with a non-None entry now raises error on a rank mismatch.

third_party/CMakeLists.txt
sshonTT: Yes but required changes are uplifted so drop it.

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
sshonTT: right. promoted to logger.warning

nsumrakTT (approved):

vvukomanTT (approved):

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2112,8 +2279,10 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
 
         kv_caches: dict[str, torch.Tensor] = {}
         for kv_cache_group in kv_cache_config.kv_cache_groups:
-            kv_cache_spec = kv_cache_group.kv_cache_spec
             for layer_name in kv_cache_group.layer_names:
+                kv_cache_spec = _get_layer_kv_cache_spec(
```
kmabeeTT, line 2283: The reason `_get_layer_kv_cache_spec` exists is that `kv_cache_group.kv_cache_spec` can be a `UniformTypeKVCacheSpecs` wrapper. But a few dozen lines above (still inside `initialize_kv_cache`) we still do `kv_cache_config.kv_cache_groups[0].kv_cache_spec.block_size` directly — and again in the
  `block_sizes=[…]` / `kernel_block_sizes=[…]` constructor args for `InputBatch`.

  If the new helper is needed here, those direct `.block_size` reads are latent bugs for the same case. If `UniformTypeKVCacheSpecs` passes `block_size` through, the helper is unnecessary. Could you reconcile?

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
```diff
@@ -135,19 +164,19 @@ def _shard_weight(self, mesh: "xs.Mesh"):
         self.q_weight = Parameter(self.q_weight.to("xla"), requires_grad=False)
         self.k_weight = Parameter(self.k_weight.to("xla"), requires_grad=False)
         self.v_weight = Parameter(self.v_weight.to("xla"), requires_grad=False)
-        xs.mark_sharding(self.q_weight, mesh, ("batch", "model"))
-        xs.mark_sharding(self.k_weight, mesh, ("batch", "model"))
-        xs.mark_sharding(self.v_weight, mesh, ("batch", "model"))
+        safe_mark_sharding(self.q_weight, mesh, ("model", "batch"))
```
kmabeeTT, line 168: Can you add something in PR desc so it's preserved about why this was needed and for what models?

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
```diff
@@ -28,6 +28,35 @@
...
+        axis_sizes = mesh.shape()
+
+    safe_spec = []
+    for i, axis in enumerate(partition_spec):
+        if axis is None or i >= tensor.ndim:
+            safe_spec.append(None)
+            continue
+        mesh_axis_size = axis_sizes.get(axis)
+        if mesh_axis_size is None or tensor.shape[i] % mesh_axis_size != 0:
+            logger.warning(
```
kmabeeTT: Hmm is this warning firing a bad sign with perf impact? Does it make sense to have a strict version of this function where it's treated as an error instead of warning for cases where we don't want to accidentally/unknowingly (warnings sometimes not easily found) fall back to replication?

integrations/vllm_plugin/vllm_tt/attention.py
```diff
@@ -550,6 +623,7 @@ def _compute_decode_attention(
             is_causal=attn_metadata.is_causal,
             attn_mask=attn_metadata.attn_mask,
             attention_sink=self.sinks,
+            scale=self.scale,
```
kmabeeTT: Sliding-window threading question: `_compute_full_attention` now passes `sliding_window_size=self.sliding_window` to the prefill SDPA op when `self.sliding_window is not None`, but the decode call here doesn't get an equivalent kwarg. For Gemma-4's sliding-window layers, decode would then look at the full KV window instead of the sliding window.

  Three possibilities — which one applies?
  1. The decode kernel handles sliding window internally (via `attn_mask` or KV cache truncation).
  2. Gemma-4's sliding-window layers don't actually hit decode.
  3. Real bug — needs the same kwarg threaded through.

  The new tests use `max_tokens=32` / `max_model_len=128`, which likely never exceeds the sliding window, so they wouldn't catch (3) even if it's the case.

kmabeeTT (approved):
Thanks! Overall seems good, plz include a link to passing vLLM nightly before merge.   I left a few comments, 2 things claude thinks might be bugs. Please add a bit more detail to PR Description about the changes what/why for things not documented since only that PR Description will persist on merge.

integrations/vllm_plugin/vllm_tt/model_runner.py
sshonTT: The two patterns are actually intentional. `UniformTypeKVCacheSpecs.from_specs` already ensures a single block_size is propagated to the wrapper, so accessing it directly is fine. The helper is really only there for per-layer attributes like num_kv_heads, head_size, or dtype which live on the individual entries, not the wrapper itself. (Same goes for page_size_bytes, which is a sum).

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
sshonTT: QKV/MergedColumn weight sharding swapped from ("batch", "model") to ("model", "batch") so num_head is sharded along the "model" axis, matches the semantics of tensor parallelism (the "batch" axis is for data parallelism). Applies to all 2D-mesh models.
As reflected in the code, this required xfailing the Mistral-Large test. likely a pre-existing latent issue that this change surfaced.

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
sshonTT: Added a strict=False parameter. pass strict=True to raise instead of warn. Current warnings are batch=1 fallback, not a perf regression.

integrations/vllm_plugin/vllm_tt/attention.py
sshonTT: confirmed bug - paged_scaled_dot_product_attention_decode accepts sliding_window_size, but the call site was missing it, so Gemma-4 sliding-window layers were looking at the full KV during decode. Fixed in attention.py to mirror the prefill path. Caveat: existing tests use max_model_len=128 ≪ sliding window, so they only verify the kwarg is plumbed, not the clipping behavior. So it needs a long-context test for full validation.

sshonTT:
@kmabeeTT nightly passed : https://github.com/tenstorrent/tt-xla/actions/runs/26533201420

integrations/vllm_plugin/vllm_tt/attention.py
```diff
@@ -512,23 +523,85 @@ def _compute_full_attention(
...
+                kv_cache[1], attn_metadata.page_table, inputs.kv_num_tokens
+            )
+            # SDPA expects [users, num_heads, tokens, head]. The gather helper
+            # already returns that layout, only Q needs the transpose.
+            query_for_sdpa = inputs.query.transpose(-3, -2)
+        else:
+            # scaled_dot_product_attention expects [B, N_tokens, N_heads, H]
+            query_for_sdpa = inputs.query.transpose(-3, -2)
+            key_for_sdpa = inputs.key.transpose(-3, -2)
+            value_for_sdpa = inputs.value.transpose(-3, -2)
```
jameszianxuTT, line 550: nit: I see this was here already, but are we just doing blind transposes with an assumption of the layout of the input? Or does the caller guarantee it somehow? 

Just wondering if the meaning of the ranks for the inputs is encoded somewhere that could be checked; in case caller behaviour changes

jameszianxuTT (approved):
small nit, approving

integrations/vllm_plugin/vllm_tt/attention.py
sshonTT: The layout is guaranteed in the same flow, not assumed. `_prepare_inputs` [asserts](https://github.com/tenstorrent/tt-xla/blob/562bce964b53048cb39b98523d92eb8dea98f6f6/integrations/vllm_plugin/vllm_tt/attention.py#L338) the input ndim is 2 or 3 and that the batch dim matches [num_users](https://github.com/tenstorrent/tt-xla/blob/562bce964b53048cb39b98523d92eb8dea98f6f6/integrations/vllm_plugin/vllm_tt/attention.py#L325), then [_reshape_to_attention_format](https://github.com/tenstorrent/tt-xla/blob/562bce964b53048cb39b98523d92eb8dea98f6f6/integrations/vllm_plugin/vllm_tt/attention.py#L443) explicitly reshapes Q/K/V to [users, tokens, heads, head_size] . 
The caller contract is the standard vLLM [num_tokens, num_heads*head_size] (or batched 3D) shape. The transpose just maps that normalized 4D layout to SDPA's [users, heads, tokens, head].

PR #4939, opened 2026-05-25
Remove model-specific install requirements from perf benchmarks

odjuricicTT:
### Problem description
The perf benchmark workflow installs Python dependencies per-test via the `pyreq` field in `perf-bench-matrix.json`, and `llm_benchmark.py` enforces a hard-coded transformers version. With transformers now uplifted to 5.5.1 (and brought in via the wheel/docker image), the per-test installs and the version check are redundant and cause maintenance overhead.

### What's changed
- Drop `check_transformers_version()` (and its caller + the now-unused `import transformers`) from `tests/benchmark/benchmarks/llm_benchmark.py`.
- Remove the `pyreq` field from every test entry in `.github/workflows/perf-bench-matrix.json` (119 entries).
- Remove the entire `Install Python dependencies` step from `.github/workflows/call-perf-test.yml`; the wheel + docker image already provide what is needed.

### CI run
Single-chip shared-runner perf benchmark on this branch: https://github.com/tenstorrent/tt-xla/actions/runs/26410094134

### Test failures observed
Six jobs failed. Two are caused by this change, the rest are pre-existing / transient:

| Test | Cause | Caused by this PR? | Proposed solution |
| --- | --- | --- | --- |
| `resnet_jax` | `ImportError: cannot import name 'FlaxResNetForImageClassification' from 'transformers'` — Flax model classes were removed from transformers in versions >= 5.x. The original `pyreq` pinned `transformers==4.57.1` + `flax==0.10.4`. | Yes | Either rewrite the test to use a current Flax/JAX ResNet implementation (e.g., from `flax.linen` or a JAX-native model zoo), or move this test into a separate workflow that installs transformers 4.57.1 + flax. |
| `bge_m3_encode` | `ImportError: cannot import name 'is_torch_fx_available' from 'transformers.utils.import_utils'` — `FlagEmbedding` depends on this symbol which was moved/removed in newer transformers. Original `pyreq` pinned `transformers==4.57.1` + `FlagEmbedding`. | Yes | Wait for a `FlagEmbedding` release compatible with transformers 5.x, or run this test on a separate worker with the older pin. |
| `ufld_v2` | `ValueError: IRD_LF_CACHE environment variable is not set.` — the model loader downloads weights from the IRD LF cache; the env var is exported in `call-test.yml` (when `shared-runners` is true) but **not** in `call-perf-test.yml`. | No (pre-existing) | Add `IRD_LF_CACHE=${{ vars.IRD_LF_CACHE }}` to `call-perf-test.yml` for the shared-runner path, mirroring what `call-test.yml` already does. |
| `tiiuae_falcon3-7b-base` | `RuntimeError: Bad StatusOr access: INTERNAL: Error code: 13` — device-side runtime error. Same failure appears on `main` in the latest nightly. | No (pre-existing) | Track in an existing falcon3-7b issue; unrelated to dependency management. |
| `mistral_7b` | `error downloading xla-whl-release-2ef6fc3: error writing zip archive: unexpected EOF` during artifact download. | No (transient) | Retry the workflow; this is an artifact-server flake. |
| `vllm_llama_3_2_1b_instruct` | Same wheel artifact download EOF as above. | No (transient) | Retry the workflow. |

### Checklist
- [x] New/Existing tests provide coverage for changes (covered by the perf benchmark CI itself)

### DIFF
--- a/.github/workflows/call-perf-test.yml
+++ b/.github/workflows/call-perf-test.yml
@@ -169,10 +169,9 @@ jobs:
       shell: bash
       run: |
         source venv/activate
-        pip install ${{ matrix.build.pyreq }}
-        # Install correct torch-xla version from requirements.txt
-        pip install $(grep "torch-xla@" python_package/requirements.txt)
-        pip install torchvision==0.25.0+cpu --index-url https://download.pytorch.org/whl/cpu
+        if [ -n "${{ matrix.build.pyreq }}" ]; then
+          pip install ${{ matrix.build.pyreq }}
+        fi
 
     - name: Create perf report directory
       run: mkdir -p ${{ steps.strings.outputs.perf_report_path }}
--- a/.github/workflows/perf-bench-matrix.json
+++ b/.github/workflows/perf-bench-matrix.json
@@ -19,67 +19,56 @@
     "tests": [
       {
         "name": "resnet_jax",
-        "pyreq": "pytest tqdm loguru requests transformers==4.57.1 datasets flax==0.10.4 torch==2.10.0",
+        "pyreq": "transformers==4.57.1 flax==0.10.4",
         "pytest": "tests/benchmark/resnet_jax_benchmark.py::test_resnet_jax"
       },
       {
         "name": "mnist",
-        "pyreq": "datasets loguru pytest requests torch==2.10.0 tqdm transformers==5.5.1",
         "pytest": "tests/benchmark/test_vision.py::test_mnist"
       },
       {
         "name": "resnet",
-        "pyreq": "datasets loguru pytest requests tabulate timm tqdm transformers==5.5.1 torch==2.10.0",
         "pytest": "tests/benchmark/test_vision.py::test_resnet50"
       },
       {
         "name": "mobilenetv2",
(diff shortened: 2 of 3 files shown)


nsumrakTT (approved):

vkovacevicTT (commented):
Nice ! 

Regarding `resnet_jax` and `bge_m3_encode` error, we could:
1. Keep `pyreq` for those models with just the missing libs (Flax,  FlagEmbedding).
2. Accumulate all additional libs and install them for every model, so we don't have per-model `pyreq` at all.

odjuricicTT:
Rebased onto main and re-ran with the two previously-failing models plus sanity-check models. All 8 jobs passed: https://github.com/tenstorrent/tt-xla/actions/runs/26569401710

- ✓ resnet_jax
- ✓ bge_m3_encode
- ✓ vllm_bge_m3_encode_batch1
- ✓ vllm_bge_m3_encode_batch32
- ✓ mnist
- ✓ vit
- ✓ llama_3_2_1b_instruct
- ✓ vllm_llama_3_2_1b_instruct

vkovacevicTT (approved):

PR #3523, opened 2026-03-02
Bring up Mixtral models in the vLLM plugin.

devisettymahidhar608:
### Ticket
[2620](https://github.com/tenstorrent/tt-xla/issues/2620#issuecomment-3949392101)

### Problem description
BringUp Mistral models in the vLLM plugin.

### What's changed
Added a **torch.nn.functional.unfold** override in python_package/tt_torch/torch_overrides.py (XLA-device only): it replaces the native op — which lowers to im2col / Tensor.unfold HLO ops unsupported on the TT backend — with an equivalent gather-based im2col. Functional equivalence with native F.unfold is verified in tests/torch/ops/test_unfold.py.

Added a new test case in **tests/integrations/vllm_plugin/generative/test_tensor_parallel_generation.py** to cover two Mistral variants: **mistralai/Mistral-Small-3.2-24B-Instruct-2506** and **mistralai/Mistral-Small-3.1-24B-Instruct-2503.**

Both models are passing successfully. The outputs from both models are as follows:

**mistralai/Mistral-Small-3.2-24B-Instruct-2506:**

[...]

PR #4947, opened 2026-05-26
[vLLM] Add data-parallel, tensor-parallel and DP+TP support across generative and pooling models

### LINKED ISSUE #4788: Add data parallel generative models support through vLLM
It was said by @mmanzoorTT we are missing support for this, and so we should add support for the following along with some models that use it:

- data parallel generative models through vllm (data parallel one dim, tensor parallel other dim)

mmanzoorTT:
### Status
Open for review. Merge blocked on next tt-mlir uplift in tt-xla main, see status comment for required tt-mlir changes. Test CI green: https://github.com/tenstorrent/tt-xla/actions/runs/27226100784

### Ticket
closes #4788

### Problem description
The vLLM plugin needs to support three parallelism modes — data-parallel (DP), tensor-parallel (TP), and combined DP+TP, across both generative and pooling models. Prior to this PR:

1. **DP decode hung with a recompile loop.** `select_hidden_states` was precompiled with an unsharded dummy `hidden_states`, but under `DATA_PARALLEL_ONLY` the backbone emits `hidden_states` sharded as `("batch", None, None)`. Every decode step missed the TT-MLIR graph cache and triggered a full recompile — producing the ~278-minute hang seen on llmbox runs.
2. **DP and DP+TP missed several batch-dim shardings.** `page_table`, `cache_position`, `inputs_embeds`, and the KV cache itself were not annotated as batch-sharded under `DATA_PARALLEL_ONLY` / `DATA_TENSOR_PARALLEL`, breaking SPMD propagation.
3. **Pooling DP regressions.** Batch padding and `dp_size` slicing were wrong, so pooling models couldn't run DP cleanly.
4. **No push-tier coverage** for DP-generative, DP+TP-generative, or DP+TP-pooling — only TP and pooling-DP existed.

### What's changed

**Decode recompile fix (`model_runner.py`):**
- `_precompile_select_hidden_states`: mark `dummy_hidden` as `("batch", None, None)` for DP mode so the warm-up compile matches what the backbone actually emits.
- `select_hidden_states`: all-gather the result back to `(None, None)` before returning — same pattern TP already uses, so `compute_logits` and `sample_from_logits` see unsharded input and hit their pre-warmed graphs.

**SPMD sharding annotations for DP and DP+TP (`model_runner.py`):**
- `_prepare_inputs` and `_dummy_run`: mark `page_table` and `cache_position` as batch-sharded for `DATA_PARALLEL_ONLY` and `DATA_TENSOR_PARALLEL`.
- `inputs_embeds` batch-sharded for multimodal DP paths.
- KV cache sharding corrected for `DATA_TENSOR_PARALLEL`.
- `max_num_reqs` rounded to a multiple of `dp_size` so per-device batches stay balanced.

**Pooling runner (`pooling_runner.py`):**
- DP-aware batch padding and `dp_size` slicing so pooling models work in DP and DP+TP modes.

**Distributed utils (`vllm_distributed_utils.py`, `vllm_utils.py`):**
- Plumbing changes needed by the DP / DP+TP paths.

**Tests added: push-tier coverage for the previously-uncovered combos:**

| Combo | Model type | New file |
|---|---|---|
| DP | generative | `test_data_parallel_generation.py` |
| DP+TP | generative | `test_data_tensor_parallel_generation.py` |
| DP+TP | pooling | `test_data_tensor_parallel_inference.py` |

TP-generative, TP-pooling, and DP-pooling are exercised through their existing test files, which now pass with the runner fixes above. Net effect: **DP, TP, and DP+TP all supported for both generative and pooling models.**

Also bumps `third_party/tt_forge_models`.

Co-authored by @ssaliceTT: added the FSDP / 2D-mesh support this builds on: batch-axis sharding of lm_head and embedding hidden dims, use_2d_mesh streamlining, and the 2D-mesh / FSDP / DP+TP benchmark configs

### Checklist
- [x] DP decode no longer triggers per-step TT-MLIR recompile
- [x] `page_table`, `cache_position`, `inputs_embeds`, KV cache annotated with batch sharding for DP / DP+TP
- [x] Pooling runner DP regressions (batch padding, `dp_size` slicing) fixed
- [x] Push tests added for DP-generative, DP+TP-generative, DP+TP-pooling
- [x] Existing TP-generative, TP-pooling, and DP-pooling tests pass with the runner changes


alinakhanTT:
### Test CI
All tests pass: https://github.com/tenstorrent/tt-xla/actions/runs/27226100784

### Blocked on tt-mlir uplift
### Required tt-mlir changes
Currently living on branch \`akhan/paged-ops-users-factor\` @ \`3db510b80\`:
1. **Users-dim Shardy factor** on the custom sharding rules for \`tt.paged_update_cache\` and \`tt.paged_scaled_dot_product_attention_decode\`. Without it, the SPMD partitioner can't keep per-replica cache updates local under DP, global reduction corrupts decode output.
2. **\`mesh_coords\` plumbing** on paged update/fill cache ops to force the \`MeshWorkloadFactory\` path under DP partitioning.
3. **Compatibility with tt-mlir #8380** (\`MeshShardType::Identity\` removal): the PJRT plugin needs to construct a \`Replicate\` sharding inline for un-annotated args. PJRT-side already implemented in this PR (commit \`4cb3b40\`).
4. **Empty Shardy mesh handling**: single-chip JAX programs emit \`sdy.mesh\` with no axes; post-#8380 tt-mlir errors out. PJRT-side Replicate fallback already implemented in this PR (commit \`b612268\`).

Once tt-mlir uplift includes the above, the rebase + merge here should be clean.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -251,19 +251,69 @@ def __init__(
...
-            mesh_shape = determine_mesh_shape(num_devices, use_2d_mesh=self.use_2d_mesh)
-            device_ids = np.array(range(num_devices))
+            if self.num_devices == 1:
+                logger.warning(
+                    "Data parallel execution is possible with multiple devices but found single device. Disabling multi device execution and proceeding with single device execution."
+                )
+                self.enable_data_parallel = False
+
+        if self.enable_data_parallel and self.enable_tensor_parallel:
+            self.parallel_mode = ParallelismMode.DATA_TENSOR_PRALLEL
```
sshonTT: nit, `DATA_TENSOR_PARALLEL`

tests/integrations/vllm_plugin/generative/test_opt_generation.py
```diff
@@ -79,6 +79,8 @@ def test_opt_generation_multibatch():
         "additional_config": {
             "enable_const_eval": False,
             "min_context_len": 32,
+            "enable_data_parallel": True,
+            "num_hidden_layers": 1,
```
mmanzoorTT: Remove these lines.

nsumrakTT (approved):

pjrt_implementation/src/api/executable_image.cc
```diff
@@ -241,6 +241,22 @@ FlatbufferExecutableImage::FlatbufferExecutableImage(
       expected_shape.push_back(2);
     }
 
+    // getOutputDimensions() collects from the StableHLO-level function result
+    // type, which is the global (pre-retype) shape. The flatbuffer carries the
+    // per-device local shape since tt-mlir #8380 retyped sharded results to
+    // local in SHLO -> TTIR conversion. Convert global to local using the
+    // sharding info before comparing.
+    const auto &sharding = output_sharding[output_index];
```
sshonTT: Could you add a guard for indexing `output_sharding[output_index]` safely? Such as output_sharding.size() == getNumOutputs()?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2079,6 +2184,8 @@ def _precompile_select_hidden_states(self) -> None:
             # graph later.
             if self.enable_tensor_parallel:
                 safe_mark_sharding(dummy_hidden, self.mesh, (None, None, "model"))
+            elif self.parallel_mode == ParallelismMode.DATA_PARALLEL_ONLY:
```
sshonTT, line 3041: Could you double-check when this elif condition can actually be hit? If we are in  `ParallelismMode.DATA_PARALLEL_ONLY`, wouldn't self.enable_tensor_parallel already be true and handled by the if block above?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2508,8 +2615,28 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
             self.kv_caches,
         )
 
-        if self.enable_tensor_parallel:
-            # Shard KV Cache — each entry is [k_cache, v_cache].
+        if self.parallel_mode == ParallelismMode.DATA_TENSOR_PRALLEL:
+            # DP+TP: leave the KV cache un-annotated, same as the DP-only
```
sshonTT: The explanation is great, but it feels a bit too specific to certain models and edge cases. Could we clean it up and keep only the core logic for the DP+TP shard spec? We can probably skip the Qwen3 example to keep it clean.

vvukomanTT (approved):

tests/integrations/vllm_plugin/generative/test_opt_generation.py
alinakhanTT: Done!

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Typo fixed

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: The elif is reachable, enable_tensor_parallel and enable_data_parallel are independent flags. In DATA_PARALLEL_ONLY mode only enable_data_parallel=True (see the mode-selection ladder around line 280), so enable_tensor_parallel=False and the first branch is skipped.

I've refactored this in 42a9fb2ef to dispatch on parallel_mode directly (matching TENSOR_PARALLEL_ONLY_1D/2D and DATA_TENSOR_PARALLEL in the if, DATA_PARALLEL_ONLY in the elif) so the reachability is now self-evident

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Done!!

pjrt_implementation/src/api/executable_image.cc
alinakhanTT: Done

integrations/vllm_plugin/vllm_tt/model_runner.py
sshonTT: Got it. Thanks for the explanation

sshonTT (approved):

alinakhanTT (approved):

.github/workflows/test-matrix-presets/basic-test-nightly.json
```diff
@@ -11,7 +11,7 @@
   {"runs-on": "n300-llmbox",         "name": "run_vllm_llmbox_tests",  "dir": "./tests/integrations/vllm_plugin",         "test-mark": "nightly and tensor_parallel and llmbox",  "shared-runners": "true", "extra-wheel": "vllm"},
   {"runs-on": "n150",                "name": "run_vllm_n150_tests",    "dir": "./tests/integrations/vllm_plugin",         "test-mark": "nightly and single_device",               "shared-runners": "true", "extra-wheel": "vllm"},
   {"runs-on": "p150",                "name": "run_vllm_p150_tests",    "dir": "./tests/integrations/vllm_plugin",         "test-mark": "nightly and single_device",               "shared-runners": "true", "extra-wheel": "vllm"},
-  {"runs-on": "n300",                "name": "run_vllm_n300_tests",    "dir": "./tests/integrations/vllm_plugin",         "test-mark": "nightly and data_parallel",               "shared-runners": "true", "extra-wheel": "vllm"},
+  {"runs-on": "n300",                "name": "run_vllm_n300_tests",    "dir": "./tests/integrations/vllm_plugin",         "test-mark": "nightly and data_parallel and not tensor_parallel",  "shared-runners": "true", "extra-wheel": "vllm"},
```
ssaliceTT, line 15: are there tensor parallel tests that aren't data parallel? Wouldn't we want to separate them?
So we'd have DP only, TP only and DP+TP tests.

.github/workflows/test-matrix-presets/basic-test.json
```diff
@@ -11,7 +11,7 @@
   {"runs-on": "n300-llmbox",        "name": "run_torch_multi_chip",   "dir": "./tests/torch",                           "test-mark": "push and llmbox",         "shared-runners": "true" },
   {"runs-on": "n150",               "name": "run_vllm_n150_tests",    "dir": "./tests/integrations/vllm_plugin",        "test-mark": "push and single_device",  "shared-runners": "true", "extra-wheel": "vllm"},
   {"runs-on": "p150",               "name": "run_vllm_p150_tests",    "dir": "./tests/integrations/vllm_plugin",        "test-mark": "push and single_device",  "shared-runners": "true", "extra-wheel": "vllm"},
-  {"runs-on": "n300",               "name": "run_vllm_n300_tests",    "dir": "./tests/integrations/vllm_plugin",        "test-mark": "push and data_parallel",  "shared-runners": "true", "extra-wheel": "vllm"},
+  {"runs-on": "n300",               "name": "run_vllm_n300_tests",    "dir": "./tests/integrations/vllm_plugin",        "test-mark": "push and data_parallel and not tensor_parallel",  "shared-runners": "true", "extra-wheel": "vllm"},
```
ssaliceTT: Also you'd want to make these changes to `vllm-model-tests.json` too so they are kept as subsets accurately. Same with `vllm-model-tests-nightly.json`

.github/workflows/test-matrix-presets/basic-test-nightly.json
alinakhanTT: Yes, generative/test_tensor_parallel_generation.py, pooling/test_tensor_parallel_inference.py, and the llmbox-parametrized cases in sampling/test_device_sampling.py/test_sampling_params.py are TP-only (no data_parallel marker). Split the matrix in 31e57849c so DP-only / TP-only / DP+TP each have their own row on llmbox

.github/workflows/test-matrix-presets/basic-test.json
alinakhanTT: Done!

integrations/vllm_plugin/vllm_tt/platform.py
```diff
@@ -111,6 +111,12 @@ class TTConfig:
     # Flag to enable 2D mesh for tensor parallel execution.
     use_2d_mesh: bool = True
 
+    # When True (default), weight partition specs include the "batch" (DP)
+    # axis — FSDP-style sharding that saves memory at the cost of extra
+    # communication.  When False, weights are sharded only on the "model"
+    # (TP) axis and replicated across DP replicas (classic DP+TP).
+    shard_weights_on_batch_axis: bool = True
```
ssaliceTT: Why is this set to true by default? It would make sense for larger models but for most of the models we have currently, is it necessary to shard weights (QKV for example) on batch as well? Doesn't it lead to more CCLs? 
Wonder if it's better to toggle only for cases where a model doesn't fit on a machine without it. @AleksKnezevic @sshonTT would appreciate thoughts on this.

sshonTT: Personally, I agree with @ssaliceTT. Since extra sharding introduces more CCL overhead in most cases, it would be better to keep it off by default and only enable a 'less CCL mode' (less sharding) initially.

alinakhanTT: flipped the default to False!

mmanzoorTT (commented):
`_pin_input_shardings` was added previously to shard inputs for TP. This needs to be updated/adjusted with current approach. Inputs should not be sharded for TP-only case.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1585,6 +1656,20 @@ def sample_tokens(
...
+                # Data-parallel: inputs carry distinct sentences per replica,
+                # so shard them on the "batch" (DP) axis. The model forward now
+                # runs inside the fused _model_* helpers (main's refactor), so
+                # these marks must be applied here, before that call. Mirrors
+                # the warmup-path marks in _dummy_run so the graphs match.
+                if input_ids is not None:
+                    xs.mark_sharding(input_ids, self.mesh, ("batch", None))
+                if inputs_embeds is not None:
+                    xs.mark_sharding(inputs_embeds, self.mesh, ("batch", None, None))
+                xs.mark_sharding(self.position_ids, self.mesh, ("batch", None))
```
mmanzoorTT: `self._pin_input_shardings` shards inputs for TP. This should be adjusted to not shard inputs in case of TP only.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1866,7 +1951,11 @@ def filtered_get_all_weights(model_config, model):
 
                 if self.enable_tensor_parallel:
                     # Apply sharding constraints to the model weights.
-                    shard_model(model, self.mesh)
+                    shard_model(
+                        model,
+                        self.mesh,
+                        self.tt_config.shard_weights_on_batch_axis,
```
mmanzoorTT: `shard_weights_on_batch_axis` value should only be used in case of `DATA_TENSOR_PARALLEL` only; otherwise it must be true. The default value of `shard_weights_on_batch_axis` is `False` causing TP-2D to only shard on `model` axis which is not right case here; we want to shard weights on both axis for TP-2D

integrations/vllm_plugin/vllm_tt/pooling_runner.py
```diff
@@ -1378,7 +1411,11 @@ def filtered_get_all_weights(model_config, model):
 
                 if self.enable_tensor_parallel:
                     # Apply sharding constraints to the model weights.
-                    shard_model(model, self.mesh)
+                    shard_model(
+                        model,
+                        self.mesh,
+                        self.tt_config.shard_weights_on_batch_axis,
```
mmanzoorTT: Same comment as in `model_runner.py`

integrations/vllm_plugin/vllm_tt/vllm_utils.py
```diff
@@ -4,22 +4,22 @@
...
     """Resolve the (batch, model) SPMD mesh shape against the device count.
 
-    'mesh_shape' is given priority over 'use_2d_mesh'.
-    If 'mesh_shape' is None and 'use_2d_mesh' is True, a pre-defined 2D mesh is used.
-    If 'mesh_shape' is None and 'use_2d_mesh' is False, [1, num_devices] is used.
+    An explicit `mesh_shape` (from config) is given priority; otherwise the
+    shape is derived from `parallel_mode`.
     """
-    # 1. If mesh_shape is provided, use it.
+    # 1. If an explicit mesh_shape is provided, validate and use it.
```
mmanzoorTT, line 22: Add verification to ensure that the given `mesh_shape` also matches with parallel_mode like (1, 4) mesh is not suitable for DP. Add a warning message for it.

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
```diff
@@ -251,25 +272,25 @@ def forward(self, input):
...
-    xla_layer = XlaQKVParallelLinear(layer, mesh)
+    xla_layer = XlaQKVParallelLinear(layer, mesh, shard_weights_on_batch_axis)
     logger.debug("Applied parallel sharding to %s", layer)
     return xla_layer
 
 
 def partition_column_parallel_linear(
-    layer: torch.nn.Module, mesh: xs.Mesh
+    layer: torch.nn.Module, mesh: xs.Mesh, shard_weights_on_batch_axis: bool = True
 ) -> torch.nn.Module:
```
ssaliceTT, line 294: shard weights on batch axis isnt being used here

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
```diff
@@ -296,31 +318,23 @@ def partition_linear(layer: torch.nn.Module, mesh: xs.Mesh) -> torch.nn.Module:
...
     assert isinstance(layer, VocabParallelEmbedding)
-    safe_mark_sharding(layer.weight, mesh, (None, "model"))
-    # Apply sharding constraint to the output. The output rank matches input
-    # rank + 1: 2D input_ids (batch, seq) → 3D output; 1D input_ids (seq,) →
-    # 2D output (e.g., during mm-encoder precompilation). Pre-compute both
-    # sharding strings to avoid per-call overhead and dispatch dynamically.
-    sdy_sharding_3d = _partition_spec_to_sdy_sharding(mesh, (None, None, None))
-    sdy_sharding_2d = _partition_spec_to_sdy_sharding(mesh, (None, None))
-
-    def rank_aware_hook(mod, input, output):
```
ssaliceTT, line 319: Was all of this meant to be overwritten? Was added with a recent PR https://github.com/tenstorrent/tt-xla/pull/3523 so just double checking

integrations/vllm_plugin/vllm_tt/pooling_runner.py
```diff
@@ -273,6 +273,19 @@ def __init__(
...
 
         self.enable_tensor_parallel = self.tt_config.enable_tensor_parallel
+        self.use_2d_mesh = self.tt_config.use_2d_mesh
+
+        if self.enable_data_parallel and self.enable_tensor_parallel:
+            self.parallel_mode = ParallelismMode.DATA_TENSOR_PARALLEL
+        elif self.enable_data_parallel:
+            self.parallel_mode = ParallelismMode.DATA_PARALLEL_ONLY
+        elif self.enable_tensor_parallel:
+            if self.use_2d_mesh:
```
mmanzoorTT: We are giving preference to `mesh_shape` flag while determining the mesh shape. we should also consider mesh_shape here. 
```suggestion
            if self.use_2d_mesh or (1 not in self.tt_config.mesh_shape):
```

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Done in 83cb3a816!

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Done

integrations/vllm_plugin/vllm_tt/vllm_utils.py
alinakhanTT: Added in 83cb3a816. After the existing shape/product checks, an explicit mesh_shape that contradicts the mode now logs a warning, e.g. (1, 4) under DATA_PARALLEL_ONLY (size-1 batch = no DP replicas), a non-(1, model) mesh under TP-1D, or a size-1 axis under TP-2D/DP+TP.

integrations/vllm_plugin/vllm_tt/pooling_runner.py
alinakhanTT: Done!

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
alinakhanTT: fixed in 83cb3a816. It now uses ("model", batch_axis), matching partition_parallel_lm_head

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
alinakhanTT: Yes. The rank-aware 3D/2D dispatch from #3523 now lives in tt_torch.sharding: sharding_constraint_hook --> _normalize_partition_spec_for_rank resizes the all-replicated (None, None, None) spec to the actual output rank, so the 1D-input (2D-output) mm-encoder precompile case is still covered. The local hook became redundant once we adopted the shared helper

integrations/vllm_plugin/vllm_tt/pooling_runner.py
alinakhanTT: Done in 83cb3a816, in both runners!

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1585,6 +1667,16 @@ def sample_tokens(
...
+            if self.parallel_mode in (
+                ParallelismMode.DATA_PARALLEL_ONLY,
+                ParallelismMode.DATA_TENSOR_PARALLEL,
+            ):
+                # _pin_input_shardings already batch-shards input_ids /
+                # inputs_embeds for the DP modes (and leaves TP-only inputs
+                # replicated). position_ids is the one model input it does not
+                # cover, so shard it here on the "batch" (DP) axis before the
+                # fused _model_* forward call.
+                safe_mark_sharding(self.position_ids, self.mesh, ("batch", None))
```
mmanzoorTT: Move `safe_mark_sharding(self.position_ids, self.mesh, ("batch", None))` to `_pin_input_shardings`. All sharding rules will be consistent for warmup/inference.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1938,16 +2041,37 @@ def _dummy_run(self, num_tokens: int, num_reqs: int, num_blocks: int) -> None:
...
+        if self.parallel_mode in (
+            ParallelismMode.DATA_PARALLEL_ONLY,
+            ParallelismMode.DATA_TENSOR_PARALLEL,
+        ):
+            # paged_update_cache requires its update-index (cache_position) and
+            # page_table to have the same per-device leading dim as the K/V
+            # input. Since inputs are sharded ("batch", None), we shard these
+            # along batch too.
+            xs.mark_sharding(page_table, self.mesh, ("batch", None))
+            xs.mark_sharding(cache_position, self.mesh, ("batch",))
```
mmanzoorTT: Not required. `_pin_input_shardings` is called later before running the model to shard inputs.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1938,16 +2041,37 @@ def _dummy_run(self, num_tokens: int, num_reqs: int, num_blocks: int) -> None:
         position_ids = torch.zeros(
             (self.max_num_reqs, num_tokens), dtype=torch.int32
         ).to(self.device)
+
+        if self.parallel_mode in (
+            ParallelismMode.DATA_PARALLEL_ONLY,
+            ParallelismMode.DATA_TENSOR_PARALLEL,
+        ):
+            if input_ids is not None:
+                xs.mark_sharding(input_ids, self.mesh, ("batch", None))
```
mmanzoorTT: prefer using `safe_mark_sharding`.

tests/integrations/vllm_plugin/generative/test_data_parallel_generation.py
```diff
@@ -0,0 +1,195 @@
...
+"""
+import pytest
+import vllm
+from conftest import assert_output_coherent, check_host_memory
+
+
+@pytest.mark.push
+@pytest.mark.data_parallel
+@pytest.mark.dual_chip
+@pytest.mark.parametrize("model_name", ["Qwen/Qwen3-0.6B"])
```
mmanzoorTT, line 18: We are using `opt_level=1` as default and `Qwen/Qwen3-0.6B` fails with `opt_level=1` for few configs. Try `opt_level=0` here for failing tests.

AleksKnezevic (approved):

.github/workflows/test-matrix-presets/basic-test.json
```diff
@@ -11,8 +11,9 @@
...
```
kmabeeTT: For here and the equivalent vllm .json file, I think this would pickup llmbox test currently unless you and "and not llmbox"

```
@pytest.mark.push
@pytest.mark.data_parallel
@pytest.mark.llmbox
@pytest.mark.parametrize("model_name", ["Qwen/Qwen3-0.6B"])
def test_data_parallel_generation_llmbox_padding(model_name: str):
```

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -691,6 +744,17 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> bool:
...
+                ParallelismMode.DATA_PARALLEL_ONLY,
+                ParallelismMode.DATA_TENSOR_PARALLEL,
+            ):
+                # KV cache is replicated across devices, but each device only
+                # writes its own slot's KV. Removing a request and re-adding
+                # it at a different slot later breaks request -> device
+                # affinity and corrupts decode output. Keep the request at
+                # its current slot; downstream consumers must tolerate a
+                # missing num_scheduled_tokens entry for it.
+                continue
```
kmabeeTT, line 888: The new `_update_states` skip keeps unscheduled requests in `input_batch` for the DP modes, and the comment notes consumers "must tolerate a missing `num_scheduled_tokens` entry." We handle that in the non-MM paths (`_prepare_inputs`, `sample_tokens` use `.get(req_id, 0)`), but `_gather_mm_embeddings` still indexes directly:

```python
max_num_scheduled_tokens = max(
    scheduler_output.num_scheduled_tokens[rid]            # ~line 1401
    for rid in self.input_batch.req_ids
)
...
num_scheduled_tokens = scheduler_output.num_scheduled_tokens[req_id]   # ~line 1416
```

So for a multimodal model in DP with mixed-length requests, once one request is descheduled but kept at its slot, this skip makes `_gather_mm_embeddings` `KeyError` on the next step. Since this PR adds `inputs_embeds` batch-sharding for the multimodal DP path, I think it's worth guarding these two spots with `.get(rid, 0)` here too (and skipping the gather for 0-token entries).

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
```diff
@@ -296,38 +326,36 @@ def partition_linear(layer: torch.nn.Module, mesh: xs.Mesh) -> torch.nn.Module:
...
-        sdy = sdy_sharding_3d if output.dim() == 3 else sdy_sharding_2d
-        return torch.ops.tt.sharding_constraint(output, sdy)
-
-    layer.register_forward_hook(rank_aware_hook)
+    batch_axis = "batch" if shard_weights_on_batch_axis else None
+    safe_mark_sharding(layer.weight, mesh, (None, batch_axis))
```
kmabeeTT: Embedding spec went `(None, "model")` → `(None, batch_axis)`, and pure TP forces `batch_axis="batch"`. On a 1D `(1, N)` mesh that axis is size-1, so the hidden dim is no longer sharded — the embedding ends up replicated on every device instead of split across `N`. `qwen3-32b-qb2-tp` is 1D TP, so its ~150k×5120 embedding is now fully replicated. Intended, or should 1D-TP keep `"model"` sharding here?

kmabeeTT (commented):
Much of this I don't have experience with, but I spent a bit of time with claude and found 3 things worth commenting on.

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
ssaliceTT: Shouldn't affect largest models we'd realistically run on qb2 since the DRAM overhead from that replication isn't much. It would remove an all-gather that would be necessary for RMS norm right after. Other larger models non-vLLM sharding specs follow the same pattern too, and this shouldn't be a bottleneck for the largest model sizes we'd want to run on QB2s. 
Also iirc later on for supporting distributed rms norm this would be necessary as well.

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: I tried this, but it regresses DP+TP. _get_dummy_inputs (the fused-prefill warmup) also calls _pin_input_shardings, and that path needs position_ids left unsharded, once _pin shards it, the fused backbone graph fails to compile with Redundant ttnn::ToLayoutOp, due to the forcing of tile/row major layouts -> ValueError: Error code: 13 at engine init. Verified on an 8-chip llmbox: test_data_tensor_parallel_generation_push fails with the consolidation and passes once the explicit per-path marks are restored. So I've kept position_ids sharded explicitly here instead of moving it into _pin. If you know a clean way to unify these that still leaves the fused precompile path unsharded, happy to take it!!

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Reverted, since position_ids can't go through _pin (it breaks the DP+TP fused precompile), these explicit marks aren't actually redundant for the warmup path, so I've restored them to match the original

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Done

tests/integrations/vllm_plugin/generative/test_data_parallel_generation.py
alinakhanTT: Done! Applied it to all four (n300 tight/wider, llmbox padding/tight)

.github/workflows/test-matrix-presets/basic-test.json
alinakhanTT: Thank you! You were right, added and not llmbox to the n300 data_parallel mark in basic-test.json and both vllm-model-tests.json / vllm-model-tests-nightly.json

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Done!

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
alinakhanTT: You're right, it was unintentionally replicating on 1D-TP. Fixed by sharding only the hidden dim!

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1460,23 +1549,27 @@ def _pin_input_shardings(
...
+        execution (sample_tokens) paths so their graphs match; otherwise warmup
+        shards these inputs while execution leaves them replicated, and the
+        backbone recompiles at the first real step for any shardable batch. Also
+        keeps large prefill traces from splitting into a secondary sync that
+        drops mhlo.spmd_output_sharding (which trips shlo_clean_for_xla_ingestion).
+
+        Only DATA_PARALLEL_ONLY and DATA_TENSOR_PARALLEL split the batch across
+        the mesh's "batch" axis. The tensor-parallel-only modes (1D and 2D/FSDP)
+        replicate the full batch to every device and shard the model weights
+        instead, so their inputs must stay unsharded here."""
```
mmanzoorTT:
```suggestion
        """Pin model input shardings (batch dimension across the mesh) in
        DATA_PARALLEL_ONLY and DATA_TENSOR_PARALLEL during both warmup and
         inference so both paths produce identical graphs and avoid recompilation."""
```

integrations/vllm_plugin/vllm_tt/model_runner.py
mmanzoorTT: Then it is big problem. you are not sharding position_ids during warmup but sharding during inference. This will trigger re-compilation and probably, you are re-compiling all the graphs again.

mmanzoorTT: Fixed in `10441e5`. 
- Moved `position_ids` to  `_pin_input_shardings`
- Added shardings for attn_metadata attributes in `_get_dummy_inputs` to match with `_prepare_inputs`. 
- Added `dp_size` for attn_metadata in `_get_dummy_inputs`.

integrations/vllm_plugin/vllm_tt/model_runner.py
mmanzoorTT: Please try to update other comments as well. specially in tests.

alinakhanTT:
### TP stop-token flake: xfail'd, tracked separately

`test_stop_token_ids[n300_llmbox]` and `test_min_tokens[n300_llmbox]` were flaking intermittently on TP-llmbox. Root-caused on device: under TP, greedy (`temperature=0`) decode is **not run-to-run reproducible at a near-tie logit**, the cross-chip fp reduction on the logit path isn't bit-reproducible, so the `argmax` winner can flip between two identical `generate()` calls (confirmed: two same-process, same-prompt greedy baselines diverged at a near-tie position; `argmax` is on-device and deterministic, so the flip is in the logits). It's a pre-existing multi-chip greedy-determinism limitation that the sharding exposes at a near-tie: orthogonal to the DP/TP feature here.

Marked both `xfail(strict=False)` on the `n300_llmbox` target only (39f213a14); other targets unchanged. `strict=False` so a run that happens to pass doesn't fail as XPASS.

Tracked in #5520 for further investigation

Issue #4491, opened 2026-04-30
[vLLM] Per-step decode cost scales with max_model_len even on short sequences


kmabeeTT:
## Context

Identical short-prompt workload on Llama-3.1-8B-Instruct via tt-inference-server, only `max_model_len` changes:

| `max_model_len` | tok/s |
|---|---|
| 128 | 14.1 |
| 4096 | 11.9 |

A ~16% drop from a config change alone. The actual sequence length is the same in both runs — the request never approaches `max_model_len`.

## Problem

Most users will hit `max_model_len=4096` (matches tt-cloud-console's default `max_tokens=4096` slider) even when their actual chat is short. So they pay the full overhead despite using only a fraction of the configured context.

It feels like there should be a way for short-actual-sequence requests to run at closer to short-`max_model_len` perf, but I don't know enough about the static-shape compile model to say what that would look like.

## Notes / observations

- `VLLM_TPU_MOST_MODEL_LEN` looks like it's meant to address this (compiles a second graph for typical-case lengths) but setting `VLLM_TPU_MOST_MODEL_LEN=256` with `max_model_len=4096, max_num_seqs=32` crashes engine init:
  ```
  TT_FATAL: input_shape[2] (2048) <= cache_shape[2] (32) * page_table_shape[1] (8)
  Input seq_len must fit in max_num_blocks_per_seq
  ```
  Looks like prefill paddings (here 2048) get compiled against the smaller most-len page table and the kernel rejects them. Possibly the env var only works for narrower configs than ours.

- vLLM's `compile_sizes` / `compile_ranges_split_points` aren't honored by the tt-xla integration as far as I can tell.

## Acceptance

A short-sequence request runs at ~the same per-token cost regardless of whether `max_model_len` is 128 or 4096. Long-context requests still work when `max_model_len=4096` is actually needed.

Found while debugging Llama-3.1-8B-Instruct on Blackhole p150 during demo prep.

kmabeeTT:
[...]

PR #4953, opened 2026-05-26
fix vllm perf's tokens / second value

jazpurTT:
### Ticket
#4491 

### Problem description
Current math for the tokens/second is missleading. When testing batch_size > 1, by just using `tokens_per_user / (max(last_token_times) - min(first_token_times))` you are mixing time windows from different users/requests. This is specifically an issue with vllm because the token gens don't seem to happen at the same exact time. We see a large delta between the time of the first token gen of the first user and the rest.
```
  Request 0: first_token_time=767.0ms
  Request 1: first_token_time=1534.0ms
  [ ... ]
  Request 31: first_token_time=1530.7ms
```
Instead for vllm I think it's better if we show avg tokens/second.

### What's changed
For vllm perf benchmark
1. Updated how we calculate the tokens/second data point and changed the name to `Avg. Sample per second` to clarify what we are displaying.
2. Added new var called `avg_decode _time` and renamed `decode_total_time` to `e2e_decode_time`
4. Added a custom measurements for `avg_decode _time` and `samples_per_sec`, and use them instead of `total_samples / total_time` (when they are part of the measuremnts) when we fetch/compare perf benchmarks.

### Checklist

### DIFF
--- a/.claude/skills/ci-benchmark-analyzer/scripts/fetch_perf_reports.py
+++ b/.claude/skills/ci-benchmark-analyzer/scripts/fetch_perf_reports.py
@@ -109,7 +109,12 @@ def extract_metrics(report: dict) -> dict:
 
     total_samples = measurements.get("total_samples", 0)
     total_time = measurements.get("total_time", 0)
-    samples_per_sec = total_samples / total_time if total_time > 0 else None
+    if "samples_per_sec" in measurements:
+        samples_per_sec = measurements["samples_per_sec"]
+    elif total_time > 0:
+        samples_per_sec = total_samples / total_time
+    else:
+        samples_per_sec = None
 
     return {
         "model": report.get("model", "unknown"),
--- a/tests/benchmark/benchmarks/vllm_benchmark.py
+++ b/tests/benchmark/benchmarks/vllm_benchmark.py
@@ -68,21 +68,18 @@ def _create_llm(config: VLLMBenchmarkConfig) -> vllm.LLM:
 
 def _extract_metrics(
     outputs: List[vllm.RequestOutput],
-    batch_size: int,
-) -> Tuple[float, int, float, float]:
+) -> Tuple[float, List[int], float, float]:
     """
     Extract per-request metrics and return aggregated per-user values.
 
     Returns:
-        (avg_ttft_ms, tokens_per_user, decode_total_time, tokens_per_sec_per_user)
+        (avg_ttft_ms, tokens_per_user, avg_decode_time_s, avg_tokens_per_sec)
     """
     ttft_values = []
-    total_gen_tokens = 0
 
     for i, output in enumerate(outputs):
         stats = output.metrics
         gen_tokens = len(output.outputs[0].token_ids)
-        total_gen_tokens += gen_tokens
 
         ttft_ms = stats.first_token_latency * 1000.0
         ttft_values.append(ttft_ms)
@@ -103,17 +100,33 @@ def _extract_metrics(
 
     avg_ttft_ms = sum(ttft_values) / len(ttft_values) if ttft_values else 0.0
 
-    decode_total_tokens = total_gen_tokens - batch_size
-    tokens_per_user = decode_total_tokens // batch_size
-
+    tokens_per_user = [len(o.outputs[0].token_ids) - 1 for o in outputs]
     first_token_times = [o.metrics.first_token_ts for o in outputs]
     last_token_times = [o.metrics.last_token_ts for o in outputs]
-    decode_total_time = max(last_token_times) - min(first_token_times)
-    tokens_per_sec_per_user = (
-        (tokens_per_user / decode_total_time) if decode_total_time > 0 else 0.0
+
+    decode_times_per_user = [
+        last_token_time - first_token_time
+        for last_token_time, first_token_time in zip(
+            last_token_times, first_token_times
+        )
+    ]
+    tokens_per_sec_per_user = [
+        tokens / decode_time if decode_time > 0 else 0.0
+        for tokens, decode_time in zip(tokens_per_user, decode_times_per_user)
+    ]
+    avg_tokens_per_sec = sum(tokens_per_sec_per_user) / len(tokens_per_sec_per_user)
+    avg_decode_time_s = (
+        sum(decode_times_per_user) / len(decode_times_per_user)
+        if decode_times_per_user
+        else 0.0
     )
 
-    return avg_ttft_ms, tokens_per_user, decode_total_time, tokens_per_sec_per_user
+    return (
+        avg_ttft_ms,
+        tokens_per_user,
+        avg_decode_time_s,
+        avg_tokens_per_sec,
+    )
 
 
 def _get_device_info(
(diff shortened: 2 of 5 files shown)


tests/benchmark/benchmarks/vllm_benchmark.py
```diff
@@ -109,11 +109,20 @@ def _extract_metrics(
     first_token_times = [o.metrics.first_token_ts for o in outputs]
     last_token_times = [o.metrics.last_token_ts for o in outputs]
     decode_total_time = max(last_token_times) - min(first_token_times)
```
kmabeeTT: Minor nit (non-blocking): `decode_total_time` is still the old all-users span and gets reported as `total_time`, which is no longer multiplicatively consistent with the new `samples_per_sec`. Could replace it with the mean per-user decode time so the two fields tell the same story.

kmabeeTT (approved):
Nice! Thanks! One small nit maybe should fix.

tests/benchmark/benchmarks/vllm_benchmark.py
jazpurTT: Yeah `decode_total_time`, later passed down as `total_time` is a shared measurement across all benchmarks, so I wanted to leave it as is. But to make it clearer I:
1. Renamed it to `e2e_decode_time`
2. Added another data point called `avg_decode_time`
3. Added a custom measurements for `avg_decode _time` and `samples_per_sec` in vllm becnhmark
4. Use `samples_per_sec` instead of `total_samples / total_time` if the custom measurement exists when we fetch/compare perf benchmarks.

kmabeeTT: Thank you!

tests/benchmark/benchmarks/vllm_benchmark.py
```diff
@@ -122,12 +109,32 @@ def _extract_metrics(
 
     first_token_times = [o.metrics.first_token_ts for o in outputs]
     last_token_times = [o.metrics.last_token_ts for o in outputs]
-    decode_total_time = max(last_token_times) - min(first_token_times)
-    tokens_per_sec_per_user = (
-        (tokens_per_user / decode_total_time) if decode_total_time > 0 else 0.0
+    e2e_decode_time = max(last_token_times) - min(first_token_times)
```
odjuricicTT: Is this still needed?

tests/benchmark/benchmarks/vllm_benchmark.py
odjuricicTT: Can you switch to having a different tokens per user for every user instead of the average?

Basically do the same as with `decode_times_per_user`.

tests/benchmark/benchmarks/vllm_benchmark.py
```diff
@@ -122,12 +109,32 @@ def _extract_metrics(
...
+
+    decode_times_per_user = [
```
odjuricicTT, line 107: Can you add a debug / verbose level print of these for every user. It would be useful to see if there is a discrepancy between different users.

tests/benchmark/benchmarks/vllm_benchmark.py
```diff
@@ -236,9 +257,9 @@ def benchmark_vllm(
         dataset_name=dataset_name,
         date=metadata["date"],
         machine_name=metadata["machine_name"],
-        total_time=decode_total_time,
+        total_time=e2e_decode_time,
```
odjuricicTT: I think it makes more sense to pass `avg_tokens_per_sec` here.

odjuricicTT (commented):
Thanks for fixing this!

I'm interested to see if there are significant result differences in t/s/u for existing vLLM models?

kmabeeTT:
About @odjuricicTT point about numbers/impact here, agreed it makes sense to share before/after this fix.  FWIW I have datapoints (Llama3.2-3B Batch32) in my adjacent ticket comment [here](https://github.com/tenstorrent/tt-xla/pull/4955#issuecomment-4550339268) which solved a prefill slowdown and made t/s/u flat across seq-len because prefill was being baked into the t/s/u calc here without this fix.   Reached 19 tok/sec/user flat without this fix, and then reported as 22 tok/sec/user with this fix.

tests/benchmark/benchmarks/vllm_benchmark.py
jazpurTT: It really depends, the reason why I kept it is because it will be passed down as `total_time` and this is a measurement that is produced by all the benchmarks, and wanted to keep the logic aligned. 

To me, `total_time` refers to the total time from the first decode to the last so `max(last_token) - min(first_token)` is technically correct, changing it to the avg on vllm seemed out of place. I added `avg_decode_time` on vllm to make sure we show there is a difference between total and avg time (unlike in the other benchmarks), and made sure `samples_per_sec` is used when comparing instead of `total_samples / total_time`.  If you'd prefer I can just remove it and use the `avg_decode_time` for vllm only.

tests/benchmark/benchmarks/vllm_benchmark.py
jazpurTT: yup will do

tests/benchmark/benchmarks/vllm_benchmark.py
jazpurTT: I think we already do that, I was looking at my logs:
```
  Request 0: gen_tokens=64, TTFT=767.0ms, decode_tokens=63, decode_time=3.223s, decode_tps=19.5
  Request 1: gen_tokens=64, TTFT=1534.0ms, decode_tokens=63, decode_time=2.455s, decode_tps=25.7
 ...
 ```
 I'll fix it though I think the math it's doing is not correct.

File: tests/benchmark/benchmarks/vllm_benchmark.py (hunk 4 above)
Comment (jazpurTT) [reply tests/benchmark/benchmarks/vllm_benchmark.py]: I think you mean `avg_decode_time`? yes I'll change it

File: tests/benchmark/benchmarks/vllm_benchmark.py (hunk 2 above)
Comment (odjuricicTT) [reply tests/benchmark/benchmarks/vllm_benchmark.py]: This var might be used for calculating the sample per second again in some downstream superset dashboards. For that purpose it would be better to have the avg time instead of total time for now.

But in the longer term it should probably be removed and the superset dashboard refactored.

File: .claude/skills/ci-benchmark-analyzer/scripts/fetch_perf_reports.py [hunk 5]
```diff
@@ -109,7 +109,12 @@ def extract_metrics(report: dict) -> dict:
 
     total_samples = measurements.get("total_samples", 0)
     total_time = measurements.get("total_time", 0)
-    samples_per_sec = total_samples / total_time if total_time > 0 else None
+    if "samples_per_sec" in measurements:
+        samples_per_sec = measurements["samples_per_sec"]
+    elif total_time > 0:
+        samples_per_sec = total_samples / total_time
+    else:
+        samples_per_sec = None
```
mmanzoorTT, line 117: Why `samples_per_sec` is set differently for `else` block in here and `compare_nightlies.py`?

mmanzoorTT (commented):
@jazpurTT Any idea why vllm perf tests are unable to fetch the previous nightly report and skipping regression check? is this PR fixing it as well?

alinakhanTT (approved):

.claude/skills/ci-benchmark-analyzer/scripts/fetch_perf_reports.py
jazpurTT: None is used in `fetch_perf_reports.py` because the value is just displayed in a report (missing = `null`), while `0` is used in `compare_nightlies.py` because it's fed into `compute_diff`'s `previous <= 0` arithmetic, where `None` would raise a `TypeError`.

jazpurTT:
> @jazpurTT Any idea why vllm perf tests are unable to fetch the previous nightly report and skipping regression check? is this PR fixing it as well?

this PR is just to get the metrics calculated correctly for vllm's perf benchmark

mmanzoorTT (approved):

odjuricicTT (approved):

AleksKnezevic (approved):

Issue #1414, opened 2025-09-17
Implement `PJRT_Device_MemoryStats`


LPanosTT:
torch-xla provides a call which allows the user to get memory information about a device. It requires that `PJRT_Device_MemoryStats` is implemented however. 

The TPU vLLM implementation uses this to determine what the max size of a new KV cache can be at any moment. It could be beneficial four ours to have this as well.

When implemented correctly we should be able to retrieve the amount of available DRAM space by calling: 
```
import torch_xla.core.xla_model as xm
m = xm.get_memory_info(device)
total_memory_size = m["bytes_limit"]
current_mem = m["bytes_used"]
```

kmabeeTT:
Parking a partial implementation here for future reference: branch `kmabee/issue_1414_dram_size_bytes_attr` (commit d600fb879).

What's in it:

- PJRT plugin exposes `dram_size_bytes` as a device attribute, sourced from the tt-mlir `ChipDesc` flatbuffer (`num_dram_channels * dram_channel_size`) that the plugin already loads at startup via `tt::runtime::getCurrentSystemDesc()`. No new tt-metal/UMD queries — just stops discarding the value.
- `integrations/vllm_plugin/vllm_tt/worker.py` reads the attribute via `xr.global_runtime_device_attributes()` and uses it for `total_memory_size` in `determine_available_memory()`, with a 12 GiB n150-safe fallback if the attribute is absent.
- Verified on p150b single-device: attribute reports 34225520640 bytes (31.88 GiB), KV-cache budget rises from 10.8 GiB → 28.7 GiB at default `gpu_memory_utilization=0.9`. Llama-3.2-3B b32 benchmark PASS with unchanged TTFT.

Why it's *not* the full fix this issue asks for:

- Only the `bytes_limit` half is implemented. `bytes_used` and `peak_bytes_used` are not — they require the full `PJRT_Device_MemoryStats` API plumbed through to tt-metal's allocator, which this branch does not do.
- Without `bytes_used`, `determine_available_memory()` still hardcodes `current_mem = 0`, so the returned KV budget does not subtract model weights. **This creates OOM risk on Blackhole for any non-trivial model** (e.g. Llama-3.1-8B weights ~16 GiB on a 32 GiB p150 leaves ~13 GiB real headroom, but the function reports 28.7 GiB → vLLM will try to allocate more KV cache than fits and OOM at engine init).
[...]

PR #4960, opened 2026-05-27
pjrt+vllm_plugin: expose dram_size_bytes; use it for KV cache sizing

### LINKED ISSUE #4958: [vLLM] hardcoded 12 GB KV cache budget under-utilizes Blackhole DRAM (~3× on p150)
### Symptom

On Blackhole boards, vLLM sees only 12 GiB of device DRAM (the n150 value) even though the hardware has 28-32 GiB. This caps achievable `batch × max_model_len` configurations on Blackhole by ~2.3-2.7× of what the hardware actually allows.

| board | physical DRAM | reported today | with fix |
|---|---:|---:|---:|
| n150 (Wormhole) | 12 GiB | 12 GiB | 12 GiB (no change) |
| p100 (Blackhole) | 28 GiB | 12 GiB | 27.9 GiB |
| p150 (Blackhole) | 32 GiB | 12 GiB | 31.9 GiB |

This value is then multiplied by vLLM's `gpu_memory_utilization` (default 0.9) to size the KV cache budget — so the actual KV budget under-utilizes the device by the same ratio.

### Root cause

`integrations/vllm_plugin/vllm_tt/worker.py:236` (in `determine_available_memory()`) hardcodes:

```python
total_memory_size = 12 * 1024**3   # m["bytes_limit"]
current_mem = 0                     # m["bytes_used"]
```

The comment explicitly notes this is a placeholder pending #1414 (`PJRT_Device_MemoryStats`). The 12 GiB value is n150-correct but wrong for every Blackhole board.

### Fix

Expose total device DRAM as a PJRT device attribute sourced from the tt-mlir `ChipDesc` (`num_dram_channels × dram_channel_size`) that the plugin already loads at startup. Read it from Python in `determine_available_memory()`, with the 12 GiB hardcode kept as a fallback.

### Caveat

This raises `bytes_limit` to the real value but does not subtract model weights — that still needs `bytes_used` from #1414. On Blackhole + larger models the over-report is now larger in absolute terms, so users may need to lower `gpu_memory_utilization` (e.g. from 0.9 → 0.45 for Llama-3.1-8B on p150) to avoid device OOM at engine init. Still a strict improvement on the status quo, which under-reports in the other direction.

### Linked issues

- #1414 — `PJRT_Device_MemoryStats`. Full fix (subtracting weights via `bytes_used`) depends on it. This issue is an independent partial improvement.

kmabeeTT:
### Ticket

Closes #4958. Related to #1414 (`PJRT_Device_MemoryStats`).

### Problem description

`worker.py:determine_available_memory()` hardcodes `total_memory_size = 12 * 1024**3` (n150 DRAM). On Blackhole (p100 = 28 GiB, p150 = 32 GiB) this caps the KV cache budget at a ~12 GiB ceiling regardless of physical DRAM — ~2.7× under-utilization on p150. The plugin already loads the tt-mlir SystemDesc at client init via `tt::runtime::getCurrentSystemDesc()`; previously every field except `arch` was discarded.

### What's changed

- `pjrt_implementation/inc/api/device_description.{h,cc}`: add `dram_size_bytes` (uint64) parameter to the constructor; add a `createInt64Attribute` helper; push the value as a PJRT device attribute alongside the existing `device_arch`.
- `pjrt_implementation/inc/api/device_instance.{h,cc}`: thread `dram_size_bytes` through `createInstance` and the private constructor.
- `pjrt_implementation/src/api/client_instance.cc`: at device populate time, compute `num_dram_channels × dram_channel_size` from the existing `chip_desc` flatbuffer entry and pass it to `DeviceInstance::createInstance`.
- `pjrt_implementation/src/api/unit_tests/device_description_tests.cc`: update `API_PJRT_DeviceDescription_Attributes` to expect 2 attributes, locate them by name (order is an implementation detail), and assert the int64 value matches.
- `integrations/vllm_plugin/vllm_tt/worker.py`: add `_get_device_dram_bytes()` helper that reads `dram_size_bytes` via `xr.global_runtime_device_attributes()` with the 12 GiB n150-safe fallback; use it for `total_memory_size` in `determine_available_memory()`; log the selected device DRAM, `gpu_memory_utilization`, and final KV cache budget at INFO so the chosen sizing is visible at engine init.

No new tt-metal/UMD queries — the SystemDesc was already loaded at startup; this just stops discarding the DRAM fields.

### Caveat

This raises `bytes_limit` to the real value but does not subtract model weights — that still needs `bytes_used` from #1414. On Blackhole + larger models, users may need to lower `gpu_memory_utilization` (e.g. from 0.9 → 0.45 for Llama-3.1-8B on p150) to avoid OOM at engine init. Strict improvement on the status quo, which under-reports in the other direction. See #4958 for the full discussion.

### Verification

Llama-3.2-3B b32, `max_model_len=128`, warmed up, p150 single device:

- `dram_size_bytes` attribute auto-detected as 34,225,520,640 bytes (31.88 GiB)
- KV cache budget rises from 10.8 GiB → 28.7 GiB at the vLLM-default `gpu_memory_utilization=0.9`
- Request 0 TTFT unchanged at 901 ms
- Benchmark PASS, no OOM
- n150 deployments see no change (their SystemDesc still reports 12 GiB)
- Example output on p150: `(EngineCore pid=246007) INFO:vllm_tt.worker:KV cache sizing: device DRAM = 31.88 GiB, gpu_memory_utilization = 0.001, KV cache budget = 0.03 GiB`

### Checklist

- [x] Tested locally on p150 (b32 PASS, no OOM at default `gpu_memory_utilization`)
- [x] No regression on n150 (SystemDesc unchanged → same budget)
- [x] Unit test updated for new 2-attribute expectation
- [x] vLLM Nightly CI run (passing) : https://github.com/tenstorrent/tt-xla/actions/runs/26494270090

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/worker.py
+++ b/integrations/vllm_plugin/vllm_tt/worker.py
@@ -182,6 +182,31 @@ class TTWorker:
             # If usage stat is enabled, collect relevant info.
             report_usage_stats(self.vllm_config)
 
+    _DEFAULT_DEVICE_DRAM_BYTES = 12 * 1024**3  # n150-safe fallback
+
+    def _get_device_dram_bytes(self) -> int:
+        """Total on-device DRAM in bytes, queried from the PJRT plugin.
+
+        The plugin exposes ``dram_size_bytes`` as a device attribute, computed
+        from the tt-mlir SystemDesc (``num_dram_channels × dram_channel_size``)
+        at device-open time.  Returns the n150 default if the attribute is
+        missing or zero, which can happen with older plugin builds.
+        """
+        try:
+            attrs = xr.global_runtime_device_attributes()
+            bytes_val = int(attrs[self.local_rank].get("dram_size_bytes", 0))
+        except Exception as e:
+            logger.warning(
+                "Could not read dram_size_bytes from PJRT device attributes "
+                "(%s); falling back to %.0f GB",
+                e,
+                self._DEFAULT_DEVICE_DRAM_BYTES / 1024**3,
+            )
+            return self._DEFAULT_DEVICE_DRAM_BYTES
+        if bytes_val <= 0:
+            return self._DEFAULT_DEVICE_DRAM_BYTES
+        return bytes_val
+
     def determine_available_memory(self) -> int:
         if self.model_config.runner_type == "pooling":
             return int(11596411699)
@@ -237,7 +262,11 @@ class TTWorker:
         # total_memory_size = m["bytes_limit"]
         # current_mem = m["bytes_used"]
         # @LPanosTT: For now we will always report that no memory has been used.
-        total_memory_size = 12 * 1024**3  # m["bytes_limit"]
+        # TTXLA: query total device DRAM via the PJRT plugin's device attribute
+        # (sourced from the tt-mlir SystemDesc that the plugin already loads at
+        # startup).  Falls back to 12 GB if the attribute is absent (e.g. older
+        # plugin build).
+        total_memory_size = self._get_device_dram_bytes()
         current_mem = 0  # m["bytes_used"]
 
         # Ideally we would use profiled = m["peak_bytes_used"] to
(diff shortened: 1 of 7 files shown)


mmanzoorTT (approved):

jameszianxuTT (approved):
lgtm!

AleksKnezevic (approved):

PR #4971, opened 2026-05-27
[CI] Multihost CI integration

### LINKED ISSUE #964: CI workflow for multihost

### LINKED ISSUE #4230: Multihost Call Test Integration

nsumrakTT:
### Ticket
closes https://github.com/tenstorrent/tt-forge/issues/964
closes https://github.com/tenstorrent/tt-xla/issues/4230

### Description and Origin
This PR originates with https://github.com/tenstorrent/tt-xla/pull/4286 by @jameszianxuTT 
The difference is in provisioning of the runner with github container hook that adds/replaces --network host to the docker image. This allows removal of exposure of docker.sock which carries some security problems with it.

Since runners requires provisioning tests for multi-host-t3000 are removed (until metal team does reprovisioning). 

### What's changed
Added multihost support with single entry and exit points for multihost runs inside general call-test.yml workflow

### Checklist
Test run: https://github.com/tenstorrent/tt-xla/actions/runs/26514672098


.github/scripts/multihost-start-workers.sh
```diff
@@ -0,0 +1,106 @@
...
+echo "All worker containers started successfully"
+
+if [[ ! -x "${REMOTE_DOCKER_SH}" ]]; then
+  echo "remote_docker.sh not found or not executable at ${REMOTE_DOCKER_SH}" >&2
+  exit 1
+fi
+
+
+echo "=== SSH connectivity check (via remote_docker.sh) ==="
+mpirun --allow-run-as-root --hostfile /etc/mpirun/hostfile \
```
vvukomanTT, line 94: We should use the already defined variable instead of the hardcoded path here: 
`--hostfile "${HOSTFILE}"`

.github/scripts/multihost-build-worker-hostfile.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
```
vvukomanTT, line 1: This `mpirun-workers-hostfile` file can be generated at provision time in ansible, right?
We haven't done it here because we do not want to break the existing multihost provisioning infra provided by tt-metal.

I would prefer it if we added this hostfile generation logic to their ansible as well, instead of doing it at CI time. (would require changes to metal's ansible repo as well, but much cleaner)

vvukomanTT (approved):

.github/scripts/multihost-build-worker-hostfile.sh
jameszianxuTT: We can open the conversation on metal-internal-workflows, I agree it would be better to generate this at provisioning, though I'd prefer to get this PR ASAP so we can get some coverage.

This script is a lot of (my) hacky code added to convert:
```
f10cs03 slots=1
f10cs04 slots=1
```
to 
```
f10cs04 slots=1
```

jameszianxuTT (approved):

tests/infra/utilities/torch_multichip_utils.py
```diff
@@ -9,6 +9,39 @@
...
+        ``(batch_dim, model_dim)`` with product equal to ``num_devices``.
+
+    Examples:
+        8 -> (2, 4)    # e.g. dual_bh_quietbox
+        16 -> (1, 16)  # e.g. dual T3K / loudbox 1x16
+        32 -> (4, 8)   # e.g. single galaxy
+        64 -> (8, 8)   # e.g. dual galaxy
+        128 -> (8, 16) # e.g. quad galaxy
+    """
+    if num_devices == 8:
```
nvukobratTT, line 29: Is this the correct way to distinguish them? E.g., the 16-chip system can be [1, 16] or [2x8]

cc @jameszianxuTT

jameszianxuTT: @nvukobratTT These are just defaults and can be overridden by specific tests.

w/r/t why 16 chips is opened as 1x16, this default is modelling a hardware and fabric constraint that is not properly controlled from upstream. We have two current multihost 16 device topologies:
- dual t3k
- dual bh lb

And both **can only be opened in 1x16 logical mesh** right now. For dual bh lb, there are unspecified hangs in non 1x16 due to metal issues that no one is debugging which I found about from slack DMs, and for dual t3k it is not physically possible to open up anything other than a 1x16 mesh device.

jameszianxuTT: From forge, we don't really have capability to query the underlying physical connectivity and determine which mesh shapes are even valid. I know from forge devs' perspective, we are assuming that as long as the product of device mesh dimensions equals num devices, we can open the mesh shape, but that is actually not generally true, which took some pain to find out during multihost bringup.

Something query tool could be made but would probably require closer integration with metal scaleout.

nsumrakTT:
@AleksKnezevic @mrakitaTT @nvukobratTT @odjuricicTT @rpavlovicTT Hey guys, this PR is open for a long time. Please review, so we can continue with multihost efforts. As this is coop effort, if you have some non-major issues lets make an issue and incrementally fix it.

nvukobratTT (approved):

PR #4975, opened 2026-05-28
Refactor Sliding Attention Overrides with Generic Model Rewrite Support

devisettymahidhar608:
### Ticket

### Problem description
The goal of this refactoring is to streamline the execution flow for sliding attention by introducing two main functions: `override_model_sliding_window_causal_mask`, which will reside in **tt-xla**, and input generation, which will reside in **tt-forge-models**. Previously, each model had its own dedicated override function, leading to duplicated logic across implementations. As part of this refactor, we plan to replace these model-specific override functions with a single generalized implementation that can support all models uniformly, while keeping only the input generation logic model-specific within **tt-forge-models**.

### What's changed
Previously, we maintained separate functions for each model, namely `override_gpt_oss_sliding_window_causal_mask`, `override_olmo3_sliding_window_causal_mask`, and `override_ministral_sliding_window_causal_mask`. As part of the refactoring effort, these model-specific implementations have been consolidated into a single generalized function, `**override_model_sliding_window_causal_mask**`, which provides a unified solution across all supported models.

To achieve the second goal, I introduced a new property **requires_model_rewrites** on the base loader in tt-forge-models (base.py) and propagate it to tt-xla. In tt-xla (dynamic_torch_model_tester.py), this flag is checked before compilation, and if it is set to true, the override_model_sliding_window_causal_mask function is invoked.

logs:
[mistral_8b.log](https://github.com/user-attachments/files/28341849/mistral_8b.log)
[olmo_7b.log](https://github.com/user-attachments/files/28341851/olmo_7b.log)
[olmo_32b.log](https://github.com/user-attachments/files/28341853/olmo_32b.log)
[gpt.log](https://github.com/user-attachments/files/28342749/gpt.log)

### Checklist

### DIFF
--- a/examples/pytorch/gpt_oss_20b.py
+++ b/examples/pytorch/gpt_oss_20b.py
@@ -27,7 +27,7 @@ from transformers.modeling_outputs import CausalLMOutputWithPast
 from tt_torch.sharding import sharding_constraint_hook
 from tt_torch.transformers_overrides import (
     override_cache_sliding_window_layers,
-    override_gpt_oss_sliding_window_causal_mask,
+    override_model_sliding_window_causal_mask,
 )
 
 DEFAULT_PROMPTS = ["Explain quantum mechanics."]
@@ -152,7 +152,7 @@ def setup_model_and_tokenizer(
         attn_implementation="eager",
     )
     # Override the gpt_oss model to use the TT-friendly sliding window causal mask
-    override_gpt_oss_sliding_window_causal_mask()
+    override_model_sliding_window_causal_mask(model)
     model = model.eval()
 
     tokenizer: PreTrainedTokenizer = AutoTokenizer.from_pretrained(model_name)
--- a/tests/runner/testers/torch/dynamic_torch_model_tester.py
+++ b/tests/runner/testers/torch/dynamic_torch_model_tester.py
@@ -148,6 +148,16 @@ class DynamicTorchModelTester(TorchModelTester):
             batch_size=batch_size,
         )
 
+        # Loaders expose `requires_model_rewrites` to request TT-friendly
+        # in-place model rewrites (e.g. sliding-window mask patches). Querying
+        # the loader property keeps the inputs dict free of non-model keys.
+        if getattr(self.dynamic_loader.loader, "requires_model_rewrites", False):
+            from tt_torch.transformers_overrides import (
+                override_model_sliding_window_causal_mask,
+            )
+
+            override_model_sliding_window_causal_mask(self._model)
+
         if self.parallelism == Parallelism.DATA_PARALLEL:
             num_devices = xr.global_runtime_device_count()
             if isinstance(inputs, collections.abc.Mapping):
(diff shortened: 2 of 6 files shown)


sonalibaskaran2499 (approved):

tests/runner/testers/torch/dynamic_torch_model_tester.py
```diff
@@ -147,6 +147,16 @@ def _get_input_activations(self):
             batch_size=batch_size,
         )
 
+        # Loaders expose `requires_model_rewrites` to request TT-friendly
+        # in-place model rewrites (e.g. sliding-window mask patches). Querying
+        # the loader property keeps the inputs dict free of non-model keys.
+        if getattr(self.dynamic_loader.loader, "requires_model_rewrites", False):
```
kmabeeTT, line 154: Similar comment in the tt-forge-models PR, in the future if we get more types of re-writes we may need to consider making this more fine grained, but this is okay for now.

examples/pytorch/gpt_oss_20b.py
```diff
@@ -244,6 +249,7 @@ def construct_inputs(
         "input_ids": inputs.input_ids,
         "past_key_values": static_cache,
         "cache_position": cache_position,
+        "position_ids": position_ids,
```
kmabeeTT, line 252: I think if all these position_ids changes are needed (?) by the transformers 5.5 uplift and the transformers 5.5 uplift broke the example tests then those changes should go in via seperate PR, and not bundled with the smaller sliding attention override changes... if this is the case, can you seperate them please?

kmabeeTT (commented):
_tt_apply_model_rewrites comment in PR Desc looks outdated, plz adjust - and I left a comment I think it's better to fix the transformers 5.5 uplift breakages first in seperate PR before merging the sliding window changes in tt-forge-models or here. If it is a real breakage can you please put up PR for that seperately first?

examples/pytorch/gpt_oss_20b.py
devisettymahidhar608: Sure @kmabeeTT , I'll raise a separate PR to address the errors caused by the Transformers 5.5 uplift. In this PR, I'll include only the changes related to sliding attention.

kmabeeTT: thank you!

kmabeeTT (approved):
Thx for trimming down - one problem is `examples/pytorch/mistral_8b.py` was recently added and refers to deleted function `override_ministral_sliding_window_causal_mask` so you'll need to update that example test too please before merge, otherwise looks good.

devisettymahidhar608:
> Thx for trimming down - one problem is examples/pytorch/mistral_8b.py was recently added and refers to deleted function override_ministral_sliding_window_causal_mask so you'll need to update that example test too please before merge, otherwise looks good.

Sorry, I forgot to add it earlier. I've added it now.
Thank You

jameszianxuTT (approved):

Issue #4761, opened 2026-05-18
[HiDream-I1-Fast] OOM in transformer


kamalrajkannan78:
### Describe the bug

- Transformers fails due to `Out of Memory: Not enough space to allocate 5368709120 B DRAM buffer across 12 banks, where each bank needs to store 447393792 B, but bank size is 1071821792 B (allocated: 726763712 B, free: 345058080 B, largest free block: 340146592 B)` - [may22_hidream_transformer_4_chips.log](https://github.com/user-attachments/files/28179550/may22_hidream_transformer_4_chips.log)

### Root cause analysis (22/05/2026)

- On bfp16, Same OOM error persists - [may22_hidream_transformer_4_chips_bfp16.log](https://github.com/user-attachments/files/28179564/may22_hidream_transformer_4_chips_bfp16.log)
- On 8 chips, model fails with ` error: ShardyToStableHLO lowering for CollectivePermuteOp is not implemented yet ` & `'sdy.collective_permute' op requires the same type for all operands and results` - [may22_hidream_transformer_8_chips.log](https://github.com/user-attachments/files/28179606/may22_hidream_transformer_8_chips.log)
- During the investigation, memory details were not getting printed even after enabling the memory logging flags and using a debug build. - https://github.com/tenstorrent/tt-xla/issues/4887
- Claude suspected OOM  triggered by this [multiplication](https://github.com/huggingface/diffusers/blob/ad3a3afc3a4d3068bbb12f58129c855087ffc6d6/src/diffusers/models/transformers/transformer_hidream_image.py#L134) operation 
- But manual bisect reveals OOM exists even before it - [may22_hidream_transformer_4_d1.log](https://github.com/user-attachments/files/28179916/may22_hidream_transformer_4_d1.log), [transformer_hidream_image.py](https://github.com/user-attachments/files/28179917/transformer_hidream_image.py)

- It's from this [reshape](https://github.com/huggingface/diffusers/blob/ad3a3afc3a4d3068bbb12f58129c855087ffc6d6/src/diffusers/models/transformers/transformer_hidream_image.py#L132) operation 
  - Return tensor before reshape - [may22_hidream_transformer_4_d3.log](https://github.com/user-attachments/files/28179936/may22_hidream_transformer_4_d3.log), [transformer_hidream_image.py](https://github.com/user-attachments/files/28179937/transformer_hidream_image.py)
[...]

PR #4985, opened 2026-05-28
[Composite] Add nn.RMSNorm module-form support

kamalrajkannan78:
### Ticket

- https://github.com/tenstorrent/tt-xla/issues/4761

### Problem description

- HiDream's transformer OOMs at `multiply.508`. Root cause is TTNN's `tile<32x32>` layout being applied to a 6D tensor whose two innermost dims are `1` and `2`. Each sub-tile dim pads up to a full 32, producing a **512× memory blowup** (10 MB logical → 5 GB padded). The multiply runs at this bloated 6D shape because the source-level `mul → reshape(...,-1,1,2)` gets rewritten during compilation to `reshape → mul`, leaving two 5 GB tensors live at once.
- `nn.RMSNorm` was being decomposed into primitives, so its tail-multiply was visible to that rewrite. Wrapping it as a `tenstorrent.rms_norm` composite hides the tail-multiply inside the composite — it can no longer be pulled out across the reshape.
- Full root-cause analysis: [#4761](https://github.com/tenstorrent/tt-xla/issues/4761)

### What's changed

- Added `nn.RMSNorm` module-form dispatch in `composite_ops.py` (mirrors existing `replace_layer_norm_module` / `replace_group_norm_module`). Functional form was already wired up in [#2213](https://github.com/tenstorrent/tt-xla/pull/2213).
- Added sanity test `test_patched_rms_norm_module` in `tests/torch/ops/test_composite_ops.py` (parametrized over `elementwise_affine` and 4 input shapes; fp32, matching the existing RMSNorm tests in the file).

### Follow-up

- After this fix, HiDream OOM moves from `multiply.508` to the downstream `concatenate.524` site. Will be addressed in a separate PR via the layout-swap approach (concat accepts row-major inputs).

### Checklist
- [x] PCC verified on both dtypes via the new sanity test.
- [x] Verified locally — composite now fires (24× `tenstorrent.rms_norm` in IR vs 0 before) in model run

### Logs

- [may29_rms_composite_cases.log](https://github.com/user-attachments/files/28399091/may29_rms_composite_cases.log)
- [may28_hidream_transformer_4_chips_before_fix.log](https://github.com/user-attachments/files/28346827/may28_hidream_transformer_4_chips.log)
- [may28_hidream_transformer_4_chips_after_fix.log](https://github.com/user-attachments/files/28346826/may28_hidream_transformer_4_chips_rms_composite.log)

### DIFF
--- a/tests/torch/ops/test_composite_ops.py
+++ b/tests/torch/ops/test_composite_ops.py
@@ -196,6 +196,37 @@ def test_composite_rms_norm(use_weight, batch_size, seq_len, hidden_size):
 @pytest.mark.single_device
 @pytest.mark.parametrize("elementwise_affine", [True, False])
 @pytest.mark.parametrize(
+    "batch_size, seq_len, hidden_size",
+    [(1, 32, 32), (1, 128, 768), (1, 1024, 768), (1, 4096, 2560)],
+)
+def test_patched_rms_norm_module(elementwise_affine, batch_size, seq_len, hidden_size):
+    class RMSNormModel(torch.nn.Module):
+        def __init__(self, hidden_size):
+            super().__init__()
+            self.rms = nn.RMSNorm(hidden_size, elementwise_affine=elementwise_affine)
+
+        def forward(self, x):
+            return self.rms(x)
+
+    options = {"tt_enable_composite_ops": True}
+
+    input_tensor = torch.randn(batch_size, seq_len, hidden_size)
+
+    model = RMSNormModel(hidden_size)
+
+    run_graph_test(
+        model,
+        [input_tensor],
+        comparison_config=ComparisonConfig(),
+        framework=Framework.TORCH,
+        torch_options=options,
+    )
+
+
+@pytest.mark.nightly
+@pytest.mark.single_device
+@pytest.mark.parametrize("elementwise_affine", [True, False])
+@pytest.mark.parametrize(
     "batch_size, seq_len, embedding_dim",
     [(1, 32, 32), (1, 197, 768), (1, 1024, 768)],
 )
--- a/python_package/tt_torch/composite_ops.py
+++ b/python_package/tt_torch/composite_ops.py
@@ -409,6 +409,52 @@ def replace_group_norm_module(
     gm.graph.erase_node(node)
 
 
+def replace_rms_norm_module(
+    gm: torch.fx.GraphModule,
+    node: torch.fx.Node,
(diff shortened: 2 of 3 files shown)


tests/torch/graphs/fusion_patterns/test_rmsnorm_module.py
```diff
@@ -0,0 +1,43 @@
...
+@pytest.mark.nightly
+@pytest.mark.single_device
+@pytest.mark.record_test_properties(category=Category.GRAPH_TEST)
+@pytest.mark.filecheck(["rms_norm.ttir.mlir"])
+@pytest.mark.parametrize(
+    "dtype",
+    [torch.float32, torch.bfloat16],
+    ids=["f32", "bf16"],
+)
+def test_rmsnorm_module(dtype, request):
```
mstojkovicTT: Since this PR is adding composite support for `nn.RMSNorm` i would add this test in `tests/torch/ops/test_composite_ops.py` instead (there are examples of other nn modules being tested there)

kamalrajkannan78: moved sanity to `tests/torch/ops/test_composite_ops.py` and updated the logs in the description

tests/torch/ops/test_composite_ops.py
```diff
@@ -192,6 +192,32 @@ def forward(self, x, weight):
         )
 
 
+@pytest.mark.nightly
+@pytest.mark.single_device
+@pytest.mark.parametrize("dtype", [torch.float32, torch.bfloat16], ids=["f32", "bf16"])
```
mstojkovicTT: Instead of doing parametrize over `dtype`, can you add more shapes of input tensor, and also add `elementwise_affine` parameter, like `patched_layer_norm` does

mstojkovicTT (approved):
Just a small nit, looks good

tests/torch/ops/test_composite_ops.py
kamalrajkannan78: Done. Initially I had the case present in my model. Added more shape cases now; also added `elementwise_affine`. All use fp32 (matches my model and the existing RMSNorm tests in this file).
[may29_rms_composite_cases.log](https://github.com/user-attachments/files/28399205/may29_rms_composite_cases.log)

sdjukicTT (approved):

Issue #3667, opened 2026-03-10
Move device specific reshape ops to mlir


sshonTT:
Discussion : https://github.com/tenstorrent/tt-xla/pull/3635#discussion_r2912231630

The frontend should remain agnostic of ttnn restrictions. I plan to remove these constraints from the frontend and move the logic to the ttir -> ttnn pass instead. This ensures a cleaner separation of concerns between the frontend and the backend-specific implementation.

PR #4988, opened 2026-05-28
MoE backend from huggingface

sshonTT:
### Ticket
N/A

### Problem description
@sgligorijevicTT [introduced](https://github.com/tenstorrent/tt-xla/pull/4424) new MoE backend integration logic from HF, giving us a more generalized way to support MoE on our stack. The previous implementation was tightly coupled to our internal API.

### What's changed
**New HF backend modules (`tt_torch`)**
- `moe_backend.py`: registers `tt_moe` (multi-chip EP) and `tt_dense` (dense bmm) as HF `ExpertsInterface` backends.
- `attention_backend.py`: registers `tt_sdpa` as an HF attention backend.
- `__init__.py`: auto-registers on import; lazy exports.

**Legacy monkey patch removal**
- `torch_overrides.py`: drop `GptOssMLP/GptOssExperts/GptOssTopKRouter.forward`  overrides — they bypassed HF's `@use_experts_implementation` dispatcher,  so `tt_moe` could not take effect on GPT-OSS.

**Bug fixes exposed by removing the monkey patch**
- `sparse_mlp.py`: `enable_sparse_mlp(mlp, ...)` was a no-op for top-level MoE input (`setattr(parent, "", x)`). Return the replacement.
- `custom_ops.py` (sparse_matmul):
  - `split_bd` was only correct for `S == 1`; restrict it.
  - Add flat-tiling fallback for `BD < M` and `S < M`.
  - CPU gate-up returns 5D `[A, B, E, M, N]` to match XLA.

**Tests / examples**
- `test_gpt_oss_mlp` parametrised over `tt_dense` / `tt_moe`.
- New `tests/torch/moe/test_hf_backends.py`.
- Benchmarks plumb `experts_implementation`; GPT-OSS TP switched to `tt_dense`.
- New OLMoE / Trinity-Nano TP+EP demos.

### Checklist

### DIFF
--- a/python_package/tt_torch/moe_backend.py
+++ b/python_package/tt_torch/moe_backend.py
@@ -1,0 +1,577 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Tenstorrent MoE experts backend for HuggingFace transformers.
+
+Registers two ``ExpertsInterface`` backends selectable via
+``experts_implementation=`` at ``from_pretrained`` time:
+
+  - ``tt_moe``   — multi-chip EP via all_to_all_dispatch / sparse_matmul / all_to_all_combine.
+  - ``tt_dense`` — dense bmm across all experts, single-device-friendly.
+
+Works with any ``@use_experts_implementation`` Experts module that exposes
+``gate_up_proj``, ``down_proj``, and ``_apply_gate``.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Callable, Dict, Optional, Tuple
+
+import torch
+import torch.nn as nn
+from torch.nn import functional as F
+from transformers.integrations.moe import ALL_EXPERTS_FUNCTIONS, ExpertsInterface
+from transformers.modeling_utils import PreTrainedModel
+
+# Ensure torch.ops.tt.* are registered.
+from . import custom_ops  # noqa: F401
+
+TT_MOE_BACKEND_NAME = "tt_moe"
+TT_DENSE_EXPERTS_BACKEND_NAME = "tt_dense"
+REDUCTION_SIZE = 32
+
+# HF built-in backend keys — patched validator falls through for these.
+_HF_BUILTIN_EXPERTS_KEYS = frozenset({"eager", "grouped_mm", "batched_mm", "deepgemm"})
+
+# Module-level EP config; set by register_tt_moe_backend().
+_config: dict = {"cluster_axis": None}
+
+
+def _resolve_cluster_axis(mesh: Any) -> int:
+    configured_axis = _config["cluster_axis"]
+    if configured_axis is not None:
+        return int(configured_axis)
+
+    for axis, size in enumerate(tuple(int(d) for d in mesh.mesh_shape)):
+        if size > 1:
+            return axis
+    return 0
+
+
+def _mesh_info() -> Tuple[int, int, Tuple[int, ...], int]:
+    """Return (total_devices, dispatch_devices_on_cluster_axis, mesh_shape, axis).
+
+    Reads the currently-set torch_xla global SPMD mesh. Returns (1, 1, (1,), 0)
+    when no mesh is registered or torch_xla is unavailable.
+    """
+    try:
+        from torch_xla.distributed.spmd import get_global_mesh
+    except ImportError:
+        return 1, 1, (1,), 0
+    mesh = get_global_mesh()
+    if mesh is None:
+        return 1, 1, (1,), 0
+    mesh_shape = tuple(int(d) for d in mesh.mesh_shape)
+    total = 1
+    for d in mesh_shape:
+        total *= d
+    ax = _resolve_cluster_axis(mesh)
+    dispatch = mesh_shape[ax] if 0 <= ax < len(mesh_shape) else 1
+    return total, dispatch, mesh_shape, ax
+
+
+def _as_sparse_matmul_weight(weight: torch.Tensor, is_transposed: bool) -> torch.Tensor:
+    """Reshape weight to `[1, E, in, out]` for `sparse_matmul`."""
+    if is_transposed:
+        return weight.unsqueeze(0).contiguous()
+    return weight.transpose(-2, -1).unsqueeze(0).contiguous()
+
+
+class _ExpertAdapter:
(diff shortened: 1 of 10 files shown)


AleksKnezevic (approved):

python_package/tt_torch/moe_backend.py
```diff
@@ -0,0 +1,577 @@
...
+    device: torch.device,
+) -> torch.Tensor:
+    """Expand top-k weights into a full `[T, E]` sparse router-scores tensor."""
+    one_hot = (
+        top_k_index.unsqueeze(-1) == torch.arange(num_experts, device=device)
+    ).to(dtype)
+    return torch.einsum("tk,tke->te", top_k_weights.to(dtype), one_hot)
+
+
+def _pad_moe_inputs(
```
odjuricicTT, line 241: Is it possible to have this as a ttnn ir level workaround for sparse matmul instead?

python_package/tt_torch/moe_backend.py
```diff
@@ -0,0 +1,577 @@
...
+    """Build `[1, 1, E, D]` one-hot expert-to-device mapping."""
+    if num_experts % num_devices != 0:
+        raise ValueError(
+            f"num_experts ({num_experts}) must be divisible by num_devices "
+            f"({num_devices}) to build a one-hot expert-to-device mapping."
+        )
+
+    experts_per_device = num_experts // num_devices
+    device_ids = (
+        torch.arange(num_experts, device=device, dtype=torch.int64)
```
odjuricicTT, line 216: If i remember correctly the moe specific ttnn ops will require uint16. Is it possible for use to use that here form the start?

python_package/tt_torch/moe_backend.py
```diff
@@ -0,0 +1,577 @@
...
+        and hasattr(module, "down_proj")
+    ):
+        return _SeparateGateUpAdapter(module)
+    raise RuntimeError(
+        f"tt_moe backend could not adapt Experts module {type(module).__name__}. "
+        "Expected fused gate_up/down or separate gate/up/down experts."
+    )
+
+
+def _expert_mapping(
```
odjuricicTT, line 202: So this assumes EP=num_devices?

For my understanding, is this validated somewhere with the provided expert weight sharding?

python_package/tt_torch/moe_backend.py
odjuricicTT: Casting these from int -> uint later on causes a hell of a lot of performance problems like having to go from / to device.

python_package/tt_torch/moe_backend.py
```diff
@@ -0,0 +1,577 @@
...
+    # Sparsity from all-gathered metadata (not local router scores).
+    _, sparsity = torch.ops.tt.moe_expert_token_remap(
+        routing_scores,
+        expert_mapping,
+        metadata,
+        num_devices=dispatch_devices,
+    )  # sparsity: [1, 1, ceil(BD*T/32), E]
+
+    if experts.has_fused_gate_up:
+        gate_up_out = torch.ops.tt.sparse_matmul(
```
odjuricicTT, line 339: This is one way of implementing the compute part of the moe block. As you know, in ttnn there are ops like moe_gpt and moe_compute that do the same math but for only work for a handful of cases. We want to be able to use both implementations.

So what i'm wondering is what do you think about wraping this in some sort of torch.ops.tt.moe_compute composite op (with sparse matmul implementation body, maybe?) and lowering this that way all the way to TTNN IR?
Then on ttnn ir level we have the option of using the constraint check for different possible ops for the composite and deciding on the best decomposition (e.g. either moe_compute or 2/3x sparse matmul etc)

Not a blocker for this PR, discussion for a followup.

cc: @sdjordjevicTT @mtopalovicTT @dmilinkovicTT

odjuricicTT (approved):
Looks good. I mostly left questions on the PR.

Also @sshonTT can provide a ttnn ir example for a MoE layer for some llm using the expert path?
All benchmarks are still on dense as i see it.

python_package/tt_torch/moe_backend.py
sshonTT: Yeah I've worked on this since a2asparse introduced : https://github.com/tenstorrent/tt-xla/issues/3667.
It's now paused to we are converting our moe logic to use new `moe_compute` op. Now that the scope of this commit is mainly on using hf logic, I can follow the issue.

python_package/tt_torch/moe_backend.py
sshonTT: torch.arange doesn't support torch.uint16 dtype, so I can do something like
```
device_ids = (
        torch.arange(num_experts, device=device, dtype=torch.int64).to(torch.uint16)
```

I am wondering they will be folded when lowering it. Let me try it while I generate ttnn graph.

python_package/tt_torch/moe_backend.py
sshonTT: Yes, it assumes EP = total mesh devices, i.e. experts compound-sharded across the full (batch, model) mesh, which is what get_tt_moe_shard_specs produces.

python_package/tt_torch/moe_backend.py
sshonTT: I have this on the roadmap but nothing concrete yet, it's gated on the ongoing moe_compute migration. Happy to have a discussion on this.

python_package/tt_torch/moe_backend.py
odjuricicTT: Please do try. Since we have control over the torch implementation here we should try to form in in a way that uses uint16 the most. And results in the least int --> uint typecasts on ttnn level.

python_package/tt_torch/moe_backend.py
odjuricicTT: Great! Since moe_compute is still a WIP. Let's proceed with a prototype for this with moe_gpt and discuss on that PR.

python_package/tt_torch/moe_backend.py
sshonTT: Yeah it works, there's no uint16 conversion round-trip in graph except const-eval one. Thanks for suggesting!

sshonTT:
@odjuricicTT [Here](https://gist.github.com/sshonTT/a734333ce05fc13a815dd962f063d741)'s single layer graph from gpt oss 20b.

Issue #4710, opened 2026-05-14
[playground-v2.5-1024px-aesthetic] OOM in VAE decoder


kamalrajkannan78:
### Describe the bug

- VAE decoder fails with `Out of Memory: Not enough space to allocate 4294967296 B DRAM buffer across 12 banks, where each bank needs to store 357916672 B, but bank size is 1071821792 B ` - [may14_vae_decoder_fp32.log](https://github.com/user-attachments/files/27772341/may14_vae_decoder_fp32.log)

### Root cause analysis (21/05/2026)

- On current main, the model fails due to OOM in both BFP16 & FP32
  - [may21_playground_v2_5_vae_decoder_bfp16.log](https://github.com/user-attachments/files/28106576/may21_playground_v2_5_vae_decoder_bfp16_r.log)
  - [may21_playground_v2_5_vae_decoder_fp32.log](https://github.com/user-attachments/files/28106577/may21_playground_v2_5_vae_decoder_fp32.log)
- All subsequent experiments use FP32.
- The issue originates from the **decomposition of GroupNorm**, specifically [`hidden_states = self.norm1(hidden_states)`](https://github.com/huggingface/diffusers/blob/ad3a3afc3a4d3068bbb12f58129c855087ffc6d6/src/diffusers/models/resnet.py#L326) when invoked from `vae.decoder.up_blocks.3.resnets.0`. The bisect experiments below confirm it:
  - **Skip** `up_blocks[3].resnets[0].norm1` (and everything after) — **PASSED** - [d1.zip](https://github.com/user-attachments/files/28107203/d1.zip)
  - Run **only** `up_blocks[3].resnets[0].norm1` (skip the rest of `up_blocks[3]` and the post-process) — **same OOM** - [d2.zip](https://github.com/user-attachments/files/28107254/d2.zip)
- A standalone test of the same `GroupNorm(32, 256)` on `(1, 256, 1024, 1024)` fp32 on a clean device **passed** without OOM — [may21_gn.log](https://github.com/user-attachments/files/28107449/may21_gn.log), [test_group_norm.py](https://github.com/user-attachments/files/28107450/test_group_norm.py). This is the key data point: *the op alone fits; it doesn't fit alongside the rest of the decoder's live working set.*
[...]

PR #4992, opened 2026-05-29
[Playground v2.5] Add end-to-end pipeline example

### LINKED ISSUE #4987: [Playground v2.5] Bringup E2E pipeline demo

kamalrajkannan78:
### Ticket

- closes https://github.com/tenstorrent/tt-xla/issues/4987

### Problem description

- All components in the model are passing now. E2E pipeline example can be added.

### What's changed

- End-to-end Playground v2.5 pipeline example on N150. All four nn modules (text_encoder, text_encoder_2, unet, vae) run on TT; scheduler / CFG / latent prep stay on CPU.
- Sequential evict: each component moves to TT right before its forward and back to CPU after, so only one model is resident at a time.
- UNet runs in bf16 on TT (fp32 OOMs at step 2 of the denoising loop); other components stay fp32.
- VAE compiles with `optimization_level=1` to keep ttnn.group_norm fused ([#4710](https://github.com/tenstorrent/tt-xla/issues/4710)).

### Validation
- CPU-only mirror of the pipeline produces **pixel-identical** output vs `diffusers.StableDiffusionXLPipeline` (same seed, prompt, steps).
- All components (incl. UNet across 50 denoising steps) achieve PCC > 0.99 vs CPU fp32 golden on real inputs.

### Checklist

### Logs

- [playground_v2_5_logs.zip](https://github.com/user-attachments/files/28385971/playground_v2_5_logs.zip)

### Outputs

**Original CPU output (fp32)**

astronaut_org_pipeline

**TT output (UNet bf16, others fp32)**

playground_v2_5_output

### DIFF
--- a/examples/pytorch/playground_v2_5_pipeline.py
+++ b/examples/pytorch/playground_v2_5_pipeline.py
@@ -1,0 +1,415 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Playground v2.5 — end-to-end TT pipeline example.
+
+Each nn.Module component (text_encoder, text_encoder_2, unet, vae) can be
+independently moved to Tenstorrent via
+`model.compile(backend="tt") + model.to(xla_device())`. Tokenizer and
+scheduler always stay on CPU. CPU→TT→CPU device switching is done inline at
+the call site of each nn component.
+"""
+
+from pathlib import Path
+from typing import Optional
+
+import torch
+import torch_xla
+import torch_xla.core.xla_model as xm
+import torch_xla.runtime as xr
+from diffusers import EDMDPMSolverMultistepScheduler
+from loguru import logger
+from PIL import Image
+from transformers import CLIPTokenizer
+
+from third_party.tt_forge_models.playground_v2_5.pytorch import (
+    ModelLoader,
+    ModelVariant,
+)
+
+MODEL_ID = "playgroundai/playground-v2.5-1024px-aesthetic"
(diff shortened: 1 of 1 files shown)


ppadjinTT (commented):
You framed this as a test but it tests only compilation. not PCC. On the otherhand that would take a lot of time on CPU.

For now I suggest the following:
1. We should test each of the components of this pipeline separately with PCC (i believe you already did this)
2. Reframe this test as an example which should work with default args (you said it compiles on n150, so i assume it should work on any TT chip)

kamalrajkannan78:
@ppadjinTT 

Thanks for the review — both points addressed.

- moved to `examples/pytorch/playground_v2_5_pipeline.py`, dropped the pytest markers, kept the same structure as `sd_v1_4_pipeline.py`
- On **PCC validation**: agreed it wasn't in the push file. Per-component PCC is already covered by the four existing tests (test_text_encoder.py, test_text_encoder_2.py, test_unet.py, test_vae_decoder.py) for random inputs. I additionally ran a full e2e PCC sweep using a [test_playground_v2_5_pipeline.py](https://github.com/user-attachments/files/28396718/test_playground_v2_5_pipeline.py) (included in the attached playground_v2_5_logs.zip) for actual inputs— all components pass PCC > 0.99 vs CPU fp32 golden on real inputs, including the UNet across all 50 denoising steps. The script is kept in the zip rather than the pushed code since running CPU twins every nightly would add hours per run. [may29_playground_v2_5_tt_run_unet_bfp16.log.zip](https://github.com/user-attachments/files/28396713/may29_playground_v2_5_tt_run_unet_bfp16.log.zip)

ppadjinTT (approved):
lgtm!

sdjukicTT (approved):

PR #4995, opened 2026-05-29
[Test] Add host RSS leak test for repeated module execution

jasonmacTT:
### Ticket
Link to Github Issue

### Problem description
More tooling is needed to measure lifetimes and memory usage of stack.

### What's changed
Add a test for checking memory leakage in repeated calls to a module.
### Checklist

### DIFF
--- a/tests/torch/graphs/test_tensor_persistence.py
+++ b/tests/torch/graphs/test_tensor_persistence.py
@@ -18,6 +18,7 @@ import os
 import threading
 
 import numpy as np
+import psutil
 import pytest
 import torch
 import torch_xla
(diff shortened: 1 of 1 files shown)


tests/torch/graphs/test_tensor_persistence.py
```diff
@@ -806,3 +807,114 @@ def forward(self, a, b, c, d, e, f, g):
...
+
+    def get_vmhwm_mib() -> float:
+        """
+        Helper to get this process's lifetime peak RSS, in MiB.
+        """
+        with open("/proc/self/status") as f:
+            for line in f:
+                if line.startswith("VmHWM:"):
+                    return int(line.split()[1]) / 1024
+        return float("nan")
```
jameszianxuTT, line 876: Because this test will be potentially executed in the same process with other tests due to test infra details (unless run with pytest --forked), I think it would be good to calculate the initial VmHWM at test start and subtract it from all the subsequent readings to get an absolute baseline

tests/torch/graphs/test_tensor_persistence.py
```diff
@@ -806,3 +807,114 @@ def forward(self, a, b, c, d, e, f, g):
...
+
+    Asserts both steady drift (slope) and step jumps (absolute growth).
+    """
+
+    # Threshold values for assertion
+    MAX_RSS_SLOPE_MIB_PER_ITER = 1.0
+    MAX_RSS_GROWTH_MIB = 50.0
+    WARMUP_ITERS = 3
+
+    assert iters > WARMUP_ITERS + 3, "need a few samples after warmup to fit a slope"
```
jameszianxuTT: Why parametrize by iters at all if you will keep it at 15 for all runs? It seems like you could just place it as a constant inside the test

tests/torch/graphs/test_tensor_persistence.py
```diff
@@ -806,3 +807,114 @@ def forward(self, a, b, c, d, e, f, g):
...
+        (1024, 15),
+        (2048, 15),
+    ],
+)
+def test_no_rss_leak_across_executions(dim: int, iters: int, record_property):
+    """
+    Run the same matmul `iters` times, drop the first WARMUP_ITERS samples,
+    and assert that post-warmup RSS is stable.
+
+    Asserts both steady drift (slope) and step jumps (absolute growth).
```
jameszianxuTT: Can you clarify exactly what is being asserted about these two variables?

tests/torch/graphs/test_tensor_persistence.py
```diff
@@ -806,3 +807,114 @@ def forward(self, a, b, c, d, e, f, g):
...
+    for it, rss, vmhwm in samples:
+        tag = "W " if it < WARMUP_ITERS else "  "
+        print(f"  {tag}iter {it:3d}: rss={rss:8.1f} MiB  vmhwm={vmhwm:8.1f} MiB")
+    print(
+        f"  post-warmup slope:  {slope:+.3f} MiB/iter (threshold {MAX_RSS_SLOPE_MIB_PER_ITER})"
+    )
+    print(
+        f"  post-warmup growth: {growth:+.2f} MiB     (threshold {MAX_RSS_GROWTH_MIB})"
+    )
+    print(f"  checksum:  {checksum.item():.4f}")
```
jameszianxuTT: What is the purpose of this checksum? It seems like you could use it as an accuracy test (eg. checksum = torch.ones * iters) but it doesn't look like that's happening?

tests/torch/graphs/test_tensor_persistence.py
```diff
@@ -806,3 +807,149 @@ def forward(self, a, b, c, d, e, f, g):
...
+    xs_pts = [it for it, _, _ in measured]
+    ys_pts = [rss for _, rss, _ in measured]
+    slope = linear_slope(xs_pts, ys_pts)
+    growth = ys_pts[-1] - ys_pts[0]
+
+    record_property("rss_slope_mib_per_iter", round(slope, 4))
+    record_property("rss_growth_mib", round(growth, 3))
+    record_property("rss_first_post_warmup_mib", round(ys_pts[0], 3))
+    record_property("rss_last_post_warmup_mib", round(ys_pts[-1], 3))
+    record_property("vmhwm_final_mib", round(samples[-1][2], 3))
```
jameszianxuTT, line 922: Don't add record_property here, these are written to a test report with a statically defined schema.

tests/torch/graphs/test_tensor_persistence.py
```diff
@@ -806,3 +807,149 @@ def forward(self, a, b, c, d, e, f, g):
...
+          sample (last RSS - first RSS) to not exceed the threshold. Catches
+          a sudden step jump that a slope could average away.
+
+    Asserts every iteration's output sum matches iteration 0
+        - The run is deterministic, each iteration must produce the same output.
+    """
+
+    # Threshold values for assertion
+    MAX_RSS_SLOPE_MIB_PER_ITER = 1.0
+    MAX_RSS_GROWTH_MIB = 50.0
```
jameszianxuTT, line 850: Are these bounds constant no matter the size of the input tensor?

Shouldn't MAX_RSS_GROWTH_MIB be dependent on the number of iterations you run? eg. a small leak over a large number of iterations may be acceptable, but  a large leak over a small number of iterations is not. It seems to me like the max growth should either be normalized to `(final - initial)/num_iter`, or the metric should be maximum change between any two samples.

Also 50MB is quite a bit, considering the size of your inputs -- 2048x2048x2 (bf16) is only 8M.

tests/torch/graphs/test_tensor_persistence.py
```diff
@@ -806,3 +807,149 @@ def forward(self, a, b, c, d, e, f, g):
...
+
+
+@pytest.mark.push
+@pytest.mark.single_device
+@pytest.mark.parametrize(
+    "dim",
+    [
+        512,
+        1024,
+        2048,
```
jameszianxuTT, line 819: Can we try some larger tensor sizes too? Maybe up to 32K if that doesn't introduce issues? Realistically we see much larger tensors in some models.

tests/torch/graphs/test_tensor_persistence.py
jasonmacTT: I think maximum change between any two samples sounds reasonable instead of collecting overall growth.
I could assert on something like for some chosen delta between two samples, ```max_change <= THRESHOLD```.
Unless you were thinking something different. 

I still think that the threshold values should be constant independent of the input size. 
My assumption before doing any iterations is that the found slope across all iterations should, in principle, more or less be zero, regardless of input size. Asserting the slope against 1.0 is a value that I don't think would make sense to scale with the input size, unless you want to scale it inversely proportional to input size.

If by asserting on ```max_change <= THRESHOLD``` as mentioned, I am on the fence to scale that as well and should just be some chosen value prior to execution. My thinking is that a leak of, for example, 20.0M is bad regardless of input size. If the 20.0M leak persists across iterations and occurs each time, then at 25 iterations we have already leaked 500M. Although, I do agree that a leak of 20.0M is unrealistic for input size of 8M.

PR #5004, opened 2026-05-29
[vllm] Sharded topk sampling for 2D mesh TP inference

### LINKED ISSUE #4494: move topk chunking for sampling into tt-mlir

alinakhanTT:
### Ticket
closes #4494

### Problem description
vLLM sampling on multi-chip TP currently all-gathers the full vocab logits (~128k tokens) across devices before running topk. This is expensive, communicating the full vocab per decode step.

### What's changed

Depends on tt-mlir PR [tenstorrent/tt-mlir#8601](https://github.com/tenstorrent/tt-mlir/pull/8601).

For the 2D mesh TP case (`is_sharded_compute_logits=True`):

1. **`compute_logits`** :stop all-gathering vocab before sampling. The logits leave this graph vocab-sharded.

2. **`sample_from_logits`** : annotate incoming logits as `(None, "model")` so the compiler knows they're vocab-sharded (needed because this is a separately compiled graph).

3. **Greedy argmax path** : replace `torch.argmax` with `composite_topk(k=1)` when sharded. `torch.argmax` on a vocab-sharded tensor returns a local index; routing through `tenstorrent.topk` composite lets tt-mlir's `kCompositesWithCustomSharding` handle the distributed reduction and return the correct global token index.

### Checklist
- [x] `compute_logits` no longer all-gathers for 2D mesh TP
- [x] `sample_from_logits` annotates vocab-sharded logits at entry
- [x] Greedy path uses `composite_topk(k=1)` when sharded
- [x] perf benchmark run green - https://github.com/tenstorrent/tt-xla/actions/runs/29757678343

### Validation

End-to-end with tt-mlir PR #8601: [Run #26830482012](https://github.com/tenstorrent/tt-xla/actions/runs/26830482012) - `vllm-model-tests.json` against `mlir_override=e88509dd8` (tt-mlir PR commits on top of pinned SHA). All matrix jobs passed: n150 / p150 single_device, n300 dual_chip TP, n300-llmbox 4-chip 2D mesh TP.

Both code paths exercised in the same run:
- `use_2d_mesh=True` (new sharded path) - pass on Llama-3.2-3B, Qwen3-0.6B, e5-mistral-7b
- `use_2d_mesh=False` (old path, regression check) - pass on same models

Behavioral correctness checks:
- `test_greedy_determinism` - identical output across 3 runs (rules out gather races)
- `test_output_coherence` - sensible completions (`1 + 1 = 2`, `Roses are red, violets are blue`), confirms topk returns correct global vocab indices, not shard-local

### Follow up
Sharded composite extension to `gather_logprobs` tracked separately in  [issue tt-xla#5137](https://github.com/tenstorrent/tt-xla/issues/5137).


integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2390,7 +2393,12 @@ def sample_from_logits(
             and sampling_metadata.no_min_tokens
             and sampling_metadata.no_generators
         ):
-            out_tokens = torch.argmax(logits, dim=-1, keepdim=True)
+            if self.is_sharded_compute_logits and self.use_2d_mesh:
+                _, out_tokens = composite_topk(
+                    logits, k=1, dim=-1, largest=True, sorted=False
+                )
+            else:
+                out_tokens = torch.argmax(logits, dim=-1, keepdim=True)
```
kmabeeTT: This makes the greedy branch shard-aware, but the `else` branch still hands the now-vocab-sharded logits to `self.sampler`, and `chunked_topk_candidates` (the exact code #4494 is about) reduces over the sharded vocab dim with plain ops — so it returns shard-local candidates, i.e. wrong tokens. Since non-greedy is supported on TP, this is the path the ticket wanted fixed, not greedy. Can we migrate `chunked_topk_candidates` to keep logits split and do the topk via the sharding-aware composite here? Otherwise the non-greedy path regresses (previously `compute_logits` all-gathered, so the sampler saw full vocab).

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2369,7 +2370,7 @@ def compute_logits(self, sample_hidden_states: torch.Tensor) -> torch.Tensor:
         # (quant_method.apply bypasses __call__) and all_gather is a
         # no-op (world_size=1). Must be inside the compiled graph —
         # external sharding_constraint between compiled functions breaks.
-        if self.enable_tensor_parallel and self.is_sharded_compute_logits:
+        if self.enable_tensor_parallel and not self.use_2d_mesh:
```
kmabeeTT: The block comment above still explains the old is_sharded_compute_logits / "all_gather is a no-op (world_size=1)" rationale, but the condition now keys off not self.use_2d_mesh. Suggested update:

      # On 1D-mesh TP, replicate logits so the downstream sampler sees the
      # full vocab (the ParallelLMHead output is otherwise vocab-sharded).
      # On 2D-mesh TP we deliberately skip this: logits stay vocab-sharded
      # and the greedy reduction is done sharding-aware via composite_topk(k=1)
      # later (non-greedy/chunked-topk still TODO per #4494), avoiding an
      # expensive full-vocab all-gather every decode step.
      # Must be inside the compiled graph — an external sharding_constraint
      # between compiled functions breaks.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2381,6 +2382,8 @@ def sample_from_logits(
         Sample with xla-friendly function. This function is to be traced
         separately from `forward` for lighter compilation overhead.
         """
+        if self.is_sharded_compute_logits and self.use_2d_mesh:
+            logits = sharding_constraint_tensor(logits, self.mesh, (None, "model"))
```
kmabeeTT, line 4617: Nice — re-annotating `(None, "model")` at the entry of this separately-compiled graph makes sense. Worth a one-line comment on why (logits enter vocab-sharded only on 2D mesh, and the sharding doesn't carry across the compiled-graph boundary).

kmabeeTT (changes requested):
Thx for working on this! This is marked `closes #4494`, but that ticket is specifically about the non-greedy `chunked_topk_candidates` path (all-gather full vocab → re-slice into 32k chunks), with the ask being to keep logits split and reduce in tt-mlir so only the tiny top-k tensors get gathered. This PR removes the all-gather in `compute_logits` and migrates the *greedy* argmax to `composite_topk(k=1)`, but `chunked_topk_candidates` in `Sampler.sample()` is untouched. Net effect: the non-greedy path (the ticket's actual subject) now receives vocab-sharded logits and reduces them per-shard → wrong tokens.

So this looks like a prerequisite step rather than a fix for #4494... Can we either migrate `chunked_topk_candidates` to the sharding-aware composite in this PR (preferred, the main goal), or drop the `closes #4494` and track the chunked-topk migration as the real follow-up?

Curious also what TP testing was done, what perf improvements were seen, and whether greedy (temperature=0) or non-greedy sampling (temperature!=0) was used.

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Added!! comment now explains that logits enter vocab-sharded only on 2D mesh and sharding doesn't carry across the compiled-graph boundary from compute_logits

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Updated! dropped the "non-greedy still TODO" since chunked_topk_candidates is now also sharding-aware. Comment is now: "On 2D-mesh TP we deliberately skip this: logits stay vocab-sharded and the topk reductions are done sharding-aware via composite_topk later, avoiding an expensive full-vocab all-gather every decode step"

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Done!! chunked_topk_candidates now takes a vocab_sharded kwarg. When True it calls composite_topk(k=128) directly instead of the chunk-splitting path, so only the 128 merged candidates cross the wire rather than the full vocab. The flag is threaded from model_runner.sample_from_logits -> sampler.forward -> sample -> chunked_topk_candidates and greedy_sample. Tested with test_tensor_parallel_generation_llmbox_small[True-Qwen/Qwen3-0.6B] (temperature=0.8, non-greedy) passing on 2D mesh

alinakhanTT:
Testing: test_tensor_parallel_generation_llmbox_small[True-Qwen/Qwen3-0.6B] passes. Qwen3-0.6B on 2D mesh, temperature=0.8, top_p=0.95 (non-greedy), output coherent! Both paths are now covered: greedy uses composite_topk(k=1) as before, non-greedy uses composite_topk(k=128) in chunked_topk_candidates. The perf win is that the full-vocab all-gather (~152k tokens × batch) is replaced by an all-gather of 128 candidates per row, for 4 shards that's ~1200x less data on the wire per decode step. Haven't run a full Tracy perf capture yet with the non-greedy path

third_party/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 option(USE_CUSTOM_TT_MLIR_VERSION "Flag to use TT_MLIR_VERSION set by the user" OFF)
 
 if (NOT DEFINED TT_MLIR_VERSION OR NOT USE_CUSTOM_TT_MLIR_VERSION)
-    set(TT_MLIR_VERSION "c5f398432a61100da79b7b9f2941130496092287")
+    set(TT_MLIR_VERSION "bf8cc1edfb4c385f283cbb8fbe5fbd33d35e4818")
```
nsumrakTT: Just a reminder to set back mlir commit sha to current uplift before merge.

alinakhanTT:
### Validation summary

Full CI run for this PR: [actions/runs/27171038543](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543) (head `cb16f0c51d`).

Every check on the PR page that exercises this change is green. The only red checks are two `gpt_oss_20b_tp*` perf benchmarks that hang on `main` nightlies as well (links at the bottom).

#### Cross repo validation against [tt-mlir#8601](https://github.com/tenstorrent/tt-mlir/pull/8601) (commit `bf8cc1edfb4c385f283cbb8fbe5fbd33d35e4818`)

* [TT-XLA mlir-uplift-qualification](https://github.com/tenstorrent/tt-xla/runs/80218816377) ✔
* [test-uplift-extended summary](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80218796701) ✔ across [fusion patterns](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215733178), [MLP](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215733209), [RMS norm](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215733198), [n300-llmbox forge_models_multichip](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215733196), [extended_models (3)](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215733220), [extended_models (4)](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215733222)

#### End to end vLLM coverage on hardware

* [vllm_llmbox_tests tensor_parallel on n300-llmbox 2D mesh](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215737074) ✔
* [vllm_n300_tests tensor_parallel dual_chip](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215737104) ✔
* [vllm_n150_tests single_device](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215737084) ✔
* [vllm_p150_tests single_device](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215737122) ✔

#### Sharded TP accuracy (the exact paths this PR changes)

* [llama_3_1_70b_tp_accuracy on n300-llmbox](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215726587) ✔ (42m19s)
* [llama_3_1_70b_tp_galaxy_accuracy on galaxy-wh-6u](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215726564) ✔ (30m23s)
* [llama_3_1_8b_instruct_tp_accuracy on n300-llmbox](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215726553) ✔ (26m13s)

#### Forge models push including 2D mesh multi host

* [test push and multi_host_cluster on n300-llmbox](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215727972) ✔
* [test push and training on qb2-blackhole](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215727989) ✔
* All [n150 / p150 forge_models matrix entries](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543) ✔

#### Builds and aux

* [Build manylinux wheel](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80210213040), [Build release](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80210213059), [Build debug with coverage](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80210199389) ✔
* [TT-XLA Unit Tests](https://github.com/tenstorrent/tt-xla/runs/80211394070), [basic-test](https://github.com/tenstorrent/tt-xla/runs/80220906095), [model-test-push](https://github.com/tenstorrent/tt-xla/runs/80220135220) ✔

### Failing checks are not caused by this PR

Only two jobs fail on the PR page, plus `check-all-green` which fails purely as a roll up of those two:

* [perf gpt_oss_20b_tp (n300-llmbox)](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215732344) hung at the 6 hour GitHub Actions wall
* [perf gpt_oss_20b_tp_d2m (n300-llmbox)](https://github.com/tenstorrent/tt-xla/actions/runs/27171038543/job/80215732326) hung at the 6 hour GitHub Actions wall

Both reach 100% weight loading and then go silent until cancellation. The same two jobs hang the same way on recent `main` nightly runs, with no code from this branch involved:

* Nightly [actions/runs/27109585912](https://github.com/tenstorrent/tt-xla/actions/runs/27109585912) (June 8 2026): [gpt_oss_20b_tp_d2m](https://github.com/tenstorrent/tt-xla/actions/runs/27109585912/job/80004782797) 6h0m42s, [gpt_oss_20b_tp](https://github.com/tenstorrent/tt-xla/actions/runs/27109585912/job/80004782820) 6h0m57s
* Nightly [actions/runs/27078082534](https://github.com/tenstorrent/tt-xla/actions/runs/27078082534) (June 7 2026): [gpt_oss_20b_tp](https://github.com/tenstorrent/tt-xla/actions/runs/27078082534/job/79918851798) 6h0m57s, [gpt_oss_20b_tp_d2m](https://github.com/tenstorrent/tt-xla/actions/runs/27078082534/job/79918851807) 6h1m53s
* Nightly [actions/runs/27047296946](https://github.com/tenstorrent/tt-xla/actions/runs/27047296946) (June 6 2026): same two jobs fail the same way

third_party/CMakeLists.txt
alinakhanTT: Will do! Plan: Once tt-mlir#8601 merges, wait for the next uplift to land on tt-xla `main`, then rebase this branch and drop my local override on `third_party/CMakeLists.txt:8` so it matches whatever main pins. Will confirm `git diff origin/main third_party/CMakeLists.txt` is empty before merging and note the new SHA in a comment here

vvukomanTT (approved):

alinakhanTT:
Note: `gather_logprobs` still all_gathers the full vocab whenever logprobs are requested. The same composite based decomposition used in this PR (local logsumexp + sharded index gather + sharded count) would apply there and yield a comparable per step wire reduction.
Left as a follow up for two reasons:
1. **Not the common decode case.** The logprobs requested path is conditional on a user requested API parameter, so production traffic with logprobs disabled is unaffected by the remaining all_gather.
2. **Substantial additional work.** Requires two new tt-mlir composites plus custom sharding rules (`tenstorrent.logsumexp` and `tenstorrent.sharded_index_gather`), each of roughly comparable compiler effort to the `tenstorrent.topk` work that landed in [tt-mlir#8601](https://github.com/tenstorrent/tt-mlir/pull/8601). Bundling it here would significantly expand the scope of this review.
Tracked separately in [#5137](https://github.com/tenstorrent/tt-xla/issues/5137).

.github/workflows/schedule-nightly.yml
```diff
@@ -3,6 +3,7 @@ name: On nightly
 on:
   schedule:
     - cron: '0 0 * * *'
+  workflow_dispatch:
```
nsumrakTT: Please don't add workflow_dispatch to the nightly.
You can always run nightly tests with "Run Tests" / "Performance benchmark" workflows which is identical to the nightly runs, except nightly include tests, perf, release, and reporting.

nsumrakTT (changes requested):

.github/workflows/schedule-nightly.yml
alinakhanTT: Done, dropped the commit that added workflow_dispatch. Force-pushed (HEAD now 865652872). Will use the Run Tests / Performance benchmark workflows for manual nightly-equivalent runs going forward!

nsumrakTT (approved):
Thank you!

kmabeeTT (approved):
Thanks for changes - looks good!

mmanzoorTT (approved):

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2947,7 +2960,14 @@ def _model_prefill_postprocess_compiled(
...
         selected_token_ids = self.sample_from_logits(logits, sampling_metadata)
-        return logits, selected_token_ids
+        # See _model_decode_compiled: drop the dead full-vocab logits output
+        # (and its all_gather) unless logprobs are requested.
+        logits_out = (
+            logits
+            if sampling_metadata.logprobs
+            else logits.new_zeros((logits.shape[0], 1))
+        )
+        return logits_out, selected_token_ids
```
jameszianxuTT, line 3976: Is this something it would make sense to warn the user about or would that result in spam? I am not sure if it would be obvious / noticeable to a normal user flow that setting `sampling_metadata.logprobs=False` would zero out the logits return path

alinakhanTT: I don't think we need a warning, would just be noise. logits_out is an internal graph tensor, it never leaves through vLLM's public output. Its only reader is execute_model (line 1990), which only touches it inside the logprobs-gated branch, the same flag that picks the placeholder. So when logprobs is off, nothing ever reads the zeros

pjrt_implementation/src/api/module_builder/module_builder.cc
```diff
@@ -783,12 +819,21 @@ tt_pjrt_status ModuleBuilder::runCompilerStableHLOPipeline(
...
+  //      value (e.g. a const-folded torch.arange) would otherwise default to
+  //      1x1, collapse the mesh, and break the live sharded buffers
+  //      (see https://github.com/tenstorrent/tt-xla/issues/5360).
+  // A purely replicated computation that merely has input arguments (e.g. a
+  // single-device argmax) matches neither and stays on a single device.
   if (current_mesh_shape.has_value() && current_mesh_shape->size() == 2 &&
-      !moduleHasAnyFuncArguments(mlir_module) && target_num_devices > 1) {
+      (moduleHasShardingConstraints(mlir_module) ||
+       anyShardedAcrossDevices(output_shardings) ||
+       (!moduleHasAnyFuncArguments(mlir_module) && target_num_devices > 1))) {
```
jameszianxuTT, line 847: Can you elaborate on the motivation for ~narrowing~ widening this condition?

jameszianxuTT: More for myself, the logic here evolved as:

1. originally, this condition was not here at all. This is not the normal way that the device mesh is inferred, which would normally come from the inputs
2. [before this PR] we found that some graphs had no inputs, and shlo pipeline would infer a degenerate 1x1 mesh, so this whole condition was a bypass/fallback path to add a mesh shape if it couldn't be inferred the normal way (i.e. from the inputs)
3.  [in this PR] we're expanding the fallback path to cover _outputs_ being sharded across devices OR if any sharding constraint is found

I am not sure in which case we have (3) but not (2) -- eg. are we observing graphs that _do_ have inputs from which device mesh cannot be extracted and falls back to 1x1 (not sure when this happens, maybe if they are all replicated?), but by adding awareness of sharding constraints + output shardings we can go towards the correct path of using the current_mesh_shape?

alinakhanTT: sample_from_logits is exactly the "(3) but not (2)" case that motivated this. It has function arguments, so the no-input branch (2) never applies, but its sharding lives in tt.sharding_constraint / @Sharding custom calls with mesh_idx_N placeholders, not in Shardy/GSPMD input annotations. So AnalyzeMeshPass has nothing to infer the mesh from the normal way, getMeshOps is empty, no GSPMD annotations exist, and it falls back to a degenerate 1×1 with the placeholders unresolved. Detecting the sharding constraint (or a device-sharded output) is what lets us pass current_mesh_shape into that fallback (AnalyzeMesh.cpp, the else if (!automaticArgAnalysis && meshShape.size() == 2) branch) so the mesh is built correctly

It's not quite "inputs are replicated", the mesh signal just lives in the constraint placeholders rather than input annotations. And branch (3) checks for a real constraint instead of "any func argument" because a purely-replicated argmax-with-input also has no annotations but genuinely belongs on one device, forcing it onto the mesh gave a Device count mismatch at runtime

jameszianxuTT: ok, sounds good. Thanks for clarifying!

jameszianxuTT (approved):

pjrt_implementation/src/api/module_builder/module_builder.cc
jameszianxuTT: Nit: I'd say the comment above this section is a bit misleading in light of this explanation (including before your change). It would suggest that this block is the "normal" path for passing mesh shape rather than a bypass path if AnalyzeMeshPass can't infer it from inputs. I suggest that it be rewritten for clarity.

jameszianxuTT: Thanks, that's much clearer!

vkovacevicTT (approved):

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -3522,9 +3589,26 @@ def sample_from_logits(
             and sampling_metadata.no_min_tokens
             and sampling_metadata.no_generators
         ):
-            out_tokens = torch.argmax(logits, dim=-1, keepdim=True)
+            if self.is_sharded_compute_logits:
+                # Greedy: gather logits to full and take a plain argmax. Faster
```
sshonTT: Could you clarify the `plain argmax` on here? I think you override the torch.argmax with composite one in [here](https://github.com/tenstorrent/tt-xla/pull/5004/changes#diff-0270b3e16371181368f385676f3eb3c79a2d15ea2472912b7a79036500ccc6f8R609).

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2844,7 +2845,19 @@ def _model_decode_compiled(
...
+        # Only return the full-vocab logits when logprobs are requested. On a
+        # vocab-sharded graph the logits leave compute_logits split across the
+        # "model" axis; returning them as a graph output forces a full-vocab
+        # all_gather every step (and, on the sampling path, an extra all_to_all
+        # reslice around the sharded topk). They are dead downstream unless
+        # logprobs are requested (execute_model sets logprobs=None otherwise),
+        # so hand back a tiny placeholder and let DCE drop the gather.
+        logits_out = (
+            logits
+            if sampling_metadata.logprobs
```
sshonTT: Do we turn on logprobs during warmup? If not, it won't get cached.

python_package/tt_torch/composite_ops.py
```diff
@@ -264,6 +283,26 @@ def composite_topk_indices(
...
+) -> Tensor:
+    """Composite argmax with custom sharding rule for distributed argmax.
+
+    Uses ttnn.argmax (fast O(n) max reduction) instead of ttnn.topk
+    (O(n log²n) bitonic sort), yielding significant speedup for greedy
+    decoding on vocab-sharded tensors.
+    """
+    dim = _normalize_dim(dim, input.dim())
+    attrs = {"dim": dim, "keepdim": keepdim}
+    builder = StableHLOCompositeBuilder(name="tenstorrent.argmax", attr=attrs)
```
sshonTT: Why don't you use `_make_composite_builder` for here?

python_package/tt_torch/composite_ops.py
```diff
@@ -567,6 +606,7 @@ def can_apply_composite(node: torch.fx.Node) -> bool:
         frozenset({0}): composite_topk_values,
         frozenset({1}): composite_topk_indices,
     },
+    torch.argmax: composite_argmax,
```
sshonTT, line 679: Could we maybe gate this in can_apply_composite so it rejects cases where dim isn't specified?

I'm asking because torch.argmax's dim is optional. Native PyTorch flattens to a scalar when dim is missing, but composite_argmax defaults to dim=-1. So a bare torch.argmax(x) might silently end up with the wrong shape.

What do you think about handling it like _check_sdpa_constraints?

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Thanks for pointing that out, added logprobs as a warmup dimension and thread it into _get_dummy_inputs

python_package/tt_torch/composite_ops.py
alinakhanTT: Switched to _make_composite_builder!!

python_package/tt_torch/composite_ops.py
alinakhanTT: Done, handled it like _check_sdpa_constraints!!

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: You're right, it's not a plain argmax. torch.argmax is globally overridden to composite_argmax in composite_ops, so this call does still lower through the composite. The difference is the tensor it runs on, I gather the logits to full/replicated first (sharding_constraint_tensor(logits, mesh, (None, None))), so the composite runs as a local reduction over the full vocab and returns the correct global index.
Gathering for the greedy path is intention. Keeping it vocab-sharded and doing the distributed argmax isn't reliable on this graph today, and since greedy decode is host/dispatch-bound the sharded reduction wouldn't buy any speedup even once that's fixed so the gather is the simpler, correct option here. The sharded reduction is still what the sampling path below uses, where it does matter

sshonTT (approved):
thanks!

integrations/vllm_plugin/vllm_tt/sampler.py
```diff
@@ -362,22 +434,31 @@ def _ttnn_sampling_padded(
         return result[:batch].to(torch.int64)
 
 
+_SHARDED_TOPK_CANDIDATES = 128  # matches typical num_chunks * k_per_chunk
```
sgligorijevicTT, line 453: Why is this hardcoded to 128? I have seen production uses of topk values up to ~150, what happens in that case?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -3529,9 +3618,28 @@ def sample_from_logits(
             and sampling_metadata.no_min_tokens
             and sampling_metadata.no_generators
         ):
-            out_tokens = torch.argmax(logits, dim=-1, keepdim=True)
+            if self.is_sharded_compute_logits:
+                # Greedy: gather logits to full and argmax over the whole vocab.
+                # We gather (rather than stay vocab-sharded like the sampling path
+                # below) because the distributed sharded argmax mis-executes at
```
sgligorijevicTT, line 4670: for my information, why does distributed sharded argmax mis-execute? and why are greedy steps dispatch bound?

sgligorijevicTT (approved):

integrations/vllm_plugin/vllm_tt/sampler.py
alinakhanTT: 128 is the width of the pre-filtered candidate pool we feed to the fused tt::sampling kernel, not the top-k cap. The actual top-k the kernel honors is capped at _TOPK_K_PER_CHUNK = 32 (see the .clamp(max=_TOPK_K_PER_CHUNK) at line 402), the kernel writer's valid k range is 1..32 and k>32 reads OOB in L1. So top_k=150 gets clamped to 32; the 128 pool is unaffected and stays safely larger than the max honored k. That 32 cap is a kernel constraint independent of this PR

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: This gather is only the greedy branch, the sampling path stays vocab-sharded and keeps the PR's win.
Why gather here: torch.argmax on vocab-sharded logits lowers to composite_argmax, whose sharding rule should merge each shard's local max into a global index. It's correct in isolation but returns a shard-local index in the full decode graph so we get garbage greedy output. I didn't chase the tt-mlir root cause; instead I all-gather to full vocab ((None, None)) and take a plain argmax, which is trivially correct. (composite_topk uses a different, validated sharding rule, that's why sampling can stay sharded.)
Dispatch-bound: a greedy step is one reduction producing one token/row, tiny device work next to the forward pass. Wall-clock is dominated by host dispatch, not the reduction or gather, so the extra all-gather is free batch=1 matches the pre-PR baseline (~38 tok/s), no regression

.github/workflows/perf-bench-matrix.json
```diff
@@ -674,6 +674,16 @@
         "skip-device-perf": true,
         "runs-on": "galaxy-wh-6u"
       },
+      {
+        "name": "vllm_llama_3_1_8b_sharded_sampling_galaxy_tp",
+        "pytest": "tests/benchmark/test_vllm_benchmarks.py::test_vllm_tp_benchmark[llama-3.1-8b-sharded-sampling-galaxy-tp]",
+        "vllm": true,
+        "skip-stablehlo-dump": true,
+        "skip-ttir-dump": true,
+        "skip-ttnn-dump": true,
```
bmalesevicTT: I now see this is applied to some other model tests too, is there a specific reason why? These dumps are often useful to us for debugging when a regression occurs, and analyzing in general

tests/benchmark/test_vllm_benchmarks.py
```diff
@@ -128,6 +128,28 @@ def _tp_config(
     )
 
 
+def _sharded_sampling_tp_config(
+    model: str, batch_size: int, *, mesh_shape, temperature: float = 0.8
+):
+    # Non-greedy device sampling on a 2D mesh: temperature>0 + cpu_sampling=False
+    # hits the vocab-sharded composite_topk path (#4494). opt-level 0 is required
+    # because device sampling under trace needs opt<1, and opt>0 would force
+    # cpu_sampling=True and bypass the composite.
```
bmalesevicTT: Wasn't this solved in https://github.com/tenstorrent/tt-xla/pull/5671 so now opt-lvl>0 can be and is used alongside trace and device sampling?

bmalesevicTT (changes requested):
Seems like a great optimization, please address these perf-benchmark related concerns.

Another concern I have is observed outputs are for Users 0 and 1 differing from others which are identical. I want to note our test infra does the PCC check for first decode on User 0 but doesn't look at the actual textual sanity of the outputs, so accuracy might be falsely positive in any test, including this one. (details on that https://github.com/tenstorrent/tt-xla/issues/5718)

.github/workflows/perf-bench-matrix.json
alinakhanTT: These follow the existing convention for every vLLM benchmark entry (bge-m3, qwen-embedding, llama-70b-qb2), introduced in #3871: vLLM runs don't emit IR to ./modules/irs/, so the dump steps find nothing, and the ttir/ttnn copy steps (which lack || true) would actually fail the job on the missing files. So no debugging artifact is being lost here; there's simply nothing to dump on the vLLM path. If we want IR dumps for vLLM benchmarks that's a broader change to how the vLLM path emits IR happy to file a follow-up if it'd be useful

tests/benchmark/test_vllm_benchmarks.py
alinakhanTT: Thanks for pointing that out. The comment was stale. Two things worth noting: _config's if optimization_level: treats 0 as falsy, so optimization_level=0 was silently dropped and the plugin defaulted to 1 anyway this benchmark was already running at opt-1. Fixed the comment and set optimization_level=1 explicitly (c943afb1c)

alinakhanTT:
@bmalesevicTT Good catch, and I agree the user-0-only PCC gate can be falsely green, so I validated this on galaxy rather than trusting it. Ran the sharded-sampling config (4,8), batch 32, and inspected all 32 completions for the same prompt:

 - Greedy (temp 0): all 32 byte-identical the composite returns correct, deterministic global vocab tokens.
 - Sampling (temp 1.0): all 32 coherent no shard-local garbage; #4494's actual requirement holds.

On the "some users differ / most identical" pattern you flagged: I isolated it with a single-device control (no tensor-parallel, no sharding, no composite in the graph at all, just the plain tt::sampling kernel). That run reproduced the same clustering and produced outright token-soup on a few rows, while the composite (4,8) run had none, all 32 coherent. So it comes from the device sampler kernel, reproducible with none of this PR's code in the graph; if anything the sharded path was cleaner

.github/workflows/perf-bench-matrix.json
bmalesevicTT: Hm, some vllm benchmarks do have the steps of dumping those artifacts, example https://github.com/tenstorrent/tt-xla/actions/runs/30227192796/job/89932417075. I am not informed on them being wrong, so please double-check this, and file an issue if these are not truly related to vllm, we should disable them all rather than have false data

tests/benchmark/test_vllm_benchmarks.py
bmalesevicTT: I don't see any logging in the benchmark run about the used opt-lvl, but the perf-report in it (https://github.com/tenstorrent/tt-xla/actions/runs/29757678343/job/88411701736 mentioned in desc) has it as 0, could very well be that is a bug, but would be useful to have that info in logs, please file an issue for that as well.
Also did you try opt-lvl=2, I think that should be the current default for vllm benchmarks as well, correct me if I am wrong please?

.github/workflows/perf-bench-matrix.json
alinakhanTT: Good catch. Those flags date to #3871 ("vLLM produces no IR artifacts"), but that's stale, _run_vllm_benchmark now sets export_path="modules", so generative vLLM benchmarks emit shlo/ttir/ttnn IR. That's why the other generative rows (single-chip + qb2 TP) already dropped them, I'd copied them off vllm_llama_3_1_70b_qb2_tp, which is a stale leftover. Removed them here so it dumps IR like the other TP rows, keeping only skip-device-perf (device perf isn't measured for vLLM). Embedding rows correctly still skip _run_vllm_embedding_benchmark sets no export_path. The 70b row is a separate pre-existing cleanup!

tests/benchmark/test_vllm_benchmarks.py
alinakhanTT: It runs at optimization_level=1 now (was 0; the cpu_sampling-at-opt>0 workaround was removed in #5671). On opt=2, the vLLM TT platform default is actually 1 (platform.py:113), and TP configs pin 1 on purpose (gemma4 hits an L1 OOM TT_FATAL at opt=2, and #5781 standardized Mistral-Small TP on opt=1), so I'd rather not flip this galaxy [4,8] sharded-topk path to opt=2 without validating it. And you're right that nothing logged it get_pjrt_compile_config built the config but never printed it, so I added a logger.info there to surface the resolved config (including optimization_level) in each run!

bmalesevicTT: While vllm platform is default opt-lvl-1, I think in benchmark tests even vllms are default 2, and the pinned ones have specific issues to solve to reenable them (example gemma4 L1 OOM). Without opt lvl 2 nothing is kept in L1, so setting it is for perf sake, however we can keep it at lvl 1 if your team prefers so in this case

bmalesevicTT (approved):

alinakhanTT:
Nightly green: https://github.com/tenstorrent/tt-xla/actions/runs/31192141101

Issue #5006, opened 2026-05-29
`experimental_kv_cache_dtype` (BFP8/BFP4) fails to compile on the vLLM decode path — paged SDPA decode requires Q/K/V same dtype


kmabeeTT:
## Problem

Enabling `experimental_kv_cache_dtype` (`bfp_bf8` or `bfp_bf4`) on the vLLM decode path fails during compilation:

```
error: 'ttnn.paged_scaled_dot_product_attention_decode' op
    Query, key, and value must have the same element type.
module_builder.cc:1172  ERR| Failed to run TTIRToTTNNCommon pipeline
```

The KV-cache dtype conversion casts the K/V cache to BFP8/BFP4 but leaves the query in bfloat16, and `ttnn.paged_scaled_dot_product_attention_decode` rejects the mismatch. Reproduced on opt-125m and Llama-3.2-3B.

## Repro test: 

`tests/integrations/vllm_plugin/generative/test_opt_generation.py::test_opt_generation_kv_cache_bfp8` (currently failing; to be `xfail`'d against this issue).

Needs debug to determine the correct fix (cast the query feeding the decode op, or relax the op to accept a lower-precision K/V cache) — likely a follow-up tt-mlir issue. tt-mlir #4140's accuracy results came from the non-paged SDPA path, which is why this wasn't caught.

AleksKnezevic:
Does this limitation come from the metal op?

kmabeeTT:
@AleksKnezevic just took a look and nope, metal op is correct and expects Query in BFLOAT16, tt-mlir verifier was wrong expecting K,Q,V all same dtype.  Quick one line fix in tt-mlir, just tested, will raise PR for fix+test.  More details:


>   No — the limitation isn't from the metal op. It was an overly-strict verifier check in tt-mlir.
> 
>   PagedScaledDotProductAttentionDecodeOp::verify() required Q/K/V to all share an element type, so a BF16 query over a BFP8 cache was rejected at compile time (Query, key, and value must have the same element type). The fix is essentially deleting that one check — the metal op already supports this case.
> 
>   sdpa_decode_device_operation.cpp accepts BFLOAT16/BFLOAT8_B/BFLOAT4_B inputs, and for GQA it requires the query to be BF16 while reading a quantized K/V cache — the opposite of a limitation; it's exactly the config we want.
> 
[...]

Issue #5005, opened 2026-05-29
vllm_plugin: enable experimental_kv_cache_dtype (BFP8/BFP4 KV cache) through vLLM + add test coverage


kmabeeTT:
The tt-mlir `experimental-kv-cache-dtype` compile option (BFP8/BFP4 KV cache) is not reachable from the vLLM plugin, and has no test coverage on the vLLM path.

Requesting:

1. Plumb `experimental_kv_cache_dtype` through `TTConfig` (`integrations/vllm_plugin/vllm_tt/platform.py`) so `additional_config={'experimental_kv_cache_dtype': 'bfp_bf8'}` reaches the compiler, mirroring `experimental_weight_dtype`.
2. Add an `xfail(strict=True)` test exercising it on a decode model.

Note: enabling it today fails to compile on the vLLM decode path (`ttnn.paged_scaled_dot_product_attention_decode` requires Q/K/V same dtype). The xfail documents this; the underlying fix will be tracked in a follow-up tt-mlir issue linked from the PR.

kmabeeTT:
Support added via #5056 and https://github.com/tenstorrent/tt-xla/pull/5007 but sub-issues still open, with #5006 being blocker for actually using this. Awaiting tt-mlir-uplift.

Edit: Uplifted, and used for some time already.

kmabeeTT:
Hey @vkovacevicTT can you or someone look into making BFP8 kv cache enabled by default in vllm benchmarks?  I'd been using it for production configs in tt-inference-server for some time, but want vllm benchmark tests to match.

ddilbazTT:
Hey @vkovacevicTT can you give an update on this issue? Thanks

kmabeeTT:
Assign to @alinakhanTT to pickup where @vkovacevicTT (ooo until end of next week) left off (he started, gave me links to below change and tests) to get this merged.

Branch: https://github.com/tenstorrent/tt-xla/tree/vkovacevic/vllm-kvcache-dtype
n150 benchmark: https://github.com/tenstorrent/tt-xla/actions/runs/29708730195
p150 benchmark: https://github.com/tenstorrent/tt-xla/actions/runs/29708738696

PR #5007, opened 2026-05-29
[vLLM] Expose experimental_kv_cache_dtype + add xfailed BFP8 repro test

kmabeeTT:
### Ticket

#5005 #5006

### Problem description

The tt-mlir `experimental-kv-cache-dtype` compile option (BFP8 KV cache) is now forwarded from `TTConfig`, but vLLM's KV-cache budgeting was unaware that the cache is a compile-time conversion to block-float — it still sized blocks as bf16. So the memory saving was invisible to the engine: longer contexts were rejected at init identically with and without the flag, delivering no benefit. The flag also had no test coverage on the vLLM path.

(There is also a tt-mlir-side decode compile failure — the cache is cast to BFP8 but the query stays bf16, and `ttnn.paged_scaled_dot_product_attention_decode` requires Q/K/V to share an element type. That fix isn't uplifting soon; the xfail test below guards it.)

### What's changed

- `integrations/vllm_plugin/vllm_tt/model_runner.py`: introduce `kv_cache_spec_dtype` — when the flag is set, vLLM's KV-cache spec/budgeting reports a 1-byte dtype (`uint8`) to match the real on-device BFP8 footprint, while the staged/transferred buffer stays bf16 (the dtype the compiler converts to block-float on device). Identity-preserving when the flag is unset. `bfp_bf4` is rejected pending sub-byte page accounting (#5011).
- `integrations/vllm_plugin/vllm_tt/worker.py`: the `determine_available_memory` placeholder tensor uses the real transferable cache dtype rather than the (now 1-byte) spec dtype.
- `tests/benchmark/test_vllm_benchmarks.py`: add a `TT_BENCHMARK_KV_CACHE_DTYPE` knob to exercise the flag and KV budget from the existing matrix.
- `tests/integrations/vllm_plugin/generative/test_opt_generation.py`: add `test_opt_generation_kv_cache_bfp8` (opt-125m, flag on), `xfail(strict=False)` against #5006. Kept non-strict so it won't churn the pipeline whenever the tt-mlir decode fix eventually lands.

### Result (the win)

Llama-3.1-8B, batch 1, single device, `gpu_memory_utilization=0.05` (KV budget 1.59 GiB):

| max_model_len | KV dtype | Result | KV capacity |
|--------------:|----------|--------|-------------|
| 16K | bf16 (default) | ❌ fails at init (needs 2.0 GiB > 1.59) | 13,056 tok |
| 16K | **bfp_bf8** | ✅ fits + runs, coherent output, no preemptions | 26,112 tok |

A 16K context that the bf16 cache rejects now serves end-to-end with the BFP8 cache — confirming the budgeting change makes the device-side saving visible to the engine.

### Checklist

- [x] Llama-3.1-8B 16K fits with bfp_bf8, fails with bf16 at the same budget
- [x] New opt-125m test reports `xfailed`; decode SDPA fix tracked in #5006

### DIFF
--- a/tests/integrations/vllm_plugin/generative/test_opt_generation.py
+++ b/tests/integrations/vllm_plugin/generative/test_opt_generation.py
@@ -31,6 +31,39 @@ def test_opt_generation():
 
 @pytest.mark.push
 @pytest.mark.single_device
+@pytest.mark.xfail(
+    reason="experimental_kv_cache_dtype casts the KV cache to BFP8 but leaves the "
+    "query in bf16, and ttnn.paged_scaled_dot_product_attention_decode requires "
+    "Q/K/V to share an element type. See tt-xla #5006.",
+    strict=False,
+)
+def test_opt_generation_kv_cache_bfp8():
+    # Reproduces the experimental_kv_cache_dtype failure on the vLLM decode
+    # path; remove the xfail once tt-xla #5006 is fixed.
+    prompts = [
+        "Hello, my name is",
+    ]
+    sampling_params = vllm.SamplingParams(temperature=0.8, top_p=0.95, max_tokens=32)
+    llm_args = {
+        "model": "facebook/opt-125m",
+        "max_num_batched_tokens": 128,
+        "max_num_seqs": 1,
+        "max_model_len": 128,
+        "gpu_memory_utilization": 0.001,
+        "additional_config": {
+            "enable_const_eval": False,
+            "min_context_len": 32,
+            "experimental_kv_cache_dtype": "bfp_bf8",
+        },
+    }
+    llm = vllm.LLM(**llm_args)
+
+    output_text = llm.generate(prompts, sampling_params)[0].outputs[0].text
+    print(f"prompt: {prompts[0]}, output: {output_text}")
+
+
+@pytest.mark.push
+@pytest.mark.single_device
 def test_opt_generation_multibatch():
     prompts = [
         "Hello, my name is",
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -282,6 +282,17 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                 self.kv_cache_dtype = model_dtype
         else:
             self.kv_cache_dtype = TPU_STR_DTYPE_TO_TORCH_DTYPE[cache_config.cache_dtype]
+        if self.tt_config.experimental_kv_cache_dtype == "bfp_bf8":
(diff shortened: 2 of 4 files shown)


kmabeeTT:
The fix needed in tt-mlir got merged today via https://github.com/tenstorrent/tt-mlir/pull/8668 so I'll remove the xfail from this test and hold off on merging this until tt-mlir uplift lands. Still valid for review in the meantime, but test will fail until then.

kmabeeTT:
Update - Rebased on latest main since @vkovacevicTT made some of the same changes released todday via https://github.com/tenstorrent/tt-xla/pull/5056 I could drop a bunch, but the kv-cache budgeting change is still necessary.  Will update PR Description, and I marked the new test `strict_xfail=False` since tt-mlir uplifts are still WIP with no ETA to bring https://github.com/tenstorrent/tt-xla/issues/5006 fix, there is advantage to getting this kv-cache-budgeting bug fix in sooner since it's already verified than letting it sit for unknown amount of time :)

Double checked test suite can run fine with 1 xfailed test:

```
========== 5 passed, 1 xfailed, 29 warnings in 498.43s (0:08:18) ==========
```

Also FYI - Vladan, confirmed same failure in vllm benchmark tests when trying BFP8 KV Cache, so avoid it until tt-mlir uplift lands.

```
loc("custom-call.192"): error: 'ttnn.paged_scaled_dot_product_attention_decode' op Query, key, and value must have the same element type.

FAILED tests/benchmark/test_vllm_benchmarks.py::test_vllm_benchmark[llama-3.2-3b] - RuntimeError: Engine core initialization failed. See root cause above. Failed core proc(s): {}```

File: tests/integrations/vllm_plugin/generative/test_opt_generation.py [hunk 1]
```diff
@@ -29,6 +29,39 @@ def test_opt_generation():
     print(f"prompt: {prompts[0]}, output: {output_text}")
 
 
+@pytest.mark.push
+@pytest.mark.single_device
+@pytest.mark.xfail(
```
mmanzoorTT, line 34: This test will work once we uplift tt-mlir or do we need any other fix?

kmabeeTT: Yep correct works after tt-mlir uplift, have tested it (and vllm benchmarks with same BFP8 KV Cache) locally with the fix cherry picked. Will circle back and remove xfail shortly after uplift lands (but since uplifts are nightmare already for folks, didn't want to cause them to have to rerun once more to remove the xfail tag, if I left it strict=True)

mmanzoorTT: Totally agree.

mmanzoorTT (approved):

Issue #5015, opened 2026-06-01
[vLLM]: drop pad_attention_heads_force_equal workaround once tt-metal concat program-cache bug is fixed


kmabeeTT:
## Issue

`pad_attention_heads_force_equal=True` is required to run Gemma-4-31B on n300_llmbox 1D 8-chip mesh (added under #5003). Without it, `_compute_head_pad_decision` picks the cheaper c=1 (no replication, just zero-pad) layout for Gemma's full-attention layers, which produces an unequal-sized `[Q;K;V]` concat along the sharded axis inside `XlaQKVParallelLinear`. That concat trips a tt-metal program-cache assertion:

```
TT_FATAL: Cannot get runtime args for kernel reader_concat_interleaved_start_id
that is not placed on core 2-0
  at tt-metal/tt_metal/impl/kernels/kernel.cpp:572
```

This issue tracks dropping the `force_equal` workaround once the underlying tt-metal bug is fixed.

### Root cause

`ConcatProgramFactory::create_descriptor` sets the kernel's `core_ranges = all_cores` (from `split_work_to_cores`) but emplaces runtime args on `grid_to_cores(num_cores)`. The mesh device-operation cache-hit slow path (`apply_descriptor_runtime_args` in `mesh_device_operation_adapter.hpp:553`) re-runs `create_descriptor` on hit and applies the fresh descriptor's runtime_args against the cached program's compiled kernel placement. When the cached call had a smaller `num_output_pages` than the current call, the fresh runtime_args include cores the cached kernel was never placed on → assertion fires.

### Current workaround

`pad_attention_heads_force_equal=True` forces `c=k` replication so padded Q == padded K == padded V, which sidesteps the unequal-concat path. It works for Gemma-4-31B (~5 tok/s) but a) costs extra memory / compute on the zero-padded heads, and b) introduces a separate hang-at-first-decode failure mode for configs that need zero-padding on top of c=k replication (e.g. Qwen2.5-7B with c=7, padded_kv=32). That hang is gated by `NotImplementedError` in `_compute_head_pad_decision`.

### Starting-point prototype (not a finished fix)

This is a **starting point** from an initial debug pass — not a vetted fix. Documenting where the trail currently is so the next person can pick it up.

Branch: `tenstorrent/tt-metal` → [`kmabee/concat-program-factory-cache-fix-prototype`][branch] (commit [`3681758c785`][commit]). Two-part change in `concat_program_factory.cpp`:

[...]

PR #5017, opened 2026-06-01
[vLLM]: pad_attention_heads + Gemma-4-31B on n300_llmbox 1D 8-chip mesh

### LINKED ISSUE #5003: padded attention heads through vLLM
We should support attention head padding through vLLM so that models can be sharded and any grid size. 

Initial PR [here](https://github.com/tenstorrent/tt-xla/pull/4881)

kmabeeTT:
### Ticket
Closes #5003. Continues from #4881 Initial PR.

### Problem description
Gemma-4-31B requires attention-head padding to run on n300_llmbox's 1D 8-chip TP mesh: its `num_global_key_value_heads=4` doesn't divide the 8-way model axis, and Shardy asserts on the non-divisible factor. Aleks's initial `pad_attention_heads` work (#4881) handles single-config models like Qwen2.5-7B but doesn't cover Gemma-4's two distinct attention configurations (sliding vs full layers) or its multimodal `text_config` nesting. A tt-metal concat program-cache bug also surfaces along the way when padded `[Q;K;V]` are unequal-sized on the sharded axis.

### What's changed

Three commits:

1. **`pad_attention_heads` (Aleks's original).** Pad `num_attention_heads` / `num_key_value_heads` so `num_kv_heads` is a multiple of the TP axis size. Two strategies (min-cost zero-pad and c=k replication); cost function picks the cheapest. Surgery patches `qkv_proj`/`o_proj`/`Attention` post-load.
2. **Extend for Gemma-4-31B on llmbox.** Per-attention-configuration spec dispatch (Gemma-4 has both sliding and full layers with different `num_kv_heads`/`head_dim`), `text_config` fallback for multimodal HF configs, new `pad_attention_heads_force_equal` flag that forces `c=k` to sidestep the tt-metal concat bug on unequal-sized `[Q;K;V]` (tracked in #5015), and a refactor pass on the surgery code (-70 lines, no behavior change).
3. **Gate hang-prone `force_equal` configs (#5015).** Raise `NotImplementedError` at engine init when `force_equal=True` would zero-pad on top of c=k replication (`padded_kv > orig_kv * c`). The only known instance is Qwen2.5-7B, which hangs at first decode on llmbox without the gate. Gemma-4-31B stays clear because both its configs have `padded_kv == orig_kv * c`.

Padding is opt-in via `pad_attention_heads` (default `False`); models that don't use it see zero behavior change.

### Known limitations

- `pad_attention_heads_force_equal` is a workaround for the tt-metal concat bug (#5015), not the final state. It costs extra memory/compute on the zero-padded heads. The real fix is upstream in tt-metal; the flag is meant to be deleted once that lands.
- `qwen2.5-7b-force-equal` push parametrization is `pytest.mark.skip`'d (gated by #5015). The gate is exercised via a fast unit test instead. Skip rather than xfail because `NotImplementedError` gets wrapped as `RuntimeError` in the engine subprocess and doesn't compose cleanly with strict xfail.
- The extension targets the specific shape of Gemma-4-31B on llmbox; other multi-attention-config or multimodal models may surface adjacent issues that aren't covered here.

### Testing

- **Push**:
  - `test_tensor_parallel_generation_llmbox_pad[llama-3.2-3b]` — pad-is-no-op path.
  - `test_tensor_parallel_generation_llmbox_pad[qwen2.5-7b]` — min-cost pad path.
  - `test_pad_decision_gates_force_equal_with_zero_pad` — fast unit test on the gate (~3s, no hardware).
- **Nightly**:
  - `test_tensor_parallel_generation_llmbox_gemma4_31b` — Gemma-4-31B end-to-end on n300_llmbox 1D 8-chip.
- **Benchmark**:
  - `test_vllm_tp_benchmark[gemma4-31b-it-tp]` — runs at ~5 tok/sec, ~10 min compile.
- **Branch CI runs**:
  - vLLM Push (passing): https://github.com/tenstorrent/tt-xla/actions/runs/26730172197
  - vLLM Nightly (passing): https://github.com/tenstorrent/tt-xla/actions/runs/26734989772

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -590,6 +590,18 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         if padded_q == orig_q and padded_kv == orig_kv:
             return None
 
+        # force_equal + padded_kv > orig_kv*c (i.e. c=k replication doesn't
+        # fill padded_kv and we'd zero-pad on top) hangs at first decode on
+        # llmbox; fail fast instead. See #5015.
+        if force_equal and padded_kv > orig_kv * best["c"]:
+            raise NotImplementedError(
+                f"pad_attention_heads_force_equal=True with orig_kv={orig_kv}, "
+                f"c={best['c']}, heads_axis={heads_axis} would zero-pad "
+                f"{padded_kv - orig_kv * best['c']} KV heads on top of c=k "
+                f"replication (padded_kv={padded_kv}). Known to hang at first "
+                f"decode on llmbox (#5015)."
+            )
+
         return {
             "orig_q": orig_q,
             "orig_kv": orig_kv,
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -17,9 +17,10 @@ import torch_xla
 import torch_xla.core.xla_model as xm
 import torch_xla.distributed.spmd as xs
 import torch_xla.runtime as xr
-import vllm.envs as envs
 from tt_torch.sharding import sharding_constraint_tensor
 from tt_torch.utils import torch_dynamo_tt_device_compatibility
+
+import vllm.envs as envs
 from vllm.compilation.wrapper import TorchCompileWithNoGuardsWrapper
 from vllm.config import (
     ParallelConfig,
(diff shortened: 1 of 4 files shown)


mmanzoorTT (approved):

kmabeeTT:
Rebased on latest main to solve `tests/benchmark/test_vllm_benchmarks.py` conflict with QB2 version of gemma4-31b-it that just got merged via https://github.com/tenstorrent/tt-xla/pull/5012. Previous onPR CI run link here had a pre-commit [fail](https://github.com/tenstorrent/tt-xla/actions/runs/26737082392/job/78873769190), otherwise was good. New run should be green.

kmabeeTT:
Worked to get model running on WH, but I don't understand the code, so closing and will defer to someone else.

Issue #4702, opened 2026-05-14
Multimodal support with Gemma 4 model


sshonTT:
The current version of Gemma 4 only supports text. However, since it has multimodal capabilities, we should be able to run it with image, audio, and video inputs.

sshonTT:
I have a working branch for image : https://github.com/tenstorrent/tt-xla/tree/sshon/multimodal-gemma4

sshonTT:
logits Softcap is required for audio

PR #5029, opened 2026-06-01
Add Gemma-4 multimodal image input for vLLM

sshonTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/4702)

### Problem description
There's not yet support for multimodal model in vLLM

### What's changed
Enable the multimodal encoder path in the TT model runner: activate supports_mm_inputs, execute/gather the mm encoder (handle the is_embed mask for models that interleave marker tokens), precompile embed_input_ids, and build the dummy mm batch via the current vLLM API.

overrides: route HF vision/audio tower attention through the fused TT SDPA op (eager attention can't lower over the device mesh); replace vLLM's masked_scatter_-based _merge_multimodal_embeddings with a static-shape cumsum+gather+where impl (the dynamic mask scatter is rejected by tt-mlir's Shardy pass); promote pre-allocated plain-tensor attrs to non-persistent  buffers so model.to(device) relocates them.

ascend_scheduler: schedule multimodal encoder inputs (the base scheduler does this inline with KV allocation; do it post-hoc to stay close to AscendScheduler.schedule()).

Add a BHQB 31B image-input test.

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/overrides.py
+++ b/integrations/vllm_plugin/vllm_tt/overrides.py
@@ -5,16 +5,20 @@
 from typing import OrderedDict
 
 import torch
-from vllm.model_executor.layers.layernorm import RMSNorm
 from vllm.model_executor.layers.rotary_embedding.base import RotaryEmbedding
 from vllm.model_executor.layers.rotary_embedding.mrope import MRotaryEmbedding
 
+from .layers.mm_embeddings import install_static_shape_merge_multimodal_embeddings
 from .layers.mrope import override_mrope_module
+from .layers.multimodal_attention import override_vision_attention
 from .layers.rmsnorm import override_rmsnorm_module
 from .layers.rotary_embedding import override_rotary_embedding_module
 from .logger import tt_init_logger
 
 logger = tt_init_logger(__name__)
+
+# Patch vLLM's multimodal embedding merge to a static-shape impl at import time.
+install_static_shape_merge_multimodal_embeddings()
 
 
 def get_fqn(module):
@@ -32,6 +36,29 @@ ISINSTANCE_OVERRIDES = [
     (MRotaryEmbedding, override_mrope_module),
     (RotaryEmbedding, override_rotary_embedding_module),
 ]
+
+
+def _promote_pre_allocated_attrs_to_buffers(model: torch.nn.Module) -> None:
+    """Re-register plain torch.Tensor attributes as buffers so .to() moves them.
+
+    Some multimodal models pre-allocate ``self.per_layer_embeddings`` as a
+    plain attribute (not a registered buffer) on CPU. ``model.to(device)``
+    only relocates parameters and registered buffers, leaving it stranded on
+    CPU; a later add against an XLA tensor then trips dynamo's mixed-device
+    check. Promote such attributes to non-persistent buffers so .to() follows
+    them.
+    """
+    pre_allocated_attrs = ("per_layer_embeddings",)
+    for attr in pre_allocated_attrs:
+        if not hasattr(model, attr):
+            continue
+        t = getattr(model, attr)
+        if not isinstance(t, torch.Tensor):
+            continue
+        if attr in dict(model.named_buffers(recurse=False)):
+            continue
+        delattr(model, attr)
+        model.register_buffer(attr, t, persistent=False)
 
 
 def replace_modules(model: torch.nn.Module) -> None:
@@ -62,6 +89,8 @@ def replace_modules(model: torch.nn.Module) -> None:
             _process_module(child_module, child_name, module)
 
     _process_module(model)
+    override_vision_attention(model)
+    _promote_pre_allocated_attrs_to_buffers(model)
 
 
 def repair_stale_moe_closures(model: torch.nn.Module) -> None:
--- a/tests/integrations/vllm_plugin/generative/test_gemma4_generation.py
+++ b/tests/integrations/vllm_plugin/generative/test_gemma4_generation.py
@@ -1,9 +1,13 @@
 # SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
 #
 # SPDX-License-Identifier: Apache-2.0
+import base64
+import io
+
 import pytest
 import vllm
 from conftest import assert_output_coherent, check_host_memory
+from PIL import Image, ImageDraw
 
 
 @pytest.mark.nightly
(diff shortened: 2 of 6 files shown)


integrations/vllm_plugin/vllm_tt/overrides.py
```diff
@@ -15,6 +16,95 @@
...
+
+    if pad_q > 0:
+        attn_output = attn_output[:, :, :seq_q, :]
+
+    attn_output = attn_output.transpose(1, 2).contiguous()
+    return attn_output, None
+
+
+# Idempotently register so reloads (e.g. pytest reruns) don't raise.
+if "tt" not in ALL_ATTENTION_FUNCTIONS:
```
AleksKnezevic: what's this?

integrations/vllm_plugin/vllm_tt/overrides.py
```diff
@@ -168,6 +258,90 @@ def tt_rotary_embedding_module(layer: torch.nn.Module) -> torch.nn.Module:
...
+    logger.info(
+        "Installed static-shape _merge_multimodal_embeddings (replaces "
+        "masked_scatter_-based default that emits dynamic shapes)."
+    )
+
+
+_install_static_shape_merge_multimodal_embeddings()
+
+
+def _promote_pre_allocated_attrs_to_buffers(model: torch.nn.Module) -> None:
```
AleksKnezevic, line 41: why is this needed?

integrations/vllm_plugin/vllm_tt/overrides.py
```diff
@@ -168,6 +258,90 @@ def tt_rotary_embedding_module(layer: torch.nn.Module) -> torch.nn.Module:
 ]
 
 
+def _route_vision_attention_to_tt(model: torch.nn.Module) -> None:
```
AleksKnezevic: why is this needed?

integrations/vllm_plugin/vllm_tt/overrides.py
sshonTT: The Gemma4 vision tower is instantiated as a plain HF module (AutoModel.from_config) and run eager attention. vLLM has no attention path for it (unlike the language model, which goes through vLLM's attention backend). 
So to make its attention use our SDPA, we register the op into HF's ALL_ATTENTION_FUNCTIONS registry under "tt", and _route_vision_attention_to_tt points the tower's _attn_implementation at it.

integrations/vllm_plugin/vllm_tt/overrides.py
sshonTT: Please see this [one](https://github.com/tenstorrent/tt-xla/pull/5029#discussion_r3335521893)

integrations/vllm_plugin/vllm_tt/overrides.py
sshonTT: We don't need this right now. Will remove it. 
For quick explanation, it's for e4b model which has per layer embeddings which gemma4 model pre-allocates as a plain tensor attribute, not a registered buffer. 
model.to(device) only moves parameters and registered buffers, so it's stranded on CPU and a later op mixes it with on-device tensors, so dynamo fails on mixed-device fake tensors. Re-registering it as a non-persistent buffer makes .to(device) follow it.

sshonTT: Actually, we need it rn as I've added e4b test to PR. I had to debug kv sharing feature in e4b and now I found a fix.

sshonTT:
nightly passes : https://github.com/tenstorrent/tt-xla/actions/runs/26772188227

sshonTT:
@kmabeeTT, @mmanzoorTT could you ptal at this PR? Thanks!

integrations/vllm_plugin/vllm_tt/overrides.py
```diff
@@ -17,6 +18,94 @@
 logger = tt_init_logger(__name__)
 
 
+def tt_sdpa_attention_forward(
```
mmanzoorTT: Suggestion: Move this definition to separate file and use `overrides.py` to just calling override function like we are doing for other layers.

sshonTT: That's a fair point. I've moved the definition. Thanks!

tests/integrations/vllm_plugin/generative/test_gemma4_generation.py
```diff
@@ -70,3 +74,104 @@ def test_generation_bhqb_gemma4_26b_a4b():
...
+        "model": model_name,
+        "limit_mm_per_prompt": {"image": 1, "video": 0, "audio": 0},
+        "max_num_batched_tokens": 2560,
+        "max_num_seqs": 1,
+        "max_model_len": 1024,
+        "gpu_memory_utilization": 0.1,
+        "additional_config": {
+            "enable_const_eval": True,
+            "min_context_len": 32,
+            "enable_tensor_parallel": True,
```
mmanzoorTT, line 171: It is using `enable_tensor_parallel=True` but not marked with `@pytest.mark.tensor_parallel`; is this intentional?

tests/integrations/vllm_plugin/generative/test_gemma4_generation.py
```diff
@@ -70,3 +74,104 @@ def test_generation_bhqb_gemma4_26b_a4b():
...
+    output_text = llm.chat(messages, sampling_params)[0].outputs[0].text
+    print(f"output: {output_text}")
+    assert_output_coherent(output_text)
+
+    check_host_memory(model_name)
+
+
+@pytest.mark.nightly
+@pytest.mark.bhqb
+def test_generation_bhqb_gemma4_26b_a4b_image():
```
mmanzoorTT: I suggest to move `test_tensor_parallel_generation_bhqb_gemma4_31b_image` here and parametrize it with `test_generation_bhqb_gemma4_26b_a4b_image`. The only difference is the model being used and different mesh shape (1D vs 2D mesh).

mmanzoorTT (approved):

tests/integrations/vllm_plugin/generative/test_gemma4_generation.py
sshonTT: Ah missed this. Thanks!

tests/integrations/vllm_plugin/generative/test_gemma4_generation.py
sshonTT: Yeah that makes sense. Moved the test

mmanzoorTT (approved):

PR #5033, opened 2026-06-01
[Test] Add CPU compile-only tests

### LINKED ISSUE #4996: Compile-only test infra
Compile only tests require further productization and CI tests on cpu-only machines (TBD - [ref in slack](https://tenstorrent.slack.com/archives/C08D5PF9ZJN/p1779950554134359?thread_ts=1779377089.553329&cid=C08D5PF9ZJN))

Part of this is creating CPU compile-only tests that prove that CPU compile-only feature works without any TT hardware; so running a few simple tests similar to the system_desc example should be added.

Tests should be multidevice mesh-shape agnostic; eg. can support any device mesh even on 1x NUM_DEVICES mesh size -- several examples of this repo like exmaples llama.py. Also for singledevice systemdesc, we should also test that singledevice IR generation works

The main example of system desc generation and compile-only execution is done in [this example](https://github.com/tenstorrent/tt-xla/blob/e875a19671ab539b88d6667f8e60556b15fd0cb9/examples/pytorch/system_desc.py). Please extend it to run compile only tests on multiple hardware arch backed by real system descriptors generated on those hardware. 

System desc may change without warning and cause backcompat issues; so some way of saving an up to date system desc to be loaded at test time and making sure it is up to date is required. Some ideas:

- Check system desc into the repo, create scheduled job to run on each architecture and assert on the validity of the system desc by regenerating it on real hardware and comparing with checked in artifact. Also, if there are differences, to account for the ones that don't matter; so a simple byte equality check may not be sufficient. 
    - For regeneration -- could do something like the update_test_durations workflow which makes a commit + push for a specific metadata file in the repo, using the repo as authoritative filestore
- Generate system desc at nightly and use github artifacts / upload and download artifact for storage; using github artifact storage as a sidecar
- Anything else you can think of

jasonmacTT:
### Ticket
Refs #4996

### Problem description
The CPU compile-only feature lets us compile a model for a target system
without any TT hardware, by reading the chip spec from a saved system
descriptor instead of opening a real chip. There was no test coverage proving
this works end-to-end, and no mechanism for keeping per-arch system
descriptors available at test time.

### What's changed
- Add `tests/cpu_compile_only/test_cpu_compile_only.py`: compiles a simple
  model on CPU against checked-in `*_system_desc.ttsys` descriptors, one per
  arch (n150, n300, n300-llmbox, p150), asserting the expected compiled
  artifacts are produced. No TT hardware required.
- Check in per-arch system descriptors under `tests/cpu_compile_only/system_descs/`,
  plus `save_system_desc.py` to regenerate them on real hardware.
- Register the `cpu` pytest marker and add a `nightly and cpu` job to the
  nightly matrix, run with `forked: true` (each arch compiles in its own
  process, since torch_xla's singleton compilation cache can't be reused
  in-process).

### Follow-up
- Keeping the checked-in descriptors up to date, some form of a workflow that
  regenerates each arch's descriptor on real hardware and validates it against
  the committed copy (per #4996).

### Checklist


tests/compile_only/test_compile_only.py
```diff
@@ -0,0 +1,135 @@
...
+def _assert_compiled_artifacts(output_dir: Path, label: str = ""):
+    """Assert compilation produced the expected artifacts."""
+    # Artifacts the compile step is expected to emit.
+    EXPECTED_ARTIFACTS = ["model.ttnn", "model_ttnn.mlir", "model_ttir.mlir"]
+    suffix = f" for {label}" if label else ""
+    for f in EXPECTED_ARTIFACTS:
+        assert (output_dir / f).exists(), f"Expected {f} was not created{suffix}"
+
+
+def _discover_fixtures():
```
jameszianxuTT: Fixtures has a meaning in pytorch, which I don't think this really meets. Also, why not hardcode the available system desc list? I don't think autodiscovery really helps us here, and makes it less explicit about what we are testing

tests/compile_only/test_compile_only.py
```diff
@@ -0,0 +1,135 @@
...
+        cwd=tmp_path,
+        check=True,
+    )
+
+    _assert_compiled_artifacts(tmp_path / "output", label=system_desc_path.name)
+
+
+if __name__ == "__main__":
+    if len(sys.argv) == 3 and sys.argv[1] == "save":
+        _save_system_desc(sys.argv[2])
```
jameszianxuTT: I would rather this file be a standalone pytest, not a CLI tool as well. If you need a `save` tool, then it can be moved to a separate file unless there's a good reason not to

jameszianxuTT:
I think it makes sense to merge in these static tests and system desc, eventually we may be able to add some `@CPU_ONLY(arch)` decorator that allows us to graft compile-only mode to any test, but it would be nice to merge this in in stages to work out the kinks in cpu only execution infrastructure before generalizing too much

tests/compile_only/test_compile_only.py
```diff
@@ -0,0 +1,135 @@
...
+                ),
+            )
+        ]
+    # id = arch name (strip the `_system_desc.ttsys` suffix) for readable test ids.
+    return [
+        pytest.param(f, id=f.name.removesuffix("_system_desc.ttsys")) for f in fixtures
+    ]
+
+
+@pytest.mark.nightly
```
jameszianxuTT, line 71: Currently we don't have a pytest mark for cpu only, @vvukomanTT is working on that infra but once that is available we want to mark this test for CPU runners. 

Without an hardware specifying mark, I think this will probably not run at all, let's wait for the above to get merged

jasonmacTT:
Things to consider.
 1) ### Generating fresh system desc based on schema hash.
    The static-file system desc updates could be regenerated when the
    runtime's compiled schema hash and the hardware system desc's schema hash disagree. 
On an attempt to run this ```test_cpu_compile_only``` with a system desc externally generated for exabox_dual,
    the runtime rejected the descriptor on a schema_hash mismatch.
In particular, the following error was given: 
```error: 'builtin.module' op system desc schema mismatch, please collect a system desc with a runtime compiled with the same schema version```
    The runtime check that enforces this (tt-mlir, pinned commit c5f3984):
    https://github.com/tenstorrent/tt-mlir/blob/c5f398432a61100da79b7b9f2941130496092287/lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp#L367-L372

2) ### Creating a workflow to run the test only when descriptors are up to date
Ideally the compile-only test runs after the system descriptors are regenerated,
so it never validates against stale static files. 
Within a single workflow this is just a job dependency e.g.
```yaml
jobs:
  update_system_descs:
    # regenerate <arch>_system_desc.ttsys on hardware
    ...

  test_compile_only:
    needs: update_system_descs   # only runs once descriptors are up to date
    ...
```
This workflow would probably be separate from what is already in the infra, and feels out of sync with what is already built, any recommendations is appreciated.

jameszianxuTT:
To address issues:
1. We need to figure out how to invalidate and update the system desc for sure. Probably another simple pytest (or standalone script) could be used here that _does_ need a device and uses `ttrt query` to extract the current system desc then compare it with the local one. 
2. This workflow for compile only can be a rough best-effort check. Let's just keep it as simple as possible - compile only checks run using in-repo system desc on cpu runners (using @vvukomanTT CPU-only runner target) in _nightly_ and you can merge this PR as is. 

We can create and schedule a new workflow to run on nightly approximately before actual nightly (or maybe in uplift onpush?) to check and update the system desc if needed in a separate PR. The system desc schema does not change often so a manual fallback is fine for now, and if these tests fail once in a while it's really not a big deal.

jasonmacTT:
[Passing test](https://github.com/tenstorrent/tt-xla/actions/runs/27157632574) using cpu and nightly marker on cpu compile only test.

tests/cpu_compile_only/test_cpu_compile_only.py
```diff
@@ -0,0 +1,111 @@
...
+    Compiled artifacts land in tmp_path/output and are auto-cleaned by pytest;
+    they're disposable, the assertions are what prove the compile succeeded.
+    """
+    # Checked-in `<arch>_system_desc.ttsys` descriptors live in system_descs/.
+    system_desc_dir = Path(__file__).resolve().parent / "system_descs"
+    system_desc_path = system_desc_dir / f"{arch}_system_desc.ttsys"
+    assert system_desc_path.exists(), f"missing system descriptor: {system_desc_path}"
+
+    script = str(Path(__file__).resolve())
+    subprocess.run(
```
jameszianxuTT: What is the purpose of running this in a subprocess? It seems like a lot of indirection

tests/cpu_compile_only/test_cpu_compile_only.py
```diff
@@ -0,0 +1,111 @@
...
+        [sys.executable, script, "compile", str(system_desc_path)],
+        cwd=tmp_path,
+        check=True,
+    )
+
+    _assert_compiled_artifacts(tmp_path / "output", label=arch)
+
+
+if __name__ == "__main__":
```
jameszianxuTT: related to prev. comment about subprocess, do we need this? most of our pytests do not have both a main entrypoint for direct invocation, and `test_*` functions for pytest to call. 

This only needs to be a standalone pytest, so adding this entrypoint seems unnecessary

pytest.ini
```diff
@@ -32,3 +32,5 @@ markers =
     multi_host_cluster: marks test for multi-host setups (llmbox)
     # FileCheck
     filecheck: marks test with FileCheck pattern files to validate generated MLIR
+    # Cpu
+    cpu: marks test as cpu
```
jameszianxuTT, line 36: Please clarify -- marks test to be run on a cpu-only runner without TT hardware

jameszianxuTT (changes requested):

jameszianxuTT:
Overall looks good, a few questions about subprocessing that I'd like to be resolved. Make sure to rebase after #5127 is merged and that should clear up the CPU-only runner infra changes in this PR.

tests/cpu_compile_only/test_cpu_compile_only.py
jasonmacTT: I believe I could add ```"forked": true``` to the nightly json file for the entry of this test and remove the indirection. Running this test without some form of a fork call fails.

jameszianxuTT: > Running this test without some form of a fork call fails.

Wait, why? What happens if you try that

jameszianxuTT: Just for context, ideally we don't run tests under subprocess because it adds complexity to a lot of places (logging, singelton object lifetimes, env variable management). `--forked` is kind of a hack to get better test isolation, but we pay for it too

jasonmacTT: An assertion error is received, namely
```AssertionError: Computation cache has already been initialized```
There seems to be an API call ```xr.clear_computation_cache()```
that I tried adding to the end of the test, but the same error occurs. 
In the original system_desc.py file there was a comment stating the PJRT client cannot be reconfigured once initialized, and is why they run the compile in a forked call.

File: tests/cpu_compile_only/test_cpu_compile_only.py [hunk 7]
```diff
@@ -0,0 +1,111 @@
...
+
+
+@pytest.mark.nightly
+@pytest.mark.cpu
+@pytest.mark.parametrize(
+    # One arch per checked-in descriptor. Each descriptor is generated on real
+    # hardware via save_system_desc.py and committed to system_descs/. Add a new
+    # arch here once its descriptor is committed.
+    "arch",
+    ["n150", "n300", "n300-llmbox", "p150"],
```
jameszianxuTT: nit: IMO this array would be better exposed as a global at the top of the file so a reader of this test can get a quick idea of what architectures are supported without having to scroll to this parameterization

tests/cpu_compile_only/test_cpu_compile_only.py
```diff
@@ -0,0 +1,100 @@
...
+@pytest.mark.parametrize(
+    "arch",
+    ARCHS_WITH_SYSTEM_DESC,
+)
+def test_cpu_compile_only_per_arch(arch, tmp_path, monkeypatch):
+    """CPU-only: compile a simple model against a checked-in `.ttsys` descriptor.
+
+    No TT hardware required -- TT_COMPILE_ONLY_SYSTEM_DESC tells the plugin to
+    read the chip spec from a file instead of opening a real chip. The PJRT
+    client cannot be reconfigured once initialized (one client per process), so
```
jameszianxuTT: Please don't repeat this justification, and put up an issue tracking this limitation

The real justification is like
> compile-only feature relies on torch-xla singleton compilation cache that does not permit output folder relocation or reuse in-process.

jameszianxuTT (approved):
some minor nits but otherwise LGTM

vvukomanTT (approved):

nsumrakTT (approved):

tests/cpu_compile_only/system_descs/n300-llmbox_system_desc.ttsys
sgligorijevicTT, line 1: I don't like the idea of commiting system descriptors to the repo, can we do better?

jasonmacTT: https://github.com/tenstorrent/tt-xla/actions/runs/27046256460/workflow

This is a sample workflow I used to initially test the compile only feature in CI.
It performed the following: 
1) For each hardware, generate the system descriptors and upload them as artifacts.
2) On a "cpu-only" runner, download the artifacts and perform the test. 

The job for running the tests would depend on the jobs for generating the hardware descriptors.

In principle, this workflow does not require the system descriptors to be checked into the repo prior, nor does it make any permanent changes to the repo.

Thoughts on if something like this should be integrated or how to integrate it?

sgligorijevicTT: That addresses the uploading of descriptors, but I feel like making a CPU job wait on a machine job just for the descriptor somewhat defeats the point
@nsumrakTT @vvukomanTT thought on uploading the descriptors to some cloud bucket and downloading them as part of checkout? Also, how tedious would it be to have some job (nightly) keep the bucket up to date?
If this is not easy to do perhaps it's a lesser evil to just check them into the repo.. I generally don't like checking in binary files, especially those that might need to change, as what can happen is that the number of files grows and suddenly the total repo size explodes because of git history. But also, I don't want to block over a small bump in size

vvukomanTT: In general that's the right way of thinking, but since these files are small and there's a few of them I wouldn't introduce bigger infra changes to handle this. My 2 cents.

jameszianxuTT: We're planning to validate/update checked-in system descs with something like `update-test-durations` workflow, which will regenerate system descs on real HW and compare + make PR if the schema changes.

We could also use a better versioned storage to map like (system desc version + architecture) <-> (system desc artifact), but it is a lot of complexity / overengineering that will bloat this PR.

This is a small test to validate cpu-only feature on nightly only so it won't block on-pr or anything like that.

jameszianxuTT: >The job for running the tests would depend on the jobs for generating the hardware descriptors.

>That addresses the uploading of descriptors, but I feel like making a CPU job wait on a machine job just for the descriptor somewhat defeats the point

They won't be waiting, we don't need to make it a strict dependency. 

I think we may trigger the machine job under various circumstances 
 - after mlir uplift
 - some random time around nightly with no guarantee it will finish queues before nightly starts
 - iff the cpu-only tests fail due to schema mismatch

Also, if the cpu-only tests fail on one night due to system desc mismatch it's not a big deal. IME the schema also changes relatively infrequently (once every few weeks?). 

These files are a few KB each so I don't think we need to worry too much about storage

tests/cpu_compile_only/test_cpu_compile_only.py
jameszianxuTT: Discussed offline

tests/cpu_compile_only/system_descs/n300-llmbox_system_desc.ttsys
nsumrakTT: @sgligorijevicTT You are right about binary files, but I agree with @vvukomanTT , adding additional infra steps (with risk to fail) because of several files with just a few KB in size is an overkill imho.

PR #5035, opened 2026-06-01
[vLLM] Pin input shardings in the execution path to match warmup

sshonTT:
### Ticket
N/A

### Problem description
With a shardable batch (max_num_reqs > 1) the warmup graph shards these inputs on the model axis while execution leaves them replicated, so the backbone recompiles at the first real step (a size-1 batch hides this: the mark is a no-op). 

### What's changed
_dummy_run pins the input_ids/position_ids/inputs_embeds shardings before the model call, but the execution path (sample_tokens) did not. 
Extract the pin into a shared _pin_preflatten_input_shardings helper and call it from both paths so warmup and execution produce one graph.

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -1368,6 +1368,31 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
             # then the embedding layer is not included in the CUDA graph.
             return input_ids, None
 
+    def _pin_input_shardings(
+        self,
+        input_ids: torch.Tensor | None,
+        position_ids: torch.Tensor,
+        inputs_embeds: torch.Tensor | None,
+    ) -> None:
+        """Pin the model input shardings (batch dim across the mesh). Applied in
+        BOTH the warmup (_dummy_run) and execution (sample_tokens) paths so their
+        graphs match;
+        otherwise warmup shards these inputs while execution leaves them
+        replicated, and the backbone recompiles at the first real step for any
+        shardable batch. Also keeps large prefill traces from splitting into a
+        secondary sync that drops mhlo.spmd_output_sharding (which trips
+        shlo_clean_for_xla_ingestion)."""
+        if not self.enable_tensor_parallel:
+            return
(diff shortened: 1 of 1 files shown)


mmanzoorTT (approved):

sshonTT:
nightly pass : https://github.com/tenstorrent/tt-xla/actions/runs/26791377105

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1368,6 +1368,31 @@ def _get_model_inputs(
...
+        otherwise warmup shards these inputs while execution leaves them
+        replicated, and the backbone recompiles at the first real step for any
+        shardable batch. Also keeps large prefill traces from splitting into a
+        secondary sync that drops mhlo.spmd_output_sharding (which trips
+        shlo_clean_for_xla_ingestion)."""
+        if not self.enable_tensor_parallel:
+            return
+        # 2D mesh: model batch dim -> "batch" axis (data-parallel).
+        # 1D mesh (1, N): "batch" axis is size 1, fall back to "model".
+        batch_axis = "batch" if self.use_2d_mesh else "model"
```
AleksKnezevic, line 1389: shouldn't batch always be on "batch" axis?

mmanzoorTT: 1D sharding is using `(Batch, model)=[1, n]`; Sharding on `batch` axis will have no effect for 1D mesh. That's why we are sharding on `model` axis. 
Once we completely add DP/TP support; then 1D mesh for DP will use (n, 1) shape and we can always use `batch` axis for sharding inputs. .

sshonTT: When we use a 1D mesh but still want to shard on the batch dimension, we can use the 'model' axis. That way, we can also shard the KV cache.

However, these contextual axis names (like 'model' or 'batch') can be misleading, as they imply a restriction to specific tensor types. In practice, we should be able to shard any resource dynamically based on our needs. Therefore, I suggest switching to neutral names like 'axis0' or 'axis1' to avoid this confusion. Let me know your thoughts, and I can open a following PR.

PR #5043, opened 2026-06-02
Add example for stable diffusion 1 5 model

### LINKED ISSUE #5042: Add demo for stable diffusion 1.5 model
stable diffusion 1.5 model is running e2e so we need to add example for the model

ctr-lelanchelianTT:
### Ticket
closes https://github.com/tenstorrent/tt-xla/issues/5042

### Problem description
Stable diffusion 1 5 model is passing e2e so we need to add example for this model 

### What's changed
- Adds `examples/pytorch/sd_v1_5_pipeline.py` for Stable Diffusion 1.5 (`stable-diffusion-v1-5/stable-diffusion-v1-5`), mirroring the existing `sd_v1_4_pipeline.py`.
- UNet runs on the TT backend; VAE + PNDM scheduler on CPU; CFG denoising loop; 512x512 output.

### Checklist

I have attached the log for your reference:
[sd_v1_5.log](https://github.com/user-attachments/files/28495121/sd_v1_5.log)

output image:
test_sd15_output

### DIFF
--- a/examples/pytorch/sd_v1_5_pipeline.py
+++ b/examples/pytorch/sd_v1_5_pipeline.py
@@ -1,0 +1,48 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Runnable Stable Diffusion 1.5 text-to-image example on Tenstorrent.
+
+The reusable pipeline implementation is shared with the image-gen benchmark
+(``tests/benchmark/test_imagegen.py``) and the nightly test, and lives in
+``tt_forge_models`` — this script is just a thin runnable demo so the
+implementation isn't duplicated in ``examples/``.
+
+The UNet (the heavy net) runs on the Tenstorrent backend via
+``torch.compile(backend="tt")``; the precision-sensitive CLIP text encoder, the
+scheduler and the VAE run on CPU.
+"""
+
+from third_party.tt_forge_models.stable_diffusion_1_5.pytorch.pipeline import (
+    SD15Config,
+    SD15Pipeline,
+    save_image,
+)
+
+PROMPT = "a photo of a cat"
+NEGATIVE_PROMPT = ""
+CFG_SCALE = 7.5
+NUM_INFERENCE_STEPS = 50
+SEED = 42
+OUTPUT_PATH = "sd_v1_5_output.png"
+
+
+def main():
+    pipeline = SD15Pipeline(config=SD15Config())
+    pipeline.setup()
+
+    image = pipeline.generate(
+        prompt=PROMPT,
+        negative_prompt=NEGATIVE_PROMPT,
+        cfg_scale=CFG_SCALE,
+        num_inference_steps=NUM_INFERENCE_STEPS,
+        seed=SEED,
+    )
+
+    save_image(image, OUTPUT_PATH)
+    print(f"Saved output image to {OUTPUT_PATH}")
+
+
+if __name__ == "__main__":
+    main()


meenakshiramanathan1 (approved):

examples/pytorch/sd_v1_5_pipeline.py
```diff
@@ -0,0 +1,263 @@
...
+    """Pipeline for text-to-image generation with Stable Diffusion 1.5."""
+
+    def __init__(self, config: SD15Config):
+        self.config = config
+        self.device = config.device
+        self.model_id = config.model_id
+        self.latents_width = config.latents_width
+        self.latents_height = config.latents_height
+        self.vae_on_tt = config.vae_on_tt
+        self.clip_on_tt = config.clip_on_tt
```
vkovacevicTT: Why do we unpack these when we have them all in config?

examples/pytorch/sd_v1_5_pipeline.py
```diff
@@ -0,0 +1,263 @@
...
+    def load_models(self):
+        self.vae = AutoencoderKL.from_pretrained(
+            self.model_id,
+            subfolder="vae",
+            torch_dtype=torch.float32,
+            device_map=self.device,
+        )
+        if self.vae_on_tt:
+            self.vae.compile(backend="tt")
+            self.vae = self.vae.to(xm.xla_device())
```
vkovacevicTT: `self.device` here is a bit misleading.
We default it to "cpu", we have `self.vae_on_tt` to signal we want it on TT device, and then we use `xm.xla_device()`, which beats the purpose of having a device attribute.

examples/pytorch/sd_v1_5_pipeline.py
```diff
@@ -0,0 +1,263 @@
...
+import torch_xla
+import torch_xla.core.xla_model as xm
+import torch_xla.runtime as xr
+from diffusers import AutoencoderKL, PNDMScheduler, UNet2DConditionModel
+from PIL import Image
+from transformers import CLIPTextModel, CLIPTokenizer
+
+
+class SD15Config:
+    def __init__(self, device="cpu", vae_on_tt=False, clip_on_tt=True):
```
vkovacevicTT: Do we want `clip_on_tt=True` by default?

examples/pytorch/sd_v1_5_pipeline.py
```diff
@@ -0,0 +1,263 @@
...
+
+            # --- Text encoding (CLIP) ---
+            negative_prompt = negative_prompt or ""
+
+            cond_tokens = self.tokenizer(
+                [prompt], padding="max_length", max_length=77
+            ).input_ids
+            uncond_tokens = self.tokenizer(
+                [negative_prompt], padding="max_length", max_length=77
+            ).input_ids
```
vkovacevicTT: Does this work when batch_size > 1?

examples/pytorch/sd_v1_5_pipeline.py
```diff
@@ -0,0 +1,263 @@
...
+        """Generate an image from a text prompt. Returns tensor (B, 3, H, W)."""
+
+        batch_size = 1 if isinstance(prompt, str) else len(prompt)
+
+        tt_cast = lambda x: (
+            x.to(dtype=torch.bfloat16).to(device=xm.xla_device())
+            if x.device == torch.device("cpu")
+            else x.to(dtype=torch.bfloat16)
+        )
+        cpu_cast = lambda x: x.to("cpu").to(dtype=torch.float16)
```
vkovacevicTT: Do we need this if in `tt_cast`?

examples/pytorch/sd_v1_5_pipeline.py
```diff
@@ -0,0 +1,263 @@
...
+            # --- Denoising loop (UNet on TT) ---
+            for i, timestep in enumerate(self.scheduler.timesteps):
+
+                model_input = torch.cat([latents] * 2)
+                model_input = self.scheduler.scale_model_input(model_input, timestep)
+
+                # CPU → TT
+                model_input = tt_cast(model_input)
+                timestep_tt = tt_cast(timestep.unsqueeze(0))
+                encoder_hidden_states = tt_cast(encoder_hidden_states)
```
vkovacevicTT: Do `encoder_hidden_states` change across loops?

vkovacevicTT (commented):
I don't mind this being in the `examples/` dir as is, but since two benchmark PRs (https://github.com/tenstorrent/tt-xla/pull/5085, https://github.com/tenstorrent/tt-xla/pull/5044) depend on this implementation I would suggest some changes.

First, there is a lot of code duplication between this PR and https://github.com/tenstorrent/tt-xla/pull/5075 and https://github.com/tenstorrent/tt-xla/pull/5074. 
Let's consider restructuring them and extracting utilities.

Second, we should consider moving the model pipeline implementations to `tt_forge_models`. 
Keep in mind that we do need some flexibility, like passing compile configs e.g. optimization-level specified [here](https://github.com/tenstorrent/tt-xla/pull/5044/changes#diff-7e62f3e7e051e48ba52d7a5d66608661cb4f42a3dc29a91b04ccc0fb0a87fc49R350-R356). 
If moving the whole pipeline to `tt_forge_models` makes things more complicated, we could implement some main model pipeline functionalities (like `generate` function) in the benchmarks (and examples).
I just think it's important we don't import anything from `examples/` dir in the benchmarks.

examples/pytorch/sd_v1_5_pipeline.py
sgligorijevicTT: I 100% agree, please address this

examples/pytorch/sd_v1_5_pipeline.py
sgligorijevicTT: This should be hoisted out of the loop, and perhaps some of these casts can be avoided to make the code cleaner

examples/pytorch/sd_v1_5_pipeline.py
```diff
@@ -0,0 +1,263 @@
...
+                    timestep_tt,
+                    encoder_hidden_states,
+                ).sample
+
+                # TT → CPU
+                unet_output = cpu_cast(unet_output)
+
+                # CFG blending (CPU)
+                uncond_output, cond_output = unet_output.chunk(2)
+                model_output = uncond_output + (cond_output - uncond_output) * cfg_scale
```
sgligorijevicTT: Is there a need to do CFG on the CPU?

examples/pytorch/sd_v1_5_pipeline.py
```diff
@@ -0,0 +1,263 @@
...
+
+                # Scheduler step (CPU)
+                latents = cpu_cast(latents)
+                latents = self.scheduler.step(
+                    model_output, timestep, latents
```
sgligorijevicTT: For my understanding, why does the scheduler step need to take in latents?
Also, does this have to happen on the CPU?

examples/pytorch/sd_v1_5_pipeline.py
ctr-lelanchelianTT: Restructured: the implementation now lives in `tt_forge_models` (companion tenstorrent/tt-forge-models#720) and this example is a thin wrapper that imports it. In the canonical pipeline I dropped the misleading `device` unpack; the remaining locals are just readability aliases for config fields.

examples/pytorch/sd_v1_5_pipeline.py
ctr-lelanchelianTT: Addressed in the `tt_forge_models` pipeline: removed the `device` config attr entirely and use the explicit `"cpu"` literal for host placement. TT placement stays separate via `.compile(backend="tt")` + `xm.xla_device()`, so there is no longer a misleading "device" attribute.

examples/pytorch/sd_v1_5_pipeline.py
ctr-lelanchelianTT: Good catch — flipped the default to `clip_on_tt=False`. CLIP text embeddings are precision-sensitive and running CLIP on TT in bf16 can wash out the prompt conditioning, so it is now opt-in.

examples/pytorch/sd_v1_5_pipeline.py
ctr-lelanchelianTT: It is written/validated for single-prompt generation only. Rather than silently mishandle batches, I added an explicit `assert batch_size == 1` in the pipeline.

examples/pytorch/sd_v1_5_pipeline.py
ctr-lelanchelianTT: Right — removed the redundant cast; `tt_cast` already handles the dtype + device move.

examples/pytorch/sd_v1_5_pipeline.py
ctr-lelanchelianTT: They do not change across the loop — hoisted the `encoder_hidden_states` cast out of the denoising loop so it runs once before the loop.

examples/pytorch/sd_v1_5_pipeline.py
ctr-lelanchelianTT: Done — removed the `device` attr in the `tt_forge_models` pipeline and use the explicit `"cpu"` literal for host placement.

examples/pytorch/sd_v1_5_pipeline.py
ctr-lelanchelianTT: Hoisted the constant `encoder_hidden_states` cast out of the loop (runs once now) and removed the redundant per-step latents cast.

examples/pytorch/sd_v1_5_pipeline.py
ctr-lelanchelianTT: CFG blending is a cheap elementwise op on the small noise-pred tensors; running it on CPU between the TT UNet forwards avoids round-tripping tiny tensors to the device. It can live on TT but the gain is negligible, so it stays on CPU (added a comment noting this).

examples/pytorch/sd_v1_5_pipeline.py
ctr-lelanchelianTT: The diffusers scheduler (PNDM here) is stateful and runs on CPU; `step(model_output, t, latents)` is the upstream API — it consumes the current latents to produce the next sample. Keeping it on CPU matches upstream numerics exactly; only the heavy UNet forward is offloaded to TT.

ctr-lelanchelianTT:
Thanks for the review  addressed the comments

sgligorijevicTT (approved):

PR #5044, opened 2026-06-02
[Playground v2.5] Wire e2e pipeline into nightly and benchmark CI

### LINKED ISSUE #5047: [Playground v2.5,SD3 and Bria] Wire e2e pipeline into nightly and benchmark CI

kamalrajkannan78:
### Ticket

- closes https://github.com/tenstorrent/tt-xla/issues/5047

### Problem description

Image-gen Model that passes bringup should be added to nightly + benchmark CI. Playground v2.5's all 4 components are passing but the e2e pipeline is still a manually-invoked example with no CI coverage.

### What's changed

Benchmark structure mirrors the `imagegen_benchmark` pattern introduced in https://github.com/tenstorrent/tt-xla/pull/5085 (per @VladanKovacevic's review): shared harness in `benchmarks/`, per-model config in `test_imagegen.py`.

- **Nightly** (`tests/torch/models/playground_v2_5/test_playground_v2_5_pipeline.py`): self-contained — pipeline class + `run_playground_v25_pipeline` helper + test function all inline (no import from `examples/`). Markers: `nightly`, `model_test`, `single_device`, `large`, `record_test_properties`.
- **Benchmark** (`tests/benchmark/test_imagegen.py::test_playground_v2_5`): config-driven entry; all 4 components on TT; `optimization_level=0` (with VAE switching to 1 inline, **merged with harness options** instead of overwriting). Two-pass harness: warmup (1 step — triggers compile of every component) + steady-state (full 50 steps, all cache hits). Emits per-component steady-state measurements: `text_encoder_1_s`, `text_encoder_2_s`, `unet_step_mean_s`, `vae_s`, `cpu_overhead_s`, `e2e_latency`, `images_per_second`.
- **Benchmark pipeline class** (`tests/benchmark/benchmarks/playground_v2_5_pipeline.py`): per @VladanKovacevic's review, the benchmark has its own pipeline (no import from `examples/`). The same self-contained pattern is applied to the nightly test; the previous `examples/pytorch/playground_v2_5_pipeline.py` demo file is removed since neither test path imports from it anymore.
- **Matrix entry** added to `.github/workflows/perf-bench-matrix.json`.
- **xfail marker** (both nightly + benchmark) for https://github.com/tenstorrent/tt-xla/issues/5176 — VAE compile TT_FATAL on current main after the recent uplift. PR code verified correct on a pre-uplift commit: https://github.com/tenstorrent/tt-xla/actions/runs/27344277516/job/80792035664.

### Checklist
- [x] Validate benchmark test on N150 — https://github.com/tenstorrent/tt-xla/actions/runs/27337975698

### DIFF
--- a/tests/benchmark/benchmarks/imagegen_benchmark.py
+++ b/tests/benchmark/benchmarks/imagegen_benchmark.py
@@ -1,0 +1,198 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Generic text-to-image (diffusion) benchmark harness for torch-xla / TT.
+
+Mirrors the structure of ``vision_benchmark.py``: the per-model configuration
+lives in ``test_imagegen.py`` and this module owns the reusable measurement
+logic. Diffusion pipelines don't fit the single-forward vision harness — each
+generation is a multi-step denoising loop — so this harness uses a two-pass
+scheme:
+
+  - Pass 1 (warmup): a single-step ``generate()`` call — enough to trigger
+    the first-forward compile of every component.
+  - Pass 2 (steady-state): a full ``generate(num_inference_steps)`` call;
+    every forward is a cache hit. This is the pass whose image is saved and
+    whose latency drives the reported throughput.
+
+Per-model wiring provides a ``build_pipeline_fn`` that returns
+``(pipeline, generate_fn)``; this module sets the XLA compile options,
+builds the pipeline (which compiles the heavy net for TT), runs the two passes
+and emits a standardized benchmark result.
+"""
+
+import socket
+import time
+
+import torch_xla
+import torch_xla.runtime as xr
+from utils import (
+    build_xla_export_name,
+    create_benchmark_result,
+    get_benchmark_metadata,
+    get_xla_device_arch,
+    print_benchmark_results,
+    save_image,
+)
+
+xr.set_device_type("TT")
+
+MODULE_EXPORT_PATH = "modules"
+
+
+def benchmark_imagegen_torch_xla(
+    build_pipeline_fn,
+    model_info_name,
+    prompt,
+    num_inference_steps,
+    height,
+    width,
+    optimization_level,
+    trace_enabled,
+    ttnn_perf_metrics_output_file,
+    display_name=None,
+    output_image_path=None,
+):
+    """Benchmark a text-to-image diffusion pipeline on the TT backend.
+
+    Args:
+        build_pipeline_fn: ``build_pipeline_fn(compile_options) -> (pipeline, generate_fn)``.
+            ``compile_options`` is forwarded so the pipeline can merge instead
+            of overwriting if it needs to switch any option inline.
+            ``generate_fn(prompt, num_inference_steps) -> image tensor (B, 3, H, W)``
+            runs one full text-to-image generation.
+        model_info_name: Model name for identification and reporting.
+        prompt: Text prompt to generate from.
+        num_inference_steps: Number of denoising steps per generation.
+        height, width: Output image dimensions.
+        optimization_level: tt-mlir optimization level for compilation.
+        trace_enabled: Whether to enable tracing.
+        ttnn_perf_metrics_output_file: Base path for TTNN perf metrics files.
+        display_name: Display name used for export naming / dashboard.
+        output_image_path: If set, the steady-state image is saved here.
+
+    Returns:
+        Standardized benchmark result dict (see ``create_benchmark_result``).
+    """
+    export_model_name = build_xla_export_name(
+        model_name=display_name or model_info_name,
+        num_layers=None,
+        batch_size=1,
+        input_sequence_length=None,
+    )
+
+    options = {
+        "optimization_level": optimization_level,
+        "export_path": MODULE_EXPORT_PATH,
+        "export_model_name": export_model_name,
+        "ttnn_perf_metrics_enabled": True,
+        "ttnn_perf_metrics_output_file": ttnn_perf_metrics_output_file,
+        "enable_trace": trace_enabled,
+    }
+    torch_xla.set_custom_compile_options(options)
+
+    # Build + compile the pipeline (heavy net registers the "tt" backend and is
+    # moved to the XLA device here; actual kernel compilation happens lazily on
+    # the first forward, i.e. during the warmup pass below).
+    pipeline, generate_fn = build_pipeline_fn(options)
+
+    # Pass 1 (warmup): 1 step is enough to trigger the first-forward compile
+    # of every component.
+    print("Starting warmup pass (includes compile)...")
+    warmup_start = time.perf_counter()
+    generate_fn(prompt, 1)
+    warmup_time = time.perf_counter() - warmup_start
+    print(f"Warmup pass: {warmup_time:.3f}s")
+
+    # Pass 2 (steady-state): steady-state generation; this image is the saved one.
+    print("Starting steady-state pass...")
+    steady_state_start = time.perf_counter()
+    steady_state_image = generate_fn(prompt, num_inference_steps)
+    steady_state_time = time.perf_counter() - steady_state_start
+    print(f"Steady-state pass: {steady_state_time:.3f}s")
+
+    if output_image_path is not None:
+        save_image(steady_state_image, output_image_path)
+        print(f"Saved output image to {output_image_path}")
+
+    # Throughput is reported on the steady-state pass. One image per run.
+    total_samples = 1
+    samples_per_sec = total_samples / steady_state_time
+
+    # Per-component forward+sync times from the pipeline's own instrumentation
+    # (steady-state pass).
+    perf = pipeline._perf
+    unet_steps = perf["unet_steps"]
+    unet_step_mean_s = sum(unet_steps) / len(unet_steps)
+    tt_components_total = perf["te1"] + perf["te2"] + sum(unet_steps) + perf["vae"]
+    cpu_overhead = max(0.0, perf["total"] - tt_components_total)
+
+    metadata = get_benchmark_metadata()
+    full_model_name = model_info_name
+    model_type = "Image Generation, Text-to-Image"
+    dataset_name = "Text Prompt"
+    input_size = (3, height, width)
+
+    print_benchmark_results(
+        model_title=full_model_name,
+        full_model_name=full_model_name,
+        model_type=model_type,
+        dataset_name=dataset_name,
+        date=metadata["date"],
+        machine_name=metadata["machine_name"],
+        total_time=steady_state_time,
+        total_samples=total_samples,
+        samples_per_sec=samples_per_sec,
+        evaluation_score=0.0,
+        batch_size=1,
+        data_format="bfloat16",
+        input_size=input_size,
+    )
+    print(
+        f"| Num inference steps: {num_inference_steps}\n"
+        f"| Steady-state:\n"
+        f"|   Text encoder 1 (s):  {perf['te1']:.3f}\n"
+        f"|   Text encoder 2 (s):  {perf['te2']:.3f}\n"
+        f"|   UNet step mean (s):  {unet_step_mean_s:.3f}\n"
+        f"|   VAE (s):             {perf['vae']:.3f}\n"
+        f"|   CPU overhead (s):    {cpu_overhead:.3f}"
+    )
+
+    custom_measurements = [
+        {"measurement_name": "images_per_second", "value": samples_per_sec},
+        {"measurement_name": "e2e_latency", "value": steady_state_time},
+        {"measurement_name": "text_encoder_1_s", "value": perf["te1"]},
+        {"measurement_name": "text_encoder_2_s", "value": perf["te2"]},
+        {"measurement_name": "unet_step_mean_s", "value": unet_step_mean_s},
+        {"measurement_name": "vae_s", "value": perf["vae"]},
+        {"measurement_name": "cpu_overhead_s", "value": cpu_overhead},
+    ]
+
+    result = create_benchmark_result(
+        full_model_name=full_model_name,
+        model_type=model_type,
+        dataset_name=dataset_name,
+        num_layers=-1,
+        batch_size=1,
+        input_size=input_size,
+        loop_count=num_inference_steps,
+        data_format="bfloat16",
+        total_time=steady_state_time,
+        total_samples=total_samples,
+        evaluation_score=0.0,
+        custom_measurements=custom_measurements,
+        optimization_level=optimization_level,
+        program_cache_enabled=True,
+        trace_enabled=trace_enabled,
+        model_info=model_info_name,
+        display_name=display_name,
+        torch_xla_enabled=True,
+        backend="tt",
+        device_name=socket.gethostname(),
+        arch=get_xla_device_arch(),
+        device_count=xr.global_runtime_device_count(),
+        input_is_image=True,
+    )
+
+    return result
--- a/tests/benchmark/benchmarks/playground_v2_5_pipeline.py
+++ b/tests/benchmark/benchmarks/playground_v2_5_pipeline.py
@@ -2,25 +2,24 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-"""Playground v2.5 — end-to-end TT pipeline example.
+"""Playground v2.5 — benchmark-side pipeline for the imagegen harness.
 
 Each nn.Module component (text_encoder, text_encoder_2, unet, vae) can be
 independently moved to Tenstorrent via
 `model.compile(backend="tt") + model.to(xla_device())`. Tokenizer and
 scheduler always stay on CPU. CPU→TT→CPU device switching is done inline at
-the call site of each nn component.
+the call site of each nn component. Per-component forward+sync times are
+collected into `self._perf` for the harness to read after each generate() call.
 """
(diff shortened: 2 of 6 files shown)


vvukomanTT:
More of a general test infra question, do we want to use parts of the `examples/` test infra in benchmarks?
@kamalrajkannan78 @vkovacevicTT

nsumrakTT (approved):

kamalrajkannan78:
> More of a general test infra question, do we want to use parts of the `examples/` test infra in benchmarks? @kamalrajkannan78 @vkovacevicTT

I followed the same structure as @ctr-lelanchelianTT 's PR #5085 (which you approved) — that PR establishes the pattern of importing the pipeline class from `examples/pytorch/<model>_pipeline.py` for image-gen benchmarks.

The individual components could be loaded from `third_party/tt_forge_models/playground_v2_5/...` directly , but the pipeline's orchestration (CFG, scheduler stepping, CPU↔TT moves per component, opt_level switching for VAE) already lives in the demo. Importing the pipeline class avoids duplicating that orchestration logic in the benchmark, and keeps the demo and benchmark in sync as a single source of truth.

Happy to switch to a different structure if you'd prefer a stricter separation between `examples/` and benchmark code.

vvukomanTT:
> > More of a general test infra question, do we want to use parts of the `examples/` test infra in benchmarks? @kamalrajkannan78 @vkovacevicTT
> 
> I followed the same structure as @ctr-lelanchelianTT 's PR #5085 (which you approved) — that PR establishes the pattern of importing the pipeline class from `examples/pytorch/<model>_pipeline.py` for image-gen benchmarks.
> 
> The individual components could be loaded from `third_party/tt_forge_models/playground_v2_5/...` directly , but the pipeline's orchestration (CFG, scheduler stepping, CPU↔TT moves per component, opt_level switching for VAE) already lives in the demo. Importing the pipeline class avoids duplicating that orchestration logic in the benchmark, and keeps the demo and benchmark in sync as a single source of truth.
> 
> Happy to switch to a different structure if you'd prefer a stricter separation between `examples/` and benchmark code.

No need, this was just a thought, something to think about. 
From the CI perspective this looks good 👍

vvukomanTT (approved):

vkovacevicTT (changes requested):
@kamalrajkannan78 @vvukomanTT 
I am also against importing model implementations from examples to benchmarks.
Checkout my comment on one of the `examples/` implementation dependencies for more details: https://github.com/tenstorrent/tt-xla/pull/5043#pullrequestreview-4453984071

kamalrajkannan78:
> @kamalrajkannan78 @vvukomanTT I am also against importing model implementations from examples to benchmarks. Checkout my comment on one of the `examples/` implementation dependencies for more details: [#5043 (review)](https://github.com/tenstorrent/tt-xla/pull/5043#pullrequestreview-4453984071)

Thanks @VladanKovacevic — agreed. Moved the pipeline class into `tests/benchmark/benchmarks/playground_v2_5_pipeline.py` so the benchmark no longer imports from `examples/`. Also applied your line-level cleanups from #5043 (config-field unpacking, batch_size assert, hoisting loop-invariant casts). PR updated.

tests/benchmark/test_imagegen.py
```diff
@@ -0,0 +1,136 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Text-to-image (diffusion) benchmarks.
+
+Config-driven entry points (one ``test_<model>`` per model) that reuse the
+end-to-end pipelines from ``examples/pytorch`` and drive them through the shared
```
vkovacevicTT: Pipeline implementation is moved to benchmarks, so we don't reuse anything from `examples/pytorch` anymore. Please modify this.

tests/benchmark/test_imagegen.py
```diff
@@ -0,0 +1,136 @@
...
+    prompt = "Astronaut in a jungle, cold color palette, muted colors, detailed, 8k"
+    num_inference_steps = 50
+    height = width = 1024
+
+    def build_pipeline_fn():
+        # All four components (text_encoder, text_encoder_2, unet, vae) run on TT.
+        pipeline = PlaygroundV25Pipeline(config=PlaygroundV25Config(device="cpu"))
+        pipeline.setup()
+
+        def generate_fn(steps):
```
vkovacevicTT: Add prompt to func arguments `generate_fn(prompt, steps)` so the default prompt above doesn't get baked into it.

tests/benchmark/test_imagegen.py
```diff
@@ -0,0 +1,136 @@
...
+            return pipeline.generate(
+                prompt=prompt,
+                negative_prompt=None,
+                cfg_scale=3.0,
+                num_inference_steps=steps,
+                seed=DEFAULT_SEED,
+            )
+
+        return pipeline, generate_fn, save_image
```
vkovacevicTT: If `save_image` is a generic function same for all models, we can move it to utility script and import it from there, not return it as a part of `build_pipeline_fn`.

tests/benchmark/benchmarks/imagegen_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+
+    options = {
+        "optimization_level": optimization_level,
+        "export_path": MODULE_EXPORT_PATH,
+        "export_model_name": export_model_name,
+        "ttnn_perf_metrics_enabled": True,
+        "ttnn_perf_metrics_output_file": ttnn_perf_metrics_output_file,
+        "enable_trace": trace_enabled,
+    }
+    torch_xla.set_custom_compile_options(options)
```
vkovacevicTT, line 93: Careful, this gets overwritten in pipeline if `vae_on_tt` is true!

tests/benchmark/benchmarks/imagegen_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+
+    # Build + compile the pipeline (heavy net registers the "tt" backend and is
+    # moved to the XLA device here; actual kernel compilation happens lazily on
+    # the first forward, i.e. during the cold pass below).
+    pipeline, generate_fn, save_fn = build_pipeline_fn()
+
+    # Pass 1 (cold): first generation, includes first-forward compile.
```
vkovacevicTT: nit: Rename "cold" and "warm", since "warm" can get mixed with "warmup" (which is actually "cold" first generation) :D

tests/benchmark/benchmarks/imagegen_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+    print("Starting cold pass (includes compile)...")
+    cold_start = time.perf_counter()
+    generate_fn(num_inference_steps)
```
vkovacevicTT: This is heavy work, let's not run it `num_inference_steps` here.
1 step should be sufficient.

tests/benchmark/benchmarks/imagegen_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+
+    # Pass 2 (warm): steady-state generation; this image is the saved one.
+    print("Starting warm pass (steady-state)...")
+    warm_start = time.perf_counter()
+    warm_image = generate_fn(num_inference_steps)
+    warm_time = time.perf_counter() - warm_start
+    print(f"Warm pass: {warm_time:.3f}s")
+
+    if output_image_path is not None and save_fn is not None:
+        save_fn(warm_image, output_image_path)
```
vkovacevicTT: Same comment as in test_imagegen.py - let's have save_fn as a utility if it's the same save_image function for all models.

tests/benchmark/benchmarks/imagegen_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+        print(f"Saved output image to {output_image_path}")
+
+    # Throughput is reported on the steady-state (warm) pass. One image per run.
+    total_samples = 1
+    samples_per_sec = total_samples / warm_time
+    per_step_latency_ms = (warm_time / num_inference_steps) * 1000.0
```
vkovacevicTT: This is an average of all inference steps + encoders + vae work.
Let's obtain these numbers from `generate` function for more fine-grained measurements.

tests/benchmark/benchmarks/imagegen_benchmark.py
```diff
@@ -0,0 +1,185 @@
...
+    # First-forward compile overhead, approximated as the cold/warm delta.
+    compile_time = max(0.0, cold_time - warm_time)
```
vkovacevicTT: `cold_time` contains other stuff apart from model compilation, so I think this is not a good approximation.

Let's remove this, since we don't measure and report it anywhere else.

tests/benchmark/benchmarks/playground_v2_5_pipeline.py
```diff
@@ -0,0 +1,427 @@
...
+SEED = 42
+GUIDANCE_SCALE = 3.0
+HEIGHT = 1024
+WIDTH = 1024
+
+
+class PlaygroundV25Config:
+    def __init__(
+        self,
+        device: str = "cpu",
```
vkovacevicTT: Let's remove this, since we can always explicitly use "cpu" when we need to.
device should refer to a tt device.

tests/benchmark/benchmarks/playground_v2_5_pipeline.py
```diff
@@ -0,0 +1,427 @@
...
+        text_encoder_on_tt: bool = True,
+        text_encoder_2_on_tt: bool = True,
+        unet_on_tt: bool = True,
+        vae_on_tt: bool = True,
```
vkovacevicTT, line 41: Do we want all of this on tt device? Other similar benchmarks mentioned in the description https://github.com/tenstorrent/tt-xla/pull/5085 default everything to cpu except the heavy unet work.

tests/benchmark/benchmarks/playground_v2_5_pipeline.py
```diff
@@ -0,0 +1,427 @@
...
+        num_inference_steps: int = 50,
+        seed: Optional[int] = None,
+    ) -> torch.Tensor:
+        assert isinstance(
+            prompt, str
+        ), "Only single-prompt generation (batch_size=1) is tested for now"
+        batch_size = 1
+
+        tt_cast = lambda x: x.to(device=xm.xla_device())
+        cpu_cast = lambda x: x.to("cpu")
```
vkovacevicTT: We can drop this and use .to(device) and .to("cpu") when needed (device here meaning tt device).

tests/benchmark/benchmarks/playground_v2_5_pipeline.py
```diff
@@ -0,0 +1,427 @@
...
+                if self.config.unet_on_tt:
+                    noise_pred = cpu_cast(noise_pred).to(torch.float32)
+
+                # CFG + scheduler step
+                uncond, text = noise_pred.chunk(2)
+                noise_pred = uncond + cfg_scale * (text - uncond)
+
+                latents = self.scheduler.step(
+                    noise_pred, t, latents, return_dict=False
+                )[0]
```
vkovacevicTT, line 325: Let's measure per-iteration times here and return them, for more fine-grained perf reports and averaging.

tests/benchmark/benchmarks/playground_v2_5_pipeline.py
```diff
@@ -0,0 +1,427 @@
...
+                .to(latents.device, latents.dtype)
+            )
+            scaling_factor = self.vae.vae.config.scaling_factor
+            latents = latents * latents_std / scaling_factor + latents_mean
+
+            # opt_level=1 keeps ttir.group_norm -> ttnn.group_norm; opt_level=0
+            # decomposes GroupNorm which can OOM on the VAE.
+            # See: https://github.com/tenstorrent/tt-xla/issues/4710
+            if self.config.vae_on_tt:
+                torch_xla.set_custom_compile_options({"optimization_level": 1})
```
vkovacevicTT: This overwrites other compile options that are set in benchmarks:
https://github.com/tenstorrent/tt-xla/pull/5044/changes#diff-738c8132015b5f24120f5b9495fada5f63e04ff0362e42ddd3cf7643880911f1R85-R92

vkovacevicTT (changes requested):
Nice! 
Great to see benchmark passing in CI :)

Let's address some few minor issues before proceeding.

tests/benchmark/benchmarks/playground_v2_5_pipeline.py
kamalrajkannan78: @nvukobratTT  said heavy-component-only on TT is fine for starters, but since all 4 Playground v2.5 components are functionally passing now, I'd prefer to keep them all on TT for the benchmark — it gives a true "end-to-end TT" measurement

tests/benchmark/benchmarks/imagegen_benchmark.py
kamalrajkannan78: Addressed — pipeline now merges via `{**self.config.compile_options, "optimization_level": 1}` instead of overwriting. Restored after VAE.

tests/benchmark/benchmarks/playground_v2_5_pipeline.py
kamalrajkannan78: Addressed — per-iteration times are measured into `self._perf["unet_steps"]` (a list, populated every loop iteration), and the harness uses them for both averaging (`unet_step_mean_ms`) and per-component compile (`unet_compile_s = warmup.unet_steps[0] − mean(steady.unet_steps)`).

tests/benchmark/benchmarks/imagegen_benchmark.py
kamalrajkannan78: Context: @nvukobratTT asked for overall compile time + per-component steady-state numbers. I think it's worth surfacing both in the benchmark run — please let me know if you'd prefer to keep just one.

**Per-component steady-state.** Since all 4 components run on TT, I extended the per-component steady-state breakdown to cover all of them — text_encoder 1, text_encoder 2, UNet (per-step mean), and VAE — not just UNet alone.

**Compile time methodology.** Addressed your "not a good approximation" concern. The old formula was `warmup_total − steady_state_total`, which compared two different workloads (1-step warmup vs 50-step steady-state) and was effectively `compile − 49 steady-state UNet steps`.

New approach — per-component compile via like-for-like subtraction:

```python
text_encoder_1_compile = warmup.te1 − steady.te1
text_encoder_2_compile = warmup.te2 − steady.te2
unet_compile          = warmup.unet_steps[0] − mean(steady.unet_steps)
vae_compile           = warmup.vae − steady.vae
compile_time          = sum of the above
```

Each subtraction now compares the same workload (one warmup forward vs one steady-state forward of the same component), so no step-count mismatch. The harness snapshots pipeline._perf after the warmup pass — before the steady-state pass resets it — then computes the deltas.

tests/benchmark/benchmarks/playground_v2_5_pipeline.py
vkovacevicTT: Got it, thanks :)

tests/benchmark/benchmarks/imagegen_benchmark.py
vkovacevicTT: Yes, but a number of things happen in warmup apart from compilation - transfers and allocations (weights, kv_cache), trace capture, program cache population etc.

If you want to measure all of that, I'm fine with it, I just don't think we should report it as "compile_time".

@nvukobratTT what is the intention here?

vkovacevicTT (approved):
Nice!

tests/benchmark/benchmarks/imagegen_benchmark.py
nvukobratTT: @kamalrajkannan78, there should be a cleaner way to track compile time. MLIR has some internal systems for it. For now, the training team is exploring how to easily enable these across the line, so for now we can remove this logic from the benchmarking. Let's keep it simple and aligned with current benchmarking infra.

kamalrajkannan78:
FYI,  @nvukobratTT @vkovacevicTT @vvukomanTT 

Current main: pipeline fails with #5176 — recent uplift is most likely the cause.

Verified the PR code itself is correct by running the current changes on top of a pre-uplift commit — passes cleanly:
https://github.com/tenstorrent/tt-xla/actions/runs/27344277516/job/80792035664

Added `xfail` markers (nightly + benchmark) linking to #5176.

tests/benchmark/benchmarks/imagegen_benchmark.py
kamalrajkannan78: Sounds good — removed the compile_time logic. Kept only the per-component steady-state metrics, aligned with current benchmarking infra.

nvukobratTT (approved):

tests/torch/models/playground_v2_5/test_playground_v2_5_pipeline.py
```diff
@@ -0,0 +1,67 @@
...
+from pathlib import Path
+
+import pytest
+import torch_xla.runtime as xr
+from infra import RunMode
+from loguru import logger
+from PIL import Image
+from utils import BringupStatus, Category, ModelGroup
+
+from examples.pytorch.playground_v2_5_pipeline import (
```
sgligorijevicTT: I don't like the idea of importing from examples for the purposes of testing

sgligorijevicTT (approved):

PR #5053, opened 2026-06-02
Fix runtime errors encountered due to transformers uplift.

devisettymahidhar608:
### Ticket
Link to Github Issue

### Problem description
Fix runtime errors caused by the transformers uplift

The recent transformers uplift changed the HF call signatures and decode-time behavior for gpt_oss and olmo3, which surfaced two runtime problems in our PyTorch examples and the TT mask override

### What's changed
Updated the sliding-window mask override for the new HF signature. Newer transformers modeling code passes position_ids (instead of, or without, cache_position) when building attention masks. The override now accepts the new signature and gracefully derives the cache position when needed.

Log:
[mistral_8b.log](https://github.com/user-attachments/files/28341849/mistral_8b.log)
[olmo_7b.log](https://github.com/user-attachments/files/28341851/olmo_7b.log)
[olmo_32b.log](https://github.com/user-attachments/files/28341853/olmo_32b.log)
[gpt.log](https://github.com/user-attachments/files/28342749/gpt.log)

### Checklist

### DIFF
--- a/examples/pytorch/gpt_oss_20b.py
+++ b/examples/pytorch/gpt_oss_20b.py
@@ -231,6 +231,11 @@ def construct_inputs(
     override_cache_sliding_window_layers(static_cache, max_cache_len, sliding_window)
 
     cache_position: torch.Tensor = torch.arange(0, inputs.input_ids.shape[1])
+    # Pass position_ids explicitly so gpt_oss's forward never enters the
+    # "position_ids is None" branch (which calls past_key_values.get_seq_length()
+    # and bakes a per-step Python int into the graph, causing dynamo to
+    # recompile on every decode step).
+    position_ids = cache_position.unsqueeze(0)
 
     # Attention mask is needed to ignore padding tokens in left-padded batches. The mask should match max_cache_len
     # to prevent recompilation or implicit padding by transformers, which can cause degenerate output.
@@ -244,6 +249,7 @@ def construct_inputs(
         "input_ids": inputs.input_ids,
         "past_key_values": static_cache,
         "cache_position": cache_position,
+        "position_ids": position_ids,
         "use_cache": True,
         "attention_mask": full_attention_mask,
     }
@@ -282,8 +288,19 @@ def transfer_to_device(
     for layer in input_args["past_key_values"].layers:
         layer.keys = layer.keys.to(device)
         layer.values = layer.values.to(device)
+        # StaticLayer.update builds a fresh `cache_position` each call as
+        # `torch.arange(kv_length, device=self.device) + self.cumulative_length`,
+        # then passes it to `self.keys.index_copy_(2, cache_position, ...)`.
+        # If `self.device` is still "cpu" or `cumulative_length` is a CPU tensor,
+        # the resulting index is CPU and `index_copy_` on an XLA `keys` tensor
+        # raises `Check failed: xtensor: Input tensor is not an XLA tensor`.
+        if isinstance(getattr(layer, "cumulative_length", None), torch.Tensor):
+            layer.cumulative_length = layer.cumulative_length.to(device)
+        if hasattr(layer, "device"):
+            layer.device = device
     input_args["input_ids"] = input_args["input_ids"].to(device)
     input_args["cache_position"] = input_args["cache_position"].to(device)
+    input_args["position_ids"] = input_args["position_ids"].to(device)
     input_args["attention_mask"] = input_args["attention_mask"].to(device)
 
     model = model.to(device)
@@ -398,6 +415,8 @@ def run_generate(
             host_cache_pos = input_args["cache_position"].to("cpu")
             host_cache_pos = torch.tensor([host_cache_pos[-1:] + 1])
             input_args["cache_position"] = host_cache_pos.to(device)
+            # keep position_ids in sync with cache_position (see construct_inputs)
+            input_args["position_ids"] = host_cache_pos.unsqueeze(0).to(device)
 
     print()
     if not is_interactive:
--- a/python_package/tt_torch/transformers_overrides.py
+++ b/python_package/tt_torch/transformers_overrides.py
@@ -77,9 +77,10 @@ def override_cache_sliding_window_layers(
 def tt_create_sliding_window_causal_mask(
     config,
     inputs_embeds: torch.Tensor,
-    attention_mask: Optional[torch.Tensor],
-    cache_position: torch.Tensor,
+    attention_mask: Optional[torch.Tensor] = None,
+    cache_position: Optional[torch.Tensor] = None,
     past_key_values=None,
+    position_ids: Optional[torch.Tensor] = None,
     **kwargs,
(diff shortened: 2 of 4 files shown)


sonalibaskaran2499 (approved):

examples/pytorch/gpt_oss_20b.py
```diff
@@ -398,6 +415,11 @@ def run_generate(
             host_cache_pos = input_args["cache_position"].to("cpu")
             host_cache_pos = torch.tensor([host_cache_pos[-1:] + 1])
             input_args["cache_position"] = host_cache_pos.to(device)
+            # Mirror cache_position into position_ids so gpt_oss's forward skips
+            # its `position_ids is None` branch — that branch derives
+            # position_ids from past_key_values.get_seq_length() (a Python int)
+            # which dynamo specializes on, forcing a recompile every step.
+            input_args["position_ids"] = host_cache_pos.unsqueeze(0).to(device)
```
kmabeeTT, line 419: Same thing for the other 2 examples too - host_cache_pos is already 2D [1,1] here (from torch.tensor([... + 1])), so .unsqueeze(0) makes this [1,1,1]. Prefill builds position_ids as 2D [1, seq_len] — these should match. Should this just be host_cache_pos.to(device)? Works at batch=1 by broadcast luck, but would break for batch>1 I think

examples/pytorch/gpt_oss_20b.py
```diff
@@ -398,6 +415,11 @@ def run_generate(
             host_cache_pos = input_args["cache_position"].to("cpu")
             host_cache_pos = torch.tensor([host_cache_pos[-1:] + 1])
             input_args["cache_position"] = host_cache_pos.to(device)
+            # Mirror cache_position into position_ids so gpt_oss's forward skips
```
kmabeeTT: This block comment repeats the explanation already in _prepare_inputs. Can we keep the full "why" there and shorten this to # keep position_ids in sync with cache_position (see _prepare_inputs)? Same goes for the copy across the other two examples.

kmabeeTT (approved):
Approve but left 2 comments, neither blocking, but worth fixing.

examples/pytorch/gpt_oss_20b.py
devisettymahidhar608: addressed @kmabeeTT

examples/pytorch/gpt_oss_20b.py
jameszianxuTT: +1, I don't understand what this unsqueeze is for either

python_package/tt_torch/transformers_overrides.py
```diff
@@ -89,17 +90,42 @@ def tt_create_sliding_window_causal_mask(
...
     hardware as part of the model forward graph.
+
+    Accepts either ``cache_position`` (older HF call signature) or
+    ``position_ids`` (newer HF olmo3 / mistral / gpt_oss signature where
+    ``mask_kwargs`` no longer includes ``cache_position``). When only
+    ``position_ids`` is supplied, it is used as the cache position.
     """
+    if cache_position is None and position_ids is not None:
+        cache_position = position_ids[0] if position_ids.ndim == 2 else position_ids
+
```
jameszianxuTT, line 101: > When only `position_ids` is supplied, it is used as the cache position.

Is this a realistic scenario? It seems like a dangerous workaround

python_package/tt_torch/transformers_overrides.py
```diff
@@ -89,17 +90,42 @@ def tt_create_sliding_window_causal_mask(
...
     if past_key_values is None:
         return create_sliding_window_causal_mask(
             config,
             inputs_embeds,
             attention_mask,
             cache_position,
             past_key_values,
+            position_ids=position_ids,
```
jameszianxuTT, line 109: is it safe to call this function with positional arguments? it seems like the API is changed (not sure if this is call to upstream transfomers impl)

https://github.com/huggingface/transformers/blob/d3f05911abb216047a00aba79c8543c73633db05/src/transformers/masking_utils.py#L1113-L1120 ?

python_package/tt_torch/transformers_overrides.py
```diff
@@ -89,17 +90,42 @@ def tt_create_sliding_window_causal_mask(
...
             **kwargs,
         )
 
+    # Some HF modeling files build `mask_kwargs` without either
+    # `cache_position` or `position_ids`, so mirror upstream
+    # `_preprocess_mask_arguments` and derive an absolute position from
```
jameszianxuTT: Can you call this directly?

examples/pytorch/gpt_oss_20b.py
devisettymahidhar608: Hi @kmabeeTT  @jameszianxuTT 
Initially, cache_position is created via torch.arange(0, seq_len), which is a 1‑D tensor. In the decode loop, when we read it back into host_cache_pos (and take the next position with reshape(-1)[-1:] + 1), it stays 1‑D [1]. Since the model's rotary embedding expects position_ids to be 2‑D [batch, seq], we apply unsqueeze(0) to turn the 1‑D [1] into a 2‑D [1, 1].

Thank You.

python_package/tt_torch/transformers_overrides.py
devisettymahidhar608: Hi @jameszianxuTT 

It's a real path, not hypothetical: olmo3/ministral build mask_kwargs with position_ids and no cache_position (gpt_oss passes neither, so it uses the get_seq_length() fallback). It's also the safer source here — the mask math only needs the query tokens' absolute positions, which position_ids carries exactly. Deriving from get_seq_length() is actually less accurate for our always-roll sliding cache, since get_seq_length() saturates at sliding_window once the buffer fills, while position_ids stays correct past the window. This also matches upstream, which marks cache_position as deprecated/unused and derives masks from position_ids. We take position_ids[0] because positions are shared across the batch for this single-buffer cache.

Thank You

python_package/tt_torch/transformers_overrides.py
devisettymahidhar608: Hi @jameszianxuTT 

We can't usefully call _preprocess_mask_arguments directly here: it's private, requires a layer_idx, returns a 7-tuple, and never returns cache_position — it only computes q_offset internally via past_key_values.get_seq_length(). So we'd still have to build arange(q_length) + q_offset ourselves, plus take on a dependency on a private API whose signature shifts across versions (same fragility as the create_sliding_window_causal_mask issue). This block already calls the relevant primitive — get_seq_length() — directly; I reworded the comment that misleadingly said it "mirrors" the upstream function.

Thank You

devisettymahidhar608:
Hi @kmabeeTT  @jameszianxuTT 
I have addressed all the comments. Could you please take another look?
Thank You

jameszianxuTT (approved):
Thanks for responses to comments.

PR #5070, opened 2026-06-03
Add prefill benchmark testing

agobeljicTT:
### Ticket
N/A

### Problem description
Our LLM benchmarks measured decode (with prefill folded in) but had no dedicated prefill-only benchmark. We want an isolated prefill latency (TTFT) + PCC measurement that can be configured and tracked separately from decode.

### What's changed
- Added a prefill benchmark (`benchmark_llm_prefill_torch_xla`) with `test_llm_prefill` / `test_llm_prefill_tp` drivers and per-model tests (Llama 3.2 1B, 3.1 8B, 3.1 70B TP, Phi-1, Qwen3 8B). Each model is swept over batch size × sequence length × optimization level in `perf-bench-matrix.json`.
- Prefill is a single from-scratch forward into a fresh **StaticCache** (the same cache type decode and deployment use), built via the existing `construct_inputs(input_prompt_tokens=...)` path, so latency and PCC are apples-to-apples with production. The cache is rebuilt at `cumulative_length=0` each run for a genuine first prefill.
- TTFT = min over N timed device runs (default `DEFAULT_PREFILL_LOOP_COUNT=10`, overridable via `-lp`). The timed graph returns only the next-token ids, skipping the logits host transfer; PCC is validated on a separate logits graph against a CPU reference.
- Prefill input is real text from the Tale of Two Cities corpus (human-readable, realistic token distribution; falls back to deterministic random tokens if the corpus is missing). Prefill latency is data-independent, so this doesn't affect TTFT.
- Configurable via `--input-sequence-length`, `--skip-pcc` (log PCC instead of asserting), and `--enable-trace on/off` (run a sweep with/without trace without editing tests).
- Runs **weekly, not nightly/PR**: prefill entries are tagged `"perf-prefill": true`; the nightly filter excludes them and a new weekly job runs only them, so the large sweep doesn't run every night.
- Fixed perf-metrics file handling for **both** decode and prefill: metrics are now written to a unique per-run directory (pytest `tmp_path`) instead of the CWD, so leftover JSONs from previous runs no longer pollute the current run's graph count / op metrics.
- Documented the `--op-support-count` profiler-buffer-overflow failure mode in `PROFILING.md`.

### Results
https://github.com/tenstorrent/tt-blacksmith/discussions/592

### Checklist


tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -870,3 +902,450 @@ def benchmark_llm_torch_xla(
...
+
+    # Resolve the device arch up front (before SPMD is enabled below) WITHOUT
+    # initializing the default XLA device. NOTE: get_xla_device_arch() calls
+    # xm.xla_device(), which initializes a *non-virtual* device; doing that
+    # before xr.use_spmd() prevents SPMD from virtualizing tensors, so unmarked
+    # inputs/buffers stay plain PjRtData and segfault inside ExecuteReplicated's
+    # dynamic_cast<PjRtShardedData*>. Reading runtime device attributes does not
+    # create the default device, so it is safe to call here.
+    _arch_name = xr.global_runtime_device_attributes()[0].get("device_arch", "").lower()
+    arch = next((a for a in ("wormhole", "blackhole") if a in _arch_name), "")
```
agobeljicTT: @vkovacevicTT we have this issue for decode as well. not the PjRt segfault, but the SPMD:0 not being resolved.

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -45,6 +46,15 @@
 
 MIN_STEPS = 16
 
+# Default number of times the prefill is re-run (post-warmup) when measuring its latency.
+# We keep the MINIMUM across runs to report the "happy path" device time, stripping
+# out host-side jitter and one-off noise. Overridable per run via -lp / loop_count.
+DEFAULT_PREFILL_PERF_RUNS = 30
```
pglusacTT: No need for the comment, naming should be enough. Maybe rename it to `DEFAULT_PREFILL_LOOP_COUNT`.

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -870,3 +880,471 @@ def benchmark_llm_torch_xla(
...
+    task,
+    data_format,
+    input_sequence_length,
+    experimental_weight_dtype,
+    experimental_enable_permute_matmul_fusion,
+    ttnn_perf_metrics_output_file,
+    read_logits_fn,
+    mesh_config_fn,
+    shard_spec_fn,
+    required_pcc,
```
pglusacTT, line 969: Can you reorder the params of this functions such that they are logically next to one another? Model info, test knobs, compiler knobs...

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -870,3 +880,471 @@ def benchmark_llm_torch_xla(
...
+    # inputs/buffers stay plain PjRtData and segfault inside ExecuteReplicated's
+    # dynamic_cast<PjRtShardedData*>. Reading runtime device attributes does not
+    # create the default device, so it is safe to call here.
+    _arch_name = xr.global_runtime_device_attributes()[0].get("device_arch", "").lower()
+    arch = next((a for a in ("wormhole", "blackhole") if a in _arch_name), "")
+
+    # Set up for multi-chip if applicable
+    if mesh_config_fn is not None and shard_spec_fn is not None:
+        is_multichip = xr.global_runtime_device_count() > 1
+        if is_multichip:
```
pglusacTT, line 1048: I don't like that this silently becomes a single chip test but I passed in some shard spec... Maybe add a warning.

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -870,3 +880,471 @@ def benchmark_llm_torch_xla(
...
+    # Instantiate model and tokenizer
+    model, tokenizer = setup_model_and_tokenizer(model_loader, model_variant)
+    full_model_name = model_loader.get_model_info(variant=model_variant).name
+
+    # Random prompt of exactly input_sequence_length tokens so the measured prefill
+    # scales with -isl (1D; construct_inputs expands it across the batch via
+    # input_prompt_tokens). Generated once and reused by the CPU reference and the device
+    # prefill so PCC compares the same inputs. Seeded via a local generator so runs are
+    # reproducible without perturbing the global RNG state.
+    rng = torch.Generator().manual_seed(PREFILL_RANDOM_SEED)
```
pglusacTT: No need for the comment.

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -45,6 +46,15 @@
...
 MIN_STEPS = 16
 
+# Default number of times the prefill is re-run (post-warmup) when measuring its latency.
+# We keep the MINIMUM across runs to report the "happy path" device time, stripping
+# out host-side jitter and one-off noise. Overridable per run via -lp / loop_count.
+DEFAULT_PREFILL_PERF_RUNS = 30
+
+# Seed for the random prefill prompt so a benchmark run is reproducible. Drawn through a
+# local torch.Generator (below) so the global RNG state is left untouched.
+PREFILL_RANDOM_SEED = 0
```
pglusacTT: Ditto.

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -870,3 +880,471 @@ def benchmark_llm_torch_xla(
         )
 
     return result
+
+
+def benchmark_llm_torch_xla_prefill(
```
pglusacTT: Maybe name it `benchmark_llm_prefill_torch_xla` instead.

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -870,3 +880,471 @@ def benchmark_llm_torch_xla(
...
+        "export_path": MODULE_EXPORT_PATH,
+        "export_model_name": export_model_name,
+        "ttnn_perf_metrics_enabled": True,
+        "ttnn_perf_metrics_output_file": ttnn_perf_metrics_output_file,
+        "experimental_enable_permute_matmul_fusion": experimental_enable_permute_matmul_fusion,
+    }
+    # Prefill runs in bf16 (experimental_weight_dtype=""). Only pass the option when a
+    # real block format is requested: feeding an empty string into the compile options
+    # crashes the tt-mlir backend during lowering (segfault in the warmup compile). The
+    # working decode path only ever passes a valid format (e.g. "bfp_bf8").
```
pglusacTT: Overly long comment.

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -870,3 +880,471 @@ def benchmark_llm_torch_xla(
...
+
+    # ========================================================
+    # PERFORMANCE + PCC BENCHMARK
+    # ========================================================
+
+    # Two graphs are compiled (only the first when skip_pcc):
+    #   * Perf/warmup graph (return_logits=False) — the timed graph. It outputs only the
+    #     argmax over the last position (tiny), NOT the logits. Keeping the full
+    #     (batch, seq, vocab) logits off the output means the timed runs never pay the
+    #     ~GB logits host transfer, so TTFT stays a clean device-latency number (see the
```
pglusacTT: The ~GB?

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -870,3 +880,471 @@ def benchmark_llm_torch_xla(
...
+            _shard_kv_cache(ia["past_key_values"], mesh, kv_cache_sharding_spec)
+        if input_output_sharding_spec:
+            xs.mark_sharding(ia["input_ids"], mesh, input_output_sharding_spec)
+        return ia
+
+    # Warmup compiles the prefill graph, so the measured runs below are pure device
+    # executions of the same graph.
+    input_args = _make_prefill_inputs_on_device()
+
+    print("Warming up...")
```
pglusacTT, line 1211: Print instead of logger? Is logging done like this throughout the perf infra?

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -870,3 +880,471 @@ def benchmark_llm_torch_xla(
...
+        input_args,
+        device,
+        1,  # only need to warm up for prefill
+        verbose=False,
+        collect_logits=False,
+        tokenizer=tokenizer,
+    )
+    print("Warmup complete")
+
+    tracy.signpost("warmup_complete")
```
pglusacTT, line 1223: Is there a signpost for warmup start? This is a nit but it's cleaner if it's surrounded by signposts.

tests/benchmark/benchmarks/llm_benchmark.py
pglusacTT: Link the issue in the comment.

agobeljicTT: there is no issue yet

tests/benchmark/benchmarks/llm_benchmark.py
agobeljicTT: this is just a print, no need to log this

tests/benchmark/benchmarks/llm_benchmark.py
agobeljicTT: I think we start with warmup so not necessary and it's like decode, can add if you insist.

.github/scripts/filter-test-matrix.py
```diff
@@ -37,6 +37,7 @@ def filter_matrix_adv(matrix, adv_filter):
       - "runs-on": machine(s) on which the test should run. If ommited, condition will be applied to all unskipped machines. Parameter can be string or array of strings.
       - "filter": string that should be present in the test name.
       - "accuracy-testing": whether to include accuracy testing or not.
+      - "perf-prefill": match tests by their "perf-prefill" property. Tests marked "perf-prefill": true are the prefill perf sweep, run only weekly (not nightly). Set this to false in the nightly filter to exclude them, or true in the weekly filter to include only them. Tests default to "perf-prefill": false.
```
vvukomanTT: Let's simplify the description and not mention specific workflows where these types of tests are ran or not, the script doesn't need that context.
e.g.
```
- "perf-prefill": whether to include preffil tests or not
```

.github/workflows/schedule-nightly.yml
```diff
@@ -89,7 +89,7 @@ jobs:
       sh-runner: false
       skip-device-perf: true
       regression_check: true
-      adv_filter: '[ {"accuracy-testing": false}, {"runs-on": "n150"}, {"runs-on": "p150-perf"}, {"runs-on": "n300-llmbox"}, {"runs-on": "galaxy-wh-6u"}, {"runs-on": "qb2-blackhole"} ]'
+      adv_filter: '[ {"accuracy-testing": false}, {"perf-prefill": false}, {"runs-on": "n150"}, {"runs-on": "p150-perf"}, {"runs-on": "n300-llmbox"}, {"runs-on": "galaxy-wh-6u"}, {"runs-on": "qb2-blackhole"} ]'
```
vvukomanTT: This is good but let's add `{"perf-prefill": false}` in other places where `call-filtered-tests.yml` is triggered but we don't want the prefill tests ran.

vvukomanTT (approved):

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -895,3 +940,434 @@ def benchmark_llm_torch_xla(
...
+    """
+    Benchmark the PREFILL of an LLM using PyTorch and torch-xla.
+
+    This function loads an LLM, compiles it with torch-xla for the Tenstorrent backend,
+    and measures the latency of a single prefill forward over a prompt of
+    input_sequence_length tokens. Output correctness is validated via PCC (Pearson
+    Correlation Coefficient) of the prefill logits against a CPU reference.
+
+    Args:
+        model_loader: Model loader instance for loading the LLM
```
agobeljicTT, line 992: reorder this to match args

tests/benchmark/test_llms.py
```diff
@@ -224,6 +260,177 @@ def test_llm(
...
+                "Skipping perf metrics."
+            )
+            results["config"]["ttnn_num_graphs"] = 0
+
+        with open(output_file, "w") as file:
+            json.dump(results, file, indent=2)
+
+
+# Generic driver invoked by the per-model prefill tests, not a test itself.
+test_llm_prefill.__test__ = False
```
vkovacevicTT, line 454: nit: We always explicitly call pytest test functions in benchmarks, but if we are adding this, let's do it for regular `test_llms` functions too.

tests/benchmark/test_llms.py
```diff
@@ -303,6 +549,47 @@ def test_llama_3_2_1b(
...
+        ),
+        optimization_level=(
+            optimization_level
+            if optimization_level is not None
+            else DEFAULT_OPTIMIZATION_LEVEL
+        ),
+        experimental_weight_dtype=None,  # prefill needs to be bf16
+        do_apply_weight_dtype_overrides=False,  # prefill needs to be bf16
+        trace_enabled=(trace_override if trace_override is not None else True),
+        skip_pcc=skip_pcc,
```
vkovacevicTT: Don't we already handle defaults in `test_llm_prefill` function for `optimization_level`, `input_sequence_length` and `trace_enabled`?

Same for other added models below.

tests/benchmark/conftest.py
```diff
@@ -87,13 +87,39 @@ def pytest_addoption(parser):
...
+        default=None,
+        type=make_validator_positive_int("--input-sequence-length"),
+        help="Input (prefill) sequence length in tokens. Overrides model default.",
+    )
+
+    parser.addoption(
+        "--skip-pcc",
+        action="store_true",
+        default=False,
+        help="Log the PCC value instead of asserting it. Lets a run complete (and write "
```
vkovacevicTT: This flag doesn't just omit PCC assert, it skips the whole PCC benchmark setup and run.
Please modify the help string.

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -436,10 +486,8 @@ def benchmark_llm_torch_xla(
 
         cpu_output_logits = cpu_prefill_logits + cpu_decode_logits
 
-    # Transfer model to device
     model = model.to(device, dtype=torch.bfloat16)
 
-    # Shard model if shard spec function is provided
```
vkovacevicTT, line 443: nit: You can leave this comment.

tests/benchmark/benchmarks/llm_benchmark.py
vkovacevicTT: Please remove this, it's solved here:

https://github.com/tenstorrent/tt-xla/blob/a97e3c74f7aff652a236c44e124f4e756aa43e58/tests/benchmark/utils.py#L37-L49

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -896,3 +943,434 @@ def benchmark_llm_torch_xla(
...
+
+    def _make_prefill_inputs_on_device():
+        """Fresh prefill inputs on device for one from-scratch prefill.
+
+        Builds a fresh StaticCache via construct_inputs (the same cache type the decode
+        benchmark and deployment use) and transfers it to device. Rebuilt per call so
+        every run starts from cumulative_length=0 - i.e. a genuine first prefill - instead
+        of accumulating into a carried-over cache.
+        """
+        ia = construct_inputs(
```
vkovacevicTT: nit: Rename ia to something more meaningful e.g. `input_args` or `input`

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -896,3 +943,434 @@ def benchmark_llm_torch_xla(
...
+            if input_output_sharding_spec:
+                xs.mark_sharding(ia["input_ids"], mesh, input_output_sharding_spec)
+        return ia
+
+    input_args = _make_prefill_inputs_on_device()
+
+    print("Warming up...")
+    _, _ = generate_and_benchmark(
+        compiled_perf_model,
+        dict(input_args),
```
vkovacevicTT, line 1214: Why `dict` ? Do we need to make a copy of it or?

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -896,3 +943,434 @@ def benchmark_llm_torch_xla(
...
+    tracy.signpost("warmup_complete")
+
+    print(f"\nStarting performance benchmark...")
+    prefill_times_ns = []
+    with torch.no_grad():
+        for run_idx in range(prefill_perf_runs):
+            # Fresh StaticCache inputs per run (rebuilt at cumulative_length=0), so each
+            # run is a genuine from-scratch first prefill rather than appending to a
+            # carried-over cache.
+            input_args = _make_prefill_inputs_on_device()
```
vkovacevicTT, line 1232: Do we want to use the same inputs in every iteration? 

Also, be advised, CPU work outside of start-end time measurement loop can cause time leakage if a blocking call is omitted.
In this case we have an explicit  `next_token_ids_replicated.to("cpu")` so we should be safe.

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -896,3 +943,434 @@ def benchmark_llm_torch_xla(
...
+        output_logits, _ = generate_and_benchmark(
+            compiled_logits,
+            input_args,
+            device,
+            1,
+            verbose=False,
+            collect_logits=True,
+        )
+
+    ttft_ns = min(prefill_times_ns)
```
vkovacevicTT, line 1273: What's the idea of using min prefill time?

.github/workflows/schedule-weekly.yml
```diff
@@ -51,11 +51,29 @@ jobs:
       artifact_suffix: ${{ needs.build-ttxla.outputs.artifact_suffix }}
       wheel_build: manylinux
 
+  # Prefill perf sweep. These tests are marked "perf-prefill" and are skipped
+  # in nightly; the {"perf-prefill": true} filter runs only them here, so the
+  # regular perf tests (already covered nightly) are not re-run.
+  perf-benchmark:
```
vkovacevicTT: Let's rename this so that it's clear it's a prefill only perf benchark.

.github/workflows/schedule-weekly.yml
```diff
@@ -51,11 +51,29 @@ jobs:
...
+    uses: ./.github/workflows/call-filtered-perf-tests.yml
+    needs: [ build-image, build-ttxla ]
+    secrets: inherit
+    with:
+      docker_image: ${{ needs.build-image.outputs.docker-image-base }}
+      sh-runner: false
+      skip-device-perf: true
+      regression_check: true
```
vkovacevicTT: @vvukomanTT Does regression check logic extend to weekly runs? If not, I would disable it here.

.github/workflows/schedule-weekly.yml
```diff
@@ -51,11 +51,29 @@ jobs:
...
+      adv_filter: '[ {"perf-prefill": true}, {"runs-on": "n150"}, {"runs-on": "p150-perf"}, {"runs-on": "n300-llmbox"}, {"runs-on": "galaxy-wh-6u"}, {"runs-on": "qb2-blackhole"} ]'
```
vkovacevicTT, line 68: nit: Can we group this in a single "runs-on" list? Same for other workflows.

.github/workflows/perf-bench-matrix.json
```diff
@@ -89,6 +89,54 @@
...
         "pytest": "tests/benchmark/test_llms.py::test_llama_3_2_1b"
       },
+      {
+        "name": "llama_3_2_1b_instruct_prefill_bs1_sl1024_opt0",
+        "perf-prefill": true,
+        "pytest": "tests/benchmark/test_llms.py::test_llama_3_2_1b_prefill",
+        "batch-size": 1,
+        "input-sequence-length": 1024,
+        "optimization-level": 0
+      },
```
vkovacevicTT: Let's try specifying pytest params in test scripts instead of making individual test entries for prefill tests.

vkovacevicTT (commented):
Looks good in general, some minor changes required.

Also, run the prefill benchmark in CI and attach links to the PR descriptions Checklist.
Try a sanity check for a couple of models first before running the whole benchmark.

.github/workflows/perf-bench-matrix.json
agobeljicTT: where exactly?
we need these to run with all these params, how can we parametrize otherwise?

.github/workflows/schedule-weekly.yml
agobeljicTT: @vvukomanTT

tests/benchmark/benchmarks/llm_benchmark.py
agobeljicTT: done

tests/benchmark/benchmarks/llm_benchmark.py
agobeljicTT: well I had issues where this `generate_and_benchmark` modifies input_args, so I used a copy instead. now i regenerate inputs each time so perhaps i don't need this, but doesn't hurt...

tests/benchmark/benchmarks/llm_benchmark.py
agobeljicTT: The idea is to include transfer to device in E2E so this should be safe.

tests/benchmark/benchmarks/llm_benchmark.py
agobeljicTT: removing host randomness, we want best case scenario

tests/benchmark/conftest.py
agobeljicTT: good catch

tests/benchmark/test_llms.py
agobeljicTT: yeah... perhaps not necessary...

.github/workflows/schedule-weekly.yml
vvukomanTT: Correct, it doesn't support weekly runs regressions (it will still check in the last nightly), so makes little sense here.

agobeljicTT: on it

.github/workflows/schedule-weekly.yml
nsumrakTT: I don't think that is within scope of this PR.

nsumrakTT (approved):

tests/benchmark/benchmarks/llm_benchmark.py
agobeljicTT: done

.github/workflows/schedule-weekly.yml
agobeljicTT: done

.github/scripts/filter-test-matrix.py
```diff
@@ -15,11 +15,33 @@ def flatten_matrix(data):
         for test in proj.get("tests", []):
             merged_test = {**test_defaults, **test, "project": proj["project"]}
 
-            runs_on = merged_test.get("runs-on", [])
-            if isinstance(runs_on, list):
-                matrix.extend({**merged_test, "runs-on": runner} for runner in runs_on)
+            # Expand a "pytest-sweep" (list of pytest parametrization ids) into one
```
vkovacevicTT: From an offline discussion:
CI workflow shouldn't have any notion of model parameters and configs.
We should parametrize this internally in benchmarks `tests/benchmark/test_llms.py`.

Use `@pytest.mark.parametrize` for a single model and just call it like other models in workflow with pytest.

PR #5080, opened 2026-06-04
Implementation of PJRT_Client_CreateUninitializedBuffer

### LINKED ISSUE #4970: `PJRT_Client_CreateUninitializedBuffer` not implemented
## Summary

tt-xla's PJRT plugin does not bind `PJRT_Client_CreateUninitializedBuffer`. The PJRT C API function-pointer slot is left null. The only path to obtain a device buffer is `PJRT_Client_BufferFromHostBuffer`.

acicovicTT:
This PR adds initial support for the PJRT_Client_CreateUninitializedBuffer API signature.


sgligorijevicTT (approved):

pjrt_implementation/src/api/client_instance.cc
```diff
@@ -333,6 +333,8 @@ void ClientInstance::bindApi(PJRT_Api *api) {
   api->PJRT_Client_DefaultDeviceAssignment =
       internal::onClientDefaultDeviceAssignment;
   api->PJRT_Client_BufferFromHostBuffer = internal::onBufferFromHostBuffer;
+  api->PJRT_Client_CreateUninitializedBuffer =
+      internal::onCreateUninitializedBuffer;
```
jameszianxuTT, line 339: Does anything currently call this API?

acicovicTT: It was requested through a community issue, otherwise no.

pjrt_implementation/src/api/client_instance.cc
```diff
@@ -947,6 +949,87 @@ onBufferFromHostBuffer(PJRT_Client_BufferFromHostBuffer_Args *args) {
...
+  if (!memory_instance) {
+    LOG_F(ERROR, "Memory space is not set either in the memory argument nor in "
+                 "device from the device argument");
+    return *ErrorInstance::makeError(tt_pjrt_status::kInvalidArgument)
+                .release();
+  }
+  if (memory_instance->isHostMemory()) {
+    LOG_F(ERROR, "Buffer creation is supported only in device memory");
+    return *ErrorInstance::makeError(tt_pjrt_status::kUnimplemented).release();
+  }
```
pilkicTT: we have this logic also in `onBufferFromHostBuffer`:
- i think we can simplify this a little bit
- maybe also add a helper function for this.

pjrt_implementation/src/api/client_instance.cc
pilkicTT: did you investigate if jax calls this in any situation? i assume torch xla doesn't call this, since from what i saw they use less of the PJRT surface area...

the main problem is that we don't test this path in any way, so it would be good if we can at least find some framework call that indirectly calls this...

acicovicTT: I ran ~10 experiments based on an AI-generated list of most likely JAX scenarios/APIs (e.g. buffer donation) that would potentially use this function, but I had no luck (with this JAX version at least, maybe it would be different if we were on a higher version.) Some of those experiments, also, were inconclusive because they hit other unimplemented PJRT functions. Since the core logic is shared with `onBufferFromHostBuffer` I don't worry too much about this. I validated this one with a custom C++ program.

sgligorijevicTT: My blind guess is that `CreateUninitializedBuffer` was added for the purposes of Pallas calls. As we have no pallas integration it's possible we might not hit this code path, ever. Unless functions like `empty_like` are special cased (now or in the future) to invoke this during eager...

PR #5092, opened 2026-06-04
Add per-weight sensitivity score heuristic for mixed precision quantization 

kdimicTT:
### Problem description

Currently, all LLMs use bfp8 as the default data type for the weights. Selecting which weights to quantize to lower data types (bfp_bf4) for mixed precision inference is currently done manually, usually by choosing only MLP weighs. The goal of this PR is to introduce quantitative metric called sensitivity of Loss with respect to weight tensor. Each weight tensor is characterised with a number called sensitivity. It represents how much end model loss degrades when that specific weight tensor is quantized to lower data format (bfp_bf4).
Based on this metrics we can sort weights from least sensitive to most sensitive and use this order as heuristics for lowering weight dtypes.
The idea behing this gready search is that weight tensor with lower sensitivity are more robust to quantization and should be lowered first.

### What's changed
Implements a two-stage per-tensor sensitivity scoring pipeline based on diagonal Fisher Information. Both stages run on the GPU machine — **no TT device is required** (see Stage 2).

#### Stage 1 — Fisher (GPU)
Computes approximate Fisher scores via accumulated squared gradients over calibration data from the C4 dataset. Measures how much the loss changes when a specific weight is modified. The calibration objective is next-token cross-entropy on the model logits, so the standard path targets causal LMs by objective (`AutoModelForCausalLM`, generic across LM architectures).
Supports two modes:
  - **Standard**: loads the full model into GPU memory.
  - **Offloaded** (`--offload-layers`): streams weights from disk one layer at a time for models too large to fit in GPU memory, with multi-GPU sample parallelism. Supports Llama-style decoder-only architectures (validated up front) and VLMs whose text config wraps a causal LM (e.g. Gemma); handles tied `lm_head`/`embed_tokens` weights.

#### Stage 2 — Sensitivity scores (host, no TT device)
Calculates the per-weight quantization error `Q` and the final sensitivity score
**S(T) = (1/N) · Σᵢ Fᵢᵢ · (wᵢ − Q(wᵢ))²** (per-element mean), outputting a JSON file with weights ranked by sensitivity.

BFP4 quantization is now done **host-side via `ttnn.typecast`** instead of on a TT device. This is more accurate for BFP4 conversion than the on-device path and, crucially, lets Stage 2 run on the same GPU machine as Stage 1 — avoiding the transfer of the large squared-gradient tensors from the GPU machine to a TT machine. The host BFP4 packing is pure CPU math; it only needs a cluster descriptor (not a physical chip), supplied via a mock cluster descriptor so the stage runs with no TT hardware present.

#### Testing
Adds `mixed_precision/test_fisher_pipeline.py`, a pure-CPU regression test run in the Push CI pipeline. It exercises the standard pipeline end-to-end (forward → next-token loss → backward → squared-gradient accumulation, plus host-side quantization and score aggregation) to guard against breakage of the surrounding boilerplate. It runs on a device-less CI runner via the mock cluster descriptor. (The offload path is CUDA + threads and is not covered by this CPU test.)

The offload utils are separated into `offload_fisher.py`, keeping `sensitivity_score.py` focused on the two-stage workflow, with shared helpers in `utils.py`.


mixed_precision/offload_fisher.py
```diff
@@ -0,0 +1,423 @@
...
+
+def _make_fisher_hook(n, acc, on_cpu=True):
+    """Return a post-accumulate-grad hook that accumulates squared gradients.
+
+    on_cpu=True (default): moves gradient to CPU before accumulating.
+    on_cpu=False: accumulates on the same device as the gradient.
+    """
+
+    def hook(p):
+        if p.grad is not None:
```
dgolubovicTT: Early return is a bit cleaner:
```
if p.grad is None:
    return
```

mixed_precision/offload_fisher.py
```diff
@@ -0,0 +1,423 @@
...
+            g2 = p.grad.detach().float().pow_(2)
```
dgolubovicTT: rename g2 to something more informative like `grad_sq`

mixed_precision/offload_fisher.py
```diff
@@ -0,0 +1,423 @@
...
+import threading
+
+import torch
+import torch.nn as nn
+from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer
+
+NUM_SAMPLES = 100
+
+
+def _make_fisher_hook(n, acc, on_cpu=True):
```
dgolubovicTT: Consider renaming n to param_name and acc to something more descriptive like fisher_accumulator or fisher_stats. It makes the hook easier to understand without needing to infer what those variables represent.

mixed_precision/offload_fisher.py
```diff
@@ -0,0 +1,423 @@
...
+            p.grad = None
+
+    return hook
+
+
+def _resolve_model_path(model_name_or_path):
+    """Resolve a HF model name or local path to a directory with safetensors files."""
+    if os.path.isdir(model_name_or_path):
+        return model_name_or_path
+    from transformers.utils import cached_file
```
dgolubovicTT: I prefer all imports being in file header.

mixed_precision/offload_fisher.py
```diff
@@ -0,0 +1,423 @@
...
+    except Exception as e:
+        raise RuntimeError(
+            f"Cannot find model at '{model_name_or_path}'. "
+            "Pass a local directory path or ensure the model is in the HF cache."
+        ) from e
+
+
+def _build_weight_map(model_path):
+    """Return {tensor_name: shard_filename} for all tensors in the model."""
+    from safetensors import safe_open
```
dgolubovicTT: Same here, avoid having imports inside functions

mixed_precision/offload_fisher.py
```diff
@@ -0,0 +1,423 @@
...
+            gpu_id,
+            out_dir,
+            f"chunk_layer_{k:04d}.pt",
+            layer_acc,
+        )
+        if gpu_id == 0:
+            print(f"  Offload Fisher: layer {k + 1}/{N} backward", flush=True)
+
+
+def fisher_thread_worker(
```
dgolubovicTT, line 251: The offload workers sync on a threading. Barrier with no timeout, and a failure in any one thread (OOM, bad model layout, disk error) leaves the others blocked at barrier.wait() indefinitely while main() hangs in join() with no error reported. 

Can we wrap fisher_thread_worker in a try/except that calls barrier.abort() on failure (so the others get BrokenBarrierError and exit), capture the exception per-thread, and re-raise it in the main thread after join()? Right now a single-thread failure looks like a hang, not an error.

mixed_precision/sensitivity_score.py
```diff
@@ -0,0 +1,397 @@
...
+    print("Computing sensitivity scores on TT device...")
+    tt_device = ttnn.open_device(device_id=0)
+    scores = {}
+
+    try:
+        for name, fii_tensor in iter_fisher(fii_source):
+            w = weight_params_dict[name].data.cpu().float()
+            q = quantize_via_ttnn(w, ttnn.bfloat4_b, tt_device)
+            quant_err = (w - q.float()) ** 2
+            score = (fii_tensor.float() * quant_err).sum().item() / fii_tensor.numel()
```
dgolubovicTT: The docstring and PR describe S(T) = Σ Fᵢᵢ·(wᵢ−Q(wᵢ))² but the code divides the sum by fii_tensor.numel(), so it's actually a per-element mean. That changes the cross-tensor ranking (sum favors big tensors, mean cancels size out) — and the ranking is the whole output. 

After doing experiments with both sum and mean we decided to go for a mean. Let's make the code and the docstring agree.

mixed_precision/sensitivity_score.py
```diff
@@ -0,0 +1,397 @@
...
+
+
+def run_fisher_stage(args):
+    if args.offload_layers:
+        _run_fisher_offload(args)
+    else:
+        _run_fisher_standard(args)
+
+
+def run_scores_stage(args):
```
dgolubovicTT, line 391: --save-path is respected in the fisher stage but run_scores_stage hardcodes get_fii_path(args.model) and ignores it. So --stage scores --save-path X silently reads the default path instead of X - and could load a stale fii.pt from a prior run. Should the scores stage take the same --save-path override?

mixed_precision/offload_fisher.py
```diff
@@ -0,0 +1,423 @@
...
+    """Compute loss backward through norm+lm_head; return (lm_head_acc, grad_outs).
+
+    lm_head_acc holds Fisher accumulators for lm_head weights.
+    """
+    N = len(boundary_acts[0]) - 1
+
+    norm_gpu = copy.deepcopy(model.model.norm)
+    load_tensors_to_layer(norm_gpu, "model.norm.", weight_map, model_path, device)
+    lm_head_gpu = copy.deepcopy(model.lm_head)
+    load_tensors_to_layer(lm_head_gpu, "lm_head.", weight_map, model_path, device)
```
dgolubovicTT: For tied-embedding models, model.safetensors.index.json has no lm_head.weight entry (it's shared with embed_tokens), so load_tensors_to_layer(lm_head_gpu, "lm_head.", ...) never populates it and lm_head stays a meta tensor in the offload path. Can we detect tying (config.tie_word_embeddings) and reuse the embedding weight for lm_head here?

mixed_precision/offload_fisher.py
```diff
@@ -0,0 +1,423 @@
...
+
+def _run_layer(layer, hidden_states, position_ids, position_embeddings=None):
+    """Run one transformer block, return the hidden_states tensor."""
+    kwargs = {}
+    if position_embeddings is not None:
+        kwargs["position_embeddings"] = position_embeddings
+    try:
+        out = layer(hidden_states, position_ids=position_ids, **kwargs)
+    except TypeError:
+        out = layer(hidden_states, **kwargs)
```
dgolubovicTT: Why do we have this except TypeError? Is it just to handle when layer() doesn't have position_ids as attribute?

mixed_precision/offload_fisher.py
```diff
@@ -0,0 +1,423 @@
...
+    return out[0] if isinstance(out, (tuple, list)) else out
+
+
+def _compute_position_embeddings(base_model, seq_len, position_ids, device):
```
dgolubovicTT: This function assumes model has RoPE. What about models that have non-rope positional embeddings?

mixed_precision/offload_fisher.py
```diff
@@ -0,0 +1,423 @@
...
+):
+    """Run embed + all transformer layers without gradients, return boundary activations.
+
+    boundary_acts[s][k] is the CPU tensor output of layer k-1 (input to layer k).
+    Index 0 is the embedding output; index N is the last layer's output.
+    """
+    N, num_s = len(layer_list), len(samples)
+    boundary_acts = [[] for _ in range(num_s)]
+
+    embed = copy.deepcopy(model.model.embed_tokens)
```
dgolubovicTT, line 89: The offload path assumes a Llama-style tree (model.model.layers, model.model.norm, model.lm_head, model.model.rotary_emb). On any other architecture this throws AttributeError inside a worker thread, which (see the barrier comment) becomes a hang. Can we validate these attributes exist up front and raise a clear "unsupported architecture" error, and document the supported-architecture assumption?

Same thing on lines: [259](https://github.com/tenstorrent/tt-xla/pull/5092/changes#diff-f3ebe94fb4ed7559070b4d26022a596ab2ccea551e4159dd1bf4fd40bb968d16R259), [261](https://github.com/tenstorrent/tt-xla/pull/5092/changes#diff-f3ebe94fb4ed7559070b4d26022a596ab2ccea551e4159dd1bf4fd40bb968d16R261),[373](https://github.com/tenstorrent/tt-xla/pull/5092/changes#diff-f3ebe94fb4ed7559070b4d26022a596ab2ccea551e4159dd1bf4fd40bb968d16R373),[158](https://github.com/tenstorrent/tt-xla/pull/5092/changes#diff-f3ebe94fb4ed7559070b4d26022a596ab2ccea551e4159dd1bf4fd40bb968d16R158)

dgolubovicTT: So we are overfitted in this offloaded case to llama architectures. 
I would add an explicit up-front check that the expected attributes exist and raise a clear "offload path currently supports Llama-style decoder-only models" error, and document that constraint.

dgolubovicTT (changes requested):

mixed_precision/offload_fisher.py
kdimicTT: I handled this in the new commit. lm_head_tied is detected from the weight map and the embed_tokens weight is loaded in its place

mixed_precision/sensitivity_score.py
kdimicTT: Thanks, it should be used in both stages

mixed_precision/offload_fisher.py
kdimicTT: The `hasattr(base_model, "rotary_emb")` guard returns None when the base model does not expose a top-level `rotary_emb` module. This covers both models that do not use RoPE and implementations where RoPE is computed within each attention layer rather than at the model level. In both cases `_run_layer` does not pass `position_embeddings`, allowing the layer to handle positional encoding internally if needed.

mixed_precision/offload_fisher.py
kdimicTT: Yes, since some layers don't accept position_ids in their forward signature (e.g. Llama 3.1+ where RoPE is pre-computed at the model level)

dgolubovicTT: Ok that explains the intent. However, this` except TypeError` will catch other types of errors and blindly rerun without position_ids. You can infer if layer has position_ids in signature something like this:

```
import inspect

def _run_layer(layer, hidden_states, position_ids, position_embeddings=None):
    """Run one transformer block, return the hidden_states tensor."""

    kwargs = {}
    params = inspect.signature(layer.forward).parameters

    if "position_ids" in params:
        kwargs["position_ids"] = position_ids
    if position_embeddings is not None and "position_embeddings" in params:
        kwargs["position_embeddings"] = position_embeddings

    out = layer(hidden_states, **kwargs)
    return out[0] if isinstance(out, (tuple, list)) else out

```

mixed_precision/utils.py
```diff
@@ -0,0 +1,203 @@
...
+    return model, tokenizer, weight_map, model_path, lm_base, lm_head_base
+
+
+def _run_layer(layer, hidden_states, position_ids, position_embeddings=None):
+    """Run one transformer block, return the hidden_states tensor."""
+    kwargs = {}
+    if position_embeddings is not None:
+        kwargs["position_embeddings"] = position_embeddings
+    try:
+        out = layer(hidden_states, position_ids=position_ids, **kwargs)
```
dgolubovicTT: _run_layer passes no attention_mask. In a normal forward the model builds a causal mask and hands it to each layer — without it the offloaded Fisher may be computed with non-causal attention, diverging from the standard path. Can you confirm the calibration forward is actually causal (e.g. SDPA inferring is_causal), or pass an explicit causal mask?

dgolubovicTT (approved):

mixed_precision/utils.py
```diff
@@ -0,0 +1,217 @@
...
+    rotary_emb = rotary_cls(config=base_model.config, device=device)
+    dummy = torch.zeros(1, seq_len, 1, dtype=torch.bfloat16, device=device)
+    with torch.no_grad():
+        position_embeddings = rotary_emb(dummy, position_ids)
+    del rotary_emb, dummy
+    return position_embeddings
+
+
+def collect_weights(model):
+    """Return [(name, param)] for all quantizable weight tensors."""
```
dgolubovicTT: "quantizable" - not the best explanation of what we want to collect here. Edit the description...

mixed_precision/utils.py
```diff
@@ -0,0 +1,217 @@
...
+        for pname, param in module.named_parameters(recurse=False)
+        if param.ndim >= 2
+        and not isinstance(module, nn.Embedding)
+        and "norm" not in name
+        and "router" not in name
+        and pname != "bias"
+    ]
+
+
+def get_fii_path(model_name):
```
dgolubovicTT: f_ii just means fischer score on main diagonal of element i. Shouldn't be part of function name...

mixed_precision/utils.py
```diff
@@ -0,0 +1,217 @@
...
+    if "attention_mask" in params:
+        kwargs["attention_mask"] = causal_mask
+
+    out = layer(hidden_states, **kwargs)
+    return out[0] if isinstance(out, (tuple, list)) else out
+
+
+def _compute_position_embeddings(base_model, seq_len, position_ids, device):
+    """Compute RoPE cos/sin for models with a model-level rotary_emb."""
+    if not hasattr(base_model, "rotary_emb"):
```
dgolubovicTT, line 196: This cumputation depends on model having exact `rotary_emb` attribute. If we can't make this more general, we should at least explain it in docstring

mixed_precision/utils.py
```diff
@@ -0,0 +1,217 @@
...
+        )
+    model.eval()
+
+    weight_map = _build_weight_map(model_path)
+    lm_base, lm_head_base = _detect_weight_paths(weight_map)
+    return model, tokenizer, weight_map, model_path, lm_base, lm_head_base
+
+
+def _run_layer(layer, hidden_states, position_ids, position_embeddings=None):
+    seq_len = hidden_states.shape[1]
```
dgolubovicTT, line 163: Add a docstring explaining what is this fucntion doing

mixed_precision/utils.py
```diff
@@ -0,0 +1,217 @@
...
+    )
+
+    kwargs = {}
+    params = inspect.signature(layer.forward).parameters
+    if "position_ids" in params:
+        kwargs["position_ids"] = position_ids
+    if position_embeddings is not None and "position_embeddings" in params:
+        kwargs["position_embeddings"] = position_embeddings
+    if "attention_mask" in params:
+        kwargs["attention_mask"] = causal_mask
```
dgolubovicTT, line 181: Add comment why we need to supply layer with additional tensors when we do forward layer by layer. When we do forward on the model as a whole, all necessary attention_mask,position_embeddings, position_ids are generated in model forward. When we execute part by part - we need to handle these outside

mixed_precision/sensitivity_score.py
```diff
@@ -0,0 +1,401 @@
...
+import torch
+import torch.nn as nn
+from datasets import load_dataset
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+try:
+    import ttnn
+
+    HAS_TTNN = True
+except ImportError:
```
dgolubovicTT, line 34: do we really need this try except :)

mixed_precision/sensitivity_score.py
```diff
@@ -0,0 +1,401 @@
...
+
+def quantize_via_ttnn(tensor, dtype, device):
+    """Roundtrip tensor through TT device at target dtype to get quantized values."""
+    orig_shape = tensor.shape
+    if tensor.ndim > 2:
+        tensor = tensor.reshape(-1, tensor.shape[-1])
+    tt = ttnn.from_torch(
+        tensor, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device
+    )
+    tt = ttnn.typecast(tt, dtype)
```
dgolubovicTT, line 55: we can do ttnn typecast on host, so we don't need a device for this. Also, turns out that ttnn typecast is more accurate on host than on device for to bfp4 conversion. Doing quantization on host (ttnn.typecast) would enable us to do both fischer information and final sensitivity calculation on GPU machine. Therefore skipping the part where we transfer huge squared gradient tensors from GPU machine to TT machine...
So even with host typecast being slower, we gain both accuracy and save time.

mixed_precision/sensitivity_score.py
```diff
@@ -0,0 +1,401 @@
...
+        help="Layer-wise CPU offloading for models too large to fit in GPU memory. "
+        "Loads one transformer block at a time and iterates layer-wise."
+        "Accumulators always live on CPU.",
+    )
+    return parser.parse_args()
+
+
+def load_model_and_tokenizer(model_name, device_map=None):
+    tokenizer = AutoTokenizer.from_pretrained(model_name)
+    model = AutoModelForCausalLM.from_pretrained(
```
dgolubovicTT, line 150: This implies that model is LLM. We want this to be general...

mixed_precision/offload_fisher.py
nvukobratTT: Not blocking, but let's iterate if this part can be a bit more generic or more intuitive for use if some customization needs to be done. However, there is no need to pre-optimize before we hit this issue.

Similar comment applies pretty much to most of the LLM-specific functions as well.

nvukobratTT (approved):

nsumrakTT (approved):

vvukomanTT (approved):

PR #5094, opened 2026-06-04
[Nightly] Fix BH galaxy nightly failure

sshonTT:
### Ticket
N/A

### Problem description
There's mesh / topology issue in tt-metal side that occurs `MGD` failure when using bh-glx machine. See https://github.com/tenstorrent/tt-xla/actions/runs/26922426056/job/79478267032#step:19:44

### What's changed
- client_instance.cc: BH galaxy opens a 2D (4,8) parent mesh + forces FABRIC_1D, gated by TT_RUNTIME_USING_BH_GALAXY (avoids the MGD topology-mapper failure on the UBB galaxy).
- call-test.yml: set that env for the galaxy-bh job; honor a per-job "timeout" (minutes) at job + step level (lifts the 240/360 default).
- basic-test-nightly.json: 480-min timeout for run_torch_bh_galaxy.
- test_deepseek_v4_e2e_streaming.py: load the tokenizer via PreTrainedTokenizerFast (avoids the transformers 5.5.1 deepseek_v4 crash).

### Checklist

### DIFF
--- a/.github/workflows/test-matrix-presets/basic-test-nightly-experimental.json
+++ b/.github/workflows/test-matrix-presets/basic-test-nightly-experimental.json
@@ -1,0 +1,3 @@
+[
+  { "runs-on": "galaxy-bh",           "name": "run_torch_bh_galaxy",   "dir": "./tests/torch",                            "test-mark": "nightly and bh_galaxy", "forked": true }
+]
--- a/.github/workflows/test-matrix-presets/basic-test-nightly.json
+++ b/.github/workflows/test-matrix-presets/basic-test-nightly.json
@@ -16,6 +16,5 @@
   {"runs-on": "qb2-blackhole",       "name": "run_vllm_bhqb_tests",    "dir": "./tests/integrations/vllm_plugin",         "test-mark": "nightly and bhqb",                        "extra-wheel": "vllm"},
   { "runs-on": "galaxy-wh-6u",       "name": "run_torch_galaxy",       "dir": "./tests/torch",                            "test-mark": "nightly and galaxy" },
   { "runs-on": "galaxy-wh-6u",        "name": "run_vllm_galaxy_wh_6u_tests", "dir": "./tests/integrations/vllm_plugin", "test-mark": "nightly and tensor_parallel and galaxy_wh_6u", "extra-wheel": "vllm" },
-  { "runs-on": "galaxy-bh",           "name": "run_torch_bh_galaxy",   "dir": "./tests/torch",                            "test-mark": "nightly and bh_galaxy" },
   { "runs-on": "cpu", "name": "run_cpu_compile_only", "dir": "./tests/cpu_compile_only", "test-mark": "nightly and cpu", "shared-runners": "true", "forked": "true" }
 ]
--- a/pjrt_implementation/src/api/client_instance.cc
+++ b/pjrt_implementation/src/api/client_instance.cc
@@ -407,8 +407,19 @@ tt_pjrt_status ClientInstance::populateDevices() {
 
   // Mesh device requires physical hardware; skip in compile-only mode.
   if (!m_compile_only) {
-    m_parent_mesh =
-        getOrCreateMeshDevice({1, static_cast<uint32_t>(m_devices.size())});
+    // [Workaround] On a 32-device galaxy (UBB) a 1D {1, N} parent mesh can't be
+    // reshaped to the 2D (4, 8) executable mesh (the MGD solver fails); open
+    // (4, 8) directly.
+    if (m_devices.size() == 32) {
+      LOG_F(WARNING, "32-device galaxy: opening the parent mesh directly as "
+                     "(4, 8); the 1D {1, N} -> (4, 8) reshape fails in the MGD "
+                     "solver. See "
+                     "https://github.com/tenstorrent/tt-metal/issues/43210");
+      m_parent_mesh = getOrCreateMeshDevice({4, 8});
+    } else {
+      m_parent_mesh =
+          getOrCreateMeshDevice({1, static_cast<uint32_t>(m_devices.size())});
+    }
   }
 
   return tt_pjrt_status::kSuccess;
@@ -509,6 +520,21 @@ ClientInstance::computeFabricConfig(const std::vector<uint32_t> &mesh_shape) {
         num_devices > 1 ? tt::runtime::FabricConfig::FABRIC_1D
                         : tt::runtime::FabricConfig::DISABLED;
     return tt::runtime::MeshFabricConfig{global, {}};
+  }
+  // [Workaround] The Blackhole galaxy (UBB) lacks a both-axis wrap, so
+  // computeMeshFabricConfig's RING_RING is rejected as TORUS_XY by the
+  // TopologyMapper; force FABRIC_1D there. Other 32-device galaxies (e.g.
+  // Wormhole) have the wrap, so keep their auto-detected fabric.
+  bool is_blackhole = m_system_descriptor->chip_descs()->size() > 0 &&
+                      m_system_descriptor->chip_descs()->Get(0)->arch() ==
+                          ::tt::target::Arch::Blackhole;
+  if (is_blackhole && m_devices.size() == 32) {
+    LOG_F(WARNING,
+          "Auto-overriding fabric config to FABRIC_1D for the "
+          "32-device Blackhole galaxy (UBB); this is a workaround, not "
+          "expected behaviour.");
+    return tt::runtime::MeshFabricConfig{tt::runtime::FabricConfig::FABRIC_1D,
+                                         {}};
   }
   return tt::runtime::computeMeshFabricConfig(m_system_descriptor, mesh_shape);
 }
--- a/.github/workflows/manual-test.yml
+++ b/.github/workflows/manual-test.yml
@@ -13,6 +13,7 @@ on:
         options:
           - 'basic-test.json'
           - 'basic-test-nightly.json'
+          - 'basic-test-nightly-experimental.json'
           - 'model-test-push.json'
           - 'model-test-full.json'
           - 'on-pr.json'
--- a/.github/workflows/schedule-nightly-experimental.yml
+++ b/.github/workflows/schedule-nightly-experimental.yml
@@ -49,11 +49,24 @@ jobs:
       artifact_suffix: ${{ needs.build-ttxla.outputs.artifact_suffix }}
       include_passed: true
 
+  test_bh_galaxy:
+    uses: ./.github/workflows/call-test.yml
+    secrets: inherit
+    needs: [ build-image, build-ttxla ]
+    if: success() || failure()
+    with:
+      test_suite: basic-test-nightly-experimental.json
+      docker_image: ${{ needs.build-image.outputs.docker-image }}
+      artifact_release_run_id: ${{ needs.build-ttxla.outputs.artifacts_run_id }}
+      artifact_suffix: ${{ needs.build-ttxla.outputs.artifact_suffix }}
+      include_passed: true
+
   fail-notify:
     if: always()
     needs:
       - test_forge_models_experimental
       - test_forge_models_lb_blackhole
+      - test_bh_galaxy
       - build-image
       - build-ttxla
     runs-on: Ubuntu-latest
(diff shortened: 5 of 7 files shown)


sshonTT:
Shrinked bh-glx test (only test dsv4 streaming) passed : https://github.com/tenstorrent/tt-xla/actions/runs/26953060019

pjrt_implementation/src/api/client_instance.cc
```diff
@@ -405,8 +405,16 @@ tt_pjrt_status ClientInstance::populateDevices() {
 
   // Mesh device requires physical hardware; skip in compile-only mode.
   if (!m_compile_only) {
-    m_parent_mesh =
-        getOrCreateMeshDevice({1, static_cast<uint32_t>(m_devices.size())});
+    // [Workaround] BH galaxy (UBB): a 1D {1, N} parent mesh can't be reshaped
+    // to the 2D (4, 8) executable mesh (MGD solver fails); open (4, 8)
+    // directly.
+    if (std::getenv("TT_RUNTIME_USING_BH_GALAXY") != nullptr &&
```
AleksKnezevic: instead of env var could we default to 4x8 whenever there are 32 devices, otherwise 1xn. The env vars can get tricky to manage in all the various test jsons.

sshonTT: Yeah good point. It's now 4x8 by default

AleksKnezevic (approved):
Thanks!

pjrt_implementation/src/api/client_instance.cc
sshonTT: Since enabling the topology might cause performance regressions, I made it conditional based on the chip architecture. You can check the implementation : [here](https://github.com/tenstorrent/tt-xla/pull/5094/changes#diff-8222eedf41ac666575a5f299aab5715d143e9c4d52c0a5cf3961fc3d3c272fbbR525)

nsumrakTT (approved):

vmilosevic (approved):

.github/workflows/test-matrix-presets/basic-test-nightly.json
```diff
@@ -16,5 +16,5 @@
   {"runs-on": "qb2-blackhole",       "name": "run_vllm_bhqb_tests",    "dir": "./tests/integrations/vllm_plugin",         "test-mark": "nightly and bhqb",                        "extra-wheel": "vllm"},
   { "runs-on": "galaxy-wh-6u",       "name": "run_torch_galaxy",       "dir": "./tests/torch",                            "test-mark": "nightly and galaxy" },
   { "runs-on": "galaxy-wh-6u",        "name": "run_vllm_galaxy_wh_6u_tests", "dir": "./tests/integrations/vllm_plugin", "test-mark": "nightly and tensor_parallel and galaxy_wh_6u", "extra-wheel": "vllm" },
-  { "runs-on": "galaxy-bh",           "name": "run_torch_bh_galaxy",   "dir": "./tests/torch",                            "test-mark": "nightly and bh_galaxy" }
+  { "runs-on": "galaxy-bh",           "name": "run_torch_bh_galaxy",   "dir": "./tests/torch",                            "test-mark": "nightly and bh_galaxy", "timeout": 480 }
```
vvukomanTT: Before we increase the job timeout, do you maybe have an example of this job running in CI for longer than 4/6 hours?

I'd first like to see if there is a real need for this override

vvukomanTT (changes requested):
Marking as request changes, just so the previous comment doesn't get lost.

.github/workflows/test-matrix-presets/basic-test-nightly.json
sshonTT: @vvukomanTT Thanks for asking! I've run a nightly test : https://github.com/tenstorrent/tt-xla/actions/runs/27034880598/job/79801280291 
Still need to investigate what occurs oom issue, it took almost 5hrs until failure.

sshonTT: @vvukomanTT I finally got ci pass : https://github.com/tenstorrent/tt-xla/actions/runs/27177934453/job/80231833627
It took 5hrs.

vvukomanTT: This is great @sshonTT , thanks.
One big concern I have is that we only have 1 blackhole galaxy in the CI currently and this job takes 5 hours. This can be a problem if it delays nightly workflows even more. What do you think about adding this test to the `On nightly Experimental` workflow instead?

sshonTT: This is a valid concern, and I think it depends on how we want to prioritize these tests.
To be honest, ensuring the sanity of DSv4 is highly critical for us right now. In fact, it's already hitting a [timeout](https://github.com/tenstorrent/tt-xla/actions/runs/27175829468/job/80224463623) even with a single test that calculates PCC.
As a compromise, I can consider moving other test, like the weight streaming one, to the On nightly Experimental workflow. But it takes 50min, so you think if it's still valuable to move it, I can work on it.

Also for next step, I'd like to introduce pcc check on  weight streaming feature so we can move this test to On Nightly and remove existing e2e test.

vvukomanTT: Hm, I understand that the model is a priority right now, but I'm wondering what's the difference if it's test is ran in the experimental nightly workflow? The cadence is the same and we can use it in the same way to catch regressions?

sshonTT: Thanks for the clarification. I actually had the impression that the experimental workflow had lower priority, meaning regressions caught there might be treated as less critical compared to the formal nightly build.
If the cadence is indeed the same and we track/address regressions with the same level of urgency in both workflows, then moving it to experimental makes total sense to me.

Just to align on expectations, could you clarify what the main criteria or structural differences are between the formal nightly and experimental workflows?

vvukomanTT: Okay, so let me clarify what I meant.
Functionally both workflows are identical, they run on a nightly cadence. But the priorities definitely are different, tests in the main nightly workflow are fixed with more urgency than those in the experimental nightly workflow.

Let's do it like this, since the 1 blackhole galaxy we have in CI isn't utilised much currently I'm fine with this being merged. But, if more bh galaxy tests are added/people start blasting the only bh galaxy we have more, we will have to discuss what's more important and remove stuff from the CI then (or move to less vital workflows than the main nightly).

vvukomanTT: Let's discuss the timeout now, the successfull run lasted 5 hours which means we can just increase the test's timeout info in the `.test_durations` file to 6 hours, that should be enough for now. 

No need to add new timeouts flag logic.

nsumrakTT: I would like to add additional clarification about nightly as it consists of:
* full test suite
* performance benchmarks
* daily release process
* reporting (to management dashboard).
Because of all these stuff inside nightly, it can run for 12h, sometimes even more as it shares machines with all other workflows and our "night" is quite short because of different time zones we work in.
Also, because of reporting, nightly is often rerun (failed tests) . Sheer size of the nightly means that network throttling, machine disk/ram space, ... increase possibility that some jobs fail because of these infrastructure problems rather than failed tests, and rerun will "fix" this in reports.
Now, as we currently have 1 galaxy, I wonder what happens if this 5h test fails and it is rerun? My guess is that galaxy won't be available at all for any other purpose. The second problem is that nightly can easily tip to the next day and our daily processes will be ruined.
That is why we created nighty experimental  which automatically runs tests but doesn't have burden of nighly (release, reporting), as well as weekly runs (and weekly experimental) that has identical full reporting but on a lower cadence running on weekends when the pressure on machines is much lower.

Also, combination is also possible, eg. if you need reporting (it is based on a rolling 7 day window) it might be wise to include it in the weekly run and nightly experimental run. In such way, you can track it daily and have it in reports.
Anyway, this comment is just clarification, it is up to you to decide based on preference/priority/potential for fire emergency...

sshonTT: @vvukomanTT , @nsumrakTT Thanks for valid clarification! I've changed workflow to experimental. Also removed unncessary timeout module and move it to duration. Now that we only have 2 tests in the workflow we can check it easily. After sanity update on weight streaming, I will convert this test to formal nightly.

sshonTT: https://github.com/tenstorrent/tt-xla/actions/runs/27285101825/job/80594828958 nightly experimental pass

vvukomanTT (approved):
Looks great, thank you 👍

.github/workflows/test-matrix-presets/basic-test-nightly-experimental.json
```diff
@@ -0,0 +1,3 @@
+[
+  { "runs-on": "galaxy-bh",           "name": "run_torch_bh_galaxy",   "dir": "./tests/torch",                            "test-mark": "nightly and bh_galaxy", "forked": true }
+]
```
jameszianxuTT, line 3: Is it safe/necessary to run this in forked mode?

pjrt_implementation/src/api/client_instance.cc
```diff
@@ -510,6 +517,17 @@ ClientInstance::computeFabricConfig(const std::vector<uint32_t> &mesh_shape) {
                         : tt::runtime::FabricConfig::DISABLED;
     return tt::runtime::MeshFabricConfig{global, {}};
   }
+  // [Workaround] The Blackhole galaxy (UBB) lacks a both-axis wrap, so
```
jameszianxuTT, line 524: I think this might be a constraint not for blackhole galaxy in general but for the specific way the blackhole galaxy we have in CI is cabled. 

Anyways, doesn't matter -- I'm working on getting a better solution from scaleout than adding these w/a, but I'm okay with w/a for now

pjrt_implementation/src/api/client_instance.cc
```diff
@@ -510,6 +517,17 @@ ClientInstance::computeFabricConfig(const std::vector<uint32_t> &mesh_shape) {
...
+  // computeMeshFabricConfig's RING_RING is rejected as TORUS_XY by the
+  // TopologyMapper; force FABRIC_1D there. Other 32-device galaxies (e.g.
+  // Wormhole) have the wrap, so keep their auto-detected fabric.
+  bool is_blackhole = m_system_descriptor->chip_descs()->size() > 0 &&
+                      m_system_descriptor->chip_descs()->Get(0)->arch() ==
+                          ::tt::target::Arch::Blackhole;
+  if (is_blackhole && m_devices.size() == 32) {
+    return tt::runtime::MeshFabricConfig{tt::runtime::FabricConfig::FABRIC_1D,
```
jameszianxuTT, line 536: I really suggest adding a warning log here, overriding fabric config automatically is not expected behaviour

pjrt_implementation/src/api/client_instance.cc
```diff
@@ -407,8 +407,15 @@ tt_pjrt_status ClientInstance::populateDevices() {
 
   // Mesh device requires physical hardware; skip in compile-only mode.
   if (!m_compile_only) {
-    m_parent_mesh =
-        getOrCreateMeshDevice({1, static_cast<uint32_t>(m_devices.size())});
+    // [Workaround] On a 32-device galaxy (UBB) a 1D {1, N} parent mesh can't be
+    // reshaped to the 2D (4, 8) executable mesh (the MGD solver fails); open
```
jameszianxuTT, line 411: I believe this is a fabric issue and not a fundamental constraint? And also possibly a regression, since we never had problems doing this before

A warning here + link to issue (maybe this one? https://github.com/tenstorrent/tt-metal/issues/43210) would be good

jameszianxuTT (approved):
some nits - I am concerned about the lack of visible warnings when overriding default behaviour 

Otherwise, thanks for finding this workaround

.github/workflows/test-matrix-presets/basic-test-nightly-experimental.json
sshonTT: Yes, it's necessary. Without it, the dsv4 streaming run will leave DRAM data uncleared, and I wanted to avoid adding cleanup logic in the test. It should be safe since we already have a passing [record](https://github.com/tenstorrent/tt-xla/actions/runs/27285101825/job/80594828958)

pjrt_implementation/src/api/client_instance.cc
sshonTT: Added warning log.

pjrt_implementation/src/api/client_instance.cc
sshonTT: Added warning log and link

PR #5095, opened 2026-06-04
[vLLM] Skip KV cache initialization

mmanzoorTT:
### Problem description
KV cache initialization crashes if no layer requires kv cache. This may occur if we are running model partially by setting num_hidden_layers.

### What's changed
Skip KV cache initialization if not required.


kmabeeTT (approved):
Nice, looks legit.

Issue #5096, opened 2026-06-04
Low pcc for deepseekv3 and glm4 benchmarks


gengelageTT:
Debug the following:
### Deepseek-v3.1
`test_deepseek_v3_1_tp_galaxy_4_layers` sees  ([run here](https://github.com/tenstorrent/tt-xla/actions/runs/26897471605/job/79346326710#step:14:284))
```
Prefill PCC verification passed with PCC=0.916316
First decode PCC verification passed with PCC=0.965493
```
but was seeing `0.998` for both prefill and decode ~ 2 weeks ago (regression). When running more layers (30) this drops to a pcc below `0.10`. 

### GLM4

`test_glm_4_5_tp_galaxy_4_layers` and `test_glm_4_7_tp_galaxy_4_layers` both see low pcc as well ([run here](https://github.com/tenstorrent/tt-xla/actions/runs/26847561519/job/79174846841#step:14:261)):
```
Prefill PCC verification passed with PCC=0.933172
First decode PCC verification passed with PCC=0.884639
```
and
```
Prefill PCC verification passed with PCC=0.941311
First decode PCC verification passed with PCC=0.860738
```

The decode pcc is the main concern for these tests since it's below `0.90` but since these are only testing 4 layers it's likely the prefill pcc will continue to drop when more layers (GLM4.7 is a 92 layer model) are run. Would be great to get both prefill and decode to `0.99`.

dgolubovicTT:
@ndrakulicTT is it currently possible to run single layer of GLM/Deepseek through chisel and get report on what operations lower the accuracy the most?

gengelageTT:
@dgolubovicTT I've tried chisel for deepseek-v3.1 4 layers and it successfully generated a report for both prefill and decode. The first op failure I looked into was `ttnn.softmax` but that ended up being a bug in chisel that has already been resolved in a newer version of tt-mlir. Rebasing this test to latest tt-xla included an MoE change that improved the prefill pcc for deepseek so I'll re-run with chisel including those changes and continue to debug.

dgolubovicTT:
Can you share what is the current PCC on single layer of deepseek and glm4?
We can have 2 sources of accuracy errors:
- single op introducing mismatch - we can catch that using chisel on 1 decoder layer.
- error accumulation: all ops in isolation ok, but error accumulates over layers. Harder to fix

Let's see if we have one or the other, or both...

gengelageTT:
[...]

Issue #4822, opened 2026-05-21
Add BH galaxy benchmark tests


odjuricicTT:
## Problem

We need benchmark coverage on BH Galaxy for all big LLMs that run on it (gpt-oss, DS V4 flash, GLM 4.7, DS V3, Kimi k2.5/k2.6, GLM 5.1, ...).

Scope:
- Identify the set of big LLMs that currently run on BH Galaxy (see #3927)
- For each model, define a sharding strategy that follows the gpt-oss WH Galaxy throughput reference in `test_gpt_oss_120b_tp_dp_galaxy_batch_size_128` (`tests/benchmark/test_llms.py`, `_moe_throughput_galaxy_shard_spec_fn`):
    - mesh axes: `"model"` (TP) and `"batch"` (DP)
    - attention: q/k/v sharded along `"model"` on the head dim, o_proj sharded along `"model"` on the input dim
    - MoE experts: `gate_up_proj = ("model", "batch", None)`, `down_proj = ("model", None, "batch")` — model dim across TP, expert dim across DP as a stand-in until EP=32 lands
    - input/output sharding: `("batch", None)`
    - KV cache: `("batch", "model", None, None)`
    - trace enabled, batch_size=128
- Shard specs should be written out explicitly in each test (not inferred)
- Add both a standard benchmark test and, where feasible, a vLLM benchmark variant per model

Related:
- #3927 — Run GLM 4.7 and deepseek on BH Galaxy
- #3905 — Add kimi-k2, deepseek v3.2 exp, glm 4.5 to benchmark tests
- #4081 — Change gpt oss 120b sharding on galaxy

Issue #4565, opened 2026-05-07
Debug kimi-k2 benchmark hang for batch size 128


gengelageTT:
Kimi-k2 was originally passing for a batch size of 128 ([run here](https://github.com/tenstorrent/tt-xla/actions/runs/24572332012)). Now it hangs in warmup decode both in CI ([run here](https://github.com/tenstorrent/tt-xla/actions/runs/25335234331/job/74280870283)) and locally. The test passes for a batch size of 64.

Test was last passing [here](https://github.com/tenstorrent/tt-xla/actions/runs/25013267793/job/73256457093#step:15:492) with fix to avoid OOM in L1 buffer cache. The tt-mlir uplift on May 4th likely contains the source of this hang.

gengelageTT:
This hang also shows up for kimi-k2.5 with a batch size of 128 - [see run here](https://github.com/tenstorrent/tt-xla/actions/runs/26106849408/job/76782974786#step:15:48).

sshonTT:
https://github.com/tenstorrent/tt-mlir/pull/8099 This patch could be a culprit for hang. There is an pcc issue in decode step though.

sshonTT:
https://github.com/tenstorrent/tt-mlir/pull/8590 Add workaround to avoid hang.

gengelageTT:
Kimi-k2 and deepseek-v3 no longer hang for batch 128. However, batch 128 causes poor decode pcc for kimi-k2 variants which is tracked in https://github.com/tenstorrent/tt-xla/issues/5558.

PR #5097, opened 2026-06-04
Add deepseek-v3.1 and glm4.7 benchmarks

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/4822
https://github.com/tenstorrent/tt-xla/issues/3905 

### Problem description
We want to add e2e bh_galaxy benchmark tests for these models. With the current strategy of using swap to load large models, each model takes 4+ hours to run. The models can load much faster using streaming. Until streaming support is added to the benchmark tester the 4 layer tests for these models can be added. 

### What's changed
Added changes to llm_benchmark since glm4 doesn't have a `"dense"` option for `set_experts_implementation` and some of the model loaders call `get_mesh_config` internally to enable sparse moe so overrides should be applied.

Added benchmarks for deepseek-v3.1, ~~glm4.5~~ and glm4.7 for 4 layers on wh galaxy. Once the bh galaxy in CI is working in nightly these tests could be added/moved to that device. There are a few issues with these tests:
- They hang for `batch_size=128`. Should be related to https://github.com/tenstorrent/tt-xla/issues/4565 
- The pcc is low (`0.86` is the lowest but pcc is very low for more layers) - tracking here: https://github.com/tenstorrent/tt-xla/issues/5096 

Removing the existing 4 layers tests for these models from test/torch/models. 

### Checklist


gengelageTT:
Tests added are passing here: https://github.com/tenstorrent/tt-xla/actions/runs/26959846674

vvukomanTT (approved):

vmilosevic (approved):

nsumrakTT (approved):

.github/workflows/perf-bench-matrix.json
```diff
@@ -344,11 +344,26 @@
         "name": "unet_for_conditional_generation",
         "pytest": "tests/benchmark/test_encoders.py::test_unet_for_conditional_generation"
       },
+      {
+        "name": "deepseek_v3_1_tp_galaxy_4_layers",
+        "pytest": "tests/benchmark/test_llms.py::test_deepseek_v3_1_tp_galaxy_4_layers",
+        "runs-on": "galaxy-wh-6u"
```
odjuricicTT, line 347: To my knowledge we have a BH galaxy in CI: `galaxy-bh`

Can you add the tests to both galaxies?

.github/workflows/perf-bench-matrix.json
```diff
@@ -344,11 +344,26 @@
...
+      },
       {
         "name": "deepseek_v3_2_exp_tp_galaxy_2_layers",
         "pytest": "tests/benchmark/test_llms.py::test_deepseek_v3_2_exp_tp_galaxy_2_layers",
         "runs-on": "galaxy-wh-6u"
       },
+      {
+        "name": "glm_4_5_tp_galaxy_4_layers",
```
odjuricicTT: If the only difference between 4.5 and 4.7 is in the weights, than i would skip adding 4.5 altogether. Please check this.

tests/benchmark/test_llms.py
```diff
@@ -2551,3 +2550,108 @@ def test_gpt_oss_20b_tp_qb2(
         shard_spec_fn=_gpt_oss_20b_shard_spec_fn,
         optimization_level=2,
     )
+
+
+# This test only runs 2 layers so we expect to see incoherent output
```
odjuricicTT: *4 layers

tests/benchmark/benchmarks/llm_benchmark.py
```diff
@@ -74,7 +74,10 @@ def setup_model_and_tokenizer(
     # Use static dense experts forward to avoid graph breaks from data-dependent
     # loops in the original experts and _grouped_mm CPU crashes.
     if hasattr(model.config, "_experts_implementation"):
-        model.config._experts_implementation = "dense"
+        try:
+            model.set_experts_implementation("dense")
+        except (ValueError, KeyError):
+            pass  # model doesn't expose "dense"; leave config alone
```
odjuricicTT: Can we pass this as pa param from the test itself instead of the try except?

e.g. expert_implementation='dense' / None

odjuricicTT (approved):

.github/workflows/perf-bench-matrix.json
gengelageTT: > Once the bh galaxy in CI is working in nightly these tests could be added/moved to that device. 

The BH galaxy is in CI but not working in nightly - [every nightly run fails like this](https://github.com/tenstorrent/tt-xla/actions/runs/26988003374/job/79704454441#step:19:44).

@hshahTT and @vmilosevic are working on getting it to work

.github/workflows/perf-bench-matrix.json
gengelageTT: The only difference is the weights (graphs are the same) so I'll skip adding 4.5 👍

gengelageTT:
Re-running tests after rebase and small changes here: https://github.com/tenstorrent/tt-xla/actions/runs/27033833987

AleksKnezevic (approved):

gengelageTT:
Fixed how I set expert implementation and re-running deepseek-v3.1 here: https://github.com/tenstorrent/tt-xla/actions/runs/27045761539

gengelageTT:
`expert_implementation` param added conflicted with `experts_implementation` param recently added with MoE backend changes. Rebased, removed duplicate `expert_implementation` and re-running tests here: https://github.com/tenstorrent/tt-xla/actions/runs/27141917161

tests/benchmark/test_llms.py
```diff
@@ -2579,3 +2582,73 @@ def test_gpt_oss_20b_tp_qb2(
...
+        ModelLoader,
+        ModelVariant,
+    )
+
+    variant = ModelVariant.DEEPSEEK_V3_1_MODIFIED
+    test_llm_tp(
+        ModelLoader,
+        variant,
+        output_file,
+        num_layers=4,
```
mvasiljevicTT: Can we have something like `num_layers=num_layers if num_layers is not None else 4` here so that the test can be used with `--num-layers 1`?

gengelageTT: For these models, there are 3 dense layers before the 4th one which is sparse so 4 layers need to be run to test the sparse MoE. The kimi-k2 tests are similar but only have 1 dense layer so need 2 layers to test the sparse MoE. 

What are the `--num-layers 1` tests used for? I can make this change but 1-3 layers isn't sufficient to check if compile will work for example.

gengelageTT: If it would be helpful I can make changes to the model loaders so that if `num_layers` is 2-3 there will be at least 1 sparse layer. This would allow you to run 2 layers tests and still test the sparse MoE. But in that case I'd prefer to make a follow up PR since I'd need to wait for the change to get uplifted from tt-forge-models.

mvasiljevicTT: > What are the --num-layers 1 tests used for? I can make this change but 1-3 layers isn't sufficient to check if compile will work for example.

I wanted to see attention part of the model for which any layer is ok. It is also easier to extrapolate perf for full model when we can run it with different num of layers so that we can make a formula directly like (t4-t3)*n + t3

> If it would be helpful I can make changes to the model loaders so that if num_layers is 2-3 there will be at least 1 sparse layer. This would allow you to run 2 layers tests and still test the sparse MoE. But in that case I'd prefer to make a follow up PR since I'd need to wait for the change to get uplifted from tt-forge-models.

that would be awesome!

gengelageTT: Ok that makes sense, I'll make that change in this pr then so attention testing can be done. I can let you know once the sparse MoE is run in a 2 layer test

mvasiljevicTT: thanks!

Issue #5102, opened 2026-06-05
[pjrt] gemma_1_1_2b: "Device count mismatch: 1 vs N" on multi-device systems — identity mesh_shard on scalar normalizer graph inflates num_devices_to_utilize


rpavlovicTT:
## Summary

`tests/benchmark/test_llms.py::test_gemma_1_1_2b` fails on multi-device Blackhole systems (e.g. QB2, 4 chips) with:

```
flatbuffer_loaded_executable_instance.cc:108  ERR| Device count mismatch: 1 vs 4
RuntimeError: Bad StatusOr access: INTERNAL: Error code: 13
```

It runs **single-device** (`is_multichip=False`), and other single-chip LLMs (qwen, llama, mistral, phi) pass on the same box. It is **gemma-specific**. (Reproduces on QB2 hardware; n150/single-device-systems are unaffected since the mesh resolves to `1x1`.)

## Root cause

1. **The check** — `pjrt_implementation/src/api/flatbuffer_loaded_executable_instance.cc:108`:
   ```cpp
   if (args->num_devices != m_executable_image->getNumDevicesToUtilize()) { ... }
   ```
   `args->num_devices = 1` (single-device execution) but `getNumDevicesToUtilize() = 4`.

2. **Where the 4 comes from** — `module_builder.cc` `collectNumDevicesToUtilize` → `collectMeshShape`. For torch_xla (`mhlo.num_partitions`/`num_replicas` unset) it returns the module-level `ttcore.meshes` shape and `num_devices = product(shape)`.

3. **The offending graph** — gemma emits a tiny standalone scalar-constant graph (`SyncTensorsGraph.3`, graph `g0`):
   ```mlir
   module @SyncTensorsGraph.3 attributes {..., ttcore.meshes = #ttcore.meshes<[<"mesh" = 1x4>]>} {
     func.func @main() -> tensor<bf16> {
       %0 = "ttir.constant"() <{value = dense<4.525000e+01> : tensor<bf16>}>            // 45.25 == sqrt(2048)
       %1 = "ttir.mesh_shard"(%0) <{shard_type = #ttcore.shard_type<identity>,          // no-op / replicate
                                    shard_direction = shard_to_full, shard_shape = [1]}>
       return %1
     }
   }
   ```
[...]

PR #5103, opened 2026-06-05
[perf] Run single-chip benchmarks on stable qb2-blackhole via p150-perf label

rpavlovicTT:
## What

Introduce a logical **`p150-perf`** `runs-on` label for single-chip Blackhole perf workloads and make it the default single-chip runner (`n150` + `p150-perf`), retiring the flaky standalone `p150` from the perf matrix. `p150-perf` is aliased to the physical **`qb2-blackhole`** runner, so single-chip perf runs on a stable QB2 host while staying cleanly separable from the multi-chip tensor-parallel (`qb2-blackhole`) tests.

## Why

CI single-chip perf on the standalone `p150` runners was unstable (flaky/queueing nodes, weaker hosts → noisy, ~2× lower numbers). QB2 boxes are stable and faster. But QB2 is a *multi-chip* system, so we needed a way to designate "single-chip workload on QB2" that is distinct from the multi-chip `_tp_qb2` tests **without** name-substring hacks or per-test duplication.

The clean solution is the existing `runs-on` filter dimension: single-chip tests use the logical label `p150-perf`, TP tests use `qb2-blackhole`, and both map to the same physical box.

## How

- **`perf-bench-matrix.json`** — default `runs-on`: `["n150", "p150-perf"]` (drops the flaky standalone `p150`).
- **`filter-test-matrix.py`** — `update_runners` gains an unconditional alias `p150-perf → qb2-blackhole`. Filtering runs *before* the remap, so `runs-on-filter=p150-perf` selects exactly the single-chip tests and `=qb2-blackhole` selects exactly the TP tests; both execute on QB2.
- **`call-perf-test.yml`** — job name uses `runs-on-original`, so single-chip jobs display as `(p150-perf)` and TP as `(qb2-blackhole)`.
- **`manual-benchmark.yml` / `schedule-nightly.yml`** — `p150` → `p150-perf` in the runner filters.

TP/galaxy/vllm tests keep their explicit `runs-on` and are unaffected. Other workflows' `p150` usage (model-test presets) is untouched.

## Validation (on QB2 hardware)

- Filter dry-run: `p150-perf` → single-chip jobs, all routed to physical `qb2-blackhole`, labeled `(p150-perf)`; `qb2-blackhole` → only the `_tp_qb2`/vllm tests. No collisions, no duplication.
- Live runs on QB2: jobs show `(p150-perf)` on `120-qb2-*` runners. **Vision** is genuinely single-chip (`torch_xla.device()`); confirmed via compiled IR (`meshShape=1x1, chipIds=[0]`). **Single-chip LLMs** also run single-chip and pass (e.g. qwen_3_8b IR `meshShape=1x1`).

## Known QB2 failures (follow-ups, not caused by this change)

- **`google_gemma-1.1-2b-it`** — `Device count mismatch: 1 vs 4` on multi-device boxes (its `sqrt(hidden_size)` scalar graph carries a spurious identity 1×4 mesh). Root-caused and tracked in **#5102**.
- **`ufld_v2`** — known Blackhole quirk (needs `IRD_LF_CACHE`).
- **`bge_m3_encode`** — pre-existing QB2 failure (needs triage).

Reviewers may want to `xfail`/exclude these from `p150-perf` before this lands in nightly.

## Note on load

This routes all single-chip perf tests (~57 jobs) onto QB2 hardware per nightly. Flag if QB2 runner capacity is a concern — we can scope the set down.

### DIFF
--- a/.github/scripts/filter-test-matrix.py
+++ b/.github/scripts/filter-test-matrix.py
@@ -112,7 +112,12 @@ def filter_matrix_adv(matrix, adv_filter):
 
 def update_runners(matrix, sh_runner):
     """Update runner names based on shared runner flag."""
-    runner_map = {"p150": "p150b"} if sh_runner else {"n150": "n150-perf"}
+    # "p150-perf" is a logical single-chip Blackhole perf label that always maps to
+    # the physical qb2-blackhole box (kept distinct from the multi-chip qb2-blackhole
+    # tests so the runs-on filter can separate them). The shared-runner-mode remap is
+    # layered on top.
+    runner_map = {"p150-perf": "qb2-blackhole"}
+    runner_map.update({"p150": "p150b"} if sh_runner else {"n150": "n150-perf"})
 
     for item in matrix:
         item["runs-on-original"] = item.get("runs-on")
--- a/.github/workflows/call-perf-test.yml
+++ b/.github/workflows/call-perf-test.yml
@@ -61,7 +61,7 @@ jobs:
 
     runs-on: ${{ !inputs.sh-runner && fromJson(format('["{0}", "in-service"]', matrix.build.runs-on)) || format('tt-ubuntu-2204-{0}-stable', matrix.build.runs-on) }}
 
-    name: "perf ${{ matrix.build.name }} (${{ matrix.build.runs-on }})"
+    name: "perf ${{ matrix.build.name }} (${{ matrix.build.runs-on-original || matrix.build.runs-on }})"
 
     container:
       image: ${{ !inputs.sh-runner && inputs.docker_image || format('harbor.ci.tenstorrent.net/{0}', inputs.docker_image) }}
--- a/.github/workflows/manual-benchmark.yml
+++ b/.github/workflows/manual-benchmark.yml
@@ -18,7 +18,7 @@ on:
         options:
           - All
           - n150
-          - p150
+          - p150-perf
           - n300-llmbox
           - galaxy-wh-6u
           - qb2-blackhole
--- a/.github/workflows/perf-bench-matrix.json
+++ b/.github/workflows/perf-bench-matrix.json
@@ -4,7 +4,7 @@
     "test-defaults": {
       "runs-on": [
         "n150",
-        "p150"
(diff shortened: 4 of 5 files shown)


nsumrakTT (approved):

.github/scripts/filter-test-matrix.py
```diff
@@ -112,7 +112,12 @@ def filter_matrix_adv(matrix, adv_filter):
 
 def update_runners(matrix, sh_runner):
     """Update runner names based on shared runner flag."""
-    runner_map = {"p150": "p150b"} if sh_runner else {"n150": "n150-perf"}
+    # "p150-perf" is a logical single-chip Blackhole perf label that always maps to
+    # the physical qb2-blackhole box (kept distinct from the multi-chip qb2-blackhole
+    # tests so the runs-on filter can separate them). The shared-runner-mode remap is
+    # layered on top.
+    runner_map = {"p150-perf": "qb2-blackhole"}
```
vvukomanTT, line 119: I'm not a fan of the logical mapping of labels in workflow files.
The general idea is good and the rest of the changes as well, but let's remove the mapping of p150-perf -> qb2-blackhole.
I'll just add the p150-perf label to the qb2 runners.

rpavlovicTT: But we want to have separate runners for p150 (single blackhole) and QB2 (multi chip) tests. So this was an easy solution. We can improve it further.

vvukomanTT: We can still have separate runners, it's just that you can request the QB2 runner with a `p150-perf` label.
So: 
- `p150` label for a p150 runner 
- `qb2-blackhole` or `p150-perf` label for a qb2 runner.

rpavlovicTT: Sounds good, let me know when we have `p150-perf` label in place

vvukomanTT: I've added the labels and created a [PR](https://github.com/tenstorrent/tt-xla/pull/5126) that removes this mapping.

Issue #5032, opened 2026-06-01
Get batch 32 vLLM benchmark tests working and added


kmabeeTT:
Get batch-32 working for TP models in vLLM benchmark suite starting with gemma4-31b-it and qwen3-32.

Undecided if we want to replace existing b1 tests with b32, but can use non-vLLM benchmark suite for inspiration.  Depends on runtime.

Currently only run batch1 tests for some reason. I heard from @ssaliceTT that issues were seen with b32 when bringing up qwen3-32b via https://github.com/tenstorrent/tt-xla/pull/5030

```
% grep _tp_config tests/benchmark/test_vllm_benchmarks.py
def _tp_config(
def _gemma4_tp_config(model: str, batch_size: int):
    pytest.param(_tp_config("tiiuae/Falcon3-7B-Base", 1), id="falcon3-7b-tp"),
    pytest.param(_tp_config("tiiuae/Falcon3-10B-Base", 1), id="falcon3-10b-tp"),
    pytest.param(_tp_config("Qwen/Qwen3-8B", 1), id="qwen3-8b-tp"),
        _tp_config("Qwen/Qwen3-8B", 1, optimization_level=1),
    pytest.param(_tp_config("Qwen/Qwen3-14B", 1), id="qwen3-14b-tp"),
    pytest.param(_tp_config("Qwen/Qwen3-32B", 1), id="qwen3-32b-tp"),
    pytest.param(_gemma4_tp_config("google/gemma-4-31B-it", 1), id="gemma4-31b-it-tp"),
        _tp_config("Qwen/Qwen2.5-14B-Instruct", 1), id="qwen2.5-14b-instruct-tp"
        _tp_config("Qwen/Qwen2.5-Coder-32B-Instruct", 1),
        _tp_config("mistralai/Ministral-8B-Instruct-2410", 1), id="ministral-8b-tp"
        _tp_config("mistralai/Mistral-Nemo-Instruct-2407", 1),
        _tp_config("mistralai/Mistral-Small-24B-Instruct-2501", 1),
```

ssaliceTT:
Sharing what I found bringing up **qwen3-32b at batch 32 on a 1×4 QB2 mesh** (via #5030). Root cause is upstream of our code, in how torch_xla emits the prefill graph.

### Root cause: `XLA_PARAMETER_WRAPPING_THREADSHOLD`

[...]

PR #4955, opened 2026-05-27
[vLLM] batch paged_fill_cache across users in prefill (compile/perf improvement)

kmabeeTT:
### Ticket

- Closes/relates to [#4880](https://github.com/tenstorrent/tt-xla/issues/4880)
- Depends on [tt-mlir #8613](https://github.com/tenstorrent/tt-mlir/pull/8613) (verifier relaxation) and [tt-metal #45117](https://github.com/tenstorrent/tt-metal/pull/45117) (kernel relaxation); both arrive via regular tt-metal -> tt-mlir -> tt-xla uplift.

### Problem

- Motivated by **compile time and prefill performance**. `_handle_paged_attention`'s prefill branch ran a Python `for batch_idx in range(inputs.users):` loop, calling `paged_fill_cache` once per user per K/V per layer. At trace time the loop unrolls into the graph, so b32 with 32 layers produced 2048 `paged_fill_cache` ops plus ~7000 supporting `slice` / `to_device` / `get_device` ops in the compiled prefill graph (vs. 64 at b1), driving both engine init and per-prefill latency.

### What's changed

- Single hunk in `integrations/vllm_plugin/vllm_tt/attention.py` (+18 / -18). The per-user loop is replaced with one batched call per K/V: `batch_idxs = torch.arange(key_for_update.shape[0], dtype=int32, ...)` produces an N-element `batch_idx_tensor` and the op is invoked once with the full batched K/V instead of N times with per-row slices.

- b1 behavior is unchanged (`arange(1)` collapses to the same single-row call the old loop produced). Multi-user prefill is what changes.

### Testing

End-to-end coverage is unchanged:
- `tests/integrations/vllm_plugin/generative/test_kv_cache_bleed.py` exercises 4-way concurrent prefill (`max_num_seqs=4`) on TinyLlama and asserts no cross-user contamination.
- `tests/benchmark/test_vllm_benchmarks.py` Llama-3.2-3B b32 sweep is the perf source of truth.

Headlines from the b32 sweep on p150 (full numbers and raw logs in the follow-up comment):

- **Wall-clock 2–3x faster**: seq128 153 s -> 70 s, seq1024 476 s -> 147 s, seq2048 610 s -> 196 s.
- **Samples/sec is now flat across seq_len**: was 16.3 / 17.1 / 18.5 at seq 2048 / 1024 / 128 (dropping with longer seq because the compiled graph grew with `seq_len`-dependent fan-out); now ~19.0–19.3 at all three.
[...]

PR #5108, opened 2026-06-05
[vLLM] Fix qwen3-32b batch-32 TP prefill sharding collapse; enable bs32 benchmark

ssaliceTT:
### Ticket
#5032

### Problem description
The qwen3-32b vLLM benchmark fails at **batch 32** on a 1×4 tensor-parallel mesh (QB2 / 4× Blackhole).

The prefill KV-cache fill loop materializes a fresh `torch.tensor([batch_idx])` per batch, per K/V, per layer. `torch` unrolls this Python loop, lifting `batch_size × 2 × num_layers` scalar `int32` constants as graph parameters, which is ~4096 for batch 32 / 64 layers. That pushes the prefill graph's parameter count over torch_xla's parameter-wrapping threshold, so it is emitted as a thin public `@main` wrapper forwarding to a private inner function. The compiler inlines that callee but does not copy the per-argument `sdy.sharding` up to `@main`, so every entry argument is treated as `Unsharded`. The compiler then inserts a replicate `distribute_tensor` on inputs that are already mesh-distributed:

- **trace off:** runtime fatal `Can't get a single buffer from host storage distributed over mesh shape MeshShape([1,4])`
- **trace on:** `TTNNLayoutAttr` `Memory layout is required for non-SystemMemory buffer type` assertion

(decode is unaffected - at batch 1 these scalars are limited as it scales based on batch_size from my tests, hence the bs1 test has not been failing on CI.)

### What's changed
- `integrations/vllm_plugin/vllm_tt/attention.py`: hoist the per-user fill batch index into a single `torch.arrange(inputs.users)` and static-slice it (`fill_batch_idx[i:i+1]`) instead of building a `torch.tensor([batch_idx])` each iteration. A static slice lowers to compile-time slice attributes, not lifted operands, so no per-user constants are added to the graph. Prefill drops back under the parameter-wrapping threshold, the graph stays flat, `sdy.sharding` survives, and both failure modes above are resolved. `paged_fill_cache` is still single-user per call; this is purely about how the index is passed.
- `tests/benchmark/test_vllm_benchmarks.py`: convert `qwen3-32b-qb2-tp` from batch 1 to batch 32 (`use_2d_mesh=False`, `gpu_memory_utilization=0.1`) so the bs32 path is exercised.

Validated on QB2 (1×4 Blackhole) locally, will add CI test.

### Checklist
- [x] qb2 w/ 1x4 mesh benchmark: https://github.com/tenstorrent/tt-xla/actions/runs/27039167668
- [ ] n300-llmbox w/ 1x8 mesh benchmark: https://github.com/tenstorrent/tt-xla/actions/runs/27039142903
- [ ] Ensure falcon 3 7B and 10B is fixed from above llmbox test.

_Note: the branch used for the tests was a custom branch with the fix but with all tests set to `use_2d_mesh=False` for illustrative purposes to show proof of support. If needed we can set them in a later PR._

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/attention.py
+++ b/integrations/vllm_plugin/vllm_tt/attention.py
@@ -487,22 +487,35 @@ class TTAttentionBackendImpl(AttentionImpl):
             key_for_update = inputs.key.transpose(1, 2)
             value_for_update = inputs.value.transpose(1, 2)
 
+            # Hoist the per-user batch index into a single tensor and
+            # static-slice it per iteration, instead of materializing a fresh
+            # torch.tensor([batch_idx]) constant each loop. torch.export unrolls
+            # this loop, so the old form lifted (batch_size * 2) scalar int32
+            # constants per layer as graph params (batch_size * 2 * num_layers
+            # total). At batch 32 / 64 layers that is ~4096 extra params, which
+            # trips torch_xla's parameter-wrapping threshold: the prefill graph
+            # is then emitted as a @main wrapper forwarding to an inner function,
+            # and the inliner drops the per-arg sdy.sharding, so already-mesh-
+            # distributed inputs get a spurious replicate distribute_tensor
+            # ("single buffer from host storage" fatal, trace off) or trip the
+            # TTNNLayout assertion (trace on). See issue #5032. The static slice
+            # fill_batch_idx[i:i+1] lowers to compile-time slice attributes, not
+            # lifted operands, so the prefill graph stays flat.
+            fill_batch_idx = torch.arange(
+                inputs.users, dtype=torch.int32, device=k_cache.device
+            )
             for batch_idx in range(inputs.users):
                 k_cache = torch.ops.tt.paged_fill_cache(
                     k_cache,
                     key_for_update[batch_idx : batch_idx + 1],
                     attn_metadata.fill_page_table,
-                    batch_idx=torch.tensor(
-                        [batch_idx], dtype=torch.int32, device=k_cache.device
-                    ),
+                    batch_idx=fill_batch_idx[batch_idx : batch_idx + 1],
                 )
                 v_cache = torch.ops.tt.paged_fill_cache(
                     v_cache,
                     value_for_update[batch_idx : batch_idx + 1],
                     attn_metadata.fill_page_table,
-                    batch_idx=torch.tensor(
-                        [batch_idx], dtype=torch.int32, device=v_cache.device
-                    ),
+                    batch_idx=fill_batch_idx[batch_idx : batch_idx + 1],
                 )
 
         # Preserve tensor identity so XLA reuses the traced graph.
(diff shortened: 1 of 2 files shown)


integrations/vllm_plugin/vllm_tt/attention.py
```diff
@@ -487,22 +487,35 @@ def _handle_paged_attention(
...
+            # total). At batch 32 / 64 layers that is ~4096 extra params, which
+            # trips torch_xla's parameter-wrapping threshold: the prefill graph
+            # is then emitted as a @main wrapper forwarding to an inner function,
+            # and the inliner drops the per-arg sdy.sharding, so already-mesh-
+            # distributed inputs get a spurious replicate distribute_tensor
+            # ("single buffer from host storage" fatal, trace off) or trip the
+            # TTNNLayout assertion (trace on). See issue #5032. The static slice
+            # fill_batch_idx[i:i+1] lowers to compile-time slice attributes, not
+            # lifted operands, so the prefill graph stays flat.
+            fill_batch_idx = torch.arange(
```
AleksKnezevic, line 504: Are we able to create this input once at setup or does it need to be done on each iteration? Also, I think you're better off creating it on 'cpu' and moving it to 'device'

integrations/vllm_plugin/vllm_tt/attention.py
```diff
@@ -487,22 +487,35 @@ def _handle_paged_attention(
             key_for_update = inputs.key.transpose(1, 2)
             value_for_update = inputs.value.transpose(1, 2)
 
+            # Hoist the per-user batch index into a single tensor and
```
AleksKnezevic, line 490: I don't know how others feel about this, so would love your opinion @mmanzoorTT and @kmabeeTT, but I think long comments like this belong in the PR description, not necessarily the code base.

ssaliceTT: I don't disagree, the vllm_tt files tend to be comment-heavy, so I matched the existing pattern, but you're right that the mechanism write-up belongs in the PR description (and it's already there). I'll trim in-code comments moving on.

ssaliceTT:
@AleksKnezevic @mmanzoorTT putting this PR on hold for now.
It was meant to be a quick fix ahead of a better solution with downstream tt-mlir changes, as described in my comment on #5032 and verified locally with a quick POC of a batched paged_fill_cache (my tt-mlir branch `ssalice/batched-paged-fill-cache`).
From my convo with @kmabeeTT today we've figured out that this should come with #4955 once the tt-mlir uplift goes through, and I believe it should fix this bug. Will revisit if needed.

integrations/vllm_plugin/vllm_tt/attention.py
kmabeeTT: Agreed in this case yeah, though in general I am fine with a 2 line comment if it documenting something non obvious about behavior since very few people will be browsing old PR's while looking at code. I usually tell claude to do a pass to make all remaining comments minimal and concise if any truly must remain.

integrations/vllm_plugin/vllm_tt/attention.py
mmanzoorTT: We are always using same number of users for generative models; so we can create this input once; applying batch padding for fewer requests. I have another branch where I am passing `batch_size` as `attn_metadata` attribute; we can use `batch_size` to create `fill_batch_idx` during initialization instead of runtime. I can cleanup the code and @ssaliceTT can cherry-pick and try it out.

Issue #4007, opened 2026-03-30
[gemma3]-Out of Memory: Not enough space to allocate 2904555520 B DRAM buffer across 12 banks


devisettymahidhar608:
While running the following models on the latest main branch, an out-of-memory error is encountered.

```
gemma3/multimodal/pytorch-google/gemma-3-4b-it-single_device-inference
gemma3/multimodal/pytorch-google/gemma-3-12b-it-single_device-inference
gemma3/multimodal/pytorch-google/gemma-3-27b-it-tensor_parallel-inference

```
error:
`Out of Memory: Not enough space to allocate 2904555520 B DRAM buffer across 12 banks, where each bank needs to store 242049024 B, but bank size is 1071821792 B`

ci run: https://github.com/tenstorrent/tt-xla/actions/runs/23676482960

sonalibaskaran2499:
@AleksKnezevic  @kmabeeTT 
## Summary

`aten.masked_scatter` fails to compile on TT hardware for all multimodal models (Gemma3, LLaVA, DeepSeek OCR). Depending on which lowering path is taken, the op hits one of two blockers:

1. **Shardy propagation error** — the default torch lowering produces a  dynamically shaped tensor.
2. **Out-of-Memory** — static-shape workarounds using flat prefix-sum create `[N×H]` index tensors that exceed DRAM capacity.

---

## Error Logs

**Error 1 — Shardy propagation (default torch path)**
```
loc("set-dimension-size.76"): error:
Shardy propagation only supports ranked tensors with a static shape.
type: 'tensor<?x3xi32, #stablehlo.bounds<1489152, ?>>'
```

**Error 2 — Out of Memory (flat static-shape path)**
```
TT_FATAL @ bank_manager.cpp:462
Out of Memory: Not enough space to allocate 3049783296 B DRAM buffer
across 12 banks, where each bank needs to store 254148608 B,
but bank size is 1071821792 B (allocated: 656514464 B, free: 415307328 B)
```

---

## Root Cause

Cumsum over all `B×SxH` flattened elements produces static shapes, but creates a large index tensor. TT TTNN internally tiles tensors in 512-column blocks, amplifying the allocation:

```
Gemma3 input shape:  [1, 277, 5376]  →  N=277, H=5376, N×H=1,489,152

[N×H] int64  =  1,489,152 × 8 B  =   11.93 MB  (actual)
On TT device =  512 × 11.93 MB   =    6.10 GB  ← OOM  
```

---

[...]

PR #5115, opened 2026-06-08
Adds sliding-window attention support for **Gemma3 multimodal** models

devisettymahidhar608:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/4007

### Problem description
Adds sliding-window attention support for **Gemma3 multimodal** models, extending the generic sliding-attention rewrite infra (#4975).
Unlike text-only models, multimodal Gemma3 builds its full attention mask mapping ahead of the decoder via
`create_causal_mask_mapping` -> `create_masks_for_generate` and passes it to the text model as a dict, so `create_sliding_window_causal_mask` is never called directly. This PR intercepts that path while leaving the
existing text-only behavior unchanged.

### What's changed
- `override_model_sliding_window_causal_mask`: additionally rebinds `create_masks_for_generate` -> `tt_create_masks_for_generate` when the model's modeling module exposes it. The `hasattr` guard keeps text-only
  models (Olmo3 / Mistral / GPT-OSS) on the direct-call path.
- New `tt_create_masks_for_generate`: mirrors stock `create_masks_for_generate`, routing `sliding_attention` layers to the always-roll TT mask and `full_attention` / `chunked_attention` to the stock builders, forwarding `or_mask_function` / `and_mask_function`.
- `tt_create_sliding_window_causal_mask`:
  - Overlays model-supplied `or_mask_function` / `and_mask_function`
    (Gemma3's `token_type_ids` image bidirectional mask), gated by
    `valid_kv` so invalid always-roll buffer slots (`kv_pos < 0`) are
    never re-enabled.
  - Derives `cache_position` from `position_ids`, or from the cache /
    `inputs_embeds` length when neither is supplied (the multimodal
    mask-mapping path passes `position_ids=None`).
    
 Logs:
 
[gemma3_27b.log](https://github.com/user-attachments/files/28693950/gemma3_27b.log)
[gemma3_12b.log](https://github.com/user-attachments/files/28693951/gemma3_12b.log)
[gemma3_4b_0.log](https://github.com/user-attachments/files/28694553/gemma3_4b_0.log)

### Checklist

### DIFF
--- a/python_package/tt_torch/transformers_overrides.py
+++ b/python_package/tt_torch/transformers_overrides.py
@@ -5,7 +5,11 @@ from typing import Any, Optional
 
 import torch
 from transformers.cache_utils import StaticCache, StaticLayer, StaticSlidingWindowLayer
-from transformers.masking_utils import create_sliding_window_causal_mask
+from transformers.masking_utils import (
+    create_causal_mask,
+    create_chunked_causal_mask,
+    create_sliding_window_causal_mask,
+)
 
 
 def override_model_sliding_window_causal_mask(model):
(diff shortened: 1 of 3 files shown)


sonalibaskaran2499 (approved):

jameszianxuTT (approved):
I'm not sure I have full context but the implementation of this looks plausible.

I think there is a lot of unchecked complexity building up here and there may be room for functional correctness errors. Doesn't need to be in this PR, but if possible I would like to see some pytorch-only unit tests to confirm that these causal mask generation mechanics do what they say they do under all the various overloaded configurations we have them in

kmabeeTT:
@jazpurTT can you please take a look at this PR as additional set of eyes?

jazpurTT (approved):
lgtm

Issue #3883, opened 2026-03-24
Add support for MLA in vLLM plugin


hshahTT:
### Is your feature request related to a problem? Please describe

Currently we do not support efficient MLA based models in our vLLM plugin.

### Describe the solution you'd like
- [TT-XLA] have the vLLM plugin output SHLO custom calls targeting the following TTNN ops:
    - for prefill: [flash_mla_prefill](https://github.com/tenstorrent/tt-metal/blob/1a68850e18ad4309e65417e0d44fe361b3bda335/ttnn/cpp/ttnn/operations/transformer/sdpa/sdpa.cpp#L208)
    - for decode: [flash_multi_latent_attention_decode](https://github.com/tenstorrent/tt-metal/blob/1a68850e18ad4309e65417e0d44fe361b3bda335/ttnn/cpp/ttnn/operations/transformer/sdpa_decode/sdpa_decode.cpp#L151)
- [TT-MLIR] Lower the custom call: shlo -> ttir -> ttnn

PR #5119, opened 2026-06-08
Add MLA implementation to vLLM

hshahTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3883

### Problem description
We currently do not have support for MLA models in vLLM

### What's changed
- Custom MLA attention backend & layer
    - New MLA classes (`TTMLAAttentionBackend, TTMLAAttentionBackendImpl, TTMLAAttention`) added
    - `TTMLAAttentionBackendImpl` has a single unified `forward()` function that handles both prefill and decode via Q absorption into latent space
- New custom ops added to handle MLA prefill and decode:
    - Prefill: `tt::flash_mla_prefill` (lowers to `ttnn.transformer.flash_mla_prefill`)
    - Decode: `tt::paged_flash_mla_decode` (lowers to `ttnn.transformer.paged_flash_multi_latent_attention_decode`)
- vLLM model_runner changes for MLA KV cache:
    - `MLAAttentionSpec` allocates a single concatenated latent KV tensor per slot (num_kv_heads == 1, head_size == kv_lora_rank + qk_rope_head_dim) instead of separate [k_cache, v_cache] pairs.
- Added an OOT override for the `SharedFusedMoE` class used by Deepseek models. This override just disables overlapping shared experts computation with a2a dispatch so that the existing `TTFusedMoE` override can be used.

Note: the following tt-mlir PRs must be submitted and uplifted before MLA support can be fully realized:
    - https://github.com/tenstorrent/tt-mlir/pull/8636
    - https://github.com/tenstorrent/tt-mlir/pull/8703

**Important note:** 
For prefill we route through the MQA path (i.e., all query heads (absorbed into latent space) attend against a **single** shared latent KV head via `tt::flash_mla_prefill`).
The [MLA paper](https://arxiv.org/pdf/2405.04434) describes the MHA path (where all query heads attend against all KV heads) as being better for prefill, while the MQA path is better for decode. However, the tt-metal MLA prefill op is only tested against the MQA path, so we're using this as the starting point.

### Checklist


integrations/vllm_plugin/setup.py
```diff
@@ -56,5 +56,8 @@ def get_tag(self):
     ],
     python_requires=">=3.12, <3.13",
     license="Apache-2.0",
-    entry_points={"vllm.platform_plugins": ["tt = vllm_tt:register"]},
+    entry_points={
+        "vllm.platform_plugins": ["tt = vllm_tt:register"],
+        "vllm.general_plugins": ["tt_mla = vllm_tt:register_mla_oot_layer"],
```
sshonTT: This registration looks good to use for moe backend. Thanks!

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2500,9 +2478,17 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
         if self.enable_tensor_parallel:
             # Shard KV Cache — each entry is [k_cache, v_cache].
             for kv_pair in self.kv_caches:
-                for cache in kv_pair:
+                is_pair = isinstance(entry, (list, tuple))
```
sshonTT, line 3019: Seems like `entry` is missing. maybe `for entry in self.kv_caches`? or entry->kv_pair?

integrations/vllm_plugin/vllm_tt/__init__.py
```diff
@@ -6,14 +6,22 @@
 
 from vllm.v1.attention.backends.registry import AttentionBackendEnum, register_backend
 
-# Register TT attention backend at module import time
+# Register TT attention backends at module import time
 register_backend(
     backend=AttentionBackendEnum.CUSTOM,
     class_path="vllm_tt.attention.TTAttentionBackend",
 )
+register_backend(
```
sshonTT, line 14: I see the MLA backend is registered here , but I couldn't find the routing that makes vLLM actually select it. Am I missing where the use_mla -> FLASH_ATTN_MLA routing happens?

integrations/vllm_plugin/vllm_tt/model_runner.py
hshahTT: good catch! forgot to rename kv_pair to entry. Fixed now

integrations/vllm_plugin/vllm_tt/__init__.py
hshahTT: Added the missing path in platform.py, thx for pointing this out!

This is the registration flow:
1. at vllm module import time, this `register_backend` call replaces the default CUDA `FLASH_ATTN_MLA` backend path with our own backend class. This allows vllm to find the correct implementation class to use.

2. vllm also loads the `general_plugins` entry point (the `register_mla_oot_layer` function) which just imports the attention_mla.py file. This import causes the `MultiHeadLatentAttentionWrapper.register_oot` decorator to be called, which swaps the attention class to use `TTMLAAttention` and gives us more control over the input tensors that get sent to the implementation's forward function

integrations/vllm_plugin/vllm_tt/attention_impls/attention_mla.py
```diff
@@ -0,0 +1,466 @@
...
+        Heuristic: prefill when more than one token per user.
+        """
+        if attn_metadata is None or attn_metadata.cache_position is None:
+            # Treat profiling runs as prefill
+            return True
+        users = attn_metadata.cache_position.shape[0]
+        if users == 0:
+            return True
+        total_tokens = q_nope.shape[0]
+        return (total_tokens // users) > 1
```
jameszianxuTT, line 256: I'm not sure what the cache_position dims are supposed to be here, are you sure this is a sensible measurement for prefill (users == 0?) 

Also q_nope is `q_nope:      [tokens, num_heads, qk_nope_head_dim]` -> The tokens is flattened for batch x seqlen right? -- so what is the purpose of taking floor division? It would seem that infer_is_prefill will be an odd heuristic, unless the flattened inputs are already padded in some way. 

eg. 3 users with seqlen 1,1,3 -> 5//3 = 1 but 1,1,4 -> 6//3 = 2  so is_prefill bounces between true/false depending on particular seqlen per user?

(Sorry if this makes no sense as my vllm exposure is low)

sshonTT (approved):

integrations/vllm_plugin/vllm_tt/attention_impls/attention_mla.py
hshahTT: good catch on the `users == 0`. I should def be asserting on that instead of assuming prefill.

Re: the floor division stuff, I talked to @mmanzoorTT  and he confirmed that our plugin's scheduler will schedule requests such that the `tokens` in the forward function consist of either all prefill or all decode. Furthermore, we pad our prefill requests to the max_seq_len in the request. So cases like 1,1,3 or 1,1,4 are not possible. In the prefill case it'll be something like 3,3,3; or 1,1,1 in the decode case. I'll make that clearer in the function description.

jameszianxuTT (approved):

tests/torch/ops/test_custom_ops.py
```diff
@@ -229,3 +229,129 @@ def test_paged_scaled_dot_product_attention_decode(
...
+    query = torch.randn(batch_size, num_heads, seq_len, head_size, dtype=torch.bfloat16)
+    key = torch.randn(
+        batch_size, num_kv_heads, seq_len, head_size, dtype=torch.bfloat16
+    )
+    value = (
+        torch.randn(batch_size, num_kv_heads, seq_len, head_dim_v, dtype=torch.bfloat16)
+        if has_value
+        else None
+    )
+    attn_mask = (
```
mmanzoorTT, line 272: Currently, `model_runner` always uses `is_causal=True` and `attn_mask=None`. Is there any case where we do need attn_mask for mla_attention layer?

mmanzoorTT (commented):
Can you add a model for testing? You can only run first few layers with `num_hidden_layers=N`; this will only load weights for specified layer(s) and you can run subset of a model on a small device. This suggestion is for running something end-to-end.

tests/torch/ops/test_custom_ops.py
hshahTT: Not at the moment, but for DSA I will need to add an attention mask as input to the MLA custom ops.

hshahTT:
> Can you add a model for testing?

@mmanzoorTT I need these 2 tt-mlir PRs to be submitted and uplifted before I can do that (in addition to @sshonTT's MoE [PR](https://github.com/tenstorrent/tt-xla/pull/5141)):
- https://github.com/tenstorrent/tt-mlir/pull/8636
- https://github.com/tenstorrent/tt-mlir/pull/8703

Once that's done I can create another PR with an E2E full model test (I've already tested it on a separate branch). However, I'd like to get this PR submitted first so I can be unblocked for my follow up DSA PR

integrations/vllm_plugin/vllm_tt/attention_impls/attention_mla.py
```diff
@@ -0,0 +1,467 @@
...
+        k_for_kernel = k_lat.transpose(1, 2).contiguous()  # [b, 1, S, L+R]
+
+        out_lat = torch.ops.tt.flash_mla_prefill(
+            query=q_for_kernel,
+            key=k_for_kernel,
+            head_dim_v=self.kv_lora_rank,
+            value=None,
+            is_causal=attn_metadata.is_causal if attn_metadata is not None else True,
+            scale=self.scale,
+        )  # [b, N, S, L]
```
mmanzoorTT, line 282: You should pass `attn_mask` to the op.

integrations/vllm_plugin/vllm_tt/attention_impls/attention_mla.py
```diff
@@ -0,0 +1,467 @@
...
+            out_lat,
+            layer.W_UV.to(device=device),
+        ).to(
+            act_dtype
+        )  # [b, N, S, V]
+
+        # Reshape to vLLM's output contract: [tokens, N * V]
+        out = out.transpose(1, 2).reshape(
+            users * seq_len, self.num_heads * self.v_head_dim
+        )
```
mmanzoorTT, line 296: Are you planning to run the MLA attention with flattened inputs? If not then this may be problematic. We will be starting with batched inputs, this transpose will make them flattened inputs. This will cause inconsistent execution and may not work at all. 
Or you are relying on the wrapper to flatten the inputs and then reshape the output to batched?

integrations/vllm_plugin/vllm_tt/attention_impls/attention_mla.py
```diff
@@ -0,0 +1,467 @@
...
+                k_lat_for_update,
+                attn_metadata.cache_position,
+                attn_metadata.page_table,
+            )
+            kv_cache.copy_(updated_cache)
+
+        # Call paged MLA decode kernel.
+        # It expects query tensor to be of shape [1, users, N, L+R] and reads K/V
+        # straight from the paged cache.
+        is_causal = attn_metadata.is_causal if attn_metadata is not None else True
```
mmanzoorTT, line 359: Add an assert if `is_causal` is `False` and `attn_mask` is `None`. Same `assert` should be added for `prefill` op.

mmanzoorTT (commented):
Please create another folder for `test_mla_attention_impl.py`. I also suggest to create a directory for attention layers implementation and move both `attention.py` and `attention_mla.py`.

integrations/vllm_plugin/vllm_tt/__init__.py
```diff
@@ -19,8 +23,12 @@ def register():
...
+def register_mla_oot_layer():
+    from . import attention_mla  # noqa: F401
+
+
 def register_moe_oot_layer():
     # OOT-registers TTFusedMoE (CustomOp.register_oot) so Gemma-4's FusedMoE
     # uses our dense / expert-parallel routing path under XLA SPMD. Mirrors the
     # MLA backend's register_*_oot_layer + vllm.general_plugins pattern.
-    from .layers.fused_moe import TTFusedMoE  # noqa: F401
+    from .layers.fused_moe import TTFusedMoE, TTSharedFusedMoE  # noqa: F401
```
mmanzoorTT: What do you think about register all OOT together in a single function?

tests/integrations/vllm_plugin/oot_backends/test_mla_attention_impl.py
```diff
@@ -0,0 +1,452 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```
mmanzoorTT, line 1: Do we still need these tests? `test_tensor_parallel_mla_prefill_decode` is running the model end-to-end and `test_custom_ops.py` is covering the custom op implementation?

tests/integrations/vllm_plugin/generative/test_tensor_parallel_generation.py
```diff
@@ -75,6 +75,40 @@ def test_tensor_parallel_generation_llmbox_small(
     check_host_memory(model_name)
 
 
+@pytest.mark.nightly
+@pytest.mark.tensor_parallel
+@pytest.mark.llmbox
+@pytest.mark.parametrize("model_name", ["deepseek-ai/DeepSeek-V2-Lite"])
+def test_tensor_parallel_mla_prefill_decode(model_name: str):
```
mmanzoorTT: You can add this test with `test_tensor_parallel_generation_llmbox_large`; You can use `mesh_shape:[2, 4]` instead of `use_2d_mesh:True`, add `flat_mode_io` parameter. You will also need `enable_const_evel:False` if this model does not work with constant evaluation.

integrations/vllm_plugin/vllm_tt/attention_impls/attention_mla.py
hshahTT: Done

integrations/vllm_plugin/vllm_tt/attention_impls/attention_mla.py
hshahTT: Done. I added the assert to the custom_ops.py file for both the prefill and decode ops

integrations/vllm_plugin/vllm_tt/__init__.py
hshahTT: Good idea. Done

tests/integrations/vllm_plugin/oot_backends/test_mla_attention_impl.py
hshahTT: I'd prefer to keep them to help ensure that the MLA backend implementation is always as expected. It provides a nice middle ground between the individual custom op tests and the full E2E tests imo.

tests/integrations/vllm_plugin/generative/test_tensor_parallel_generation.py
hshahTT: Done, thx for the pointer

integrations/vllm_plugin/vllm_tt/layers/fused_moe.py
mmanzoorTT:
```suggestion
``@CustomOp.register_oot``; the import is fired from ``register_oot_layer``
```

integrations/vllm_plugin/vllm_tt/attention_impls/__init__.py
```diff
@@ -0,0 +1,3 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
mmanzoorTT:
```suggestion
# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```

mmanzoorTT (approved):

PR #5125, opened 2026-06-08
[vLLM] Fix num_hidden_layers override for text-only models

### LINKED ISSUE #5124: [vLLM] num_hidder_layer override is not working for text-only LLMs
vLLM is applying num_hidder_layer override to multi-modal models but not for text-only models.

mmanzoorTT:
### Ticket
closes #5124

### Problem description
num_hidden_layers override is not being applied for text-only models

### What's changed
- Apply num_hidden_layers override for both text-only and multi-modal models
- Code refactoring and reusing shard override logic across both runners
- Remove locally applied torch_dynamo_tt_device_compatibility patch

### Checklist


kmabeeTT (approved):
Thanks, looks good. Had to refresh my memory, removal of `torch_dynamo_tt_device_compatibility()` calls safe because of global import time call of `apply_dynamo_compatibility_patches()` inside `python_package/tt_torch/__init__.py`

PR #5127, opened 2026-06-08
[CI] Add CPU only tests

vvukomanTT:
### Ticket
Slack issue

### Problem description
There is no option to run tests on CPU only runners in `call-test.yml`.

### What's changed
Add an option to run tests on CPU only runners.

### Checklist


jameszianxuTT (approved):
thank you! cc @jasonmacTT

nsumrakTT (approved):

PR #5129, opened 2026-06-08
[vLLM] Warmup phase optimization

### LINKED ISSUE #5267: [vLLM plugin] Graph Optimization & Reduction
### Known problems and issues
1. Multiple paths/implementations for same post-processing task; which can be problematic to maintain. 
2. JIT-compiled graphs during inference/serve outside of warmup phase (for more details: https://github.com/tenstorrent/tt-xla/issues/5247)
3. Multiple graphs execution for decode. 
4. Code refactoring/cleanup required.

mmanzoorTT:
### Ticket
closes #5267 

### Problem description
Graphs are JIT-compiled at inference time outside of warmup phase and multiple implementations for post-processing tasks.

### What's changed
- Uses the same post-processing code/implementation for fused and unfused paths.
- Fused Decode graphs: Creates 4 fused decode graphs (model backbone + 4 variations of greedy/non-greedy sampling and grammar/no-grammar combinations).
- Fused Prefill graphs: Creates 1 model backbone and 4 post-processing graphs for each input prefill length.
- Flag 'enable_decode_fused_graphs' [Default: False]: Whether to follow fused decode graphs path (4 graphs; combined model forward and post-processing graphs) or fused prefill graphs path (5 graphs; separate model forward and post-processing graphs).
- Unfused graphs. All graphs are separate and this path is activated when cpu_sampling is True.
- Avoid graph creation/compilation during inference. Prefill graphs are generated during inference currently. 
- Add test asserting no graph recompiles at inference/serve time.

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -251,6 +251,7 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         self.device = device
         self.check_recompilation = envs.VLLM_XLA_CHECK_RECOMPILATION
         self.use_flat_model_io = self.tt_config.flat_model_io
+        self.enable_decode_fused_graphs = self.tt_config.enable_decode_fused_graphs
 
         # SPMD Related
         self.enable_tensor_parallel = self.tt_config.enable_tensor_parallel
@@ -1496,6 +1497,7 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
     def sample_tokens(
         self, grammar_output: "GrammarOutput | None"
     ) -> ModelRunnerOutput:
+        logger.info(f"sample_tokens called with grammar_output")
         if self.scheduler_output is None:
             # Nothing to do (PP non-final rank case), output isn't used.
             return None  # type: ignore[return-value]
@@ -1572,7 +1574,7 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                     )
                 )
             else:
-                if self.position_ids.shape[-1] == 1:
+                if self.position_ids.shape[-1] == 1 and self.enable_decode_fused_graphs:
                     hidden_states, logits, selected_token_ids, kv_connector_output = (
                         self._model_decode(
                             input_ids,
@@ -2122,7 +2124,7 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                 dummy_grammar_bitmask,
                 dummy_bitmasks,
             ) = self._get_dummy_inputs(config)
-            if config["num_tokens"] == 1:
+            if config["num_tokens"] == 1 and self.enable_decode_fused_graphs:
                 _, _, _, _ = self._model_decode(
                     dummy_inputs,
                     dummy_positions,
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -1548,7 +1548,9 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                     require_struct_decoding,
                     grammar_bitmask_padded,
                     bitmask_arange,
-                ) = self.prepare_structured_decoding_input(grammar_output, num_reqs)
+                ) = self.prepare_structured_decoding_input(
+                    grammar_output, self.max_num_reqs
+                )
             torch_xla.sync(wait=False)
 
             # Decode path
@@ -1570,7 +1572,7 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                     )
                 )
             else:
-                if input_ids is not None and input_ids.shape[1] == 1:
+                if self.position_ids.shape[-1] == 1:
                     hidden_states, logits, selected_token_ids, kv_connector_output = (
                         self._model_decode(
                             input_ids,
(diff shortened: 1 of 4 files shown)


integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1565,68 +1522,83 @@ def sample_tokens(
...
+            apply_grammar = grammar_output is not None
+            require_struct_decoding = None
+            grammar_bitmask_padded = None
+            bitmask_arange = None
+            if apply_grammar == True:
+                (
+                    require_struct_decoding,
+                    grammar_bitmask_padded,
+                    bitmask_arange,
+                ) = self.prepare_structured_decoding_input(grammar_output, num_reqs)
```
kmabeeTT: This passes the *actual* `num_reqs`, but the old code always sized these tensors to `max_num_reqs` (it derived `num_reqs` from `logits.shape[0]`, and `logits_indices`/`logits` are always padded to `max_num_reqs`).

  Downstream, `apply_grammar_bitmask` does `(...).reshape(logits.shape[0], -1)` where `logits.shape[0] == max_num_reqs` (logits come from `select_hidden_states` on the padded `logits_indices`). With grammar tensors sized to `num_reqs`, that reshape will fail (or misalign) whenever `num_reqs < max_num_reqs` — i.e. any structured-output request on a non-full batch. This hits both the fused and `_model_unfused` paths.

  Should this be `self.max_num_reqs` instead of `num_reqs`?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2193,97 +2581,38 @@ def _precompile_gather_logprobs(self) -> None:
...
-            self._precompile_backbone()
-            if self.tt_config.decode_only:
-                return
-            self._precompile_mm_encoder()
             if not self.tt_config.cpu_sampling:
-                # decode_postprocess handles both decode (num_tokens=1) and
-                # prefill (last token pre-selected eagerly before the call).
-                self._precompile_decode_postprocess()
+                logger.info("Warmup mode: fused runtime path")
+                self._precompile_model_fused()
```
kmabeeTT, line 2616: The fused `decode_only` path now falls through to `_precompile_mm_encoder()` and `_precompile_gather_logprobs()`, whereas the old code returned right after the backbone (skipping both). The unfused branch still returns early at 2607, so the two are asymmetric. Intentional? In particular, is warming the MM encoder wanted for an MM model in `decode_only` mode?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1565,68 +1522,83 @@ def sample_tokens(
...
-                    batch_idx = torch.arange(
-                        self.max_num_reqs, dtype=torch.int32, device=self.device
-                    )
-                    hidden_states = hidden_states[
-                        batch_idx, logits_indices, :
-                    ].unsqueeze(1)
-                    if self.enable_tensor_parallel and self.use_2d_mesh:
-                        xs.mark_sharding(
-                            hidden_states, self.mesh, (None, None, "model")
+                if input_ids is not None and input_ids.shape[1] == 1:
```
kmabeeTT: Decode is selected via `input_ids is not None and input_ids.shape[1] == 1`. When `input_ids` is `None` (pure inputs_embeds), a 1-token step falls through to `_model_prefill`, but warmup only compiles `_model_prefill` for `num_tokens > 1` — so this would trigger a cold compile during inference. Likely fine if decode always has `input_ids`, but worth confirming the guard can't send a decode step down the prefill branch.

kmabeeTT (commented):
Overall seems good, left few things that claude helped me find questions on.

integrations/vllm_plugin/vllm_tt/model_runner.py
mmanzoorTT: Thanks for pointing it out. I have updated as suggested.

integrations/vllm_plugin/vllm_tt/model_runner.py
mmanzoorTT: You are right. Added early return for fused path in case of `decode_only` graph generation.

integrations/vllm_plugin/vllm_tt/model_runner.py
mmanzoorTT: I was planning to use `position_ids` but forgot. `positioin_ids` is always available and `input_ids` will be None incase of multi-modal inputs.

tests/integrations/vllm_plugin/generative/test_tensor_parallel_generation.py
```diff
@@ -25,6 +25,8 @@ def test_tensor_parallel_generation_n300(model_name: str):
             "enable_const_eval": False,
             "min_context_len": 32,
             "enable_tensor_parallel": True,
+            "num_hidden_layers": 1,
```
AleksKnezevic: was this intentional?

AleksKnezevic (approved):

tests/integrations/vllm_plugin/generative/test_tensor_parallel_generation.py
mmanzoorTT: I added them for local testing. Removed.

kmabeeTT (approved):
Thx for feedback changes, if you include regression test to catch when JIT compiles happen (from issue https://github.com/tenstorrent/tt-xla/issues/5247#issuecomment-4733544064) here that is now passing with these changes would be great!   Should also make obvious from PR Desc (and adjust title) that by default w/o `enable_decode_fused_graphs` this change will actually increase number of graphs (from 12 to 37 according to my test in that issue) rather than reduce....

PR #5140, opened 2026-06-09
[FX Fusing] Add RMSNorm fusion patterns for Llama, GPT-OSS, Gemma family, and vLLM

alinakhanTT:
## Summary

Extends `RMSNormFusionProvider` with patterns for additional RMSNorm variants encountered in production models:

- **Llama** (existing): `weight * hidden_states.to(input_dtype)` - cast before weight multiply
- **GPT-OSS** (new): `(weight * hidden_states).to(input_dtype)` - cast after weight multiply  
- **Gemma / Gemma2 / Gemma3** (new): weight upcast to fp32 and shifted by `+1` before multiply
- **Gemma4** (new): `torch.pow(x, -0.5)` instead of `torch.rsqrt`, no +1 shift
- **Gemma4 no_scale** (new): same as Gemma4 but without learned weight

For **vLLM**, flips `TTRMSNorm`'s operand order from `x * weight` → `weight * x` so the existing Llama pattern matches the vLLM path directly, no dedicated pattern needed.

Also fixes `TTRMSNorm.forward` to use positional args in `.mean(-1, keepdim=True)` matching the pattern's traced form.

## Test plan

- [x] `test_llama_rms_norm_fusion` : verifies Llama pattern produces `rms_norm` op in TTIR
- [x] `test_gpt_oss_20b_rms_norm_fusion` : verifies GPT-OSS pattern
- [x] `test_gemma_rms_norm_fusion` : verifies Gemma family pattern
- [x] `test_gemma4_rms_norm_fusion` (parametrized `with_scale=[True, False]`) : verifies Gemma4 variants
- [x] Verified via dynamo tracing that `TTRMSNorm` (vLLM) fuses correctly with existing Llama pattern
- [x] Nightly + model-test-passing workflows show no regressions from this PR


alinakhanTT:
Perf validated on n150: **+28.15% TPS** on llama-3.1-8B vs main nightly (14.72 -> 18.86 sps). Run: https://github.com/tenstorrent/tt-xla/actions/runs/27356477853

generated/watcher/kernel_names.txt
```diff
@@ -0,0 +1,3 @@
+0: blank
```
AleksKnezevic: all these generated files should not be part of the PR I think.

pjrt_implementation/src/api/buffer_instance.cc
```diff
@@ -222,7 +222,7 @@ void BufferInstance::copyFromHost(
       tt::runtime::utils::dataTypeElementSize(runtime_data_type);
   std::vector<std::uint32_t> shape =
       calculateShape(dims, num_dims, m_data_type);
-  std::vector<std::int64_t> strides =
+  std::vector<std::int64_t> strides_int64 =
```
AleksKnezevic: we typically don't append dtype to variable names, you should remove this change.

pjrt_implementation/src/api/flatbuffer_loaded_executable_instance.cc
```diff
@@ -119,8 +119,10 @@ void FlatbufferLoadedExecutableInstance::materializeShellTensors(
     // copies the bytes from the client's host_base pointer. Subsequent
     // buffers in the group get unsafe-borrowed tensors aliasing that owned
     // tensor, so we only hold one copy of the data on the worker side.
+    std::vector<std::int64_t> strides_int64(shell->strides.begin(),
```
AleksKnezevic: same comment as above, you should remove the variable renames

AleksKnezevic (approved):
Couple nits re unneeded files and unnecessary variable name changes, but otherwise LGM

generated/inspector/mesh_devices_log.yaml
```diff
@@ -0,0 +1,11 @@
...
+    mesh_id: 0
+    timestamp_ns: 1781031257354514172
+    devices: [1, 4]
+    shape: [1, 2]
+- mesh_device_initialized:
+    mesh_id: 0
+    timestamp_ns: 1781031257355179976
+- mesh_device_destroyed:
+    mesh_id: 0
+    timestamp_ns: 1781031257366578722
```
mmanzoorTT: `generated` directory should not be included. `generated` is added in `.gitignore` https://github.com/tenstorrent/tt-xla/pull/5192. Please remove them.

pjrt_implementation/src/api/buffer_instance.cc
mmanzoorTT: These variable name changes are irrelevant and should be reverted back.

generated/watcher/kernel_names.txt
alinakhanTT: Removed

pjrt_implementation/src/api/buffer_instance.cc
alinakhanTT: Reverted back

pjrt_implementation/src/api/flatbuffer_loaded_executable_instance.cc
alinakhanTT: Done!

generated/inspector/mesh_devices_log.yaml
alinakhanTT: Done

pjrt_implementation/src/api/flatbuffer_loaded_executable_instance.cc
```diff
@@ -119,8 +119,10 @@ void FlatbufferLoadedExecutableInstance::materializeShellTensors(
     // copies the bytes from the client's host_base pointer. Subsequent
     // buffers in the group get unsafe-borrowed tensors aliasing that owned
     // tensor, so we only hold one copy of the data on the worker side.
+    std::vector<std::int64_t> strides(shell->strides.begin(),
```
mmanzoorTT: `shell->strides` is already defined as `std::vector<std::int64_t>` in https://github.com/tenstorrent/tt-xla/blob/786b7e1e309f044e07af89e113dd0c01ad10e97d/pjrt_implementation/inc/api/tensor.h#L57 these changes here seems redundant.

pjrt_implementation/src/api/buffer_instance.cc
```diff
@@ -282,17 +248,18 @@ void BufferInstance::copyFromHost(
...
     TT_ASSERT(host_buffer_semantics !=
                   PJRT_HostBufferSemantics_kImmutableOnlyDuringCall,
               "In distributed mode we unsafely borrow data from host_buffer; "
               "this is okay iff caller guarantees that host_buffer lifetime is "
               "safe until execution.");
+    std::vector<std::uint32_t> shell_strides(strides.begin(), strides.end());
+
     PjrtTensor::HostTensorShell host_tensor_shell = PjrtTensor::HostTensorShell{
-        host_buffer, shape, strides, element_size, runtime_data_type};
+        host_buffer, shape, shell_strides, element_size, runtime_data_type};
```
mmanzoorTT: Why do we need `uint32_t` for `strides` here? `PjrtTensor::HostTensorShell` struct is expecting `int64_t`. 
https://github.com/tenstorrent/tt-xla/blob/786b7e1e309f044e07af89e113dd0c01ad10e97d/pjrt_implementation/inc/api/tensor.h#L54

jameszianxuTT: Please rebase! This was from a previous merge timing issue that @kmabeeTT found and was patched in #5156

alinakhanTT: Done!

pjrt_implementation/src/api/flatbuffer_loaded_executable_instance.cc
alinakhanTT: After rebasing on main these C++ changes are no longer part of the PR

alinakhanTT:
Fusion verified on CI: benchmark run with --check-fusions passed on n150, confirms ttnn.rms_norm is present in all TTNN IR graphs for Llama-3.1-8B via vLLM AOT path.
Run: https://github.com/tenstorrent/tt-xla/actions/runs/27426991917

tests/torch/graphs/fusion_patterns/test_rmsnorm.py
```diff
@@ -25,3 +25,91 @@ def rmsnorm(x: torch.Tensor, gamma: torch.Tensor) -> torch.Tensor:
         framework=Framework.TORCH,
         request=request,
     )
+
```
mstojkovicTT: Can you add these tests instead to `tests/torch/ops/test_fusion_ops.py`, and also import RMSNorm module like it is done for `llama` and `gptoss` from hugging face? Also, my suggestion is to drop aot tests.

mstojkovicTT (changes requested):
Thanks for adding new patterns for RMSNorm @alinakhanTT! 

One question i have is, where do we use `AOTAutograd` compile right now? 
Also, if `aotautograd` is concern for something upcoming, we have #3795 that will enable it by default, and will make fusion pattern run only on pre-aten graph (so the duplicates introduced here would be redundant).

python_package/tt_torch/fusion_providers.py
```diff
@@ -166,15 +188,135 @@ def pattern_cast_after_mul(
...
+        variance_eps = torch.ops.aten.add.Tensor(variance, eps)
+        rsqrt_var = torch.ops.aten.rsqrt.default(variance_eps)
+        hidden_normalized = torch.ops.aten.mul.Tensor(h_fp32, rsqrt_var)
+        weighted = torch.ops.aten.mul.Tensor(weight, hidden_normalized)
+        return torch.ops.aten._to_copy.default(weighted, dtype=dtype)
+
     @staticmethod
     def replacement(hidden_states: Tensor, weight: Tensor, eps: float, dtype) -> Tensor:
-        """Shared replacement for both RMS norm pattern variants."""
+        """Shared replacement for the Llama/GPT-OSS pattern variants."""
```
mstojkovicTT: nit: you can just drop `the Llama/GPT-OSS` and keep it as `Shared replacement for RMS norm pattern variants`.

python_package/tt_torch/fusion_providers.py
```diff
@@ -166,15 +188,135 @@ def pattern_cast_after_mul(
...
+        shifted_w = torch.ops.aten.add.Tensor(weight_fp32, 1.0)
+        return torch.ops.aten.mul.Tensor(normed, shifted_w)
+
+    @staticmethod
+    def replacement_gemma(hidden_states: Tensor, weight: Tensor, eps: float) -> Tensor:
+        """Replacement: rms_norm with effective weight = 1 + weight.float()."""
+        return torch.nn.functional.rms_norm(
+            hidden_states,
+            normalized_shape=weight.shape,
+            weight=1.0 + weight.float(),
```
mstojkovicTT, line 218: how is this resolved in graph?

python_package/tt_torch/fusion_providers.py
```diff
@@ -166,15 +188,135 @@ def pattern_cast_after_mul(
...
+            eps=eps,
+        )
+
+    # ----- Gemma4 (no +1, pow(-0.5) instead of rsqrt) -----
+
+    @staticmethod
+    def pattern_gemma4(hidden_states: Tensor, weight: Tensor, eps: float) -> Tensor:
```
mstojkovicTT, line 225: From [gemma4 impl](https://github.com/huggingface/transformers/blob/d242bb790bcbbe6c9a20e46cf9d70648739a90bf/src/transformers/models/gemma4/modeling_gemma4.py#L193), it seems that we can have variant that has `torch.pow(mean_squared, -0.5)`, but without the weight at all (`with_scale = false`), which will stay unmatched ([example](https://github.com/huggingface/transformers/blob/d242bb790bcbbe6c9a20e46cf9d70648739a90bf/src/transformers/models/gemma4/modeling_gemma4.py#L931)). I recommend here to make a test with this module, and go over potential graphs and be sure we covered all patterns :))

python_package/tt_torch/fusion_providers.py
```diff
@@ -166,15 +188,135 @@ def pattern_cast_after_mul(
...
+        weight_fp32 = torch.ops.aten._to_copy.default(weight, dtype=torch.float32)
+        return torch.ops.aten.mul.Tensor(normed, weight_fp32)
+
+    @staticmethod
+    def replacement_gemma4(hidden_states: Tensor, weight: Tensor, eps: float) -> Tensor:
+        """Replacement: rms_norm with weight upcast to fp32."""
+        return torch.nn.functional.rms_norm(
+            hidden_states,
+            normalized_shape=weight.shape,
+            weight=weight.float(),
```
mstojkovicTT, line 245: how is this resolved in graph?

python_package/tt_torch/fusion_providers.py
alinakhanTT: Captured the post-fusion FX graph for `GemmaRMSNorm`. The `1.0 + weight.float()` expression in the replacement materializes as two persistent nodes:

```
float_3 = l__self___weight.float()        # call_method
add_2   = 1.0 + float_3                   # call_function (operator.add)
```

`add_2` then feeds the `weight` kwarg of `rms_norm`. The transformation is not eliminated, it is moved from inline in the RMSNorm forward (where it produced ~10 unfused ops) to a single arg prep before the fused op. Since `weight` is an `nn.Parameter` (constant during inference), the cast and scalar add are constant foldable downstream. The new `test_gemma_rms_norm_fusion` in `tests/torch/ops/test_fusion_ops.py` confirms the final MLIR is a single `ttir.rms_norm` via the `filecheck` assertion (passing for all 6 parametrized cases).

python_package/tt_torch/fusion_providers.py
alinakhanTT: Same shape as the Gemma case, one fewer node. The post-fusion graph for `Gemma4RMSNorm(with_scale=True)` contains a single extra node:

```
float_3 = l__self___weight.float()        # call_method
```

feeding the `weight` kwarg of `rms_norm`. The upcast is preserved (semantically required) but folds away on a constant parameter. `test_gemma4_rms_norm_fusion` confirms a single `ttir.rms_norm` in the lowered MLIR.

python_package/tt_torch/fusion_providers.py
alinakhanTT: Done

tests/torch/graphs/fusion_patterns/test_rmsnorm.py
alinakhanTT: Done in 581f60568. Regarding dropping AOT: verified empirically that vLLM does not exercise AOTAutograd (zero references to aot_autograd in integrations/vllm_plugin), and #3795 will move fusion to run pre aten, making the aten form patterns unreachable. So I dropped the AOT tests AND the four AOT patterns (pattern_aot, pattern_aot_cast_after_mul, pattern_gemma_aot, pattern_gemma4_aot). The real cause of the vLLM-Llama fusion miss turned out to be operand order on the trailing multiply (x * weight versus HF Llama's weight * x), now covered by the new pattern_x_weight_order!!

python_package/tt_torch/fusion_providers.py
alinakhanTT: Thank you for that catch! Added in 581f60568:
  1. New pattern_gemma4_no_scale plus replacement_gemma4_no_scale in fusion_providers.py. The pattern matches the prefix up to x_fp32.mul(inv_std) (no trailing weighted multiply); the replacement passes weight=None and derives normalized_shape from hidden_states.shape[-1:].
   2. New test_gemma4_no_scale_rms_norm_fusion in test_fusion_ops.py that instantiates Gemma4RMSNorm(hidden_size, with_scale=False) directly from HF transformers. All 6 parametrized cases pass with the rms_norm.ttir.mlir filecheck. I also captured FX graphs for Gemma4RMSNorm(with_scale=True) and (with_scale=False) to confirm both variants are matched, and the pattern_x_weight_order for the vLLM x * weight order which was the other gap

tests/torch/ops/test_fusion_ops.py
```diff
@@ -66,3 +92,125 @@ def test_gpt_oss_20b_rms_norm_fusion(batch_size, seq_len, hidden_size, dtype, re
...
+def test_gemma4_rms_norm_fusion(batch_size, seq_len, hidden_size, dtype, request):
+    """Gemma4 default (with_scale=True): weight upcast (no +1 shift),
+    torch.pow(mean_squared, -0.5) in place of torch.rsqrt."""
+
+    options = {
+        "tt_enable_torch_fx_fusion_pass": True,
+        "tt_enable_composite_ops": True,
+    }
+
+    model = Gemma4RMSNorm(hidden_size, with_scale=True)
```
mstojkovicTT: You can parametrize over `with_scale`, so we dont duplicate code

tests/torch/graphs/fusion_patterns/test_rmsnorm.py
```diff
@@ -1,27 +0,0 @@
-# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```
mstojkovicTT, line 1: you can leave this file as is

tests/torch/ops/test_fusion_ops.py
```diff
@@ -4,14 +4,40 @@
...
 
 from tests.infra.evaluators import ComparisonConfig
 from tests.infra.testers.single_chip.graph.graph_tester import run_graph_test
 
 
+class _VLLMRMSNormShape(nn.Module):
+    """Mirror of integrations/vllm_plugin/vllm_tt/layers/rmsnorm.py TTRMSNorm
+    (no residual / no variance_size_override paths).
+
+    Defined inline so this file does not pull in the vllm package at collection
```
mstojkovicTT: Do we have some problems with importing `vllm` here? I would rather make a separate file `test_vllm_fusion_ops` that will import this class, then implementing it by the hand because we can easily miss the change in their code.

python_package/tt_torch/fusion_providers.py
```diff
@@ -166,15 +179,127 @@ def pattern_cast_after_mul(
         result = weight.mul(hidden_normalized)
         return result.to(dtype)
 
+    @staticmethod
+    def pattern_x_weight_order(
+        hidden_states: Tensor, weight: Tensor, eps: float, dtype
+    ) -> Tensor:
+        """
+        vLLM TTRMSNorm variant: cast happens before multiply with weight, and the
```
mstojkovicTT: Name `TTRMSNorm` indicates to me that this is our implementation of RMSNorm. Maybe its easier to just switch that back in vllm? (tagging @AleksKnezevic for more context on this)

tests/torch/ops/test_fusion_ops.py
alinakhanTT: Done

tests/torch/graphs/fusion_patterns/test_rmsnorm.py
alinakhanTT: Done

tests/torch/ops/test_fusion_ops.py
alinakhanTT: Done in fe1d34171. New file tests/torch/ops/test_vllm_fusion_ops.py imports the real TTRMSNorm from vllm_tt.layers.rmsnorm wrapped around a real vllm.model_executor.layers.layernorm.RMSNorm instance, so we'll catch any drift in vLLM's forward. Guarded with pytest.importorskip("vllm") / importorskip("vllm_tt.layers.rmsnorm") at module level so the file skips cleanly on dev envs without vLLM installed (vLLM isn't a hard dep of the dev venv). Removed the inline _VLLMRMSNormShape mirror from test_fusion_ops.py!

python_package/tt_torch/fusion_providers.py
mmanzoorTT: yes, TTRMSNorm is our [implementation](https://github.com/tenstorrent/tt-xla/blob/5d20786bd2ae86d4326d4092cd92f7c9580d5733/integrations/vllm_plugin/vllm_tt/layers/rmsnorm.py#L10) and we can use `composite_rms_norm` op directly. However, fusion pattern provides global coverage and it will be useful if any other model is using similar RMSNorm pattern.

mstojkovicTT: I didnt mean to use composite op directly, just changing the order of multiply will be enough for us to not need another pattern.

mmanzoorTT: got it. @alinakhanTT Please try it out.

alinakhanTT:
## Nightly / model-test-passing results

Ran both workflows on this branch:
- **vllm-nightly** ([run 27631187514](https://github.com/tenstorrent/tt-xla/actions/runs/27631187514)): 4 failure ,all pre-existing on main (same `Non-hoistable op found in middle of hoistable ops` tt-mlir compiler error).
- **model-test-passing** ([run 27638961461](https://github.com/tenstorrent/tt-xla/actions/runs/27638961461)): 20 failures, cross-checked against main's latest nightly, all identical. No regressions from this PR.

CI (`push` suite) passes cleanly.

python_package/tt_torch/fusion_providers.py
alinakhanTT: Thanks for the feedback! I've addressed this:
vLLM path: Flipped `TTRMSNorm`'s operand order (`weight * x`) so it matches the existing Llama pattern directly. No dedicated vLLM pattern needed

mstojkovicTT (approved):

integrations/vllm_plugin/vllm_tt/layers/rmsnorm.py
```diff
@@ -65,12 +65,12 @@ def forward(
 
             x_var = x[:, :, : self.variance_size_override]
 
-        variance = x_var.pow(2).mean(dim=-1, keepdim=True)
+        variance = x_var.pow(2).mean(-1, keepdim=True)
```
mmanzoorTT, line 68: This change doesn't seem relevant. Please verify and revert it.

mmanzoorTT (approved):

PR #5142, opened 2026-06-10
Remove dead optimizer-submesh device code

rpavlovicTT:
## What

Removes the now-unused optimizer-submesh device machinery from `ClientInstance`:
- `getOrCreateOptimizerSubmesh()`
- `closeOptimizerSubmesh()`
- `m_optimizer_submesh` member
- the orphaned `closeOptimizerSubmesh()` call left in `module_builder.cc`

## Why

The functional change — not passing a real submesh `MeshDevice` ptr to the optimizer via `options.devicePtr` — already landed on `uplift` in `21726fffc` ("Remove optimizer-related device pointer option from module builder"). After the tt-mlir MetalEnv migration ([#8737](https://github.com/tenstorrent/tt-mlir/pull/8737)), the optimizer/OpModel opens a **mock device** by default, so these helpers have no remaining callers and `closeOptimizerSubmesh()` is a permanent no-op.

This also removes the source of the `releaseSubMeshDevice` lifecycle crash seen on multi-device (tensor-parallel) runs:
```
TT_THROW: MeshDevice cq ID 0 is in use by parent mesh ID 2 during close of mesh ID 4
 --- tt::runtime::releaseSubMeshDevice(...)
```

## Fallback note

The real-device path is **not** lost: tt-mlir still honors `options.devicePtr` (the `if (externalDevice)` branch in `DevicePassesWrapper`). If the mock device ever regresses, re-passing a device ptr is a small, well-scoped change recoverable from git history.

## Scope

Pure dead-code removal — no behavior change (the `devicePtr` was already not being set). 59 deletions across 3 files.

### DIFF
--- a/pjrt_implementation/inc/api/client_instance.h
+++ b/pjrt_implementation/inc/api/client_instance.h
@@ -143,24 +143,11 @@ public:
     return m_parent_mesh;
   };
 
-  // Closes currently opened mesh device and submesh device, if any.
+  // Closes currently opened mesh device, if any.
   void closeMeshDevice();
-
-  // Returns the optimizer submesh device of the provided shape. If there is
-  // already opened optimizer submesh and its shape matches the provided shape,
-  // it is returned. Otherwise, we close any previously opened optimizer submesh
-  // and create a new one with the provided shape.
-  //
-  // NOTE: this method is not thread-safe and we will need to revisit this when
-  // adding support for parallel execution.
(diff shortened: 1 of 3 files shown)


rpavlovicTT:
Superseded by the main-targeted PR (full removal incl. the active devicePtr passing). The functional disable already exists on uplift via 21726fffc; landing the complete removal on main and letting it flow up.

PR #5146, opened 2026-06-10
Remove optimizer submesh device; let optimizer use mock device

rpavlovicTT:
## What

Stops creating an "optimizer submesh" and passing its `MeshDevice` ptr to tt-mlir via `options.devicePtr`. Removes the supporting machinery from `ClientInstance`:
- `getOrCreateOptimizerSubmesh()`
- `closeOptimizerSubmesh()`
- `m_optimizer_submesh` member
- the `devicePtr` block + `closeOptimizerSubmesh()` call in `module_builder.cc`

## Why

With no external device passed, the optimizer falls back to the **mock device** — `DevicePassesWrapper`'s `else { ... openMockDevice(); }` branch, which tt-mlir already supports. The optimizer doesn't need a real on-silicon submesh to run op-model constraint queries.

This also removes the source of the `releaseSubMeshDevice` lifecycle crash that surfaces on multi-device (tensor-parallel) meshes once tt-metal is uplifted:
```
TT_THROW: MeshDevice cq ID 0 is in use by parent mesh ID 2 during close of mesh ID 4
 --- tt::runtime::releaseSubMeshDevice(...)
```
The optimizer submesh aliased the parent mesh's physical command queues; releasing it while the parent's CQ was in use tripped tt-metal's "one owner per CQ at teardown" guard.

## Notes

- The same functional change already landed on `uplift` (`21726fffc`); this brings `main` in line and additionally removes the now-dead helpers.
- The real-device path is preserved in tt-mlir (`if (externalDevice) ...` in `DevicePassesWrapper`), so it can be re-enabled from git history if the mock path ever regresses.

### DIFF
--- a/pjrt_implementation/inc/api/client_instance.h
+++ b/pjrt_implementation/inc/api/client_instance.h
@@ -143,24 +143,11 @@ public:
     return m_parent_mesh;
   };
 
-  // Closes currently opened mesh device and submesh device, if any.
+  // Closes currently opened mesh device, if any.
   void closeMeshDevice();
-
-  // Returns the optimizer submesh device of the provided shape. If there is
-  // already opened optimizer submesh and its shape matches the provided shape,
-  // it is returned. Otherwise, we close any previously opened optimizer submesh
-  // and create a new one with the provided shape.
-  //
-  // NOTE: this method is not thread-safe and we will need to revisit this when
-  // adding support for parallel execution.
-  tt::runtime::Device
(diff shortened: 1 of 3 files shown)


rpavlovicTT:
🚧 **Do not merge until the uplift branch is merged into `main`.** Marked as draft to enforce this. Once the tt-metal uplift lands on main, this can go in.

sgligorijevicTT (approved):

nvukobratTT (approved):

Issue #5083, opened 2026-06-04
Multichip benchmark LLMs fail with batch_size=32 on vLLM benchmark


vkovacevicTT:
Multichip benchmark LLMs through vLLM fail with batch_size=32 with error:
```
loc("custom-call.137"): error: 'ttir.paged_update_cache' op Input tensor must have shape equal to the number of users (determined by update index shape): 32, got 16
2026-06-02 22:31:24.737 ( 523.732s) [         4E8C300]      module_builder.cc:913    ERR| Failed to convert from SHLO to TTIR module
```

Current multichip models in vLLM benchmark run with batch_size=1.

Tested locally on WH-QB-03:
`falcon3_7b_tp`, `falcon3_10b_tp`, `qwen3_14b_tp` failed.
`llama_3_1_8b_instruct_tp` pased.

To reproduce the error: build tt-xla, install python requirements and run 
```
TT_BENCHMARK_BATCH_SIZE=32 pytest -svv tests/benchmark/test_vllm_benchmarks.py::test_vllm_tp_benchmark[falcon3-7b-tp] 
```

When setting `use_2d_mesh=False` on `falcon3_7b_tp`, we get:
```
loc("transpose.141"): error: Could not apply propagated tensor shardings to tensor dimensions
    error: Could not update shapes based on their tensor sharding attributes
```

kmabeeTT:
FYI @ssaliceTT overlap with https://github.com/tenstorrent/tt-xla/issues/5032 though this ticket has the actual errors which I hadn't seen yet :)

ssaliceTT:
> FYI [@ssaliceTT](https://github.com/ssaliceTT) overlap with [#5032](https://github.com/tenstorrent/tt-xla/issues/5032) though this ticket has the actual errors which I hadn't seen yet :)

I was able to get `qwen3-32b` to run with b32 on a QB2 with a `1x4` mesh. Updated #5032 on the error I saw and some fixes I have.
Not sure the `paged_update_cache` issue here is related but the other one could be since shardings failing to propagate properly would be caused by the xla argument threshold detailed in the issue.

ssaliceTT:
_Update_: This PR of mine https://github.com/tenstorrent/tt-xla/pull/5159 had fixed the tests for the `paged_update_cache` issue by just replicating `position_ids` but it should get fixed when DP+TP support is added with https://github.com/tenstorrent/tt-xla/pull/4947.
[...]

PR #5159, opened 2026-06-10
[vLLM] Fix batch-32 TP benchmark failure; set all TP tests to bs32

ssaliceTT:
### Ticket
#5083

### Problem description
TP vLLM benchmarks failed at batch 32 with:
```
'ttir.paged_update_cache' op Input tensor must have shape equal to the number of users: 32, got 16
```
`position_ids` were being batch-sharded on the mesh's `"batch"` axis (size 2 on a 2×4 mesh). This shard propagates through RoPE into K, but the page table is replicated across all 32 users, so `paged_update_cache` receives K with 16 users/device while the page table expects 32.

### What's changed
- **`model_runner.py`**: stop sharding `position_ids` in `_pin_input_shardings`. All other input shardings are unchanged.
- **`test_vllm_benchmarks.py`**: set all TP benchmark configs to `batch_size=32`, raise `_tp_config` default `gpu_memory_utilization` from `0.005` to `0.1`, bump Gemma-4 from `0.1` to `0.2`, default `use_2d_mesh=True`.

### Checklist
- [x] n300-llmbox vLLM perf: https://github.com/tenstorrent/tt-xla/actions/runs/27654181919
   llama-3.1-70b needed higher gpu utilization so failed in above run, re ran here: https://github.com/tenstorrent/tt-xla/actions/runs/27718835369
- [x] qb2-blackhole vLLM perf: https://github.com/tenstorrent/tt-xla/actions/runs/27654205001

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -1417,7 +1417,6 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         batch_axis = "batch" if self.use_2d_mesh else "model"
         if input_ids is not None:
             safe_mark_sharding(input_ids, self.mesh, (batch_axis, None))
-        safe_mark_sharding(position_ids, self.mesh, (batch_axis, None))
         if inputs_embeds is not None:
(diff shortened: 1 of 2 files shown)


vkovacevicTT (approved):
Thanks for this fix! :)

AleksKnezevic (approved):

PR #5160, opened 2026-06-10
Unify TTNN codegen with normal runs: emit per-graph Python, edit, and reload via env vars + vLLM support

AleksKnezevic:
### Problem description
No way to emit TTNN and load it back up (to run), which is a useful approach for vLLM, as it emits ~tens of graphs.

### What's changed
A couple things were changed to enable this path, I will list them clearly here:
- Add another backend: `TTNNCodegenLoadPy` - enables running previously emitted code
- Decoupled `dry_run` and `target_module` options - previously `target_module` was only set when `dry_run` was set to `false` (only used for testing of emitted TTNN thru `PythonModelRunner`), but now it makes sense to decouple these options to give more flexibility to pick needed/wanted scenario
- Added emit-then-load tests, both in "regular" path and thru vLLM
- Added an interactive, multi-client demo

### Additional details
- Graphs are now emitted with an id (SHA256 of the StableHLO module text) - written to `module_key` file within each graph dir
  - Root codegen dir will contain a `manifest.json` listing module keys for all graphs
  - A small change was needed to `StableHLOCompositeBuilder` to make deterministic op names for composite ops
- Env variables `TTXLA_CODEGEN_EXPORT_DIR=<dir>` and `TTXLA_CODEGEN_LOAD_DIR=<dir>` are added as an additional mechanism to set the correct backend (useful for vLLM)
  - `TTXLA_CODEGEN_EXPORT_DIR=<dir>` emits each compiled graph into `<dir>/graph_N/`. Each dir gets a `module_key` (sha256 of the StableHLO module text — stable across runs and compile order — plus mesh shape and device count); `manifest.json` maps hashes to dirs for humans.
  - `TTXLA_CODEGEN_LOAD_DIR=<dir>` loads saved code instead of compiling: graphs are matched by hash, SHLO→TTIR→TTNN and codegen are skipped, mesh shape/device count are restored from `module_key`, and the (possibly edited) `main.py` executes. Missing graphs are a hard compile error listing available hashes. Explicit compile options always win over env vars.

### Tests
```sh
# vLLM
pytest -svv tests/integrations/vllm_plugin/codegen/test_codegen_emit_load.py

# Regular
pytest -svv tests/torch/codegen/test_codegen_emit_load.py
pytest -svv tests/torch/codegen/test_codegen_emit_load_multichip.py
```

EDIT: currently skipping tests in CI (confirmed to work locally) due to manylinux compatibility issues, more details in https://github.com/tenstorrent/tt-xla/issues/5481

### Demo
```sh
# Emit once (offline) to populate the codegen dir:
python examples/vllm/Qwen3-0.6B-codegen/qwen.py --emit

# Start the server:
./examples/vllm/Qwen3-0.6B-codegen/service.sh

# In another terminal, start a client (or multiple)
python examples/vllm/Qwen3-0.6B-codegen/client.py
```

### Checklist


pjrt_implementation/src/api/module_builder/module_builder.cc
```diff
@@ -1322,6 +1361,29 @@ ModuleBuilder::buildModuleForTTNNCodegen(
     return {status, nullptr};
   }
 
+  // Python codegen records the graph (hash + mesh) so load mode can later match
+  // it by hash.
+  if (compile_options.backend == BackendRuntime::TTNNCodegenPy &&
+      !compile_options.graph_hash.empty()) {
+    std::filesystem::path graph_dir(*compile_options.export_path);
```
hshahTT, line 1369: You should assert to make sure that `compile_options.export_path` is not `nullopt`

pjrt_implementation/src/api/module_builder/module_builder.cc
hshahTT: Nit:
```suggestion
} else if (is_codegen_cpp || is_codegen_py) {
```

pjrt_implementation/inc/api/compile_options.h
```diff
@@ -161,6 +174,10 @@ struct CompileOptions {
   // The graph number (g0, g1, etc.) is automatically appended.
   std::string export_model_name = "";
 
+  // Stable identity of the compiled graph: hash of the StableHLO module text,
+  // computed by the module builder after the frontend SHLO pipeline.
+  std::string graph_hash = "";
```
jameszianxuTT: Is there no other place this graph_hash could be stored? It looks like this string is being used as a scratch area in a mini state machine being written at compile time and read at runtime. It would be better to remove this from the compile_options interface and avoid mixing control and data in a string as it will be hard to reason about if it starts going wrong

I suggest a separate datastructure to facilitate this graph loading feature's control flow holding target_module, graph_hash and other control-related variables with a clean public interface, rather than embedding it inside compile options, which to me is for user-set configuration options.

pjrt_implementation/src/api/module_builder/module_builder.cc
```diff
@@ -317,6 +319,29 @@ ModuleBuilder::buildModule(
...
+      compile_options.backend == BackendRuntime::TTNNCodegenLoadPy;
+  if (is_codegen_py || is_codegen_cpp) {
+    const std::string codegen_export_root = *compile_options.export_path;
+    static std::mutex counters_mutex;
+    static std::unordered_map<std::string, int> counters;
+    int graph_index;
+    {
+      std::lock_guard<std::mutex> lock(counters_mutex);
+      graph_index = counters[codegen_export_root]++;
+    }
```
jameszianxuTT: Is there a reason to guard this with a mutex - ie was it observed that PJRT makes concurrent calls to buildModule?

examples/vllm/Qwen3-0.6B-codegen/client.py
```diff
@@ -0,0 +1,69 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
mmanzoorTT:
```suggestion
# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```

pjrt_implementation/src/api/module_builder/module_builder.cc
svuckovicTT: No, not really needed - removed.

pjrt_implementation/inc/api/compile_options.h
svuckovicTT: Yeah, `CompileOptions` isn't a good place for `graph_hash` to live, I'll move it elsewhere, thanks for flagging! `target_module` is a user-facing option though, so I'll leave it there.

svuckovicTT:
Pinging for reviews:

@jameszianxuTT @mmanzoorTT 

`/pjrt_implementation/...` @mrakitaTT @hshahTT @sdjukicTT @sgligorijevicTT @sshonTT @acicovicTT
`/python_package/tt_torch/` @nvukobratTT @dgolubovicTT @jameszianxuTT @AleksKnezevic @mstojkovicTT @ppadjinTT @vkovinicTT

tests/integrations/vllm_plugin/codegen/vllm_serve_helper.py
```diff
@@ -0,0 +1,42 @@
...
+def main(out_file: str) -> None:
+    llm = vllm.LLM(
+        model="meta-llama/Llama-3.2-3B",
+        max_num_batched_tokens=16,
+        max_num_seqs=1,
+        max_model_len=16,
+        gpu_memory_utilization=0.002,
+        additional_config={
+            "enable_const_eval": False,
+            "min_context_len": 16,
```
mmanzoorTT: FYI: vLLM uses `min_context_len=32`; this `16` will be padded.

examples/vllm/Qwen3-0.6B-codegen/service.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
mmanzoorTT:
```suggestion
# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```

mmanzoorTT (approved):

python_package/tt_torch/composite_ops.py
```diff
@@ -25,6 +27,40 @@
...
+# torch_xla's own _get_uuid so dynamo bakes the value in as a constant.
+_composite_id_counter = itertools.count()
+
+
+@torchdynamo.assume_constant_result
+def _deterministic_composite_id() -> str:
+    return str(next(_composite_id_counter))
+
+
+def _make_composite_builder(name, attr=None):
```
jameszianxuTT, line 58: nit: I don't think there's a great pythonic way to enforce that the below functions do not directly instantiate StableHLOCompositeBuilder, which could lead to some interesting and hard to detect bugs per the comment above. 

One idea is to move this _make_composite_builder to some common utils file and then remove the import of StableHLOCompositeBuilder in this file to prevent accidental usage. Though, it does not prevent anyone from reimporting it in the future. AI suggests a linter rule, which is interesting but we don't really have precedent for that afaik

Probably not a big deal, feel free to ignore on the grounds that this mistake is unlikely

pjrt_implementation/src/api/module_builder/module_builder.cc
```diff
@@ -339,6 +360,15 @@ ModuleBuilder::buildModule(
     return {status, nullptr};
   }
 
+  // Stable identity of the graph (hash of the SHLO module text), used by the
+  // codegen emit/load helpers to key the saved graph directory. Kept as a local
+  // -- it's derived state, not a user-set compile option, so it is passed
+  // explicitly rather than stored on CompileOptions.
```
jameszianxuTT: Very minor nit, feel free to ignore: 

>  // -- it's derived state, not a user-set compile option, so it is passed
  // explicitly rather than stored on CompileOptions.

I think this is in response to my previous comments but I don't think it requires justification to _not_ make this a compile option.

jameszianxuTT (approved):
Thanks for changes, left some minor nits but feel free to disregard!

python_package/tt_torch/composite_ops.py
svuckovicTT: Another way would be to imbue `pytorch-xla` with a change to use any type of stable ids instead of random `uuid`s. I'll follow up on that next week.

pjrt_implementation/src/api/module_builder/module_builder.cc
svuckovicTT: Ugh, missed this, will remove! LLMs sure love leaving comments on the decision making context... I'll put something in my CLAUDE.md to avoid this in the future, thanks for marking it!

pjrt_implementation/src/api/module_builder/module_builder.cc
```diff
@@ -1374,11 +1491,18 @@ ModuleBuilder::performCodegen(std::string_view ttnn_mlir,
                                  " "
                                  "tensor-load-directory='./tensors'";
 
+  // Controls wether the generated solution is designed
+  // for standalone execution(is_local=false)
```
acicovicTT, line 1494: nit: is_local no longer exists

acicovicTT (approved):

pjrt_implementation/src/api/module_builder/module_builder.cc
svuckovicTT: Thanks for the review!

It's a bit confusing in the context of the PR since the variable used to be named the same, but the `is_local` here refers to the arg name of the API being called, which remains unchanged.

python_package/tt_torch/composite_ops.py
jameszianxuTT: Yeah, that is a better solution. Looking forward to seeing it

hshahTT (approved):
Looks good overall. Could you just add a check or assert near L1368 of module_builder.cc to make sure `compile_options.export_path` is not `std::nullopt` before merging this PR in?

pjrt_implementation/src/api/module_builder/module_builder.cc
acicovicTT: Oh okay didn't notice that, then disregard this comment. Thanks for the explanation!

PR #5211, opened 2026-06-15
Clamp out-of-range negative aten.slice starts to -dim_size (#5199)

### LINKED ISSUE #5199: [DiffusionGemma] sliding-window KV-cache slice fails when seq_len < sliding_window (aten.slice start -1023 on dim 19)
### Problem description 

- `diffusiongemma/pytorch-26B-A4B-it` fails with below error on 8 chips

```
RuntimeError: Value out of range (expected to be in range of [-19, 18], but got -1023)
```

### Repro

Once this [PR](https://github.com/tenstorrent/tt-forge-models/pull/733) lands on main , run below command 

```
TT_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 TT_XLA_SPMD=1 CONVERT_SHLO_TO_SHARDY=1 \
  TT_XLA_ARCH=n300-llmbox HF_HOME=/proj_sw/user_dev/kkannan/new_hub \
  pytest -svv "tests/runner/test_models.py::test_all_models_torch[diffusiongemma/pytorch-26B-A4B-it-tensor_parallel-inference]"
```

### Logs

- [jun15_diffusongemma4_8_chips.log](https://github.com/user-attachments/files/28950169/jun15_diffusongemma4_8_chips_3.log)

kamalrajkannan78:
### Ticket

- fixes https://github.com/tenstorrent/tt-xla/issues/5199

### Problem description

- Models with a sliding-window KV cache on a short sequence (e.g. DiffusionGemma) fail at compile with `RuntimeError: Value out of range`. The cache slice `x[:, :, -sliding_window + 1:, :]` lowers to `aten.slice.Tensor(..., start=-1023)` on a dim of size 19 — CPU's `__getitem__` silently clamps an out-of-range negative start, but XLA's strict `aten.slice.Tensor` rejects it. (#5199)

### What's changed

- Implemented as a torch FX pass, per @jameszianxuTT's suggestion in https://github.com/tenstorrent/tt-xla/pull/4552#issuecomment-4413007441.
- Add `clamp_neg_slice_bounds`: rewrites a constant negative slice `start`/`end` `< -dim_size` to `-dim_size` (CPU eager semantics), only on statically-known dims — valid slices are untouched.
- Wire it into `torch_pass_pipeline` (on the exported graph, before `program.module()`) so both the legacy and experimental compile paths get it 
- Generalized from start-only to also clamp negative `end` (per the review on #5211). A four-corner check (start/end × negative/positive) confirmed only the two **negative** bounds are rejected by XLA; positive bounds are already clamped natively, so they're left untouched.
- Add `tests/torch/ops/test_slice.py` covering all four out-of-range corners — non-empty results (neg start, pos end) compared via PCC, empty results (neg end, pos start) via shape assert.
- **DiffusionGemma:** this clears the slice error — the model now traces past it and fails further downstream with a **segfault** (separate blocker, to be handled next). This PR only unblocks the slice issue.
- **krea VAE not fixed:** its failing slice (`x[:, :, -CACHE_T:, :, :]`, `autoencoder_kl_wan.py:883`) takes a different path — it runs via `__torch_function__` (eager, from a Dynamo graph break in the frame-by-frame decode loop) instead of becoming an `aten.slice.Tensor` node in the exported FX graph this pass rewrites, so the FX pass can't reach it. Handled separately.

### Checklist
- [x] Verify the changes through local testing in WH

### Logs

- [jun16_slice_4_cases_before_fix.log](https://github.com/user-attachments/files/28987725/jun16_slice_4_cases_before_fix.log)
- [jun16_slice_4_cases_after_fix.log](https://github.com/user-attachments/files/28998699/jun16_z_slice_4_cases_after_fix.log)
- [jun15_diffusiongemma4_8_chips_before_fix.log](https://github.com/user-attachments/files/28955060/jun15_diffusongemma4_8_chips_3.log)
- [jun16_diffusiongemma4_8_chips_after_fix.log](https://github.com/user-attachments/files/28998694/jun16_z_diffusongemma4_8_chips_after_fix.log)

### DIFF
--- a/python_package/tt_torch/backend/backend.py
+++ b/python_package/tt_torch/backend/backend.py
@@ -29,6 +29,7 @@ from .passes import (
     bypass_assert_tensor_metadata,
     bypass_dtype_promotion_and_redundant_cast,
     bypass_redundant_getitem,
+    clamp_neg_slice_bounds,
     handle_composite_ops,
     insert_argument_type_markers,
     rewrite_adaptive_avgpool_to_mean,
@@ -71,6 +72,13 @@ def torch_pass_pipeline(
         strict=False,
     )
     program = program.run_decompositions(decompositions)
+
+    # Clamp out-of-range negative slice start/end bounds to -dim_size (CPU
+    # __getitem__ semantics) on the exported graph, where nodes still carry shape
+    # meta (program.module() drops it). XLA's strict aten.slice otherwise rejects
+    # them. In torch_pass_pipeline so both legacy and experimental paths get it.
+    # Refs: tt-xla #5199.
+    clamp_neg_slice_bounds(program.graph_module)
 
     compiled_graph = program.module()
     # When torch.compile traces a model, it flattens the module hierarchy and
--- a/python_package/tt_torch/backend/passes.py
+++ b/python_package/tt_torch/backend/passes.py
@@ -327,6 +327,58 @@ def run_shape_prop(gm, example_inputs):
     shape_prop.run(*fake_args)
 
 
+def clamp_neg_slice_bounds(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
+    """Clamp out-of-range negative ``aten.slice.Tensor`` start/end bounds to ``-dim_size``.
+
+    CPU's ``Tensor.__getitem__`` clamps a bound ``< -dim_size`` to ``-dim_size``;
+    XLA's strict ``aten.slice.Tensor`` instead raises "Value out of range". Only a
+    constant negative start/end on a statically known dim is rewritten, so valid
+    slices are untouched. Refs: tt-xla #5199.
+    """
+    changed = False
+    for node in gm.graph.nodes:
+        if node.op != "call_function" or node.target is not torch.ops.aten.slice.Tensor:
+            continue
+        # aten.slice.Tensor(input, dim, start, end=None, step=1)
+        if len(node.args) < 3:
+            continue
+        inp, dim = node.args[0], node.args[1]
+        if not isinstance(inp, torch.fx.Node) or not isinstance(dim, int):
+            continue
+        shape = getattr(inp.meta.get("val"), "shape", None)
+        if shape is None:
+            shape = getattr(inp.meta.get("tensor_meta"), "shape", None)
+        if shape is None:
+            continue
+        ndim = len(shape)
+        d = dim + ndim if dim < 0 else dim
+        if not (0 <= d < ndim):
+            continue
+        dim_size = shape[d]
+        if not isinstance(dim_size, int):  # skip symbolic / dynamic dims
+            continue
+        # Clamp the start (args[2]) and end (args[3]) bounds independently.
+        new_args = list(node.args)
+        node_changed = False
+        for i in (2, 3):
+            bound = new_args[i] if i < len(new_args) else None
+            if isinstance(bound, int) and bound < -dim_size:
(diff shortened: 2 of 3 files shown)


vkovinicTT (changes requested):
The core logic looks right. Only suggestion before landing: this covers one of four out-of-range corners, and negative `end` has the exact same mismatch (`end < -dim_size` normalizes below zero just like `start`), so we might hit it on the next model - worth folding in now. Positive `start` is the symmetric case too, though I'd confirm it actually fails first. If you generalize, normalizing each bound and clamping into [0, dim_size] - keeping the "only rewrite when out of range" guards.

Just first confirm how cpu implementation behaves first.

kamalrajkannan78:
> The core logic looks right. Only suggestion before landing: this covers one of four out-of-range corners, and negative `end` has the exact same mismatch (`end < -dim_size` normalizes below zero just like `start`), so we might hit it on the next model - worth folding in now. Positive `start` is the symmetric case too, though I'd confirm it actually fails first. If you generalize, normalizing each bound and clamping into [0, dim_size] - keeping the "only rewrite when out of range" guards.
> 
> Just first confirm how cpu implementation behaves first.

Thanks! Confirmed CPU behavior and ran a four-corner check (start/end × negative/positive) on XLA before generalizing:

- **negative start + negative end** → both rejected by XLA with the same `Value out of range`. Folded the negative end in — the pass now clamps both bounds (renamed `clamp_neg_slice_bounds`), still only when out of range.
- **positive start + positive end** → XLA already clamps these like CPU, so no rewrite needed. (positive-end passes as-is; positive-start yields an *empty* slice, which only trips the test harness's PCC on 0 elements — not a slice rejection.)

So I scoped the clamp to the two negative corners (the only ones that actually fail), kept the "only rewrite when out of range" guard, and added `tests/torch/ops/test_slice.py` covering all four corners — 4/4 green.

### Logs

- [jun16_slice_4_cases_before_fix.log](https://github.com/user-attachments/files/28987725/jun16_slice_4_cases_before_fix.log)
- [jun16_slice_4_cases_after_fix.log](https://github.com/user-attachments/files/28987723/jun16_slice_4_cases_after_fix.log)

vkovinicTT (approved):

python_package/tt_torch/backend/backend.py
```diff
@@ -191,6 +192,10 @@ def _call_experimental_compile(self, *args):
             # All this is a problem because DynamoBridge Partitioner can get confused by wrong next nodes and partition the graph in a way which fails to execute.
             legalize_graph(program.graph_module)
 
+            # Clamp out-of-range negative slice start/end bounds to -dim_size (CPU
+            # __getitem__ semantics); the bridge otherwise rejects them. Refs: tt-xla #5199.
+            clamp_neg_slice_bounds(program.graph_module)
```
sgligorijevicTT: I think `torch_pass_pipeline` is a better place for this than `_call_experimental_compile`, that way legacy gets the benefit of that path too.

sgligorijevicTT (approved):

python_package/tt_torch/backend/backend.py
kamalrajkannan78: Good call — moved it to `torch_pass_pipeline` so legacy gets it too. One detail worth noting: it has to run on `program.graph_module` (before `program.module()`), since `module()` drops the node shape meta the pass reads to get `dim_size`. Re-validated sanity tests and model

PR #5219, opened 2026-06-15
[EmitPy] Add regression test for TP/EP when export_tensors=True

### LINKED ISSUE #5180: Add a regression test for #5177, #4100
We've seen this issue twice now, should add simple a regression test in CI.

amilovanovicTT:
### Ticket
Closes #5180 
### Problem description
When `export_tensor = True`, every sharded input or weigth is serialized to `tensors/argN.tensorbin` as a single device shard instead as a full distributed tensor. Since this bug reappeared, we need a test to easily detect it.
### What's changed
Added a simple regression test.

### Checklist

### DIFF
--- a/tests/torch/test_codegen_export_tensors_sharding.py
+++ b/tests/torch/test_codegen_export_tensors_sharding.py
@@ -1,0 +1,125 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Regression test for codegen ``export_tensors=True`` with multi-device sharded tensors.
+
+See https://github.com/tenstorrent/tt-xla/issues/5177.
+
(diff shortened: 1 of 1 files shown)


tests/torch/test_codegen_export_tensors_sharding.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def __init__(self):
+        super().__init__()
+        self.w = torch.nn.Parameter(torch.randn(ROWS, COLS, dtype=torch.bfloat16))
+
+    def forward(self, x):
+        return x + self.w
+
+
+@parametrize_arch(["dual_chip", "llmbox", "galaxy"])
```
svuckovicTT: Just one of these is enough, probaby best to use `dual_chip` only.

svuckovicTT (approved):

acicovicTT (approved):

PR #5222, opened 2026-06-15
[Benchmark] WAN 2.2 test

vkovinicTT:
### Ticket
Closing #4869, #4870,  #4871, #4872

### What's changed
Added benchmark perf test for **WAN 2.2 14B** version (there is support for `5B` as well, just didn't want to stress perf runners).

**TODO:**
- [ ] add `bh-galaxy` as tester machine once [this](https://github.com/tenstorrent/tt-metal/issues/47309) change hits the `tt-xla`.


.github/workflows/perf-bench-matrix.json
```diff
@@ -792,6 +792,69 @@
...
         "skip-device-perf": true,
         "runs-on": "n300-llmbox"
+      },
+      {
+        "name": "wan_umt5_sharded",
+        "pyreq": "accelerate diffusers==0.37.1 transformers==5.2.0",
+        "pytest": "tests/benchmark/test_wan5b.py::test_wan_umt5[sharded]",
+        "runs-on": [
+          "qb2-blackhole",
+          "galaxy-bh"
```
nvukobratTT: Do we need to test the 5B variant on Glx? Should we target only p150 or p300 for this one?

nvukobratTT (approved):

.github/workflows/perf-bench-matrix.json
vkovinicTT: This is just text encoder. It's the same for 5B and 14B model. We don't test anything from 5B specifically. I can optionally add 5B benchmark tests for other components on `p150`/`p300` if you want.

.github/workflows/perf-bench-matrix.json
```diff
@@ -792,6 +792,69 @@
         "skip-ttnn-dump": true,
         "skip-device-perf": true,
         "runs-on": "n300-llmbox"
+      },
+      {
+        "name": "wan_umt5_sharded",
+        "pyreq": "accelerate diffusers==0.37.1 transformers==5.2.0",
```
mstojkovicTT: Is there a reason why we downgrade the transformers lib?

mstojkovicTT (approved):
Looks good

.github/workflows/perf-bench-matrix.json
vkovinicTT: Removed, thanks for noticing it :))

vvukomanTT:
Two things:
1. Adding tests to `perf-bench-matrix.json` doesn't automatically make these tests run in workflows unless they fall into a group of tests that is already getting executed.

e.g. In the nightly workflow we are executing these perf tests
https://github.com/tenstorrent/tt-xla/blob/611f504778ff9934ca30e94f0f8030f164c286dd/.github/workflows/schedule-nightly.yml#L92

You need to add `{ "runs-on": "galaxy-bh" }` to the filter in order to run the bh galaxy tests in the nightly workflow as well.

2. Can you execute these tests in the CI so we can validate that they are passing and to see their execution times.

nsumrakTT (approved):
Is there a sense to run same tests on both qb and galaxy?

vkovinicTT:
> Is there a sense to run same tests on both qb and galaxy?

Yep, we want to measure perf on different architectures.

tests/benchmark/test_wan14b.py
```diff
@@ -0,0 +1,58 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```
vkovacevicTT: Can we merge test_wan14b.py and test_wan5b.py into something like "test_videogen.py" or "test_wan.py"?

tests/benchmark/test_wan14b.py
```diff
@@ -0,0 +1,58 @@
...
+    vae_decoder_patches=(_patch_wan_resample_rep_sentinel,),
+    dit_timestep=scalar_timestep,
+    safe_xla_slicing=safe_xla_slicing,
+    dit_override_disabled=torch_function_override_disabled,
+)
+
+
+# UMT5 output is resolution-independent, so only the sharding axis is varied.
+@pytest.mark.parametrize("sharded", SHARDING_VARIANTS)
+def test_wan_umt5(sharded, output_file, request):
```
vkovacevicTT: I see we test individual components, do we want to create a pipeline in the future, like for imagegen models?

tests/benchmark/benchmarks/video_gen_benchmark.py
```diff
@@ -0,0 +1,301 @@
...
+def _device_arch() -> str:
+    """SPMD-safe device architecture ("wormhole" / "blackhole").
+
+    The shared ``get_xla_device_arch`` resolves ``xm.xla_device()``, which under
+    SPMD is the virtual ``SPMD:0`` device and can't be looked up. The runtime
+    device attributes expose the physical arch in both sharded and single-device
+    modes.
+    """
+    attributes = xr.global_runtime_device_attributes()
+    return align_arch(str(attributes[0]["device_arch"]).lower())
```
vkovacevicTT: We have this implemented in utils.py:

https://github.com/tenstorrent/tt-xla/blob/3725d1c071f8097c3012ab7937b952d92e723277/tests/benchmark/utils.py#L38-L50

vkovacevicTT (commented):
In benchmarks we import all models from tt_forge_models repo. Here we import them from `tests/torch/models`.
We do checkout whole tests/ dir in CI, so this wouldn't be a problem, but let's discuss if it makes sense to transfer model implementations from `tests/torch/models` to tt_forge_models repo.

tests/benchmark/test_wan14b.py
vkovinicTT: Yep, done that.

tests/benchmark/test_wan14b.py
vkovinicTT: We might add full pipeline to benchmark later, but individual components are just fine for now.

tests/benchmark/benchmarks/video_gen_benchmark.py
vkovinicTT: Good catch.

vkovinicTT:
> In benchmarks we import all models from tt_forge_models repo. Here we import them from `tests/torch/models`. We do checkout whole tests/ dir in CI, so this wouldn't be a problem, but let's discuss if it makes sense to transfer model implementations from `tests/torch/models` to tt_forge_models repo.

You are absolutely right, this should only be temporary solution as we want to have benchmarking as fast as possible.

vmilosevic (approved):

.github/workflows/schedule-nightly.yml
```diff
@@ -89,7 +89,7 @@ jobs:
       sh-runner: false
       skip-device-perf: true
       regression_check: true
-      adv_filter: '[ {"accuracy-testing": false}, {"runs-on": "n150-perf"}, {"runs-on": "p150-perf"}, {"runs-on": "n300-llmbox"}, {"runs-on": "galaxy-wh-6u"}, {"runs-on": "qb2-blackhole"} ]'
+      adv_filter: '[ {"accuracy-testing": false}, {"runs-on": "n150-perf"}, {"runs-on": "p150-perf"}, {"runs-on": "n300-llmbox"}, {"runs-on": "galaxy-wh-6u"}, {"runs-on": "qb2-blackhole"}, {"runs-on": "galaxy-bh"} ]'
```
mvasiljevicTT, line 92: do we still need this change? same goes for `.github/workflows/manual-benchmark.yml`

vkovinicTT: hmmm honestly not sure. I know that someone recently updated it, so I think it helps on GH actions to be able to specify which device you want to run it on (like in UI).

mvasiljevicTT: do you add `galaxy-bh` tests in this PR? It seems to me that [all added tests](https://github.com/tenstorrent/tt-xla/pull/5222/changes#diff-fe71e29e1636da633c1d68458261d67e4fbb6c84c1bc49ac1e185f8c5d6fcbba) are for qb2?

vkovinicTT: yep because we are waiting for [this](https://github.com/tenstorrent/tt-metal/issues/47309) to reach the `tt-xla` and then we will enable `bh-galaxy` tests as well.

mvasiljevicTT: ok sorry my bad then, doesn't really matter if it is added now or in the next pr

sdjukicTT (approved):

vkovacevicTT (approved):
Great! 
Since two tt-mlir uplifts landed after last benchmark check for wan, let's wait for this to pass as a sanity check before merging: https://github.com/tenstorrent/tt-xla/actions/runs/28248343805

Let's follow up with tt_forge_models migration and enabling on bh_galaxy :)

Issue #5223, opened 2026-06-15
Add Gemma 4-31B through vLLM on Blackhole Galaxy


ddilbazTT:
- [x] Add batch_size 1 test: https://github.com/tenstorrent/tt-xla/commit/13626449c50dd97be614fc60842081ad099f74d9 
- [ ] Add batch size 128 data-parallel + tensor-parallel test:

ddilbazTT:
Using the following branch: `mmanzoor/vllm-data-parallel`

Added the following test:
```
@pytest.mark.nightly
@pytest.mark.tensor_parallel
@pytest.mark.parametrize(
    ["enable_const_eval", "experimental_weight_dtype"],
    [
        pytest.param(True, ""),
    ],
)
@pytest.mark.parametrize(
    "mesh_shape",
    [
        pytest.param([8, 4], marks=pytest.mark.bh_galaxy),
    ],
)
def test_data_tensor_parallel_generation_gemma4_31b(
    mesh_shape: list[int],
    enable_const_eval: bool,
    experimental_weight_dtype: str,
):

    model_name = "google/gemma-4-31B-it"

    prompts = [
        "Describe Tenstorrent in one sentence.",
        "Explain what a neural network is in one sentence.",
        "What is the capital of France?",
        "Write one sentence about the ocean.",
        "Summarize the theory of relativity in one sentence.",
        "Give me a one-sentence description of photosynthesis.",
        "What is machine learning, in one sentence?",
        "Describe the sun in one sentence.",
        "Explain gravity in one sentence.",
        "Write a single sentence about mountains.",
        "What does a CPU do, in one sentence?",
        "Describe the internet in one sentence.",
        "Summarize the water cycle in one sentence.",
        "What is a black hole, in one sentence?",
        "Give a one-sentence description of a rainforest.",
        "Explain how a battery works in one sentence.",
        "Describe music in one sentence.",
        "What is democracy, in one sentence?",
        "Write one sentence about the moon.",
        "Explain what DNA is in one sentence.",
        "Describe a thunderstorm in one sentence.",
        "What is a programming language, in one sentence?",
        "Summarize evolution in one sentence.",
        "Describe a desert in one sentence.",
        "What is electricity, in one sentence?",
        "Write a single sentence about the human heart.",
        "Explain what a vaccine is in one sentence.",
        "Describe winter in one sentence.",
        "What is the speed of light, in one sentence?",
        "Give a one-sentence description of a volcano.",
[...]

PR #5224, opened 2026-06-15
[vLLM] Add Gemma 4-31B Blackhole Galaxy test through vLLM

ddilbazTT:
### Ticket
#5223 

### Problem description
We want Gemma 4-31B through vLLM.

### What's changed
- Parameterized mesh shape
    - Mesh shapes were pre-defined, and for Galaxy it was `[4,8]`. This resulted in sharding across dim indivisible by 8. Now, mesh shape is parameterized and using `[8,4]` for Gemma 4.
- Parameterized test_tensor_parallel_generation_gemma4_31b to run on bhqh and bh galaxy

### Checklist
- [x] vLLM Model Test Nightly: https://github.com/tenstorrent/tt-xla/actions/runs/27699193354
- [x] Gemma 4 blackhole galaxy (basic test nightly experimental): https://github.com/tenstorrent/tt-xla/actions/runs/27699850119

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -257,12 +257,14 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
 
         if self.enable_tensor_parallel:
             num_devices = xr.global_runtime_device_count()
-            mesh_shape = determine_mesh_shape(num_devices, use_2d_mesh=self.use_2d_mesh)
+            mesh_shape = determine_mesh_shape(
+                num_devices, self.tt_config.use_2d_mesh, self.tt_config.mesh_shape
+            )
             device_ids = np.array(range(num_devices))
             self.mesh = xs.Mesh(device_ids, mesh_shape, ("batch", "model"))
-            # Updating the config to reflect the actual mesh shape used.
-            if self.use_2d_mesh and 1 in mesh_shape:
-                self.use_2d_mesh = False
+            # A mesh with a size-1 axis is effectively 1D; downstream sharding
+            # treats only genuinely 2D meshes as 2D.
+            self.use_2d_mesh = 1 not in mesh_shape
 
         self.enforce_eager = model_config.enforce_eager
 
--- a/integrations/vllm_plugin/vllm_tt/platform.py
+++ b/integrations/vllm_plugin/vllm_tt/platform.py
@@ -111,6 +111,11 @@ class TTConfig:
     # Flag to enable 2D mesh for tensor parallel execution.
     use_2d_mesh: bool = True
 
+    # Explicit (batch, model) SPMD mesh shape for tensor/data parallel
+    # execution. When None, use_2d_mesh is used to determine the mesh shape.
+    # When set, it overrides use_2d_mesh.
+    mesh_shape: Optional[list[int]] = None
+
     # Flatten model I/O to a flat token stream at the model-call boundary
     # (needed by HF forwards like Gemma-4's PLE path).
     flat_model_io: bool = False
--- a/integrations/vllm_plugin/vllm_tt/vllm_utils.py
+++ b/integrations/vllm_plugin/vllm_tt/vllm_utils.py
@@ -8,9 +8,39 @@ from .logger import tt_init_logger
 logger = tt_init_logger(__name__)
 
 
-def determine_mesh_shape(num_devices: int, use_2d_mesh: bool) -> tuple[int, int]:
+def determine_mesh_shape(
+    num_devices: int,
+    use_2d_mesh: bool | None = None,
+    mesh_shape: tuple[int, int] | list[int] | None = None,
+) -> tuple[int, int]:
+    """Resolve the (batch, model) SPMD mesh shape against the device count.
+
+    'mesh_shape' is given priority over 'use_2d_mesh'.
+    If 'mesh_shape' is None and 'use_2d_mesh' is True, a pre-defined 2D mesh is used.
+    If 'mesh_shape' is None and 'use_2d_mesh' is False, [1, num_devices] is used.
+    """
+    # 1. If mesh_shape is provided, use it.
+    if mesh_shape is not None:
+        dims = list(mesh_shape)
+        if len(dims) != 2:
+            raise ValueError(
+                f"mesh_shape must have exactly 2 dimensions (batch, model); got {mesh_shape}"
+            )
+        if any(d <= 0 for d in dims):
+            raise ValueError(
+                f"mesh_shape dimensions must be positive; got {mesh_shape}"
+            )
+        if dims[0] * dims[1] != num_devices:
+            raise ValueError(
+                f"mesh_shape {tuple(dims)} has product {dims[0] * dims[1]}, "
+                f"which does not match the device count {num_devices}"
+            )
+        resolved = (dims[0], dims[1])
+        logger.info(f"Using mesh shape for {num_devices} devices: {resolved}")
+        return resolved
+
+    # 2. If use_2d_mesh is True, use a pre-defined 2D mesh.
     if use_2d_mesh:
-        # Use predefined mesh shapes based on number of devices
         mesh_shapes = {
             2: (1, 2),
             4: (2, 2),
@@ -18,12 +48,12 @@ def determine_mesh_shape(num_devices: int, use_2d_mesh: bool) -> tuple[int, int]
             16: (4, 4),
             32: (4, 8),
         }
+
         if num_devices in mesh_shapes:
-            mesh_shape = mesh_shapes[num_devices]
             logger.info(
-                f"Using predefined mesh shape for {num_devices} devices: {mesh_shape}"
+                f"Using predefined mesh shape for {num_devices} devices: {mesh_shapes[num_devices]}"
             )
-            return mesh_shape
+            return mesh_shapes[num_devices]
         else:
             # Fallback to computation for unsupported device counts
             logger.warning(
@@ -36,11 +66,13 @@ def determine_mesh_shape(num_devices: int, use_2d_mesh: bool) -> tuple[int, int]
             mesh_shape = (mesh_dim1, mesh_dim2)
             logger.info(f"Computed mesh shape: {mesh_shape}")
             return mesh_shape
+    # 3. If use_2d_mesh is False, use a 1D mesh.
     else:
-        # For 1D mesh, all devices are in one dimension.
-        mesh_shape = (1, num_devices)
-        logger.info(f"Using 1D mesh shape for {num_devices} devices: {mesh_shape}")
-        return mesh_shape
+        resolved = (1, num_devices)
+        logger.info(
+            f"Using default 1D mesh shape for {num_devices} devices: {resolved}"
+        )
+        return resolved
 
 
 def prev_power_of_2(n: int) -> int:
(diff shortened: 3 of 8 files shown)


tests/integrations/vllm_plugin/generative/test_gemma4_generation.py
```diff
@@ -27,7 +27,7 @@ def test_generation_single_device_gemma4_e4b():
             "enable_const_eval": True,
             "min_context_len": 32,
             "enable_tensor_parallel": False,
-            "use_2d_mesh": False,
+            "mesh_shape": [1, 4],
```
AleksKnezevic: we can leave this as default (None), can we not?

ddilbazTT: Yes we could, it would result in the same mesh shape. I explicated the mesh shapes to avoid confusion for future users.

integrations/vllm_plugin/vllm_tt/platform.py
```diff
@@ -108,8 +108,11 @@ class TTConfig:
     # target number of layers to filter the weights accordingly during loading.
     num_hidden_layers: int = 0
 
-    # Flag to enable 2D mesh for tensor parallel execution.
-    use_2d_mesh: bool = True
+    # Explicit (batch, model) SPMD mesh shape for tensor/data parallel
+    # execution. When None, defaults to a 1D mesh (1, num_devices). When set,
+    # its two positive dimensions must multiply to the device count, e.g.
+    # [1, 4] or [8, 4].
+    mesh_shape: Optional[list[int]] = None
```
mmanzoorTT, line 117: Why not to keep both? if `mesh_shape is not None` then use provided `mesh_shape` otherwise use the old path. Secondly, it always generate 1D mesh if `mesh_shape` is not specified.

ddilbazTT: Do you mean keeping `use_2d_mesh` and `mesh_shape` both at the same time?

tests/benchmark/test_vllm_benchmarks.py
```diff
@@ -180,46 +182,56 @@ def _gemma4_tp_config(model: str, batch_size: int):
...
 
+# All _tp_config benchmarks below run on n300-llmbox (8 devices): a (2, 4) 2D
+# mesh. The qb2-blackhole config (qwen3-32b-qb2) uses a 1D mesh (mesh_shape=None).
 TP_CONFIGS = [
-    pytest.param(_tp_config("tiiuae/Falcon3-7B-Base", 1), id="falcon3-7b-tp"),
-    pytest.param(_tp_config("tiiuae/Falcon3-10B-Base", 1), id="falcon3-10b-tp"),
-    pytest.param(_tp_config("Qwen/Qwen3-8B", 1), id="qwen3-8b-tp"),
     pytest.param(
-        _tp_config("Qwen/Qwen3-8B", 1, optimization_level=1),
+        _tp_config("tiiuae/Falcon3-7B-Base", 1, mesh_shape=[2, 4]), id="falcon3-7b-tp"
```
ssaliceTT, line 190: why not just set `mesh_shape=[2,4]` in `tp_defaults`?

ddilbazTT: Now, I made mesh_shape explicit. For example, the default galaxy mesh shape was [4, 8] but that doesn't work with Gemma. [8,4] works though. Similar cases could happen with llmbox where [2,4] vs [4,2] choice matters.

integrations/vllm_plugin/vllm_tt/vllm_utils.py
```diff
@@ -8,39 +8,37 @@
...
-        # For 1D mesh, all devices are in one dimension.
-        mesh_shape = (1, num_devices)
-        logger.info(f"Using 1D mesh shape for {num_devices} devices: {mesh_shape}")
-        return mesh_shape
+def determine_mesh_shape(
+    num_devices: int, mesh_shape: tuple[int, int] | list[int] | None = None
+) -> tuple[int, int]:
+    """Resolve the (batch, model) SPMD mesh shape against the device count.
+
+    When `mesh_shape` is None, defaults to a 1D mesh `(1, num_devices)`. When
```
ssaliceTT: Had the same issue with bringing up Qwen3-32b on bh galaxy so I agree with the option to specify a mesh when needed.  
But instead of defaulting to 1D mesh why not use ideal mesh shapes from above based on num devices as defaults if `mesh_shape` isn't given? Or could having `mesh_shape` act as an override while keeping `use_2d_mesh` be an option?
I think the current auto mesh determination would reduce friction for users to get started so might be worthwhile keeping it.

ddilbazTT: Currently, the auto determination only works if `use_2d_mesh` is set to true. If not set, it was always defaulting to a 1d mesh which parallels the default for new `mesh_shape`. I assumed somebody who wants 2d mesh would also be aware of the mesh shape they want to use.

integrations/vllm_plugin/vllm_tt/platform.py
mmanzoorTT: Yes; my suggestion is to keep both options. `mesh_shape` is a good option and it should be given preference over `use_2d_mesh`. It was in my TODO but didn't get a chance.

ddilbazTT: Will do!

integrations/vllm_plugin/vllm_tt/vllm_utils.py
ssaliceTT: In `vllm_tt/platform.py`, `use_2d_mesh` is initialized to True so it would default to a 2D mesh unless specified as False through `llm_args = { "additional_config": {"use_2d_mesh"}}`. In which case the behaviour proposed would be the opposite to main.

tests/benchmark/test_vllm_benchmarks.py
ssaliceTT: Makes sense, only suggested as all the tp tests currently are 2x4 except the ones explicitly overridden

ddilbazTT: Had some discussions offline, re-introducing `use_2d_mesh` that is lesser priority than `mesh_shape` :) It seems majority wants this :D

vvukomanTT (approved):

nsumrakTT (approved):

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -252,17 +252,19 @@ def __init__(
 
         # SPMD Related
         self.enable_tensor_parallel = self.tt_config.enable_tensor_parallel
-        self.use_2d_mesh = self.tt_config.use_2d_mesh
+        self.use_2d_mesh = False
```
mmanzoorTT: We are giving priority to 2D mesh shapes in vLLM
```suggestion
        self.use_2d_mesh = True
```
Or don't change the default value in `platform.py`

integrations/vllm_plugin/vllm_tt/vllm_utils.py
```diff
@@ -8,39 +8,59 @@
...
+    if any(d <= 0 for d in dims):
+        raise ValueError(f"mesh_shape dimensions must be positive; got {mesh_shape}")
+    if dims[0] * dims[1] != num_devices:
+        raise ValueError(
+            f"mesh_shape {tuple(dims)} has product {dims[0] * dims[1]}, "
+            f"which does not match the device count {num_devices}"
+        )
+    resolved = (dims[0], dims[1])
+    logger.info(f"Using mesh shape for {num_devices} devices: {resolved}")
+    return resolved
```
mmanzoorTT: Keep fallback mechanism for unsupported/unseen device count. 
```suggestion
    if mesh_shape is not None:
        dims = list(mesh_shape)
        if len(dims) != 2:
            raise ValueError(
                f"mesh_shape must have exactly 2 dimensions (batch, model); got {mesh_shape}"
            )
        if any(d <= 0 for d in dims):
            raise ValueError(f"mesh_shape dimensions must be positive; got {mesh_shape}")
        if dims[0] * dims[1] != num_devices:
            raise ValueError(
                f"mesh_shape {tuple(dims)} has product {dims[0] * dims[1]}, "
                f"which does not match the device count {num_devices}"
            )
        resolved = (dims[0], dims[1])
        logger.info(f"Using mesh shape for {num_devices} devices: {resolved}")
        return resolved

    if use_2d_mesh:
        # Use predefined mesh shapes based on number of devices
        mesh_shapes = {
            2: (1, 2),
            4: (2, 2),
            8: (2, 4),
            16: (4, 4),
            32: (4, 8),
        }
        if num_devices in mesh_shapes:
            mesh_shape = mesh_shapes[num_devices]
            logger.info(
                f"Using predefined mesh shape for {num_devices} devices: {mesh_shape}"
            )
            return mesh_shape
        else:
            # Fallback to computation for unsupported device counts
            logger.warning(
                f"Using fallback computation for {num_devices} devices (not in predefined shapes)"
            )
            mesh_dim1 = int(num_devices**0.5)
            while num_devices % mesh_dim1 != 0:
                mesh_dim1 -= 1
            mesh_dim2 = num_devices // mesh_dim1
            mesh_shape = (mesh_dim1, mesh_dim2)
            logger.info(f"Computed mesh shape: {mesh_shape}")
            return mesh_shape
    else:
        # For 1D mesh, all devices are in one dimension.
        mesh_shape = (1, num_devices)
        logger.info(f"Using 1D mesh shape for {num_devices} devices: {mesh_shape}")
        return mesh_shape```

File: tests/integrations/vllm_plugin/generative/test_tensor_parallel_generation.py [hunk 7]
```diff
@@ -43,10 +43,10 @@ def test_tensor_parallel_generation_n300(model_name: str):
         pytest.param("Qwen/Qwen3-0.6B"),
     ],
 )
-@pytest.mark.parametrize("use_2d_mesh", [True, False])
+@pytest.mark.parametrize("mesh_shape", [[2, 4], [1, 8]])
```
mmanzoorTT: I'll suggest to keep at least few of the tests using `use_2d_mesh` option and test this configuration. Or keep using `use_2d_mesh` option for all old test and use `mesh_shape` option for the newly added test only.

integrations/vllm_plugin/vllm_tt/platform.py
```diff
@@ -109,7 +109,12 @@ class TTConfig:
     num_hidden_layers: int = 0
 
     # Flag to enable 2D mesh for tensor parallel execution.
-    use_2d_mesh: bool = True
+    use_2d_mesh: Optional[bool] = True
```
mmanzoorTT: Do we need `Optional` here?

tests/integrations/vllm_plugin/generative/test_tensor_parallel_generation.py
```diff
@@ -158,18 +158,25 @@ def test_tensor_parallel_generation_galaxy_wh_6u_large(
 
 @pytest.mark.nightly
 @pytest.mark.tensor_parallel
-@pytest.mark.bhqb
```
mmanzoorTT, line 161: Is removing `@pytest.mark.bhqb` intentional?

ddilbazTT: Yes, it's in the params:
```
@pytest.mark.parametrize(
    "mesh_shape",
    [
        pytest.param([1, 4], marks=pytest.mark.bhqb),
        pytest.param([8, 4], marks=pytest.mark.bh_galaxy),
    ],
)
```

mmanzoorTT: Thanks, I missed that.

mmanzoorTT (approved):

PR #5228, opened 2026-06-16
[vLLM] Fix cached-prefill greedy degeneration: cold-direct SDPA + TP sharding

### LINKED ISSUE #5132: Cached prefill (prefix-cache HIT) with trace produces empty greedy output (TT)
**Component:** tt-xla vLLM plugin (`integrations/vllm_plugin/vllm_tt`) — cached-prefill / paged-KV path
**Hardware:** P150 (Blackhole), single device

## Problem

Hit by all LLMs in tt-inference-server benchmarks.  With **`enable_trace=True` + prefix caching**, a greedy (`temperature=0`) completion of the json-mode prompt generates the full answer on the **first** request (prefix-cache MISS → cold prefill) but collapses to **zero output text** on any **subsequent identical** request (prefix-cache HIT → cached prefill). The cache-hit cached-prefill path is producing wrong/degenerate first-token logits. Greedy decoding of an identical prompt must be deterministic and non-empty, so this is a correctness bug in the cached-prefill path.

The trigger is the **cache hit**, not concurrency. A single cold request works (51 text chunks); the very next identical request degenerates (0 text chunks). A concurrent batch of identical prompts degenerates the same way — row 0 is the cache miss and generates, rows 1-3 hit the prefix row 0 just inserted and collapse (`text=[51,0,0,0]`).

**Not a structured-decoding/grammar bug.** The repro uses no guided decoding, and the surfacing benchmark runs at `--structured-output-ratio 0.0` (grammar *off*). Grammar masking actually hides it (the ratio=1.0 path passes); the failure is in the unconstrained greedy path. "json-mode" here is just the prompt content (it asks for JSON in plain text), not constrained decoding.

How it relates to the benchmark failure (**same underlying bug — chain verified**): the Falcon3-7B `xgrammar_bench --structured-output-ratio 0.0` run in CI reports **"Never received a valid chunk to calculate TTFT"** (tt-shield run [27117648384](https://github.com/tenstorrent/tt-shield/actions/runs/27117648384/job/80030102763), tt-inference-server [#3951](https://github.com/tenstorrent/tt-inference-server/issues/3951)). All the links match:

- **Warm trigger:** the CI server serves a suite of benchmarks first, so every json-mode request is a prefix-cache hit — exactly this repro's trigger.
- **Greedy:** the degeneration is greedy-specific (at default/random sampling the warm hit generates fine; only `temperature=0` collapses it). The CI bench is **vLLM v0.13.0**, which *defaulted to greedy* — the newer 0.19.1 used here carries the note *"vllm bench serve no longer sets temperature==0 (greedy) … for the old behavior include --temperature=0"*. So CI ran greedy and hit the greedy-only degeneration.
- **The 1-vs-0 `choices`-chunk gap is a vLLM frontend-version artifact.** That bench error fires only when the stream carries **zero** `choices` chunks (`vllm/benchmarks/lib/endpoint_request_func.py` — an empty-*text* chunk still flips `first_chunk_received=True`). For a zero-token completion the **v0.13.0** server emits *"only a usage summary and [DONE]"* (**0** choices chunks → bench fails); the **0.19.1** server here emits **1** empty-text stop chunk (→ bench reports success). Same zero-token degeneration, different streaming packaging by the two frontend versions.

Verified directly (driving the genuine `async_request_openai_completions` against a warm source server): the real bench reports `success=True` here in every case, because the 0.19.1 frontend always emits ≥1 chunk. Reproducing the literal zero-chunk error would require the old v0.13.0 *server* frontend (a version-pin lift, not worth it) — the numerical bug to fix is the warm cache-hit greedy degeneration, which is fully reproduced.

## Repro

**Branch `kmabee/issue-5132-cached-prefill-greedy-repro` has ebf5e5e59 to add below test for repro/debug.**

`tests/integrations/vllm_plugin/generative/test_prefix_cache_degeneration.py::test_cached_prefill_warm_hit_not_degenerate`

Starts a trace + prefix-caching Falcon3-7B server, sends the exact json-mode prompt once (cold, expect full output) then again (warm cache hit), and asserts the warm request is still non-empty.

```bash
pytest -svv tests/integrations/vllm_plugin/generative/test_prefix_cache_degeneration.py

cold (cache miss): {'choices_chunks': 52, 'text_chunks': 51}
warm (cache hit) : {'choices_chunks': 1, 'text_chunks': 0}
E   AssertionError: warm prefix-cache HIT produced zero output text
    (cold={'choices_chunks': 52, 'text_chunks': 51},
     warm={'choices_chunks': 1, 'text_chunks': 0}) — cached-prefill greedy degeneration
```

**Reproduces reliably and deterministically** — cold generates (51 text chunks), the next identical (warm) request produces zero output text.

## More details

- Config that triggers it: `enable_trace=True`, `optimization_level=0` (trace + opt≥1 is rejected, #4570), `experimental_weight_dtype=bfp_bf8`, prefix caching on, `min_context_len=32`.
- `enable_trace=True` is required — with trace off it is stable. This is why no-trace/offline paths and the tt-mlir/tt-metal submodule bisect (all trace-off) were inconclusive; do not trust that bisect.
- The prompt must be the **exact** json-mode-eval[0] string (967 chars) — a paraphrase does not trigger it.
- **`fp32_dest_acc_en=True` does NOT fix it** (verified: same degeneration with the flag on, confirmed plumbed through `TTConfig` → PJRT compile options). So this is **likely a different root cause than #5116**, which *was* fixed by fp32 accumulation — despite the similar "identical greedy diverges" symptom. Root cause is **open — needs debug**. The goal is a correct cached prefill, not disabling features (keep trace + prefix caching enabled).

## Cross-references

- **#5116** — batched-prefill bf16-accumulation non-determinism. Similar symptom (identical greedy diverges), but `fp32_dest_acc_en` fixes #5116 and does **not** fix this — probably a distinct root cause.
- tt-inference-server [**#3951**](https://github.com/tenstorrent/tt-inference-server/issues/3951) — the benchmark return-code-1 symptom.

alinakhanTT:
### Ticket
closes #5132.

## Problem

With `enable_trace=True` + prefix caching + greedy decoding (temperature=0), a request that **misses** the prefix cache produces correct output, but the **next identical** request (which **hits** the cache) collapses to empty/garbage output.

Root cause is the trace-once model: `torch.compile` is wrapped with `NoGuards`, so the model is traced **once** during warmup and replayed forever. Any Python `if`/`else` on cold-vs-cached prefill gets **baked at trace time**; the other branch becomes dead code. Warmup used dummy inputs with `num_computed_tokens=0`, so only the **cold-prefill** graph was recorded, cached requests then ran the cold program against offset inputs → wrong attention → degenerate first token.

Two independent defects turned out to be in play, each with a distinct symptom:

1. **Cold / single-chunk prefill degenerated to token 0** on some models (Llama-3.2-3B) because it *gathered K/V back out of the paged cache* after writing them; the write-then-read-back round-trip corrupts the freshly-computed K/V and breaks first-token attention. (The always-mask "extra softmax columns" theory and the tt-mlir `#8929` pin were both experimentally **ruled out**, each degenerated identically.)
2. **Tensor-parallel decode produced token-soup** (`meta-llama/Llama-3.2-3B`, n300): a tt-mlir regression (#8931) in the `nlp_create_qkv_heads_decode` fusion mis-routed K/V, computing `Q · rope(V)` with context from raw `K`. Fixed upstream by tt-mlir **#8960**.

## What's changed

**`attention_impls/attention.py`** : `_compute_full_attention` now selects one of three graphs by a *Python-level* trigger (never data-dependent control flow), so each traces as a distinct, stable graph:

- **chunked-prefix hit** : a chunk whose prefix is already in the paged cache attends over it via `chunked_scaled_dot_product_attention` (mask + offset internal to the kernel, reads the paged cache directly, no host gather).
- **cold prefill** (`num_computed_tokens == 0`, `attn_mask is None`) : attends its own freshly-computed `inputs.key/value` **directly** with native-causal SDPA (`is_causal=True`), like the no-paged-cache path. Skips the redundant full-slab gather whose read-back caused the token-0 degeneration; also cheaper (drops the 270 MiB cold mask). Shared-KV layers have no local K/V, so they still gather.
- **cached-prefix hit** (`attn_mask` set) : gathers the full per-user K/V slab from the paged cache and runs SDPA with an explicit `attn_mask`. Under TP, re-asserts head-axis sharding with a graph-emitted `sharding_constraint_tensor` (the gather drops the head-dim sharding the KV cache carries; `mark_sharding` can't be traced inside the fused-prefill fullgraph region).

**`model_runner.py`** : warmup now precompiles **both** the cold (`attn_mask=None`, native-causal) and cached (masked-gather) prefill graphs via a new `cold` dimension in the `_precompile_model_fused` config product, so a warm cache hit never triggers a mid-serving recompile (the original degeneration trigger).

**Tests** : new `test_prefix_cache_degeneration.py` (cold-miss vs warm-hit identical greedy output); `test_chunked_prefill.py` asserts single-chunk == multi-chunk prefix.

## Verification

- **Cached-hit repro (on-device, Wormhole):** cold (cache miss) and warm (cache hit) produce **identical** non-empty greedy output (`text_chunks=51` both).
- **Cold / single-chunk (Llama-3.2-3B, on-device):** coherent output, `ids[:8]=[578, 3925, 315, 25213, ...]` (was degenerate token 0).
- **Chunked-prefill matches single-chunk (n150, CI):** PASS.
- **Tensor-parallel generation (n300):** validated with tt-mlir #8960 (now in `main`); the token-soup was the tt-mlir fusion bug, not this diff's Python.
- **Nightly run:** confirming no new regressions https://github.com/tenstorrent/tt-xla/actions/runs/29135453400

## Notes

- On n150, the Falcon3-7B degeneration test can OOM at the forge config (hard 12 GB DRAM: weights + const-eval ~11.5 GB + warmup transients); the cold-direct fix drops the 270 MiB cold mask, but a lighter config or a larger-DRAM runner may still be needed for that param.
- Unrelated CI reds (JAX `test_dot_psum` PCC, forge-models, training, uplift-qualification) fail on `main` too.


integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
```diff
@@ -171,8 +171,15 @@ def fill_cache_workaround(
...
     cache_position: torch.Tensor
+    # Prefill SDPA mask. Shape: [num_users, 1, suffix_len, num_blocks*block_size],
+    # dtype matches Q (typically bfloat16). 0 where Q[i] should attend to K/V[j],
+    # -inf otherwise. Encodes the per-user causal pattern including any cached
+    # prefix offset (num_computed_tokens). None during decode.
     attn_mask: torch.Tensor
     page_table: torch.Tensor
+    # Only consumed by the decode kernel. SDPA prefill ignores this and reads
+    # the pattern from attn_mask instead — that branch passes is_causal=False
+    # so the mask is the single source of truth.
```
mmanzoorTT: My suggestion is to not add prefill specific comments here. `attn_mask`, `is_causal` and others are being used by pooling_runner as well. It may cause confusion. You have already explained `attn_mask` in `model_runner.py::_build_prefill_atnn_mask`.

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
```diff
@@ -569,30 +587,71 @@ def _compute_full_attention(
...
+        # freshly-computed K/V directly -- like main / the no-paged-cache path --
+        # instead of gathering the full slab back out of the paged cache. The
+        # prompt's K/V is already in inputs.key/value (and was just written to
+        # the cache for later decode), so the gather is redundant work whose
+        # full-slab read-back degenerates the first token (token 0) for some
+        # models (Llama-3.2-3B). Native causal over the aligned prompt is
+        # numerically correct AND cheaper. Shared-KV layers have no local K/V,
+        # so they must still gather. Cached-prefix hits (attn_mask set) keep the
+        # full-slab always-mask gather.
+        cold_direct = (
```
mmanzoorTT: are `cold_direct` and `is_cold_prefill` (defined in model_runner) different?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2672,6 +2824,13 @@ def _get_dummy_inputs(self, config: dict):
         ).to(self.device)
         cache_position = torch.ones((num_reqs,), dtype=torch.int32).to(self.device)
 
+        attn_mask = None
+        if num_tokens > 1 and not cold and not self.enable_data_parallel:
```
mmanzoorTT: Why do we need `self.enable_data_parallel` here again? isn't it used to determine `cold_options` at line 2697?

integrations/vllm_plugin/vllm_tt/platform.py
```diff
@@ -128,7 +128,14 @@ class TTConfig:
...
     # Enable fp32 destination accumulation in matmul/reduction kernels.
-    fp32_dest_acc_en: Optional[bool] = None
+    # Defaults to True: the always-mask prefill SDPA path (is_causal=False +
+    # explicit mask over the full padded K/V slab) normalizes the softmax over
+    # more columns than native causal, and in bf16 that rounding can flip a
+    # borderline greedy first token to EOS (empty output — reproduced on
+    # Wormhole with facebook/opt-125m via test_responses_api_instructions).
+    # fp32 dest accumulation closes that gap. Set False in additional_config to
+    # trade precision for speed.
+    fp32_dest_acc_en: Optional[bool] = True
```
mmanzoorTT: Don't change it. Please have a look at https://github.com/tenstorrent/tt-xla/pull/5599/changes#diff-2fb0263502e5033497192d3dd58db0547fb2afe6c1fd6d5438c7bb0be387f6dbR131

tests/integrations/vllm_plugin/generative/test_prefix_cache_degeneration.py
```diff
@@ -0,0 +1,283 @@
...
+frontend.
+
+The prompt must be the **exact** json-mode-eval[0] string (below) — a paraphrase
+does not trigger it. ``enable_trace=True`` is required (trace off is stable),
+and the collapse is greedy-specific (``temperature=0``).
+
+This test fails on the buggy build (the warm request produces empty output); it
+passes once the cached-prefill path is fixed. Prefix caching and trace stay
+enabled — the fix is a correct prefill, not disabling features.
+"""
```
mmanzoorTT: Please move this explanation to issue page.

mmanzoorTT (commented):
Please avoid adding similar comments/explanation at multiple places. It will become hard to keep them updated with every modification. e.g. `attn_mask` is explained at multiple locations in `attention.py` and `model_runner.py`. It should be explained once in `_build_prefill_attn_mask` which can be used as ground truth for entire plugin.

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
alinakhanTT: Done!

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
alinakhanTT: Same "no cached prefix" condition observed at two layers, not a duplicated flag. is_cold_prefill is the decision, computed from num_computed_for_reqs, model_runner signals it downstream by leaving attn_mask=None. The attention kernel never sees num_computed, so cold_direct re-derives cold-ness from attn_mask is None, narrowed by two kernel-only conditions (has_paged_cache, not shared_kv_mode). Updated the comment to mention this

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: You're right, it's redundant. Removed the  and not self.enable_data_parallel

integrations/vllm_plugin/vllm_tt/platform.py
alinakhanTT: Reverted to = None!

tests/integrations/vllm_plugin/generative/test_prefix_cache_degeneration.py
alinakhanTT: Trimmed the docstring to just the repro essentials (what degenerates, and that the exact prompt + enable_trace=True + greedy are required). Moved the full root-cause analysis and the Falcon3-7B xgrammar_bench chain to #5132

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
```diff
@@ -569,30 +580,73 @@ def _compute_full_attention(
...
+                "is_causal": attn_metadata.is_causal,
+                "attn_mask": attn_metadata.attn_mask,
+                "scale": self.scale,
+            }
 
-        sdpa_kwargs = {
-            "is_causal": attn_metadata.is_causal,
-            "attn_mask": attn_metadata.attn_mask,
-            "scale": self.scale,
-        }
```
mmanzoorTT, line 585: Why are you creating `sdpa_kwargs` for each path separately? You are already creating `attn_mask` and `is_causal` based on `is_code_prefill` in model_runner.py. Then you don't need `if cold_direct:` path.

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
```diff
@@ -569,30 +580,73 @@ def _compute_full_attention(
...
+                from tt_torch.sharding import sharding_constraint_tensor
+
+                # The paged gather (gather/view/permute/reshape) drops the
+                # head-dim sharding the KV cache carries, so under TP the
+                # gathered K/V keep full heads while Q stays sharded -> SDPA
+                # "Query num heads must be divisible by key/value num heads".
+                # Re-assert the head-axis sharding with a graph-emitted
+                # sharding_constraint (torch.compile-safe; eager mark_sharding
+                # can't be traced inside the fused-prefill fullgraph region).
+                key_for_sdpa = sharding_constraint_tensor(
```
mmanzoorTT, line 610: Can you try sharding `attn_mask` in `model_runner.py` instead of sharding key/value here?

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
alinakhanTT: Collapsed the three-way branch into must_gather (cached-prefix hit or shared-KV layer -> gather from the paged cache) vs attending inputs.key/value directly!

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
```diff
@@ -569,30 +581,54 @@ def _compute_full_attention(
...
         else:
-            # scaled_dot_product_attention expects [B, N_tokens, N_heads, H]
             query_for_sdpa = inputs.query.transpose(-3, -2)
             key_for_sdpa = inputs.key.transpose(-3, -2)
             value_for_sdpa = inputs.value.transpose(-3, -2)
+            sdpa_kwargs = {
+                "is_causal": attn_metadata.is_causal,
+                "attn_mask": attn_metadata.attn_mask,
+                "scale": self.scale,
+            }
```
mmanzoorTT, line 630: This setting `sdpa_kwargs` should be outside of if-else and use the same old code. model_runner is setting `is_causal` and `attn_mask`. Is there any case where these will not same as created in model_runner?

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
alinakhanTT: The mask can't carry this sharding, attn_mask is [num_users, 1, suffix_len, kv_len] its head dim is 1, so there's nothing to shard on the "model" axis. The failure is the SDPA head-count assert on the K/V inputs: _gather_paged_to_dense drops the head-axis sharding the KV cache carried, so the gathered K/V present full num_kv_heads while Q is sharded. Only re-annotating K/V (as here) restores the head split, the mask's sharding doesn't affect K/V head count. Kept the constraint, but happy to revisit if there's a mask layout with a real head dim I'm missing

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
kmabeeTT: Answering your "do these ever differ" question — yes, one case: a shared-KV layer in a fully-cold batch has `must_gather` true but `attn_mask` None, so it runs non-causal SDPA over the full padded slab with no mask. Your suggestion (just use `attn_metadata.is_causal`/`attn_mask`here) fixes it. Latent unless a validated model sets `kv_sharing_target_layer_name` — worth closing though.

kmabeeTT (commented):
One thing seems risky, left reply to Asif's thread.  Also PR description needs some small updates (no more mlir uplift, and fp32_dest_acc_en change in platform.py got reverted I think).

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
mmanzoorTT: Sorry, I wasn't clear. I wanted to ask that is there any case where `attn_mask` and `is_causal` differ from what we have set in `model_runner.py`? If not, then we should use them instead of hard-coded value.

kmabeeTT: Yeah claude says yes — this (else) block already uses the metadata values, so hoisting it out like you suggest is fine for cached-prefix (model_runner sets `is_causal=False` when `attn_mask` is set, matching). The divergence is in the `must_gather` branch, which hardcodes `is_causal=False`: for shared-KV cold, `attn_mask` is None so model_runner gives `is_causal=True` there. Unifying on the metadata values (as you propose) uses `True` for that case and fixes it.

alinakhanTT: Thanks Kyle!  shared-KV in a fully-cold batch does hit the gather path with attn_mask None, so today it's non-causal over the padded slab
But I tried the flip on device and it doesn't actually compile: switching the gather branch to attn_metadata.is_causal fails with Error code: 13 in _xla_warm_up_cache during prefill precompile on google/gemma-4-E4B-it. Gemma-3n sets kv_sharing_target_layer_name, so the E4B nightly already exercises this,  it's not purely latent, and hardcoded False is what compiles + generates coherent output. CI: https://github.com/tenstorrent/tt-xla/actions/runs/29363128438
Reason: the gather path runs SDPA over the padded slab (key_len > query_len), and is_causal=True there wants a bottom-right causal mask over the padding, which TT SDPA can't legalize. So attn_mask can be hoisted (same in both branches), but is_causal has to stay False here. If we want true causal semantics for shared-KV cold, that's a separate change (explicit slab-sized mask), happy to file a follow-up

tests/integrations/vllm_plugin/generative/test_chunked_prefill.py
```diff
@@ -29,6 +29,15 @@
...
 
+# bf16 query over a bfp8 KV cache trips the un-relaxed TTIR
+# PagedScaledDotProductAttentionDecodeOp verifier at engine warmup. The TTNN
+# verifier already allows it (tt-mlir #8668); the TTIR one was never mirrored.
+# Fix tracked in tt-mlir #8988; un-xfail once it lands and is pinned.
+_BFP8_KV_XFAIL = pytest.mark.xfail(
+    reason="bf16 query over bfp8 KV cache trips TTIR SDPA decode verifier (tt-mlir #8988)",
+    strict=True,
+)
+
```
mmanzoorTT: Please remove this XFAIL. tt-mlir#8988 is part of tt-xla now.

alinakhanTT: Done

mmanzoorTT (approved):

kmabeeTT:
Taking another look thx for earlier changes

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -747,6 +779,41 @@ def _alloc_dev(shape, dtype):
...
+        # Constant shape lets warmup and runtime share a single traced graph.
+        # Keyed by (num_reqs, num_tokens): num_reqs follows the request-count
+        # bucketing (min/max) so the mask batch matches the query batch
+        # (target_num_reqs), like the page_table/cache_position buffers.
+        self._prefill_attn_mask_dev_max: dict[tuple[int, int], torch.Tensor] = {}
+        self._prefill_attn_mask_dev_most: dict[tuple[int, int], torch.Tensor] = {}
+        for _nt in self.num_tokens_paddings:
+            if _nt <= 1:
+                continue
+            for _nr in {self.min_num_reqs, self.num_reqs_max_model_len}:
```
kmabeeTT, line 793: These key on `{min_num_reqs, num_reqs_max_model_len}`, but prefill row counts come from `_bucket_num_reqs`, whose middle bucket is `min(max_prefill_num_reqs, path_max)` — the `_page_table_dev_max` dicts just above use the full 3-element set. So any config with `max_prefill_num_seqs != min_num_seqs` KeyErrors here. 

`test_vllm_benchmarks.py::test_vllm_benchmark[falcon3-7b-instruct-production]` (max_num_seqs=32, max_prefill_num_seqs=16, min_num_seqs=1) gives keys `{1, 32}` while prefill runs at 16 rows, so `_get_dummy_inputs` raises during warmup; `test_max_prefill_num_seqs.py` hits it the same way. `_prepare_inputs` has the same gap at runtime. Keying off `self._max_len_num_reqs` / `_most_len_num_reqs` (as the page-table dicts do) fixes both. 

Worth adding that benchmark to whatever you re-run after the rebase — it's our one config at real production settings (32K context, 1K chunks) and it hits every path this PR touches.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2627,6 +2765,20 @@ def _dummy_run(
         )
         cache_position = torch.ones((num_reqs,), dtype=torch.int32).to(self.device)
 
+        # Match runtime: prefill buckets carry a constant-shape attn_mask so
+        # the traced graph picks up the SDPA-with-mask code path; decode
+        # bucket (num_tokens=1) leaves mask=None and uses the decode kernel.
+        # Data parallel is always cold (see _precompile_model_fused).
+        attn_mask = None
+        if num_tokens > 1 and not self.enable_data_parallel:
```
kmabeeTT, line 2773: `_dummy_run` builds a mask for every `num_tokens > 1` with no cold variant, and `_precompile_backbone` only varies `prefix_chunk` — so on the unfused path (`cpu_sampling=True`) we warm the masked graph but never the cold one, and the first real prefill compiles mid-serving. Same hazard as #5132, inverted: pre-PR warmup and runtime prefill both had attn_mask=None. `cpu_sampling=True` is used by the benchmark suite, the gemma4 image test and the TP/DP+TP tests, and the new degeneration test pins it False. Was the unfused path exercised?

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
kmabeeTT: Thanks for digging into the legalization failure — agreed `is_causal` has to stay `False` here. Yeah, could you file it as follow up issues and leave a TODO with the issue number at the hardcoded `False`? Worth noting in the issue that main trimmed this path to `kv_num_tokens` with `is_causal=True`, so shared-KV cold is a small regression rather than a pre-existing gap.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1548,12 +1634,64 @@ def _prepare_inputs(self, scheduler_output: "SchedulerOutput", start_index: int)
...
+        # Cold and cached are distinct traced graphs, both precompiled in warmup
+        # -> no runtime recompile. Decode (suffix == 1) keeps attn_mask None and
+        # the decode kernel handles causality.
+        attn_mask: torch.Tensor | None = None
+        is_cold_prefill = not bool(np.any(num_computed_for_reqs > 0))
+        # Cached-prefix prefill takes the masked full-gather path, under DP too:
+        # DP suppresses only chunk_start_idx (the chunked SDPA op), and routes
+        # the cached prefix through the masked gather instead. Cold prefill
+        # leaves attn_mask None. Both shapes are precompiled in warmup.
+        if padded_total_num_scheduled_tokens > 1 and not is_cold_prefill:
```
kmabeeTT, line 1650: This is gated only on `not is_cold_prefill`, so it also runs when `chunk_start_idx` is set — and in that case `_compute_full_attention` returns from the `chunked_prefix` branch without ever reading `attn_mask`. So the default chunked path pays a host mask build plus an H2D copy per continuation chunk for a dead tensor — at `falcon3-7b-instruct-production` (32K context, 1K chunks, 16 prefill rows) that's 1 GiB copied per prefill step that nothing reads. Can we add `chunk_start_idx is None` here?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1525,11 +1592,30 @@ def _prepare_inputs(self, scheduler_output: "SchedulerOutput", start_index: int)
...
             np.any(num_computed_for_reqs > 0)
         )
         chunk_start_idx = None
-        if prefix_chunk_step and self._chunked_sdpa_active:
+        # DP keeps prefill on the cold/direct SDPA path; do not route DP
+        # continuation chunks through chunked SDPA.
+        if (
+            prefix_chunk_step
+            and self._chunked_sdpa_active
+            and not self.enable_data_parallel
```
kmabeeTT, line 1617: `test_data_parallel_chunked_prefill_n300[Qwen/Qwen3-0.6B]` fails on the branch with degenerate output (`' ...\n...\n'`, no word characters) and failed again on the retry. It's a `push` test and it's green on main, including run 30376857319 from the same day. Its config is DP with prefill_chunk_size=128 < max_model_len=512, so this gate is what changes its behaviour: continuation chunks come off the chunked-SDPA op and onto the masked gather. Same degeneration class we're fixing in this PR, so I'd rather root-cause it here than defer — and the new `test_dp_chunked_prefill_mixed_cached_and_cold` is nightly-marked, so it won't gate it either.

kmabeeTT (changes requested):
Structurally looks good for #5132 - Requesting changes/fixes for two things, both inline: a `push` test that regresses on this head (`test_data_parallel_chunked_prefill_n300`, comment on the DP gate) and a warmup KeyError on the mask buffers. Three more inline — the unfused warmup path, the mask copied on the chunked path, and a TODO/issue ask on the shared-KV `is_causal` thread — are fine as follow-up commits or your next PR as long as each gets a filed issue.

Branch needs rebase to solve conflicts, plz rerun vllm nightly, and also `test_vllm_benchmarks.py::test_vllm_benchmark[falcon3-7b-instruct-production]` to ensure passing.

PR #5235, opened 2026-06-16
Add Lazy Execution Option in Legacy Compile Path

pglusacTT:
### Ticket
N/A

### Problem description
Legacy path always forces torch_xla sync. In training, we want more freedom to target specific tensors (like gradients and loss only) in order to produce cleaner execution graphs.

### What's changed
- Add `tt_lazy_execution` flag that is off by default and it skips the syncs in the `__call__` method.

### Checklist

### DIFF
--- a/python_package/tt_torch/backend/backend.py
+++ b/python_package/tt_torch/backend/backend.py
@@ -244,7 +244,6 @@ class XLAExecutor:
                 gm_has_functional_output_kind = False
                 break
 
-
         if self.lazy_execution:
             return output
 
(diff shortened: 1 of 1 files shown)


pglusacTT:
FYI @abogdanovicTT @agobeljicTT

python_package/tt_torch/backend/backend.py
```diff
@@ -242,6 +244,9 @@ def __call__(self, *args):
                 gm_has_functional_output_kind = False
                 break
 
+        if self.lazy_execution:
+            return output
+
```
agobeljicTT: what if `gm_has_functional_output_kind`
does the user need to manually do `torch_xla._XLAC._xla_sync_multi(output_tensors, devices, wait=False)` afterwards or?

pglusacTT: I've moved the early exit upwards so it doesn't compute `gm_has_functional_output_kind`.
Sync is left to the user, which we already do in our training scripts.

mstojkovicTT (approved):
lgtm!

vkovinicTT (approved):

Issue #5209, opened 2026-06-15
PCC failures on n300-llmbox and galaxy-wh-6u TP models


vkovacevicTT:
Several TP LLM benchmarks fail their PCC checks.

## n300-llmbox
- `test_llama_3_1_70b_tp` — `Prefill PCC failed. PCC=0.011727, Required=0.94`

## galaxy-wh-6u
- `deepseek_v3_1_tp_galaxy_4_layers` — `Prefill PCC failed. PCC=0.810351, Required=0.96`
- `glm_4_7_tp_galaxy_4_layers` — `First decode PCC failed. PCC=0.859760, Required=0.86`
- `kimi_k2_tp_galaxy_2_layers` — `Prefill PCC failed. PCC=0.048342, Required=0.94`
- `kimi_k2_5_tp_galaxy_2_layers` — `Prefill PCC failed. PCC=0.047948, Required=0.94`

---
**Run**: [nightly 27517218376](https://github.com/tenstorrent/tt-xla/actions/runs/27517218376) (2026-06-15)
**Commits**: tt-xla `98af4ea8` · tt-mlir `464a5f34` · tt-metal `68e82deb`

gengelageTT:
@sshonTT found two things causing these regressions: 

The first is a bug with the `rms_norm_pre_all_gather` op in tt-mlir. Fixing this restores GLM4.7 to what was seen before and gets deepseek-v3.1 and kimi-k2 above `0.90`. I can put the fix up in a PR once I confirm Sungjoon hasn't already done so:
```
// rms_norm_pre_all_gather emits the per-device *sum* of squares (sum(x^2)),
// not the per-device mean. The MeanOp above only divided the gathered sums by
// numDevices; to recover the global E(x^2) = sum(x^2) / H we must additionally
// divide by the per-device hidden size N (= input's last dim). Omitting this
// leaves the variance N times too large, shrinking the normalized output by
// sqrt(N).
```


The second regression is caused by `tt-metal commit 955e110`. I'm not sure about the status of a fix for this.

mvasiljevicTT:
@gengelageTT thank you for pointing this out! 

You are right, in `DistributedRMSNormDecompositionRewritePattern` there should be an additional div in a post all_gather part to fully reproduce what `rmsnorm_post_all_gather` op does. 

[...]

PR #5240, opened 2026-06-16
[Benchmark] Fix GLM4.7 and Deepseek-v3.1 MoE shard specs

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/5096

### Problem description
These models were both seeing low pcc for 4 layers that dropped significantly as more layers (4 -> 6) were run. Also, the graph for GLM4.7 contains a lot of `point_to_point` ops which is usually an indication of improper sharding.

### What's changed
Both models see a PCC improvement when the expert sharding switches from `("model", "batch")`  to `("batch", "model")`. 

**Deepseek-v3.1** with this change (before the regression tracked in https://github.com/tenstorrent/tt-xla/issues/5209) sees:
- 4 layers - `0.978 -> 0.998` prefill and `0.965 -> 0.999` decode
- 6 layers - `0.886 -> 0.992` prefill and `XX -> 0.996` decode

**GLM4.7** with this change compared to nightly:
- 4 layers - `0.935 -> 0.997` prefill and `0.859 -> 0.993` decode

I set `required_pcc=0.99`. Since this is only a 4 layer test, pcc lower than 0.99 usually indicates a larger full model regression.

Changing `kv_cache_sharding_spec=("batch", None, None, None)` to `kv_cache_sharding_spec=("batch", "model", None, None)` for GLM reduced the number of `point_to_points` from `2688` in prefill to `0` but had no effect on PCC.

**Deepseek-v3.2-exp** - fixed a typo in `required_pcc`. Nightly pcc is still above `0.92`.

### Checklist

### DIFF
--- a/tests/benchmark/test_llms.py
+++ b/tests/benchmark/test_llms.py
@@ -2768,5 +2768,5 @@ def test_glm_4_7_tp_galaxy_4_layers(
         shard_spec_fn=_glm_4_7_shard_spec_fn,
         input_output_sharding_spec=("batch", None),
         kv_cache_sharding_spec=("batch", "model", None, None),
-        required_pcc=0.99, # Since this is only a 4 layer test, pcc lower than 0.99 likely indicates a larger full model regression
+        required_pcc=0.99,
     )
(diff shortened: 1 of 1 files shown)


gengelageTT:
Ran tests in CI here: https://github.com/tenstorrent/tt-xla/actions/runs/27624289893

Deepseek is failing due to the regression. Tests [with this fix](https://github.com/tenstorrent/tt-mlir/pull/8798) for the regression are running here: https://github.com/tenstorrent/tt-xla/actions/runs/27624365728

mvasiljevicTT (approved):
Great!

As we continue adding custom shardings, we should consider refactoring this to minimize duplicated logic. Not blocking this pr, just putting my thoughts here for future changes.

Issue #5210, opened 2026-06-15
galaxy-wh-6u: fabric "Failed to add pinning constraints" (MGD vs physical topology)


vkovacevicTT:
Three galaxy-wh-6u TP models fail at fabric setup with a topology-mapping `TT_FATAL`, before model compile. Model-independent.

## Affected models
- `llama_3_1_70b_tp_galaxy`
- `test_gpt_oss_120b_tp_dp_galaxy_batch_size_128`
- `test_gpt_oss_20b_tp_galaxy_batch_size_64`

## Error
```
critical | Fabric | Pinned ASIC position (tray_id: 1, asic_location: 1) to fabric node id
  (mesh_id: 0, chip_id: 0) from MGD not found in physical topology (topology_mapper_utils.cpp:1326)
TT_FATAL: Failed to add pinning constraints (assert.hpp:104)
ValueError: TT_FATAL @ .../tt_metal/fabric/topology_mapper_utils.cpp:1342: success
```

## Context
Adjacent tt-metal topology-mapper work: [tt-metal#45759](https://github.com/tenstorrent/tt-metal/pull/45759), [tt-metal#43384](https://github.com/tenstorrent/tt-metal/pull/43384), [tt-metal#43451](https://github.com/tenstorrent/tt-metal/issues/43451).

Jobs: [llama_3_1_70b_tp_galaxy](https://github.com/tenstorrent/tt-xla/actions/runs/27517218376/job/81328034980), [test_gpt_oss_20b_tp_galaxy_batch_size_64](https://github.com/tenstorrent/tt-xla/actions/runs/27517218376/job/81328034996)

---
**Run**: [nightly 27517218376](https://github.com/tenstorrent/tt-xla/actions/runs/27517218376) (2026-06-15)
**Commits**: tt-xla `98af4ea8` · tt-mlir `464a5f34` · tt-metal `68e82deb`

jameszianxuTT:
Seems to bee a similar failure mode (but on WH glx) that this fixed -- #5094

sshonTT:
I think we might need to apply the same change to wh-glx as well. Since bh-glx is currently in the bring-up phase, performance shouldn't be an immediate issue there. However, I do have some concerns regarding wh-glx. In any case, I'll go ahead and submit the changes.

sshonTT:
I tried the fix but still see the same failure : https://github.com/tenstorrent/tt-xla/actions/runs/27626837730/job/81699851732#step:15:126

[...]

PR #5036, opened 2026-06-01
Wan14b tests

ppadjinTT:
## TL;DR

tests for components of Wan14b pipeline

### Dit
- compiles with bad PCC on QB2
- needed additional patch `_patch_wan_time_embedder_dtype_probe` because of the [pytorch uplift](https://github.com/tenstorrent/tt-xla/pull/4543/) that bumped torch version from 2.9 to 2.10 and added explicit dynamo patches
### Vae encoder and decoder
- work on singlechip, hang when sharded

### Umt5 text encoder
- works sharded

## Requirements for merging this PR
[This](https://github.com/tenstorrent/tt-mlir/commit/0590b60e3a32b443004613222ab00bd77a10005a) tt-mlir commit needs to be uplifted to tt-xla to enable vae decoder

PR #5241, opened 2026-06-16
[Nightly][Galaxy] Force FABRIC_1D on WH galaxy and add per-job timeout

sshonTT:
### Ticket
1. [Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/5210#event-26800293987), 
2. [BlackHole Galaxy nightly experimental fail](https://github.com/tenstorrent/tt-xla/actions/runs/27597367074/job/81590486274)

### Problem description
1. it hits the same TopologyMapper pinning failure (#5210) that the Blackhole path already handles, and the (4,8) parent-mesh fix alone was not enough.
2. Also, since #5036, bh nightly fails due to timeout.

### What's changed
1. Extend the 32-device FABRIC_1D workaround to the Wormhole galaxy
2. Add a per-job 'timeout' (minutes) override in call-test.yml, honored by the bh_galaxy experimental preset (420). The duration-based estimate falls back to 240 because most bh_galaxy tests lack durations, and the deepseek e2e outlier would otherwise inflate the 3x margin.

### Checklist

### DIFF
--- a/.github/workflows/test-matrix-presets/basic-test-nightly-experimental.json
+++ b/.github/workflows/test-matrix-presets/basic-test-nightly-experimental.json
@@ -1,3 +1,3 @@
 [
-  { "runs-on": "galaxy-bh",           "name": "run_torch_bh_galaxy",   "dir": "./tests/torch",                            "test-mark": "nightly and bh_galaxy", "forked": true }
+  { "runs-on": "galaxy-bh",           "name": "run_torch_bh_galaxy",   "dir": "./tests/torch",                            "test-mark": "nightly and bh_galaxy", "forked": true, "timeout": 420 }
 ]
--- a/.github/workflows/call-test.yml
+++ b/.github/workflows/call-test.yml
@@ -372,23 +372,34 @@ jobs:
     - name: Collect Tests
       id: collect-tests
       shell: bash
+      env:
+        TIMEOUT_OVERRIDE: ${{ matrix.build.timeout }}
       run: |
         source venv/activate
         echo "Collecting tests for group ${{ matrix.build.group-id || 1}}..."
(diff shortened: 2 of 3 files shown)


sshonTT:
Triggered wh-glx-6u benchmark test : https://github.com/tenstorrent/tt-xla/actions/runs/27626837730
bh galaxy nightly experimental test : https://github.com/tenstorrent/tt-xla/actions/runs/27627005603

.github/workflows/test-matrix-presets/basic-test-nightly-experimental.json
```diff
@@ -1,3 +1,3 @@
 [
-  { "runs-on": "galaxy-bh",           "name": "run_torch_bh_galaxy",   "dir": "./tests/torch",                            "test-mark": "nightly and bh_galaxy", "forked": true }
+  { "runs-on": "galaxy-bh",           "name": "run_torch_bh_galaxy",   "dir": "./tests/torch",                            "test-mark": "nightly and bh_galaxy", "forked": true, "timeout": 420 }
```
vvukomanTT, line 2: This approach introduces more parameters to test presets (which we want to avoid) and it also adds overiding of the timeout calculation logic.
Let's use a different approach for solving the timeout issue.

In `calculate_test_timeout.py` we currently do this:
1. if the duration of >= 1 tests is missing from `.test_durations` -> use the default duration
2. if no durations are missing -> use the calculated sum of test durations

Let's modify this logic so that if a duration of >=1 tests is missing we use `max(default_duration, calculated_duration)` instead of just `default_duration`.
This way we keep the existing functionality while allowing your longer tests to have longer than the default duration. 
(only thing needed is to increase the duration of at least 1 of the tests from the test suite to the desired duration in `.test_durations`, I think you've already done this)

vvukomanTT (changes requested):

Issue #5176, opened 2026-06-11
[Playground v2.5] VAE compile TT_FATAL after uplift (device_hash mismatch)


kamalrajkannan78:
## Problem description

- After the recent tt-mlir/tt-metal uplift, `tests/benchmark/test_imagegen.py::test_playground_v2_5` fails during the warmup pass when VAE is first compiled and executed. The test never reaches steady-state 

```
TT_FATAL: Enqueueing a Program across devices with different cores harvested is not supported,
unless coordinate virtualization is enabled (only enabled on Wormhole and above).
TT_FATAL @ tt_metal/impl/program/program.cpp:1927: *get_cached() == device_hash
→ RuntimeError: Bad StatusOr access: INTERNAL: Error code: 13
```

- Sanity - [jun13_check_issue_a.log](https://github.com/user-attachments/files/28921769/jun13_check_issue_a.log), [test_issue.py](https://github.com/user-attachments/files/28921768/test_issue.py)

### Root cause 

- The VAE OOMs from GroupNorm decomposition at opt_level=0; opt_level=1 enables the composite ttnn.group_norm which lets the VAE pass, but switching opt level(UNet opt 0 -> VAE opt 1) in a single session trips a device-hash mismatch 

## Repro

- Test: `tests/benchmark/test_imagegen.py::test_playground_v2_5`
- Device: N150
- Branch: `kkannan/jun2_playground_v2_5_nightly_benchmark` (PR #5044)
- CI run: https://github.com/tenstorrent/tt-xla/actions/runs/27337975698/job/80771087059
- Pre-uplift the benchmark ran cleanly.

PR #5244, opened 2026-06-16
[SDXL Lightning] Add e2e pipeline in nightly and benchmark CI

### LINKED ISSUE #5245: [SDXL Lightning] Add e2e pipeline in nightly and benchmark CI

kamalrajkannan78:
### Ticket

- Closes https://github.com/tenstorrent/tt-xla/issues/5245

### Problem description

Image-gen Model that passes bringup should be added to nightly + benchmark CI. SDXL Lightning's all 4 components are passing but the e2e pipeline is still a manually-invoked example with no CI coverage.

### What's changed
Mirrors the Playground v2.5 wiring (#5044), adapted to the SDXL-Lightning recipe (4 steps, `guidance_scale=0` → no CFG, Euler `"trailing"` scheduler, standard VAE scaling).

- **Nightly** (`tests/torch/models/sdxl_lightning/test_sdxl_lightning_pipeline.py`): self-contained e2e pipeline + test. Text encoders + UNet on TT (UNet in bf16), VAE on CPU.
- **Benchmark** (`tests/benchmark/test_imagegen.py::test_sdxl_lightning` + `benchmarks/sdxl_lightning_pipeline.py`): config-driven entry through the shared imagegen harness; warmup + steady-state passes; emits per-component metrics (`text_encoder_1_s`, `text_encoder_2_s`, `unet_step_mean_s`, `vae_s`, `cpu_overhead_s`, `e2e_latency`, `images_per_second`).
- **Matrix entry** added to `.github/workflows/perf-bench-matrix.json`.
- **VAE on CPU**: The VAE OOMs from GroupNorm decomposition at opt_level=0; opt_level=1 enables the composite ttnn.group_norm which lets the VAE pass, but switching opt level(UNet opt 0 -> VAE opt 1) trips a device-hash mismatch ([#5176](https://github.com/tenstorrent/tt-xla/issues/5176)). So keeping the VAE on CPU until the [fix](https://github.com/tenstorrent/tt-metal/pull/46959) lands.

### Internal checks

Verified per-component PCC against an fp32 CPU reference on the **real** pipeline tensors (all components on TT in one pass):

- `text_encoder` ≈ 0.99999, `text_encoder_2` ≈ 0.9973 (hidden) / 0.9995 (pooled), `unet` 0.99966–0.99997 across all 4 steps — all ≥ 0.99.
- `vae`: not measurable on TT for real inputs (the opt-level switch noted above), but its standalone component test passes PCC > 0.99 on random inputs at `opt_level=1`, and since it runs on CPU here its output already matches the reference. - [pcc_check.zip](https://github.com/user-attachments/files/29033944/pcc_check.zip)

### Checklist
- [x] Validate benchmark test on N150 - https://github.com/tenstorrent/tt-xla/actions/runs/27634277652
- [x] Validate benchmark test on P150 - https://github.com/tenstorrent/tt-xla/actions/runs/27749477316
- [x] Validate nightly test on N150 by local run - [nightly_local_run.zip](https://github.com/user-attachments/files/29033999/nightly_local_run.zip)

### DIFF
--- a/tests/benchmark/test_imagegen.py
+++ b/tests/benchmark/test_imagegen.py
@@ -39,7 +39,7 @@ def test_imagegen(
     """Run a text-to-image benchmark with the given configuration.
 
     Args:
-        build_pipeline_fn: Callable returning ``(pipeline, generate_fn, save_fn)``;
+        build_pipeline_fn: Callable returning ``(pipeline, generate_fn)``;
             see ``benchmark_imagegen_torch_xla``.
         model_info_name: Model name for identification and reporting.
         output_file: Path to save benchmark results as JSON.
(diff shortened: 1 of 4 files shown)


blozano-tt:
Can we leverage LFC, to avoid downloading 7GB of SDXL weights in every run?

See AI generated PR here for reference, but don't fully trust it: https://github.com/tenstorrent/tt-xla/pull/5248

kamalrajkannan78:
Thanks @blozano-tt, great catch — agreed that cold-downloading the SDXL-Base weights on every shared-runner run is wasteful, and thanks for the reference PR as well.

Digging in, fully routing these weights through the LFC isn't a small change in this PR. The SDXL-Base components (`text_encoder`, `text_encoder_2`, `vae`, tokenizers) are loaded inside the `tt_forge_models` submodule, which always loads them via `from_pretrained(<hardcoded HF repo id>)` — i.e. straight from HuggingFace. `from_pretrained` can take either a repo id or a local directory, so to read the LFC copy instead, the submodule loader would need to accept a local path and pass it through. That means a change in the `tt_forge_models` submodule, a submodule bump here, and a re-run of the (heavy) benchmark + nightly jobs to validate — which would delay merging this high priority demo PR.

Also, Playground v2.5 (#5044) loads its weights the same way and has the exact same issue, so the fix really belongs in one shared place rather than just here.

Given that, I'd prefer to handle the LFC wiring for both models together in a dedicated follow-up PR, so this one can land now. is it okay for you?

vvukomanTT (approved):

nsumrakTT (approved):

tests/benchmark/test_imagegen.py
vkovacevicTT: nit: It doesn't return save_fn, we moved it to utils.

vkovacevicTT (approved):
Nice!

Please run benchmark on p150 and disable it if it fails (set "runs-on": "n150" for the model entry in `.github/workflows/perf-bench-matrix.json`)

kamalrajkannan78:
> Nice!
> 
> Please run benchmark on p150 and disable it if it fails (set "runs-on": "n150" for the model entry in `.github/workflows/perf-bench-matrix.json`)

model is passing on P150 too - https://github.com/tenstorrent/tt-xla/actions/runs/27749477316/job/82099380130

sdjukicTT (approved):

PR #5261, opened 2026-06-17
[Test] Blackhole galaxy simple test

vkovinicTT:
### What's changed
We were lacking bh galaxy tests in CI, so I added one small test in nightly just so we can track whether everything works across different platforms.


vvukomanTT (approved):

vvukomanTT:
Please run this in the CI and validate that it works currently.

nsumrakTT (commented):
What is the difference running this test on bh-galaxy comparing to any other blackhole setup eg. p150 as it doesn't use any gh specific feature?

jameszianxuTT:
> What is the difference running this test on bh-galaxy comparing to any other blackhole setup eg. p150 as it doesn't use any gh specific feature?

BH galaxy is prone to failures like #5094 due to unchecked fabric bugs. Just opening mesh device may regress and block us from doing anything on device.

nsumrakTT (approved):
@jameszianxuTT Ok, thanks for explanation.

sdjukicTT (approved):

Issue #5247, opened 2026-06-16
Prefill last-token select recompiles at runtime on the fused (device-sampling) path


kmabeeTT:
**Status:** likely already addressed by #5129 (see below) — filing mainly to give that work a
ticket + a standalone repro and to have ticket to track it landing in tt-shield CI.

## What / why
On the fused / device-sampling path (`cpu_sampling=False`), engine init only precompiles the **decode** shape (`num_tokens==1`). `execute_model` selects the prefill last token with an eager gather **outside** `decode_postprocess` for `num_tokens > 1`, and that select graph is a distinct graph **per prefill token bucket** — so it JIT-compiles at runtime on the first prefill of each new length, instead of at init. It's a regression from #4789 ("Reduce vLLM decode graphs from 5 to 2"), which dropped the unconditional `_precompile_select_hidden_states()` for this path. Bucket-keyed and one-time per length (≤5 buckets), `cpu_sampling=False` only; the sharp edge is that with `VLLM_XLA_CHECK_RECOMPILATION=1` it's a hard crash.

Root cause in code (`integrations/vllm_plugin/vllm_tt/model_runner.py`):
- `_precompile_decode_postprocess`: `for num_tokens in [1]:` — only the decode shape is precompiled.
- `execute_model` (fused branch, `num_tokens > 1`): `hidden_states[batch_idx, logits_indices, :]` selects the prefill last token outside `decode_postprocess` — a distinct graph per bucket, not covered above. (Comment there claims "no extra compiled graphs needed" — not true.)

## How it was spotted
A long-prompt request on a forge server stalled/timed out on first use. Counting MLIR module dumps (`TTXLA_LOGGER_LEVEL=DEBUG`) split by phase showed the recompiles happened in the first request, not init; diffing the graphs' `@main` signatures showed they were all the last-token **select** (`(idx, 1×BUCKET×hidden) -> (idx, 1×1×hidden)`), not attention. `cpu_sampling=True` (which still precompiles the select for all buckets) showed zero recompiles — isolating it to the fused-path select precompile.

## Repro (standalone `vllm serve` + client)
Committed (cherry-pickable): `examples/vllm/Llama-3.2-3B/recompile_repro_service.sh` + `recompile_repro_client.py` — branch `kmabee/vllm-prefill-select-recompile-repro`, commit `81f6d5802a1cec3a1d6206a78e0fa76a9db586bc`.

```
# terminal 1 — stock (pre-fix) behavior:
[...]

PR #5270, opened 2026-06-18
Add vLLM regression test for prefill graph recompiles at serving time

kmabeeTT:
### Ticket
#5247

### Problem description
- There is no regression test guarding that prefill graphs are precompiled at engine init rather than JIT-compiled at serving time. When a prefill bucket isn't precompiled, the first request hitting it triggers an XLA recompile, causing a latency spike mid-serving.

### What's changed
- `tests/integrations/vllm_plugin/generative/test_prefill_recompile.py`: new test that warms up an in-process engine, then sweeps prompt lengths across every prefill token-bucket of a 4K context and asserts the XLA cached-graph count does not grow. The test re-execs itself in a throwaway child process so `VLLM_ENABLE_V1_MULTIPROCESSING=0` and the XLA computation-cache init stay confined to the child and never leak into the shared pytest worker.  See referenced ticket for before/after fix results.

### Checklist
- [x] New test added. Fails currently as expected until https://github.com/tenstorrent/tt-xla/pull/5129 is merged first. Could be cherry picked into that PR to go alongisde the fix.

### DIFF
--- a/tests/integrations/vllm_plugin/generative/test_prefill_recompile.py
+++ b/tests/integrations/vllm_plugin/generative/test_prefill_recompile.py
@@ -6,16 +6,20 @@
 Warms up, then sweeps prompt lengths across every prefill token-bucket of a 4K
 context and asserts the XLA cached-graph count does not grow — i.e. all prefill
 graphs were precompiled at init. A single transformer layer keeps compile fast.
+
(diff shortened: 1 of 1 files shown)


kmabeeTT:
merged via https://github.com/tenstorrent/tt-xla/pull/5129 instead, closing.

PR #5278, opened 2026-06-18
tracy ttrun integration

jasonmacTT:
### Ticket
Refs tenstorrent/tt-mlir#8772

## Problem description

Distributed (multi-host) runs launched via ttrun had no way to enable Tracy
profiling. tt-mlir exposes `withTracy()` / `withTracyArgs()` on
`MultiProcessArgs`, this pr wires it from tt-xla end.

## What's changed

Added one environment variable, read in `launchDistributedRuntime()`:

- `TT_DISTRIBUTED_TRACY_CMD` — on/off toggle and args in one. When set, routes
  every rank through `python -m tracy` via ttrun; contents are forwarded
  verbatim to `python -m tracy`, tokenized on whitespace.

### DIFF
--- a/pjrt_implementation/src/api/client_instance.cc
+++ b/pjrt_implementation/src/api/client_instance.cc
@@ -122,6 +122,11 @@ static tt_pjrt_status launchDistributedRuntime() {
       std::getenv("TT_DISTRIBUTED_TCP_IFACE");
   // Path to an MPI rankfile, needed for >2 hosts when using a rank binding file
   const char *rank_file_path = std::getenv("TT_DISTRIBUTED_RANK_FILE_PATH");
+  // On/off toggle for routing every rank through `python -m tracy` via ttrun
+  const char *tracy_enabled = std::getenv("TT_DISTRIBUTED_TRACY");
+  // Args forwarded verbatim to `python -m tracy`.
+  // `-r` flag is required, else rank dirs stay empty.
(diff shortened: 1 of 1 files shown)


pjrt_implementation/src/api/client_instance.cc
```diff
@@ -195,6 +200,23 @@ static tt_pjrt_status launchDistributedRuntime() {
...
+    std::vector<std::string> tracy_args;
+    if (tracy_args_env && !std::string(tracy_args_env).empty()) {
+      std::istringstream arg_stream(tracy_args_env);
+      std::string arg;
+      while (arg_stream >> arg) {
+        tracy_args.push_back(arg);
+      }
+    }
+    distributed_options.multiProcessArgs->withTracyArgs(tracy_args);
+  }
```
jameszianxuTT, line 236: LGTM! Maybe want to guard against the case where someone sets tracy args without TT_DISTRIBUTED_TRACY. 

Alternatively, we could merge this "interface" into a single env variable like TT_DISTRIBUTED_TRACY_CMD="" for enable tracy with no args, and TT_DISTRIBUTED_TRACY_CMD="-r ... other args" to pass args. I think that might be better UX

jameszianxuTT (approved):
LGTM thanks for pushing this from tt-mlir -> tt-xla side

Issue #5264, opened 2026-06-17
[DiffusionGemma] Fatal Python error: Segmentation fault


kamalrajkannan78:
### Describe the bug 

- DiffusionGemma fails with `Fatal Python error: Segmentation fault` - [jun17_diffgemma_before_fix.log](https://github.com/user-attachments/files/29061807/jun17_diffgemma_before_fix.log)

### Root cause analysis

**TL;DR:** `torch.histc` has no XLA/TTNN lowering, so torch_xla routes it to a CPU fallback. The fallback force-materializes the *in-flight, partial* XLA graph mid-stream, and compiling that partial graph segfaults inside torch_xla's `BuildStableHLOCompositePass` (in `_XLAC`).

#### 1. The trigger is `histc`

GDB backtrace at the crash (`jun18_diffgemma_gdb_info1.log`), innermost → outermost:

```
#19  at::_ops::histc::redispatch          ← histc is dispatched
#18  functionalizeFallback                ← functionalization redispatches it
#17  torch_xla::xla_fallback              ← no XLA kernel for histc → fall back
#16  at::native::cpu_fallback             ← run it on CPU
#14  at::_ops::_to_cpu::call              ← move inputs to CPU
#12  torch_xla::XLANativeFunctions::_to_cpu
     ...
#1   BuildStableHLOCompositePass::runOnOperation
#0   BuildStableHLOComposite              ← SIGSEGV
```

- `xla_fallback` / `cpu_fallback` (# 16–17) are entered **directly from `histc::redispatch`** (# 18–19) → `histc` is the op fed into the fallback, not a bystander.
- It is the **only** fallback in the crash stack:
  ```
  2 ::redispatch   2 cpu_fallback   2 xla_fallback
  ```
  (the `2` = backtrace printed twice — crash bt + `thread apply all bt` — i.e. one distinct fallback op: `histc`).
- Confirmed independently in `jun18_diffgemma_gdb_info2.log`: a breakpoint on `at::_ops::histc::redispatch` **fires** (`hit Breakpoint 1`), proving `histc` is actually executed and is the fallback entry. 

#### 2. Why the fallback crashes

[...]

PR #5280, opened 2026-06-18
Add aten.histc decomposition (one_hot + sum)

kamalrajkannan78:
### Ticket

- https://github.com/tenstorrent/tt-xla/issues/5264

### Problem description

- `diffusiongemma-26B-A4B-it` crashes with `Fatal Python error: Segmentation fault`.
- `torch.histc` has no XLA/TTNN lowering, so torch_xla routes it to a CPU fallback. To move the op's inputs to CPU, the fallback force-materializes the in-flight (partial) XLA graph mid-stream, and compiling that partial graph segfaults inside torch_xla's `BuildStableHLOCompositePass` (in `_XLAC`). Full breakdown in the root-cause analysis of #5264.

### What's changed

- Added an `aten.histc` decomposition (`one_hot + sum`) in `python_package/tt_torch/backend/decompositions.py`, so `histc` lowers statically instead of falling back to CPU — this removes the mid-stream materialization and the segfault.
- Added `tests/torch/ops/test_histc.py` covering the model case plus histogram corner cases (fractional bins, out-of-range/NaN ignored, `value == max`, negative range, data-derived range).
- After this fix the model gets further and now fails with: `error: 'sdy.all_slice' op operates on axis "_axis_0" which is already bound by a parent sdy.manual_computation op`.

### Checklist
- [x] Verify the changes through local testing in n150

### Logs

- [jun17_diffgemma_before_fix.log](https://github.com/user-attachments/files/29102629/jun17_diffgemma_before_fix.log)
- [jun18_diffgemma_after_fix.log.zip](https://github.com/user-attachments/files/29102633/jun18_diffgemma_after_fix.log.zip)
- [jun22_histc_before_fix.log](https://github.com/user-attachments/files/29214089/jun22_histc_before_fix.log)
- [jun22_histc_after_fix.log](https://github.com/user-attachments/files/29214085/jun22_histc_after_fix.log)

### DIFF
--- a/python_package/tt_torch/backend/decompositions.py
+++ b/python_package/tt_torch/backend/decompositions.py
@@ -408,6 +408,27 @@ def _get_default_decomposition_ops() -> DecompositionOpsList:
     ]
 
 
+def histc(input, bins=100, min=0, max=0):
+    """Decomposition for aten.histc via one_hot + sum."""
+    # torch.histc: equal bounds (min == max, including the 0/0 default) mean "use the
+    # tensor's own range"; if that range is degenerate (constant input), widen by 1
+    # on each side -- matching aten's reference kernel:
+    # https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/SummaryOps.cu#L331
+    if min == max:
+        lo, hi = input.min(), input.max()
+        same = (lo == hi).to(input.dtype)
+        lo, hi = lo - same, hi + same
+    else:
+        lo, hi = min, max
+    x = input.flatten()
+    # bin index of each value; clamp pulls v == hi into the last bin.
+    idx = torch.floor((x - lo) / (hi - lo) * bins).to(torch.long).clamp(0, bins - 1)
+    # values outside [lo, hi] and NaN (compares False) are not counted.
+    valid = (x >= lo) & (x <= hi)
+    onehot = torch.nn.functional.one_hot(idx, bins).to(input.dtype)
+    return (onehot * valid.unsqueeze(-1).to(input.dtype)).sum(dim=0)
+
+
 def _get_custom_decompositions() -> DecompositionTable:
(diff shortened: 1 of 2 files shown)


python_package/tt_torch/backend/decompositions.py
```diff
@@ -408,6 +408,19 @@ def _get_default_decomposition_ops() -> DecompositionOpsList:
     ]
 
 
+def histc(input, bins=100, min=0, max=0):
+    """Decomposition for aten.histc via one_hot + sum."""
+    # min == max == 0 means "use the tensor's own range" (torch.histc semantics).
+    lo, hi = (input.min(), input.max()) if min == 0 and max == 0 else (min, max)
```
vkovinicTT: Shouldn't we have more complete logic like [here](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/SummaryOps.cu#L331) where we also cover cases where `min = max` and `> 0`?

kamalrajkannan78: Thanks, great catch! Addressed — switched `min == 0 and max == 0` to `min == max` so equal nonzero bounds (e.g. `min=max=5`) also use the tensor's own range, plus the constant-input widen (±1) from the kernel you linked. Added tests for both. updated the description with latest sanity log

For context: I opened this when `histc` was the blocker; only later, on current main while debugging the persistent segfault, I find it moves one op downstream to `torch._grouped_mm` ([details](https://github.com/tenstorrent/tt-xla/issues/5264#issuecomment-4769739755)), and that the cleaner DiffusionGemma fix is routing experts through the `tt_moe` backend (#5323). Still worth landing this one though — the `histc` decomposition is general and should help other models. Thanks for the review!

vkovinicTT (approved):

sdjukicTT (approved):

Issue #5116, opened 2026-06-08
[vLLM] Per-user last-token gather index computed via bf16 matmul → wrong logits for batched users past flat-index 512


kmabeeTT:
## Problem

Greedy (`temperature=0`) decoding of **N identical prompts in one batch** does not produce identical outputs across all rows. Each row in a batch is an independent computation, so argmax decoding must be identical for every row — divergence is a correctness bug. First spotted in `tests/benchmark/test_vllm_benchmarks.py::test_vllm_benchmark[llama-3.2-3b]` locally and on Nightly CI (https://github.com/tenstorrent/tt-xla/actions/runs/27175829468/job/80224459033#step:15:123), reproducible using simpler case below.

The divergence has a clean, structural signature: the batch splits into **contiguous groups of rows** that each compute a different result (e.g. rows `0..k-1` emit one token, rows `k..N-1` emit another). It is **deterministic**, and it is **fully eliminated by enabling fp32 destination accumulation** (`fp32_dest_acc_en=True`). It is therefore a **bf16 destination-accumulation** problem in the batched prefill graph (tt-mlir compute-kernel-config family).

## Reproduction

Llama-3.2-3B-Instruct, 32 identical prompts, greedy, generating a single token (pure prefill):

```python
import vllm

PROMPT = "Here is an exhaustive list of the best practices for writing clean code:"
N = 32

llm = vllm.LLM(
    model="meta-llama/Llama-3.2-3B-Instruct",
    max_model_len=128,
    max_num_seqs=N,
    max_num_batched_tokens=N * 128,        # prefill all N prompts together
    gpu_memory_utilization=0.5,
    enable_prefix_caching=False,
    additional_config={"fp32_dest_acc_en": False},
)
sp = vllm.SamplingParams(temperature=0.0, max_tokens=1, ignore_eos=True)
outs = llm.generate([PROMPT] * N, sp)

ids = [o.outputs[0].token_ids[0] for o in outs]
for i, t in enumerate(ids):
    print(f"row {i:2d}: token {t}")
print("unique tokens:", len(set(ids)))   # bug: 2 (should be 1)
```

Observed (deterministic): rows 0–3 emit token `720`, rows 4–31 emit token `320`; `unique tokens = 2`. With `fp32_dest_acc_en=True` all 32 rows emit `720` (`unique tokens = 1`). Full captured output: https://gist.github.com/kmabeeTT/dd7524de473cd8158040fc860886cafd

[...]

PR #5283, opened 2026-06-18
[vLLM] Enable chunked prefill: decouple buckets + runtime chunked SDPA (#4986)

### LINKED ISSUE #4986: Enable chunked prefill in vllm_tt (DRAM savings + tail latency + flat prefill compile time)
## Request

Make chunked prefill actually take effect on the TT path, and — as part of the same change — decouple prefill **precompile** from `max_model_len` so engine-init/compile time stops scaling with context length.

Three coupled wins:
1. **Runtime scheduling**: slice long prefills into chunks and interleave with decodes → bounded tail latency under mixed load.
2. **Compile time**: once no single prefill chunk exceeds the chunk budget, the prefill precompile buckets can be capped at that budget instead of `max_model_len`, making compile time flat in context length.
3. **Peak prefill memory**: capping the prefill shape at the chunk budget also shrinks the prefill activation, which today is sized to `batch × max_model_len` and OOMs at larger batch×context.

## Motivation A — tail latency

Chunked prefill slices a long-prompt prefill into N-token chunks interleaved with decodes for other in-flight requests:
- A single 8K-token prefill no longer monopolizes the device and blocks short-prompt decodes queued behind it.
- P99 TTFT under mixed load stays bounded.
- vLLM ecosystem alignment — most models default `enable_chunked_prefill=True`; the absence is what triggers the startup warning below.

## Motivation B — compile time scales with `max_model_len`

Engine init (compile + warmup) grows roughly linearly with `max_model_len`, even for a short (~15-token) prompt at batch 1, where almost no real prefill work happens. Measured on Blackhole, Llama-3.2-3B, opt-1:

| `max_model_len` | engine init (compile+warmup) | decode (128 tok) | total |
|---|---|---|---|
| 128  | 141.6 s | 4.28 s | 173.7 s |
| 256  | 197.7 s | 4.87 s | 231.1 s |
| 512  | 240.8 s | 4.85 s | 274.4 s |
| 1024 | 288.0 s | 4.85 s | 321.7 s |

~+50 s per doubling, entirely on the init/compile side; decode is flat.

## Motivation C — peak prefill memory scales with `batch × max_model_len`

The prefill graph is compiled (and its activation allocated) at `max_num_batched_tokens`, which is `batch × max_model_len`. The backbone precompile materializes a `[batch × max_model_len × hidden]` activation, which OOMs at larger batch×context — independent of the actual (short) prompt lengths.

Observed on Blackhole (Llama-3.2-3B, bfp_bf8 KV, batch 32, `max_model_len=16384`, trace off): the run dies in `_precompile_backbone` trying to allocate a single **3 GB** DRAM buffer:

```
TT_FATAL: Out of Memory: Not enough space to allocate 3221225472 B DRAM buffer ...
  (largest free block: 390807616 B)
```

That tensor is exactly `[32 × 16384 × 3072] bf16 = 3 GB` — i.e. `batch × max_model_len × hidden`. The compiled prefill range confirms the driver: `compile_ranges_endpoints: [524288]` = `32 × 16384`. KV cache is not the issue here (29,824 tokens available, only 16,384 needed); the activation is.

Repro (branch `kmabee/llm_kv_cache_seq_len_work` @ `c8cc0e739`):
```
TT_BENCHMARK_TRACE=0 TT_BENCHMARK_BATCH_SIZE=32 TT_BENCHMARK_KV_CACHE_DTYPE=bfp_bf8 _BENCH_OPTIMIZATION_LEVEL=1 TT_BENCHMARK_MAX_MODEL_LEN=16384 TT_BENCHMARK_GMU=0.05 pytest -sv tests/benchmark/test_vllm_benchmarks.py::test_vllm_benchmark[llama-3.2-3b]
```

This is the same root cause as Motivation B (prefill sized to `batch × max_model_len` rather than a fixed chunk), and it's why batch 32 hits a memory wall where batch 1 does not — the prefill shape is 32× larger. Capping the prefill at the chunk budget makes peak prefill memory **batch-independent**, raising the achievable batch×context ceiling.

**Root cause:** the prefill precompile bucket ceiling is keyed directly off `max_model_len`, not off the chunk budget:

```python
# integrations/vllm_plugin/vllm_tt/model_runner.py:323
self.num_tokens_paddings = _get_token_paddings(
    min_token_size=self.tt_config.min_context_len,
    max_token_size=self.max_model_len,   # <-- ceiling is max_model_len
)
#
```

Each doubling of `max_model_len` appends one more (and larger) prefill graph to precompile — exactly the per-doubling step above (`compile_ranges_endpoints` tracks it 1:1: `[128]`→`[256]`→`[512]`→`[1024]`). With genuine chunked prefill no runtime prefill chunk exceeds `max_num_batched_tokens`, so buckets only need to reach the chunk budget — a fixed cap independent of context length.

## Where it's disabled / bypassed today

| Layer | Location | What it does |
|---|---|---|
| Custom scheduler | `vllm_tt/scheduler/ascend_scheduler.py:58-59` — `# if self.scheduler_config.chunked_prefill_enabled: # return super().schedule()` (commented out) | TT scheduler ignores the flag and always runs non-chunked scheduling |
| Runner assert | `vllm_tt/model_runner.py:301-304` — `assert max_model_len * max_num_seqs <= max_num_batched_tokens` | Bakes in the non-chunked assumption; forbids setting a small fixed chunk |
| Runner precompile | `vllm_tt/model_runner.py:323-326` — `_get_token_paddings(max_token_size=self.max_model_len)` | Precompiles prefill buckets up to `max_model_len` regardless of scheduler/chunk budget → compile scales with context len |
| Platform | `vllm_tt/platform.py:192-193` registers `AscendScheduler` as the TT scheduler | Puts the commented-out bypass on the load-bearing path |
| Platform (MLA-only) | `vllm_tt/platform.py:265-275` force-disables for MLA models | Doesn't fire for Llama/Qwen3/Falcon; covers DeepSeek-V2-style models |
| tt-media-server (downstream) | `tt-media-server/tt_model_runners/vllm_runner.py:44` + per-model siblings — `enable_chunked_prefill=False` hardcoded | Triggers the vLLM warning; removable once the scheduler honors the flag |

## Warnings / prints seen today

**When `enable_chunked_prefill=False` (tt-inference-server path):**
```
WARNING [arg_utils.py:1949] This model does not officially support disabling
chunked prefill. Disabling this manually may cause the engine to crash or
produce incorrect outputs.
```

**When left at vLLM's default `True` (direct `vllm serve` / tt-xla examples):**
```
INFO [scheduler.py:238] Chunked prefill is enabled with max_num_batched_tokens=128.
```
Misleading — vLLM thinks it's active, but `AscendScheduler.schedule()` runs the non-chunked path anyway.

## Scope of fix

1. Uncomment / honor the `chunked_prefill_enabled` branch in `AscendScheduler.schedule()`.
2. Relax the `max_num_batched_tokens ≥ max_model_len × max_num_seqs` assert so a fixed chunk budget smaller than the full context is allowed.
3. Cap the prefill precompile buckets at the chunk budget — change `_get_token_paddings(max_token_size=…)` from `self.max_model_len` to `max_num_batched_tokens`.
4. Drop the hardcoded `enable_chunked_prefill=False` in tt-media-server once 1-3 land.

## Acceptance

- [ ] Long prefills are sliced and interleaved with decodes at runtime (no single prefill monopolizes the device).
- [ ] Engine-init/compile time is flat in `max_model_len` for a fixed chunk budget (no per-doubling step).
- [ ] Peak prefill activation is flat in `batch × max_model_len` (the batch-32 / `max_model_len=16384` OOM above no longer occurs at a fixed chunk budget).
- [ ] Output parity vs. non-chunked on a decode model (Llama-3.2-3B / Qwen3).
- [ ] Startup no longer emits the misleading enabled/disabled mismatch.

## Image observed

`ghcr.io/tenstorrent/tt-shield/tt-media-inference-server-forge:2c0072ef33f24c4f51f6a2b4b11a1aa376c0939b_c323fef_78027618590`

alinakhanTT:
### Ticket
closes [#4986](https://github.com/tenstorrent/tt-xla/issues/4986)

### Problem description

`vllm_tt` sizes prefill precompilation and activation to `max_model_len`, and has no runtime mechanism to split a long prompt across steps. This causes three problems (issue #4986, Motivations A/B/C):

- **Compile time (B):** the precompile token-padding ladder scales with `max_model_len`, so large context lengths blow up compile time.
- **Peak prefill DRAM (C):** a full-length prefill materializes one large activation that OOMs on long contexts / large batches (e.g. batch 32 × 16384 → 6 GB activation OOM in `_precompile_backbone`).
- **Tail latency (A):** without chunking, a long prompt occupies the engine for one long prefill step instead of being interleaved.

### What's changed

Enables **chunked prefill end-to-end**, opt-in via `prefill_chunk_size`(`TTConfig`, default `0` = off). Two parts:

#### 1. Decouple prefill precompile / activation from `max_model_len`

Addresses **Motivation B + C** (Scope items 2 + 3):
- caps `max_num_batched_tokens` at the chunk size;
- sizes the precompile token-padding ladder (`_get_token_paddings`) by the per-step chunk budget instead of `max_model_len`;
- relaxes the `max_num_batched_tokens >= max_model_len * max_num_seqs` assertion under chunked prefill.

#### 2. Runtime chunked prefill

Addresses **Motivation A** (Scope item 1). Long prompts are split into block-aligned chunks and continued across steps; a cached-prefix chunk attends over the paged KV cache **on device** via a new `tt::chunked_scaled_dot_product_attention` op (no host gather, causal mask + prefix offset internal to the op, trace-compatible).

- **custom op** (`custom_ops.py`): XLA `stablehlo.custom_call` + CPU reference + `register_fake`, emitted on the cached-prefix path in `attention.py`;
- **`AscendScheduler`**: block-aligned chunk slicing, same-stage prefill batching, continuation of partial prefills, and abort-safety (a mid-prefill chunk aborted on client disconnect no longer crashes the EngineCore);
- **`model_runner`**: cached-prefix detection, persistent `chunk_start_idx` device buffer, chunked-SDPA usability gating + precompile, multi-shape recompile warm-up.

Adds `TT_BENCHMARK_PREFILL_CHUNK_SIZE` to drive it from the vLLM benchmark.

### Dependency / temp pin

Requires the tt-mlir `chunked_scaled_dot_product_attention` op (**#8789**, merged `ddbf1fa`). Commits `35272e850` and `96ff53b3b` are **TEMP pin bumps for on-device testing and MUST be reverted before merge** , once a nightly tt-mlir uplift brings `ddbf1fa` (or later) into main's pin, this PR carries no pin change. Bit-exact single- vs multi-chunk equivalence is additionally gated on SDPA fp32 accumulation (tt-mlir #8657); see Checklist.

### Checklist

Decouple slice (Blackhole, llama-3.2-3b, bfp_bf8 KV, opt1, GMU 0.05):
- [x] batch 32 × 16384, no chunk → 6 GB `_precompile_backbone` OOM (the bug); `chunk 2048` → passes, 8:52, no OOM (Motivation B + C)
- [x] batch 1 × 8192, chunk 2048 → passes (assert-relax path)
- [x] default path (`prefill_chunk_size` unset) unchanged

Runtime chunking (Blackhole @ pin `ddbf1fa`):
- [x] op lowers + runs on device; budget-decoupled test passes; coherent output
- [x] single- vs multi-chunk greedy: leading tokens match (full-sequence divergence is benign low-precision argmax drift, shared prefix grows 17 → 30 tokens as precision rises; bit-exact gated on tt-mlir #8657, not a bug)
- [x] abort-safety: 2 CPU scheduler tests pass; mid-prefill abort keeps the engine serving

Tests:
- [x] `tests/integrations/vllm_plugin/generative/test_chunked_prefill.py`
- [x] `tests/integrations/vllm_plugin/test_ascend_scheduler_abort.py`
- [x] `tests/integrations/vllm_plugin/test_chunked_sdpa_op_cpu.py`
- [x] Full nightly vLLM model test suite (vllm-model-tests-nightly.json) passes: [#28265321943](https://github.com/tenstorrent/tt-xla/actions/runs/28265321943)

Before merge:
- [ ] revert TEMP pin commits `35272e850` + `96ff53b3b` (PR carries no pin change)


kmabeeTT:
Awesome - thank you! Will be looking closer at this, and testing it out as soon as I can, to see how it compares in a few areas that I struggled with using my PoC implementation.

kmabeeTT:
Can you take a look at this bug @alinakhanTT? https://github.com/tenstorrent/tt-inference-server/issues/4326 

I haven't tested it locally yet to reproduce using your PR, but I think your version has same recent mistake I made which somewhat of a regression against previous/existing behavior where at the moment multiple users chunked prefills won't be batched together in the batch-32 (if max_num_seqs=32) prefill graph, causing prefills to be serialized across users: https://github.com/tenstorrent/tt-inference-server/issues/4326 - for example a chunked prefills size of 2k should handle 32 users prefills of ISL=2K simultaneously if they are all in the scheduler waiting to be serviced.  When I made the fix a couple days ago on my side, prefill performance jumped big, so we should include the same or similar fix here.

alinakhanTT:
@kmabeeTT thanks for pointing that out! The previous version had the same regression. Fixed in the latest push: the scheduler budget is now per_seq_chunk × max_num_seqs (instead of just per_seq_chunk), so same-stage prefills for all waiting users batch into one [max_num_seqs, chunk] prefill step rather than serializing one user per step.

Validated on-device (Blackhole, Llama-3.2-1B, batch 32, ISL 1382, chunk 2048, bfp_bf8):
Batched (fix): 18.2 s for 32 users -> 2,427 prefill toks/s
Serialized (before): 205 s for 32 users -> 216 prefill toks/s
11.3× throughput improvement, consistent with what you saw on your side.
At max_num_seqs=1 the budget is identical to before (no behavior change for single-user). The fix is in commit efc6f6a40 on this PR!!

integrations/vllm_plugin/vllm_tt/scheduler/ascend_scheduler.py
```diff
@@ -78,6 +78,21 @@ def schedule(self) -> SchedulerOutput:
...
+        num_partial_prefill_scheduled = 0
+
+        # Chunked prefill (tt-xla #4986): the num_computed_tokens stage of the
+        # first prefill request scheduled this step. We only batch prefill
+        # requests at the SAME stage in one step. Mixing a fresh request
+        # (num_computed=0) with a continuation (num_computed>0) forces the fresh
+        # request through the all-or-nothing cached-prefix attention path and
+        # corrupts it (see attention's chunked-prefix path). None until the first
+        # prefill is scheduled this step.
+        step_prefill_num_computed: Optional[int] = None
```
kmabeeTT: In general (not concerned about tests, just production code), the comments are a bit heavy — lots of multi-line explanation plus inline issue tags (~27 #4986/#4326 refs across these files). Could we trim to be more concise, and just keep issue comments in PR Desc? Inline they just go stale and crowd the logic.

tests/integrations/vllm_plugin/generative/test_chunked_prefill.py
```diff
@@ -0,0 +1,176 @@
...
+    correct (e.g. a stale page-table pointer or bad GQA broadcast), this test
+    catches it. Also sets VLLM_XLA_CHECK_RECOMPILATION=1 to verify that the
+    chunked-prefill path reuses compiled graphs after warm-up (no recompile).
+    """
+    monkeypatch.setenv("VLLM_XLA_CHECK_RECOMPILATION", "1")
+
+    chunk = 32
+    n_users = 4
+    max_tokens = 16
+    prompt = "The quick brown fox jumps over the lazy dog and then runs away quickly"
```
kmabeeTT: This doesn't actually chunk — ~16-token prompt vs chunk 32 is a single chunk per user. Good batch / no-recompile check, but not the multi-chunk path it looks like it guards. A prompt longer than the chunk would fix that.

tests/integrations/vllm_plugin/generative/test_chunked_prefill.py
```diff
@@ -0,0 +1,176 @@
...
+
+    The two paths are numerically different (different attention kernels and
+    prefill bucket shapes), so we assert the leading ``_MATCH_PREFIX_TOKENS``
+    greedy tokens match -- a real cached-prefix attention bug diverges from the
+    first generated token, while benign low-precision drift only flips a late
+    argmax (see _MATCH_PREFIX_TOKENS; full-sequence equality is gated on SDPA
+    fp32 accumulation, tt-mlir #8657).
+    """
+    single_text, single_ids = _generate(_MAX_MODEL_LEN)  # single chunk (oracle)
+    multi_text, multi_ids = _generate(128)  # several block-aligned chunks
```
kmabeeTT, line 120: Hmm.. two vllm.LLM in one process: the first's del llm doesn't release the device before the second starts, so it deadlocks when I ran it locally. Separate processes, or explicit teardown + assert the device is free between them needed?  Tried running this in nightly-vllm tests via CI yet?

integrations/vllm_plugin/vllm_tt/platform.py
```diff
@@ -369,6 +379,48 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+                    budget,
+                    scheduler_config.max_num_seqs,
+                    scheduler_config.max_num_seqs,
+                )
+                scheduler_config.enable_chunked_prefill = True
+                scheduler_config.chunked_prefill_enabled = True
+                # TT-internal per-sequence chunk cap, read by AscendScheduler
+                # (per-request chunk sizing) and TTModelRunner (bucket ladder).
+                scheduler_config.tt_prefill_chunk_size = per_seq_chunk
+                scheduler_config.max_num_batched_tokens = budget
```
kmabeeTT, line 449: Worth calling out in PR Desc that enabling chunked prefill makes max_num_batched_tokens a derived value (prefill_chunk_size × max_num_seqs) here, so a user-supplied value is ignored in that path...

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -330,11 +349,70 @@ def __init__(
...
+        # drops the prefix_chunk=False prefill graphs (only one prefill graph per
+        # token bucket compiles), ~halving backbone precompile at long context.
+        # Correctness relies on chunked SDPA at chunk_start_idx=0 matching
+        # standard causal SDPA over the chunk; keep off until bit-exact verified.
+        self._unify_first_chunk = self._chunked_sdpa_active and (
+            os.environ.get("TTXLA_UNIFY_FIRST_CHUNK", "0") == "1"
+        )
+
+        # The chunked SDPA op is the only cached-prefix prefill path now (the
+        # gather workaround was removed). So if chunked prefill would actually
```
kmabeeTT: References to "gather workaround" here and elsewhere should be removed completely (it was a brief workaround in my PoC branch, never released, nobody needs to know about :)

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -330,11 +349,70 @@ def __init__(
...
+        # chunk (budget < full context) but the page-table layout is unsupported
+        # (num_blocks_per_req % 8 != 0), there is no correct path -- fail loudly
+        # rather than silently dropping the cached prefix. Realistic (power-of-2)
+        # max_model_len always satisfies this; pick a multiple of 8*block_size.
+        if (
+            getattr(scheduler_config, "chunked_prefill_enabled", False)
+            and self.prefill_chunk_budget < self.max_model_len
+            and self.max_num_blocks_per_req % 8 != 0
+        ):
+            raise NotImplementedError(
```
kmabeeTT, line 369: Failing hard here while opt-in makes sense, but if/when we change to chunked prefill default on later (probably a good idea) this may break previously working configs, so it should probably fall back to standard prefill instead of throwing.  Something to consider later..

kmabeeTT (commented):
In general looks awesome, still looking a bit more and few things I'm curious about,  but here are few things I noticed, please take a look. Edit: PS the tt-mlir uplift landed, you can rebase and drop the custom pin.

integrations/vllm_plugin/vllm_tt/scheduler/ascend_scheduler.py
alinakhanTT: Done in aa5127c03!!

tests/integrations/vllm_plugin/generative/test_chunked_prefill.py
alinakhanTT: Bumped the prompt to ~110 tokens so each user's prefill splits into ~4 chunks, so it now actually exercises the multi-chunk cached-prefix path in the batch dimension. (12105839a)

integrations/vllm_plugin/vllm_tt/platform.py
alinakhanTT: Added a comment at the assignment and a note in the PR description: under chunked prefill max_num_batched_tokens is derived as chunk × max_num_seqs, so a user-supplied value on that path is intentionally overridden (also logged at init). aa5127c03

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Removed!

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Agreed. While chunked prefill is opt-in, the hard NotImplementedError is intentional and I've left it: you only reach it by explicitly setting prefill_chunk_size, so a loud error that tells you to pick a max_model_len that's a multiple of 8 × block_size is better than silently degrading a config you deliberately asked to chunk. Adding a silent fallback now would mask that, you'd ask for chunking and quietly not get it, with no signal.

I've added a TODO(default-on) at the guard (d5374792b) noting that when we flip the default, this should fall back to standard single-shot prefill instead of raising. Holding off implementing it until that switch so we don't change current opt-in semantics

tests/integrations/vllm_plugin/generative/test_chunked_prefill.py
alinakhanTT: Fixed the teardown: added a _shutdown() helper that calls llm.llm_engine.engine_core.shutdown() + gc.collect() before the next engine is constructed (the same idiom as sampling/conftest.py), a bare del llm defers teardown to weakref finalize and leaves the EngineCore subprocess holding the device, which is the deadlock you hit. Applied in both _generate (the two-engine equivalence test) and the batch test.  Running nightly-vllm CI job right now

integrations/vllm_plugin/vllm_tt/scheduler/ascend_scheduler.py
kmabeeTT: That change didn't really make the heavy comments much shorter :)

integrations/vllm_plugin/vllm_tt/platform.py
```diff
@@ -53,6 +53,14 @@ class TTConfig:
     batch_size: int = 1
     enable_precompile_all: bool = True
 
+    # Per-step prefill chunk size (caps max_num_batched_tokens); 0 (default) =
+    # opt-out. When set, long prompts split into chunks of this many tokens,
+    # bounding compile time + peak prefill DRAM by the chunk, not max_model_len.
+    # KNOWN LIMITATION: at max_num_seqs > 1 output may be inconsistent across
+    # requests (pre-existing batched-prefill precision issue, tt-mlir #8666);
+    # unaffected at max_num_seqs == 1.
```
kmabeeTT: I think we can remove this `KNOWN LIMITATION` 3 line comment, the tt-xla open issue is https://github.com/tenstorrent/tt-xla/issues/5116 and lately I found that it is resolved with opt-level=1, we can leave that issue open to do the tracking and remove this comment here.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -330,11 +344,61 @@ def __init__(
...
+        )
+
+        # Prototype (TTXLA_UNIFY_FIRST_CHUNK, default off): route the first/no-prefix
+        # chunk through the chunked op too (chunk_start_idx=0), dropping the separate
+        # standard-prefill graphs (~halves backbone precompile at long context).
+        # Keep off until chunked SDPA at chunk_start_idx=0 is bit-exact-verified
+        # against standard causal SDPA.
+        self._unify_first_chunk = self._chunked_sdpa_active and (
+            os.environ.get("TTXLA_UNIFY_FIRST_CHUNK", "0") == "1"
+        )
```
kmabeeTT: This is interesting optimization, but since it's a prototype and default off, and stands to improve compile time performance, I think we should strip the "unify first chunk" feature out of this PR and land it seperately as default enabled once it's proven good.

kmabeeTT (approved):
Thanks for changes! This will be great to land, just two non-blocking changes I'd request.

integrations/vllm_plugin/vllm_tt/platform.py
alinakhanTT: Done!

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Done!

tests/torch/ops/test_chunked_sdpa_op_cpu.py
```diff
@@ -0,0 +1,156 @@
...
+    # query row i is absolute position chunk_start + i; attends key j iff
+    # j <= chunk_start + i.
+    q_pos = chunk_start + torch.arange(chunk_len).view(chunk_len, 1)
+    k_pos = torch.arange(s_len).view(1, s_len)
+    mask = (k_pos <= q_pos).view(1, 1, chunk_len, s_len)
+    return F.scaled_dot_product_attention(query, gk, gv, attn_mask=mask, scale=scale)
+
+
+@pytest.mark.push
+@pytest.mark.cpu
```
mmanzoorTT: Do you want to run these tests on cpu? We are not running any vllm test on cpu.

integrations/vllm_plugin/vllm_tt/scheduler/ascend_scheduler.py
```diff
@@ -543,19 +662,42 @@ def finish_requests(
...
+        the *waiting-side* queues. A mid-prefill chunked request (status RUNNING)
+        lives in ``self.skipped_waiting``, not ``self.running``; the base handler
+        only removes RUNNING requests from ``self.running``, so an aborted one
+        would linger and crash the next ``schedule()`` (Invalid request status:
+        FINISHED_ABORTED). Removing it from both waiting queues prevents that.
         """
         if request_ids is None:
             return
+        # Normalize to a concrete list: a str is a single id, and a one-shot
+        # generator would be exhausted before super() could consume it.
```
jameszianxuTT: nit: Agree with Kyle on some of the excessive commenting like this one, which I think adds more confusion than it explains

integrations/vllm_plugin/vllm_tt/platform.py
```diff
@@ -369,6 +374,42 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
                 vllm_config.model_config.max_model_len,
                 vllm_config.scheduler_config.DEFAULT_MAX_NUM_BATCHED_TOKENS,
             )
+        elif model_config is not None and model_config.runner_type != "pooling":
+            # Non-MLA generative: chunked prefill is opt-in via prefill_chunk_size.
+            # NOTE: `chunked_prefill_enabled` is a TT-internal flag (not a vLLM
+            # field) set alongside the real vLLM `enable_chunked_prefill`.
```
jameszianxuTT: Is it possible to not use `chunked_prefill_enabled` as a name here to do something different than `enable_chunked_prefill`? If it's not possible to use the enable_chunked_prefill as a single source of truth, then maybe chunked_prefill_enabled could be named in a clearer way to document its distinction.

jameszianxuTT (approved):
left some nits but otherwise lgtm. I see the difficulty in creating a clean unit test for these custom ops without a simple torch golden, thanks for adding them where possible

integrations/vllm_plugin/vllm_tt/scheduler/ascend_scheduler.py
mmanzoorTT: I agree, comments are bit excessive.

tests/torch/ops/test_chunked_sdpa_op_cpu.py
```diff
@@ -0,0 +1,156 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```
mmanzoorTT, line 1: These tests can be moved to `tests/torch/ops/test_custom_ops.py`. Not directly related to vLLM.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2502,20 +2607,37 @@ def _model_prefill_postprocess_compiled(
...
                 torch_xla.sync()
+
+        for num_tokens in self.num_tokens_paddings:
+            logger.info("  -- num_tokens: %d", num_tokens)
+            # Sync per token count so prefill and decode graphs stay separate.
+            _run_backbone_dummies(num_tokens, prefix_chunk=False)
+            # Precompile the cached-prefix graph for prompt chunks
+            # (num_tokens > 1) so it isn't compiled on the first continuation.
+            if chunked and num_tokens > 1:
+                _run_backbone_dummies(num_tokens, prefix_chunk=True)
```
mmanzoorTT:
```suggestion
        # Precompile the cached-prefix (chunked SDPA op) graph only when the op is
        # usable; otherwise small configs would hit the ttnn page-table-stick assert.
        chunked = self._chunked_sdpa_active
        for num_tokens in self.num_tokens_paddings:
            prefix_chunk = chunked and num_tokens > 1
            logger.info("  -- num_tokens: %d", num_tokens)
            self._dummy_run(
                num_tokens, self.num_reqs_max_model_len, self.max_num_blocks_per_req, prefix_chunk
            )
            # Sync per token count so prefill and decode graphs stay separate.
            torch_xla.sync()
            if self.most_model_len is not None:
                self._dummy_run(
                    num_tokens,
                    self.num_reqs_most_model_len,
                    self.num_blocks_per_most_len_req,
                    prefix_chunk,
                )
                torch_xla.sync()
```

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -3382,6 +3504,18 @@ def _get_token_paddings(min_token_size: int, max_token_size: int) -> list[int]:
...
+    override = os.environ.get("TTXLA_PREFILL_PADDINGS")
+    if override:
+        paddings = sorted({int(x) for x in override.split(",") if x.strip()})
+        if 1 not in paddings:
+            paddings = [1] + paddings
+        if paddings[-1] < max_token_size:
+            paddings.append(max_token_size)
+        logger.info("TTXLA_PREFILL_PADDINGS override -> %s", paddings)
+        return paddings
+
```
mmanzoorTT: Why not to define config flag in `platform.py` instead of env variable? We need the prefill's num_tokens to be divisible by 32. We can assert here to fail early. The other option can be to use `_adjust_min_token()` and make them divisible by 32. 
@kmabeeTT what do you suggest?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1943,11 +2031,28 @@ def _dummy_run(self, num_tokens: int, num_reqs: int, num_blocks: int) -> None:
...
+        chunk_start_idx = None
+        if prefix_chunk:
+            # A continuation chunk uses a distinct fill_page_table at runtime;
+            # pass a separate tensor so the traced graph has matching input arity
+            # (else an extra graph recompiles at runtime).
+            fill_page_table = torch.zeros((num_reqs, num_blocks), dtype=torch.int32).to(
+                self.device
+            )
+            if self._chunked_sdpa_active:
+                chunk_start_idx = self._chunk_start_idx_dev
```
mmanzoorTT:
```suggestion
        if prefix_chunk:
            fill_page_table = torch.zeros((num_reqs, num_blocks), dtype=torch.int32).to(
                self.device
            )
            chunk_start_idx = torch.zeros((1,), dtype=torch.int32).to(self.device)
```

integrations/vllm_plugin/vllm_tt/scheduler/ascend_scheduler.py
alinakhanTT: Cut down on comments!

integrations/vllm_plugin/vllm_tt/platform.py
alinakhanTT: Checked vLLM 0.19.1: chunked_prefill_enabled is not a vLLM SchedulerConfig field at all, and enable_chunked_prefill can't be the single source of truth, it defaults True and stays True when we opt out (prefill_chunk_size == 0), so it doesn't reflect the TT opt-in. Renamed the TT-internal attribute to tt_chunked_prefill_enabled (matching the existing tt_prefill_chunk_size) across all readers in 4e718f64!!

tests/torch/ops/test_chunked_sdpa_op_cpu.py
alinakhanTT: Moved to tests/torch/ops/test_chunked_sdpa_op_cpu.py. Kept it as a standalone file rather than folding into test_custom_ops.py since those are on-device run_op_test cases and these target the op's CPU reference branch

tests/torch/ops/test_chunked_sdpa_op_cpu.py
alinakhanTT: tt::chunked_scaled_dot_product_attention has two branches, the xla branch that lowers to the ttnn kernel, and a hand-written CPU reference branch (custom_ops.py:964) that gathers the paged K/V and runs offset-causal SDPA. These tests pin that reference math (paged gather, GQA broadcast, chunk_start offset mask, scale/operand order) against an independent F.scaled_dot_product_attention. It's cheap (~ms) and guards the branch that doubles as the device-equivalence oracle

alinakhanTT: Happy to add the on-device equivalent test if needed

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Applied

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: The loop is already simplified (dropped the prototype branch)! I didn't take the suggestion exactly because it runs only prefix_chunk=True for num_tokens > 1, which compiles just the cached-prefix graph and skips the standard first-chunk prefill graph. At runtime the first chunk (num_computed == 0) takes the standard path, so it would recompile on the first request. The current version compiles both (standard + cached-prefix) per bucket on purpose

integrations/vllm_plugin/vllm_tt/model_runner.py
kmabeeTT: Ah I didn't spot this somehow, it's a nice experimental feature and I considered something like it too a while back (when exploring compile time improvements, but tradeoff with higher TTFT wasn't compelling). I think this could be dropped until sometime later when it wants to be explored.

kmabeeTT:
(I clicked update-branch button to rebase to get onPR/onPush fixes from main that were showing up as failures here)

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Dropped the TTXLA_PREFILL_PADDINGS experiment knob!

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1998,11 +2068,27 @@ def _dummy_run(self, num_tokens: int, num_reqs: int, num_blocks: int) -> None:
...
+        fill_page_table = None
+        chunk_start_idx = None
+        if prefix_chunk:
+            # A continuation chunk uses a distinct fill_page_table at runtime;
+            # pass a separate tensor so the traced graph has matching input arity
+            # (else an extra graph recompiles at runtime).
+            fill_page_table = torch.zeros((num_reqs, num_blocks), dtype=torch.int32).to(
+                self.device
+            )
+            chunk_start_idx = self._chunk_start_idx_dev
```
mmanzoorTT: We are not using the pre-allocated tensors for dummy run. I suggest not to use it here. 
```suggestion
            chunk_start_idx = torch.zeros((1,), dtype=torch.int32).to(self.device)
```

mmanzoorTT (commented):
You have not updated `_precompile_model_fused` warmup path. Please ensure that no compilation happens during inference.

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Fixed in 18c938e10. _precompile_model_fused now precompiles the cached-prefix chunked-SDPA graph alongside the standard one, gated on _chunked_sdpa_active, mirroring what _precompile_backbone already did

I also extended tests/integrations/vllm_plugin/generative/test_prefill_recompile.py to enable chunked prefill (prefill_chunk_size=512) so this path is actually exercised: it sweeps ISLs past the chunk size and asserts the XLA cached-graph count doesn't grow after warmup. Verified on-device, it fails without this fix (+1 recompile at the first continuation chunk, ISL=1000) and passes with it (0 new graphs during the sweep)

kmabeeTT: Thanks for update!

mmanzoorTT (approved):

PR #5288, opened 2026-06-19
Fix runner label mapping logic in perf tests

vvukomanTT:
### Ticket
/

### Problem description
Users are allowed to choose an architecture for runners that don't exist in civ2. The job will wait for a runner that doesn't exist until it times out.

### What's changed
If perf benchmark test tries to be ran on qb2-blackhole or galaxy-wh-6u on shared runners, the job will throw a warning visible in the github UI and remove the tests that use that arch from the workflow (other jobs will execute normally).

### Checklist
Note:
Ran a few test runs but all are stuck waiting on cpu runners from civ2, will wait for them to finish and be successful before merging.

### DIFF
--- a/.github/scripts/filter-test-matrix.py
+++ b/.github/scripts/filter-test-matrix.py
@@ -112,23 +112,19 @@ def filter_matrix_adv(matrix, adv_filter):
     return filtered_matrix
 
 
-def update_runners(matrix):
-    """Update runner names based on shared runner flag."""
-    runner_map = {"p150-perf": "p150b", "n150-perf": "n150"}
+def update_runners(matrix, sh_runner):
+    """Resolve each test's final ``runs-on`` label and shared-runner flag."""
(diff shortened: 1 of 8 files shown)


nsumrakTT (commented):
Maybe we should make it fail if non-existent shared runner is selected not just remove and continue (I guess no one will look at error messages produced if filter job succeed)?

vvukomanTT:
I was thinking of that as well, but then if someone runs Performance Benchmark with arch = "All" or runs only let's say llama models and one of llama tests is on qb2 it will fail the entire run.

nsumrakTT:
Could we change these test jobs not to be shared-runner?

vvukomanTT:
@nsumrakTT I like that suggestion the most and I've implemented it now.
If a user chooses shared runners to be true, all of the runner types which have civ2 alternatives will execute in civ2 and those which don't will execute on civ1 runners 👍

nsumrakTT (approved):
Thanks, LGTM!

Issue #5246, opened 2026-06-16
If TT_DISTRIBUTED_CONTROLLER_HOST_NAME is not set, default it to the calling process' host's hostname


jameszianxuTT:
This is a very small QOL thing in pure tt-xla -- while multihost supports a separate controller that doesn't live on the same host as one of the workers, that use case is not being exercised.

PR #5293, opened 2026-06-19
[Qol] Set a default controller hostname when env var not set

jasonmacTT:
### Ticket
Refs #5246

### Problem description
While multihost supports a separate controller that doesn't live on the same host as one of the workers, that use case is not being exercised.

### What's changed
In `launchDistributedRuntime()`, when `TT_DISTRIBUTED_CONTROLLER_HOST_NAME`
is not set, default the controller hostname to this process's hostname via
`gethostname()`.
Small QOL change; no behavior change when the env var is set.

### DIFF
--- a/pjrt_implementation/src/api/client_instance.cc
+++ b/pjrt_implementation/src/api/client_instance.cc
@@ -11,10 +11,15 @@
 #include "api/client_instance.h"
 
 // c++ standard library includes
+#include <cerrno>
 #include <cstddef>
+#include <cstring>
 #include <filesystem>
 #include <map>
 #include <optional>
+
+// system includes
+#include <unistd.h>
 
 // tracy includes
 #include "tracy/Tracy.hpp"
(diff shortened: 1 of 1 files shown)


pjrt_implementation/src/api/client_instance.cc
```diff
@@ -175,16 +177,37 @@ static tt_pjrt_status launchDistributedRuntime() {
...
-
+  // Controller hostname: prefer the explicit env-var override, otherwise fall
+  // back to this machine's hostname.
   if (controller_host_name) {
     distributed_options.multiProcessArgs->withControllerHostname(
         controller_host_name);
+  } else {
+    char hostname[256];
+    if (gethostname(hostname, sizeof(hostname)) != 0) {
+      DLOG_F(
```
jameszianxuTT: `LOG_F(ERROR` here? I'm not sure how we can recover from this condition

jameszianxuTT (approved):
small change, lgtm with nits

pjrt_implementation/src/api/client_instance.cc
jameszianxuTT: FYI DLOGs are compiled out if you don't compile tt-xla with -DCMAKE_BUILD_TYPE=Debug -- so we guard all info/warnings/especially errors that a user may care about with `LOG_F`

PR #5322, opened 2026-06-22
[pjrt] fix prepare inputs for codegen

### LINKED ISSUE #5177: codegen export_tensors serializes only one device shard of multi-device tensors
With `export_tensors=True` (`codegen_py` backend), every multi-device input/weight is serialized to `tensors/argN.tensorbin` as a SINGLE device shard instead of the full distributed tensor. On reload the generated `utils.load_tensor` → `ttnn.to_device(mesh)` replicates that lone shard across all devices, so every TP/EP-sharded weight silently runs with device-0's slice on every device.

## Repro
One-op model; both the input and the weight are sharded over the mesh, codegen'd with `export_tensors=True`:
```python
import glob
import os

import numpy as np
import torch
import torch_xla
import torch_xla.distributed.spmd as xs
import torch_xla.runtime as xr
from torch_xla.distributed.spmd import Mesh

from tt_torch.sharding import sharding_constraint_tensor  # registers the "tt" backend

os.environ["CONVERT_SHLO_TO_SHARDY"] = "1"

xr.set_device_type("TT")
xr.use_spmd()
dev = torch_xla.device()

n = xr.global_runtime_device_count()
mesh = Mesh(np.arange(n), (1, n), ("batch", "model"))


class M(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.w = torch.nn.Parameter(torch.randn(256, 128, dtype=torch.bfloat16))

    def forward(self, x):
        return x + self.w  # w = weight, x = input


m = M().to(dev, torch.bfloat16)
xs.mark_sharding(m.w, mesh, ("model", None))  # WEIGHT sharded over the mesh
torch_xla.set_custom_compile_options(
    {"backend": "codegen_py", "export_path": "out", "export_tensors": True}
)
m.compile(backend="tt", options={"tt_legacy_compile": True})
x = torch.randn(256, 128, dtype=torch.bfloat16).to(dev)
xs.mark_sharding(x, mesh, ("model", None))  # INPUT sharded over the mesh

with torch.no_grad():
    m(x)

import ttnn  # reload the exported args

for f in sorted(glob.glob("out/**/tensors/arg*.tensorbin", recursive=True)):
    t = ttnn.load_tensor(f)
    print(
        f,
        list(t.shape),
        "nshards=",
        len(ttnn.get_device_tensors(t)),
        "numel=",
        t.volume(),
        "bytes=",
        os.path.getsize(f),
    )

```
Output (32-device mesh, `n=32`):
```
Python codegen successful. Check out/graph_0 for the results.
out/graph_0/tensors/arg0.tensorbin [8, 128] nshards= 1 numel= 1024 bytes= 2384
out/graph_0/tensors/arg1.tensorbin [8, 128] nshards= 1 numel= 1024 bytes= 2384
```
- Actual: each arg (input `x` and weight `w`, both global `[256,128]`) loads as a single `[256/n, 128]` shard (`[8,128]` here) with `nshards=1`.
- Expected: the full `[256,128]` tensor, distributed across the mesh.

Confirmed on a real model (GLM-4.7, TP-8 / EP-32): `arg2` (`...self_attn.k_proj.weight`, global `[1024,5120]`) is 1,311,056 B = one `[128,5120]` shard (~1.3 MB, not the full ~10.5 MB); `arg77` (expert weight, global `[160,1536,5120]`) is 78.6 MB = 5/160 experts. The codegen `main.py` then scores PCC **0.013** vs **0.85** for the same graph run with live `mark_sharding`.

## Suspected cause
- `pjrt_implementation/src/api/tensor.cc:133` — `PjrtTensor::move_to_host()` sets `m_runtime_tensor = std::move(tensors[0])`, collapsing the multi-device tensor to shard 0 (the SO codegen path returns this from `prepareInputTensor`, `so_loaded_executable_instance.cc:202-204`).
- `pjrt_implementation/src/api/loaded_executable_instance.cc:95` — `dumpInputs()` then `dumpTensor`s that single-shard tensor (`tt::runtime::dumpTensor` faithfully serializes only what it is handed, `tt-mlir runtime/lib/ttnn/runtime.cpp:2307`).

Might be related to #4100.

pilkicTT:
Python codegen expects the inputs to be host side in row-major layout. When we create a `PjrtTensor` from buffers we create a host tensor in row-major layout, so the `move_to_host()` call is not needed.

The reason why this call created problems when dealing with multi-chip tensors is that this method, besides pulling the data to host, is also reconstructing the buffer instances to each have their individual shard; i.e. it is splitting our one multi-chip `PjrtTensor` into multiple ones (to match the way xla represents the tensors as sets of buffer instances). So, `tensor.move_to_host(); return tensor.runtime_tensor()` will return the runtime tensor representing a single shard.

Removing the extra `move_to_host()` and updating the regression test. Regression test was expecting `tensor.volume()` to return global volume - which is wrong, it returns the per-shard volume. Also, adding explicit sync so that we make sure everything is exported before verifying the artifacts.

Closes #5177

### DIFF
--- a/pjrt_implementation/src/api/so_loaded_executable_instance.cc
+++ b/pjrt_implementation/src/api/so_loaded_executable_instance.cc
@@ -199,8 +199,6 @@ SOLoadedExecutableInstance::prepareInputTensor(
 
   // SO (Python codegen) execution expects host row-major inputs; the generated
   // Python takes care of any device transfer and layout conversion itself.
-  tensor.move_to_host();
-
   return tensor.runtime_tensor();
 }
 
--- a/pjrt_implementation/inc/api/so_loaded_executable_instance.h
+++ b/pjrt_implementation/inc/api/so_loaded_executable_instance.h
@@ -48,8 +48,7 @@ public:
 private:
   // Returns an input tensor constructed from the provided buffer instances,
   // prepared for execution. The generated Python (SO) entrypoint expects host
-  // row-major inputs, so the tensor is moved to host (and untilized) before
-  // being returned.
+  // row-major inputs.
   std::optional<tt::runtime::Tensor>
   prepareInputTensor(const std::vector<BufferInstance *> &arg_buffers,
                      tt::runtime::Device device, size_t num_devices,
--- a/tests/torch/test_codegen_export_tensors_sharding.py
+++ b/tests/torch/test_codegen_export_tensors_sharding.py
@@ -49,17 +49,8 @@ class Model(torch.nn.Module):
         return x + self.w
 
 
-_XFAIL_REASON = (
-    "codegen with export_tensors=True does not preserve TP/EP sharding: "
-    "single shard is serialized to tensors/arg*.tensorbin instead of the "
-    "full distributed tensor, so a sharded tensor is silently replicated across the "
-    "mesh - https://github.com/tenstorrent/tt-xla/issues/5177"
-)
-
-
 @pytest.mark.push
 @pytest.mark.dual_chip
-@pytest.mark.xfail(reason=_XFAIL_REASON)
 def test_codegen_export_tensors_preserves_sharding():
     """Codegen export_tensors must serialize full sharded tensors, not one shard."""
     import ttnn
@@ -95,6 +86,10 @@ def test_codegen_export_tensors_preserves_sharding():
     with torch.no_grad():
         model(x)
 
+    # We need to sync here - in order for everything to be exported
+    # before we proceed with verifying the exported artifacts.
+    torch_xla.sync(wait=True)
+
     tensorbins = sorted(
         glob.glob(
             os.path.join(export_path, "**", "tensors", "arg*.tensorbin"),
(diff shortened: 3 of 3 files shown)


nvukobratTT (approved):

pjrt_implementation/src/api/so_loaded_executable_instance.cc
```diff
@@ -199,8 +199,6 @@ SOLoadedExecutableInstance::prepareInputTensor(
 
   // SO (Python codegen) execution expects host row-major inputs; the generated
   // Python takes care of any device transfer and layout conversion itself.
-  tensor.move_to_host();
```
sgligorijevicTT, line 202: Flagging to @amilovanovicTT and @svuckovicTT , the point of this change was to guarantee Host RM tensors to Codegen, is it valid to just drop it?

pilkicTT: the above API:
```
PjrtTensor::from_pjrt_buffers(
      arg_buffers, m_executable_image->getDevicesMeshShape(), *strategy);
 ```
 is already creating a host tensor, so the move to host is not needed

svuckovicTT (approved):

mrakitaTT (approved):

PR #5268, opened 2026-06-17
move cpu_compile_only to experimental nightly

jameszianxuTT:

PR #5326, opened 2026-06-22
[CI] Refresh checked in syst descs before nightly cpu compile only tests

jasonmacTT:
### Ticket
Refs #4996 

### Problem description

The CPU compile-only test (tests/cpu_compile_only/test_cpu_compile_only.py) validates against checked-in system descriptors (.ttsys files). When tt-mlir's schema changes, the runtime rejects descriptors with mismatched schema hashes, causing the test to fail in nightly.

### What's changed

Add schedule-refresh-system-descriptors.yml: a workflow that automatically regenerates system descriptors on real hardware (one per arch: n150, n300, n300-llmbox, p150), compares schema hashes against checked-in versions, and opens an auto-merging PR when the schema drifts. This runs 3 hours before nightly, to allow he CPU compile-only test compares against non-stale descriptors.


jameszianxuTT:
Small note here -- once this workflow has been merged and soaked for a while, we should probably disable running on-pr (or most of it) in the autogenerated changeset to save resources.

jasonmacTT:
Previously, the nightly test for cpu-compile-only failed in nightly due to a mismatch in the schema hash and as such has been moved to nightly experimental as per [#5268](https://github.com/tenstorrent/tt-xla/pull/5268). The drafted workflow aims to alleviate these failures by creating a workflow that can generate fresh system descs some time before nightly.

The drafted workflow [here](https://github.com/tenstorrent/tt-xla/actions/runs/27856652855/workflow) performs the following:

1. Generate the system descriptors per arch on TT hardware, and upload them as artifacts
2. If the generation fails, the workflow will continue anyways
3. On CPU runner, download the artifacts
4. Compare the schema hash of the downloaded system descs and their respective checked-in artifacts
5. If there exists a difference, stage them for PR
6. Merge the updated system descriptors in the main repo

Then on nightly, the cpu-compile-only tests will have fresh descriptors to perform the tests.

I have tested this workflow by doing the following:

1. Checkout a branch with the scale descriptors (that failed nightly)
2. Add a fresh system desc for n300-llmbox
3. Run cpu-compile-only tests against stale descriptors + fresh n300-llmbox (Expect failure except for n300-llmbox)
4. Run the workflow that will generate fresh descriptors and merge them into the branch
5. Run the cpu-compile-only tests again (Expect passing tests)

Results:
- Running the cpu-compile-only test against stale descriptors: [run](https://github.com/tenstorrent/tt-xla/actions/runs/27845780995) (Note: All failed except for n300-llmbox, as expected)
- Running the workflow to generate fresh system descs: [run](https://github.com/tenstorrent/tt-xla/actions/runs/27856652855)
- PR that merged in fresh system descs: [#5303](https://github.com/tenstorrent/tt-xla/pull/5303) (Note: An updated n300-llmbox was not merged in, since I checked-in a fresh one before running the workflow)
- Running the cpu-compile-only test against fresh descriptors: [run](https://github.com/tenstorrent/tt-xla/actions/runs/27869174078) (All passed, as expected)

.github/workflows/schedule-refresh-system-descriptors.yml
```diff
@@ -0,0 +1,188 @@
...
+      dockbuild: ci
+
+  build-ttxla:
+    needs: build-image
+    uses: ./.github/workflows/call-build.yml
+    secrets: inherit
+    with:
+      docker_image: ${{ needs.build-image.outputs.docker-image }}
+      docker_image_manylinux: ${{ needs.build-image.outputs.docker-image-manylinux }}
+      build_types: release,explorer
```
jameszianxuTT, line 41: nit: don't think we need explorer for this, though it probably doesn't matter

jameszianxuTT (approved):
LGTM, thanks for making this workflow.

nsumrakTT (commented):
System desc can only change when mlir changes are uplifted. So, I guess tt-mlir->tt-xla uplift is the right time to do this, not the schedule.

jasonmacTT:
@nsumrakTT I have updated the workflow file to detect when MLIR changes are uplifted by checking the hash in ```third_party/CMakeLists.txt```. It checks out HEAD and one previous and compares the MLIR hashes for any difference. Upon difference, it will perform the refresh, otherwise skip. This workflow would run on push to main and the check against uplift runs on CPU only, so it is not resource heavy in the case it does not need to refresh.

I have tested this workflow by "mocking" an uplift by making a commit with a change to the hash in ```third_party/CMakeLists.txt```.

Run that performed the refresh upon push: https://github.com/tenstorrent/tt-xla/actions/runs/28960390554
Workflow file: https://github.com/tenstorrent/tt-xla/actions/runs/28960390554/workflow

Notes: workflow file still in draft version, need to change minor things (e.g. change push to jmac/cpu-compile-only to main).

Additional thoughts: 
1) Is there a simple way to directly run the refresh _exactly_ on uplift? If so, that removes the need to check for HEAD and previous.
2) Upon some failure of updating the system descriptors, should the workflow retry up to N times? (In hindsight, I should probably add some control flow to recognize a manual running and not check MLIR hash)
3) The bash code for detection follows the regex pattern as seen in, ```.github/scripts/uplift/detect_uplift.py```, would it be preferred to use this script more directly in the workflow, to consistently match how uplift detection could change in the future?

nsumrakTT:
@jasonmacTT 
Yes there is, but you need compiled xla (with new version of tt-mlir) to generate system descriptor. Then again you need another pr to update it. Maybe it is better to go step back and do the daily scheduled uplift (as you already did) than to attach it to existing processes, as simpler process is easier to maintain.
 
I remember resistance for bringing ttsys files into repo, and looking on complexity that emerged I would say that from my point of view the best way forward is actually to go all the way back and delete system descriptors from the repo. There must be better solutions than introducing new uplift process (Use ttsim? Use real hw machine? Generate descriptor on demand? Store as repo variable? ...)

AleksKnezevic:
@nsumrakTT, the idea behind having them in the repo is that they're available for developers to run tests without having access to hardware. If we use a repo variable or some other store, we'd need to version it somehow, which git already does for us. 

I would propose we keep the schedule flow as it's least complicated and provides versioned storage.

.github/workflows/manual-test.yml
```diff
@@ -108,12 +108,3 @@ jobs:
...
       dump_irs: ${{ inputs.dump_irs }}
-
-  process-irs:
-    if: always() && !cancelled()
-    uses: ./.github/workflows/call-process-dumped-irs.yml
-    needs: [ build-image, test ]
-    secrets: inherit
-    with:
-      docker_image: ${{ needs.build-image.outputs.docker-image }}
-      dumped_irs_run_id: ${{ github.run_id }}
```
nsumrakTT, line 119: please revert changes to this file

.github/workflows/schedule-refresh-system-descriptors.yml
```diff
@@ -0,0 +1,188 @@
...
+      - name: Save system descriptor
+        shell: bash
+        run: |
+          source venv/activate
+          cd tests/cpu_compile_only
+          python save_system_desc.py system_descs/${{ matrix.arch }}
+          ls -la system_descs/
+
+      - name: Upload descriptor
+        uses: actions/upload-artifact@v4
```
nsumrakTT: Sorry, now I see depreciated actions use upload-artifact@v6

.github/workflows/schedule-refresh-system-descriptors.yml
```diff
@@ -0,0 +1,188 @@
...
+        run: |
+          gh run download ${{ needs.build-ttxla.outputs.artifacts_run_id }} \
+            --repo ${{ github.repository }} \
+            --dir wheels \
+            --name xla-whl-explorer${{ needs.build-ttxla.outputs.artifact_suffix }}
+          source venv/activate
+          uv pip install wheels/*.whl
+
+      - name: Download generated descriptors
+        uses: actions/download-artifact@v4
```
nsumrakTT: Also depreciated use download-artifact@v8

nsumrakTT (approved):
Thanks, sorry for the hassle. Please just upgrade versions on depreciated actions before merge

PR #5346, opened 2026-06-23
[CI] Op-by-op workflow

nsumrakTT:
### Ticket
https://github.com/tenstorrent/tt-forge/issues/954
https://github.com/tenstorrent/tt-forge/issues/952

### Problem description
With https://github.com/tenstorrent/tt-mlir/pull/8674 merged and uplifted to tt-xla it is possible to create op-by-op workflow.

### What's changed
- Add op-by-op-infra to explorer wheel.
- Add workflow call-process-dumped-irs.yml
- Modify manual-test.yml to perform process dumped irs and add single test json

After workflow is tested and optimized it will be added to the nightly run for Saturday runs.

### Checklist
Workflow run within manual test run (all resnet50 tests):
https://github.com/tenstorrent/tt-xla/actions/runs/28372657614/job/84061743928
Single test run:
https://github.com/tenstorrent/tt-xla/actions/runs/28379137310


nvukobratTT (approved):

.github/workflows/manual-test.yml
```diff
@@ -108,3 +108,12 @@ jobs:
       torchxla_build_run_id: ${{ inputs.torchxla_build_run_id }}
       forge_models_override: ${{ inputs.forge_models_override }}
       dump_irs: ${{ inputs.dump_irs }}
+
+  process-irs:
```
vvukomanTT, line 112: I would prefer if this job was optional because it lasts quite a while.
We can add an input flag for it and set it to false by default.

nsumrakTT: It is kinda optional because inside the job it checks if there are ir dumps first. If there are not, it will just skip everything.
This is because there is plan to use --dump-irs=on-fail so it will dump only if test fails.

vvukomanTT (approved):

PR #5356, opened 2026-06-24
Add PCC check to DeepSeek-V4 streaming test

sshonTT:
### Ticket
N/A

### Problem description
The streaming test previously ran to completion as a functionality smoke test without verifying numerical correctness. Add a CPU-reference PCC check for both prefill and decode, computed without ever re-loading weights to host.

### What's changed
The reference is computed layer-outer (one weight load per layer): while each block's weights are resident on host during the streaming pass, its CPU forward is advanced through prefill and every decode step in time order, accumulating that block's KV cache. This matches the device's time-outer run (the only cross-time state is the per-block KV cache) and requires teacher-forced decode inputs, which is fine since PCC compares CPU vs device under identical inputs.

Decode PCC is near-constant across steps, so compare only prefill + 2 decode steps. Log top-1/top-5 token agreement alongside PCC to tell a benign tail-drift (low PCC, argmax still matches) from a wrong prediction, and add a greedy free-run generation pass that prints decoded continuations so output coherence is visible in CI (the PCC pass is teacher-forced and never shows real continuations; its assert is deferred so generation runs even on a PCC failure). Prefill accumulates more dense-vs-sparse MoE drift over the full depth, so hold it to a 0.91 bar; keep decode at 0.95.

### Checklist

### DIFF
--- a/tests/torch/models/deepseek_v4/test_deepseek_v4_e2e_streaming.py
+++ b/tests/torch/models/deepseek_v4/test_deepseek_v4_e2e_streaming.py
@@ -20,13 +20,17 @@ The flow, top to bottom:
     model = torch.compile(model, backend="tt")
     prefill + decode loop
 
-This is a run-to-completion smoke test: it generates tokens and prints the
-decoded continuations, but does not check PCC against a CPU reference.
+A CPU-reference PCC check runs without re-loading weights: each block's CPU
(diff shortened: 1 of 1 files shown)


sshonTT:
Experimental nightly test : https://github.com/tenstorrent/tt-xla/actions/runs/28072622666

AleksKnezevic (approved):

PR #5291, opened 2026-06-19
[Janus-Pro] Add e2e text-to-image pipeline (1B + 7B): nightly + benchmark

ashokkumarkannan1:
### Ticket

- https://github.com/tenstorrent/tt-xla/issues/5233

### Problem description

- Janus-Pro passed component bring-up — `image_token_step`, `gen_img_embed` and `gen_vision_decode` for both the 1B and 7B variants (added in #4810) — but there was no end-to-end text-to-image pipeline wired into nightly + benchmark CI. 
- This mirrors the image-gen pattern introduced in #5044 (Playground v2.5) and extended by SDXL-Lightning: a model whose components pass bring-up should get an e2e pipeline in nightly + a perf benchmark.

### What's changed

Chains the three validated Janus-Pro components into an end-to-end text-to-image pipeline on TT, following the deepseek-ai/Janus `generation_inference.py` reference loop (classifier-free guidance + multinomial sampling, 576 image tokens, `parallel_size=1`) and the `playground_v2_5` / `sdxl_lightning` e2e-pipeline conventions.

- **Nightly** (`tests/torch/models/janus_pro/test_janus_pro_pipeline.py`): self-contained — pipeline class + `run_janus_pro_pipeline` helper + tests inline (no import from `examples/`). `test_janus_pro_pipeline` (1B) and `test_janus_pro_pipeline_7b` (7B). Markers: `nightly`, `model_test`, `single_device`, `large`, `record_test_properties`. The autoregressive step uses a `StaticCache` pre-allocated to `prompt_len + 576` so the prefill and decode graphs compile **once** and every later decode step is a cache hit (a `DynamicCache`, as in the original script, recompiles each step because the KV length changes — hours on TT). Components move to TT via `model.compile(backend="tt") + model.to(xla_device())` and are evicted back to CPU after use, keeping at most one model resident on TT DRAM.
[...]

Issue #5234, opened 2026-06-16
[FLUX.2- dev] Wire e2e pipeline into nightly and benchmark CI


ashokkumarkannan1:

ashokkumarkannan1:
https://github.com/tenstorrent/tt-xla/pull/5359 closes this

PR #5359, opened 2026-06-24
[FLUX.2] Add e2e text-to-image pipeline: nightly + benchmark

ashokkumarkannan1:
### Ticket

- https://github.com/tenstorrent/tt-xla/issues/5234

### Problem description

- FLUX.2-dev passed component bring-up (text encoder, transformer, VAE decoder — added in #5317) and the full e2e text-to-image pipeline runs on Tenstorrent, but there was no e2e pipeline wired.
- This mirrors the image-gen pattern introduced in #5044 (Playground v2.5) and extended by SDXL-Lightning (#5244) and Janus-Pro (#5291): a model whose components pass bring-up should get an e2e pipeline in nightly + a perf benchmark.
- FLUX.2 is the **first multichip (tensor-parallel) image-gen model** to be wired in: the ~24B Mistral3 text encoder and ~32B Flux2 transformer are SPMD-sharded across the mesh's `model` axis (contraction-parallel degree 4); the VAE is replicated. It targets the **4-chip Blackhole QuietBox (`qb2-blackhole`)**.

### What's changed

Wires the full FLUX.2-dev pipeline (the standard diffusers `Flux2Pipeline` orchestrates; every compute module runs on TT via `torch.compile(backend="tt")`, tensor-parallel sharded with the same SPMD shard specs as the component tests) into nightly + benchmark CI, following the `playground_v2_5` / `sdxl_lightning` e2e conventions.

- **Nightly** (`tests/torch/models/flux2/test_flux2_pipeline.py`): self-contained — pipeline class + test inline, imports shard specs from `third_party.tt_forge_models` (no import from `examples/`). Added `@pytest.mark.qb2_blackhole` so the existing standalone-models multichip nightly job collects it; markers: `tensor_parallel`, `nightly`, `model_test`, `large`, `qb2_blackhole`, `record_test_properties`. Mesh is now selected from `MESH_SHAPES[num_devices]` (model-parallel axis fixed at degree 4; no-op at 4 chips). Memory strategy: the text encoder is placed → used → evicted before the transformer is placed, and the VAE is placed lazily at first decode, so peak DRAM ≈ max(component) rather than the sum.
- **Benchmark**: the shared diffusion harness `tests/benchmark/benchmarks/imagegen_benchmark.py` is made component-set-agnostic — a model reports only the components it runs (FLUX.2 has a single text encoder, so `te2` is omitted from the breakdown + measurements), with an optional per-step label and `mesh_shape` reporting. **SDXL/Playground output is unchanged** (verified on hardware, see table). A self-contained multichip `benchmarks/flux2_pipeline.py` enables SPMD via the shared `enable_spmd()` helper (turns on SPMD + shardy annotations, so the StableHLO handed to tt-mlir carries the `@Sharding` custom calls presharded args need) and builds the mesh from `MESH_SHAPES[num_devices]`, with per-component `_perf` timing. Entry `test_flux2` in `test_imagegen.py`. Two-pass warmup + steady-state. `optimization_level=0`.
- **Matrix**: `flux2` entry added to `.github/workflows/perf-bench-matrix.json` with `runs-on: qb2-blackhole` (4-chip). A `qb2-blackhole` standalone-models nightly job (scoped to the `flux2` dir, mark `qb2_blackhole`) added to `model-test-lb-blackhole-nightly.json`.
- **Machine placement**: the component correctness tests (#5317) run on `lb-blackhole` (the functional-nightly convention); the e2e pipeline + perf benchmark run on `qb2-blackhole` — matching the perf-matrix convention (all Blackhole multichip perf runs on qb2) and co-locating the e2e with the box it was validated on.

Validated on a 4-chip qb2 Blackhole:

| Test | Result | Time |
| --- | --- | --- |
| `test_flux2_pipeline` (nightly e2e) | ✅ passed | 20m59s (cold cache) |
| `test_flux2` (benchmark) | ✅ passed | 19m54s (warmup+steady) |

Steady-state benchmark numbers (qb2-blackhole, 4 chips, opt_level=0, 50 steps):

| metric | value |
| --- | --- |
| e2e latency | 835.8s |
| warmup (incl. compile) | 336.2s |
| text encoder | 29.8s |
| transformer step mean | 13.55s |
| VAE | 6.4s |

Both the nightly and benchmark passes produce a coherent 1024×1024 image.

### Checklist

- [x] Validate nightly e2e + benchmark on 4-chip qb2 blackhole (logs - [zip](https://github.com/user-attachments/files/29304040/flux2_pr_logs.zip))

- [x] Confirm single-chip imagegen (SDXL-Lightning) benchmark unaffected by the harness change
- [x] Perf benchmark CI run (`test_flux2`, qb2-blackhole): https://github.com/tenstorrent/tt-xla/actions/runs/28118639108


vvukomanTT (approved):

nsumrakTT (approved):

kamalrajkannan78 (approved):

tests/torch/models/flux2/test_flux2_pipeline.py
```diff
@@ -0,0 +1,217 @@
...
+        te_compiled = torch.compile(encoder_wrapper, backend="tt")
+
+        with torch.no_grad():
+            prompt_embeds = te_compiled(input_ids.to(dev), attention_mask.to(dev))
+        torch_xla.sync()
+        prompt_embeds = prompt_embeds.cpu()
+
+        # Free the 24B encoder from device before placing the 32B denoiser.
+        self.pipe.text_encoder = text_encoder.to("cpu")
+        del te_compiled, encoder_wrapper
```
jameszianxuTT, line 161: Interesting, did you confirm that this actually evicts the weights of the text encoder from device DRAM? There is a long callstack through the torch-xla to conclude in actual eviction of device weights.

jameszianxuTT (approved):
LGTM! Some minor questions

tests/torch/models/flux2/test_flux2_pipeline.py
```diff
@@ -0,0 +1,217 @@
...
+
+    pipeline = Flux2TTPipeline()
+    pipeline.setup()
+    image = pipeline.generate(
+        PROMPT, num_inference_steps=NUM_INFERENCE_STEPS, seed=SEED
+    )
+    image.save(output_path)
+
+    assert output_file.exists(), f"Output image {output_path} was not created"
+    with Image.open(output_path) as img:
```
jameszianxuTT, line 215: Are there actual accuracy tests for whole pipeline or its subcomponents somewhere?

ashokkumarkannan1: Yes — each compute subcomponent has a PCC accuracy test added in #5317, in this same directory:

- `test_text_encoder_sharded` — PCC ~0.981 (threshold relaxed to 0.98 for the TP numerical gap)
- `test_transformer` — PCC ~0.9887 (threshold 0.98)
- `test_vae_decoder` — default `ComparisonConfig` (PCC 0.99) via `run_graph_test`

The full-pipeline test here intentionally does not PCC the e2e output — a 50-step diffusion run has no single deterministic golden tensor to compare against in this harness — so it validates wiring + output dimensions, with accuracy anchored at the component level and the final 1024×1024 image visually confirmed coherent.

tests/torch/models/flux2/test_flux2_pipeline.py
ashokkumarkannan1: Yes — confirmed. The confirmation is the run completing on hardware. @jameszianxuTT 

Per chip on qb2 (~31.75 GiB usable DRAM), the two weight sets sharded degree-4 are ~12 GiB (24B encoder) + ~16 GiB (32B transformer) = ~28 GiB/chip, leaving only ~3.7 GiB for the transformer's activations at 1024×1024 over 50 steps — not enough, so keeping both resident OOMs on activation headroom. The pipeline runs to completion (qb2, 20m59s, validated on hardware), which is only possible because the encoder is actually evicted before the transformer is placed, freeing its ~12 GiB/chip.

I wanted a direct DRAM readback for a hard number but couldn't find one on this path: `torch_xla.core.xla_model.get_memory_info()` is unimplemented on the TT backend (verified — raises `RuntimeError: Bad StatusOr access: UNIMPLEMENTED: Error code: 12`; the plugin has no PJRT MemoryStats handler), and the tt-mlir runtime's `dumpMemoryReport`/`getMemoryView` need a device handle the plugin doesn't surface to Python. If you know another way to log device DRAM under the PJRT path, I'm happy to add it.

tests/torch/models/flux2/test_flux2_pipeline.py
jameszianxuTT: Thanks for clarifying!

tests/torch/models/flux2/test_flux2_pipeline.py
jameszianxuTT: Sounds good. I don't suggest adding it in this PR as this should be productized separately for tt-xla or even under a separate tool

> If you know another way to log device DRAM under the PJRT path, I'm happy to add it.

FYI tt-metal added a shared memory region publishing device counter info that any process can query at runtime. 

I have some example tooling here if you are curious - https://github.com/jameszianxuTT/memory_logger/blob/main/memory_logger.py#L94-L162 - but I'd like to add overall device memory (and host memory) checking as a thought out feature eventually and not ad-hoc in random PRs :)

tests/benchmark/benchmarks/flux2_pipeline.py
```diff
@@ -0,0 +1,217 @@
...
+            xs.mark_sharding(tensor, mesh, spec)
+        self._compiled = torch.compile(transformer, backend="tt")
+
+    def __call__(self, **kwargs):
+        moved = {
+            k: (v.to(self._dev) if torch.is_tensor(v) else v) for k, v in kwargs.items()
+        }
+        t0 = time.perf_counter()
+        out = self._compiled(**moved)
+        torch_xla.sync()
```
vkovacevicTT: Do we need this sync if we cast it to cpu?

tests/benchmark/benchmarks/flux2_pipeline.py
```diff
@@ -0,0 +1,217 @@
...
+        # so it does not inflate the denoiser's peak DRAM; place it only now.
+        if self._compiled is None:
+            vae = self._vae.to(self._dev)
+            self._compiled = torch.compile(
+                lambda z: vae.decode(z, return_dict=False)[0], backend="tt"
+            )
+        t0 = time.perf_counter()
+        out = self._compiled(latents.to(self._dev))
+        torch_xla.sync()
+        image = out.cpu() if torch.is_tensor(out) else out
```
vkovacevicTT: `image = out.cpu() if torch.is_tensor(out) else out` what's the logic behind this?

And also same as above, if we cast to cpu, do we need sync?

vkovacevicTT (approved):

tests/torch/models/flux2/test_flux2_pipeline.py
ashokkumarkannan1: thanks! will follow that seperately

tests/benchmark/benchmarks/flux2_pipeline.py
ashokkumarkannan1: Good catch — no, we don't. `.cpu()` on an XLA tensor is itself a materialization barrier: it forces the pending graph to compile+execute and blocks until the result is on host, so the explicit `torch_xla.sync()` right before it was doing no work the `.cpu()` wasn't already forcing. Removed it.

tests/benchmark/benchmarks/flux2_pipeline.py
ashokkumarkannan1: The guard was dead code — the compiled callable is `lambda z: vae.decode(z, return_dict=False)[0]`, and indexing `[0]` always yields a tensor, so the `else out` branch could never fire. Collapsed it to `image = out.cpu()`. And same as above: `.cpu()` forces the graph to execute and blocks until the result lands on host, so the preceding `torch_xla.sync()` was redundant and is removed.

PR #5302, opened 2026-06-20
Uplift third_party/tt-mlir to aae51da94d33f0de5046338fa3e8a2d663e4081c 2026-06-25

vmilosevic:
This PR uplifts the third_party/tt-mlir to the aae51da94d33f0de5046338fa3e8a2d663e4081c from June 25

  - Add missing websockets dependency now required by tracy's bundled serve_wasm.py import path.
  - Fix tracy tool detection after tt-metal's capture-release → tracy-capture rename by using the bundled common.TRACY_CAPTURE_TOOL constant (avoids "Tracy tools were not found" in clean wheel envs).

Issue #5189, opened 2026-06-12
PR-related workflows should use the manylinux wheel instead of linux wheel


acicovicTT:
To avoid the manylinux-related regressions showing up for the first time in nightly, ideally our PR-related workflows (on push, on merge) should replicate the nightly setup and also use the manylinux wheel.

nsumrakTT:
Since manylinux wheel build takes longer time we decided to use ubuntu linux wheel for PRs. It is however built in on push event but it seems it is not used there for testing.
cc @vmilosevic

acicovicTT:
@nsumrakTT this seems to be the the on push event for the `main` branch meaning that it is run once the branch is already merged into main? If that is the case, I think it's still too late because not a lot of people pay attention to that workflow. Can we make the same change in "On PR" workflow, where you cannot even merge the PR into `main` unless you have a new build of manylinux wheel?

sdjordjevicTT:
Yeah, I agree with @acicovicTT. Not sure what can break the manylinux build, but if I understand correctly, can we at least run that during the tt-metal uplift? As far as I understand, most problems come from the uplift. Please correct me if I am wrong. :)

acicovicTT:
If I understand correctly, a new build of manylinux image itself is picked up through metal uplift, and changes to the environment made between two versions of the image can break the build, like a new compiler or linker version (happened two times recently here https://github.com/tenstorrent/tt-xla/issues/5067  and here https://github.com/tenstorrent/tt-xla/issues/5172.)

nsumrakTT:
@acicovicTT @sdjordjevicTT For "normal" PR I don't think it is a good idea to run manylinux as build time is 10min longer.
I guess we can set up manylinux build for uplift, or you might consider using "Run Test" workflow with the "Build and use manylinux wheel" option checked. I'm saying this because I am not sure if additional build time is worth in the long run.

nsumrakTT:
@sdjordjevicTT @acicovicTT Forgot to mention the fact that ubuntu release wheel which is used in normal PR and development will not be tested during uplift.

acicovicTT:
[...]

PR #5362, opened 2026-06-24
[build] fix manylinux linker issues

### LINKED ISSUE #5312: [Uplift] Manylinux build failing
https://github.com/tenstorrent/tt-xla/pull/5302 More info pending

pilkicTT:
With recent upgrade of gcc toolset in `manylinux` image we've encountered library load time crashes due to `patchelf` corrupting libs created with the new version of `ld` linker. `patchelf` is used by the `auditwheel repair` command, which runs it on all of the shared libs we ship. Loading any of these libraries will cause a seg fault during load time (due to the corruption).

PR #5181 attempted a fix which caused problems with current uplift of `tt-metal` & `tt-mlir`, where the `LDFLAGS` var is consumed by recently introduced `emscripten` process during build of `tt-metal` and causing failure. The uplift is happening in #5302.

The root cause of the original problem (.so corruption) is that `patchelf` is not able to properly handle binaries produced with new versions of `ld`. This change installs the `lld` linker and sets up cmake so that we use it by default - which mitigates the issue (`tt-mlir` already chooses it whenever possible and propagates it to `tt-metal`)

As follow-up items we should enforce alignment of build toolchains in `tt-metal`, `tt-mlir` and `tt-xla`.

Closes #5312


mrakitaTT (approved):
Awesome, thanks for figuring this out!

mrakitaTT:
@nsumrakTT Can you please enable uplift PR to be tested with manylinux wheel so we can catch/test these situations easier?

nsumrakTT (approved):

nsumrakTT:
@mrakitaTT Here is discussion about subject:
https://github.com/tenstorrent/tt-xla/issues/5189

vvukomanTT (approved):

pilkicTT:
basic tests using manylinux build passed: https://github.com/tenstorrent/tt-xla/actions/runs/28110854963

PR #5363, opened 2026-06-24
[vLLM] Add prefill request-count bucketing via min_num_seqs

### LINKED ISSUE #5281: Add vLLM support for lower latency b1 prefill graphs (while decode remains b32)
Follow up from https://github.com/tenstorrent/tt-xla/issues/5269 and relates to request we just discussed offline :)

## Problem:

- batch32 prefill graphs are often ~8x-15x slower in terms of TTFT than batch1 prefill graphs
- in real world we want to reduce prefill interruption to decode by having it be as fast as possible
- we want to explore some heuristics for leverging batch-1 prefill more often (when single request arrives, or even multiple if say less than break-even number of say 16x)

## Request

- Can we explore precompiling batch-1 prefill graphs too and having vllm scheduler controllable for selecting whether b1 or b32 (max_num_seqs actually, not really 32) prefill graph is picked.
- We can then do some benchmark driven profiling to help tune the right numbers or see how this helps
- Quick impl for experimentation here is fine to start, can polish later.

mmanzoorTT:
### Ticket
closes #5281 

### Problem description
vLLM runner always generates graphs with max_num_reqs and incoming requests are batch padded to match the max_num_reqs. However, batch32 prefill graphs are often ~8x-15x slower in terms of TTFT than batch1 prefill graphs.

### What's changed
- Added request-count bucketing for prefill warmup and execution in the TT vLLM runner. The backend can now distinguish between a small prefill bucket and the max request-count bucket, preallocate the matching buffers, and compile the corresponding graph shapes instead of forcing all prefills through the max batch shape.
- Added a scheduler heuristic for b1-prefill via prefill_batch_threshold. When the pending prefill count is small, scheduling can prefer the small prefill path instead of filling a large batch with mostly wasted rows, which targets better TTFT for low-concurrency prefills.
- Added a dedicated standalone b1-prefill regression/demo test to exercise the new scheduling and prefill behaviour end to end, with explicit coverage around TTFT-oriented single-request prefill execution.

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -1722,9 +1722,6 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                         )
                     )
 
-            logger.info(f"hidden_states: {hidden_states.shape} {hidden_states}")
-            logger.info(f"logits: {logits.shape} {logits}")
-            logger.info(f"selected_token_ids: {selected_token_ids}")
             # Save hidden states (before position selection) for prompt
             # logprobs.  Only extract rows for requests that actually need
             # them, keyed by batch index, so we never copy the full
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -2236,6 +2236,12 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         ]
 
         for config in configs:
+            if config["num_tokens"] == 1 and config["num_reqs"] != self.max_num_reqs:
+                logger.debug(
+                    f"Skipping config={config} because decode path only supports max_num_reqs={self.max_num_reqs}"
+                )
+                continue
+
             logger.info(f"Compiling graph for config={config}")
             (
                 dummy_inputs,
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -482,50 +482,45 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
             num_reqs: _alloc_dev((num_reqs,), torch.int32)
             for num_reqs in {self.min_num_reqs, self.max_num_reqs}
         }
-        self._page_table_dev_max = _alloc_dev(
-            (self.num_reqs_max_model_len, self.max_num_blocks_per_req),
-            self.block_table_cpu.dtype,
-        )
-        self._page_table_dev_min = _alloc_dev(
-            (self.min_num_reqs, self.max_num_blocks_per_req),
-            self.block_table_cpu.dtype,
-        )
-        self._fill_page_table_dev_max = _alloc_dev(
-            (self.num_reqs_max_model_len, self.max_num_blocks_per_req),
-            self.block_table_cpu.dtype,
-        )
-        self._fill_page_table_dev_min = _alloc_dev(
-            (self.min_num_reqs, self.max_num_blocks_per_req),
-            self.block_table_cpu.dtype,
-        )
-        self._cache_position_dev_max = _alloc_dev(
-            (self.num_reqs_max_model_len,), self.seq_lens_cpu.dtype
-        )
-        self._cache_position_dev_min = _alloc_dev(
-            (self.min_num_reqs,), self.seq_lens_cpu.dtype
-        )
+        self._page_table_dev_max = {
+            num_reqs: _alloc_dev(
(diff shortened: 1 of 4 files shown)


integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -462,46 +462,65 @@ def _alloc_dev(shape, dtype):
             return torch.zeros(shape, dtype=dtype, device="cpu").to(self.device)
 
         self._input_ids_dev = {
-            n: _alloc_dev((self.max_num_reqs, n), torch.int32)
-            for n in self.num_tokens_paddings
+            (num_reqs, num_tokens): _alloc_dev((num_reqs, num_tokens), torch.int32)
+            for num_reqs in {self.min_num_reqs, self.max_num_reqs}
```
kmabeeTT, line 466: These three buffers (`_input_ids_dev`, `_position_ids_dev`, `_logits_indices_dev`) are keyed by `{min_num_reqs, max_num_reqs}`, but `_prepare_inputs` clamps the lookup key down to the SMEM row limit: `target_num_reqs = min(target_num_reqs, num_reqs_max_model_len)`. When `num_reqs_max_model_len < max_num_reqs`, the lookup misses → `KeyError`.

The page-table buffers already account for this (keyed by `{min_num_reqs, num_reqs_max_model_len}` / `num_reqs_most_model_len` after 99adbc280) — these three just didn't get the same treatment.

Note this isn't behind the feature flag: with `min_num_seqs` unset it resolves to `max_num_seqs`, so the clamp still runs and a plain decode hits the `KeyError` whenever `max_num_seqs` exceeds the SMEM seq limit (e.g. >128K context at `max_num_seqs=32`). On main this worked since input_ids was always sized `max_num_reqs` and decoupled from the page-table row count.

Cheap fix: key these the same way as the page tables. Could also add a constructor-time assert that every reachable `target_num_reqs` is present in each buffer dict, to turn this into a clear startup error instead of a deep KeyError.

Not a hard blocker given it needs an aggressive long-context config, but worth doing before merge since it's a few lines and isn't gated behind the feature.

Was easily reproduced in a device-free test (asserting the buffer keys cover every clamped target_num_reqs), which passes with the suggested fix.

kmabeeTT (approved):
Looks good to me - thanks for working on this, I'd been using version from last week successfully in tt-inference-server side to see some huge perf prefill perf gains - there is one thing claude helped find (potential KeyError) should probably be addressed.

integrations/vllm_plugin/vllm_tt/model_runner.py
mmanzoorTT: You are right, I had also observed this inconsistency between pre-allocated buffers creation and I planned to investigate further in a subsequent task. I have removed `target_num_reqs = min(target_num_reqs, num_reqs_max_model_len)`; so we are following original code and reply on either `self.max_num_reqs` or `self.min_num_reqs`. 

Using all reachable `target_num_reqs` will also require compiling graphs for all of them which will increase the warmup time. I want to explore where we compute either `self.num_reqs_max_model_len` or `self.num_reqs_most_model_len` and use them instead of `self.max_num_reqs`; we will keep two possible `target_num_reqs` in this way. All the buffers are pre-allocated with updated `self.max_num_reqs`. What do you think?

kmabeeTT: Thx for quick reply and filing #5416 to track the full fix — deferring is fine w/ me.

But now that the clamp is gone, #5416 wording is stale and should be updated, but also for the default case (`min_num_seqs` unset) a long-context decode no longer crashes, but it builds a batch where the page table has `max_num_reqs` rows while seq_lens only has `num_reqs_max_model_len` ... so we trade a loud crash for a silent shape mismatch, which is worse than before.

Since we're deferring the real fix, consider adding assert to fire at startup in constructor... something like:

```
  # Fail fast at startup if any per-step buffer can't be looked up by a
  # row count _prepare_inputs will request (target_num_reqs is min/max_num_reqs).
  _reachable_num_reqs = {self.min_num_reqs, self.max_num_reqs}
  for _name, _keys in {
      "input_ids": {k[0] for k in self._input_ids_dev},
      "position_ids": {k[0] for k in self._position_ids_dev},
      "logits_indices": set(self._logits_indices_dev),
      "page_table_dev_max": set(self._page_table_dev_max),
      "cache_position_dev_max": set(self._cache_position_dev_max),
  }.items():
      assert _reachable_num_reqs <= _keys, (
          f"{_name} buffer keyed by {sorted(_keys)} is missing reachable "
          f"row counts {sorted(_reachable_num_reqs - _keys)}"
      )
```

mmanzoorTT: Thanks for the suggestion. Assertion added.

kmabeeTT: Thanks!

PR #5366, opened 2026-06-24
Fix device-count mismatch for no-input graphs on multi-device runs

### LINKED ISSUE #5360: [HunyuanImage-2.1-Distilled-Diffusers] Device count mismatch: 4 vs 1 in transformer
### Describe the bug

On 4 chips, `tests/torch/models/HunyuanImage_2_1/test_transformer.py::test_transformer_sharded` fails with `Device count mismatch: 4 vs 1`
Issue originates from this [arange](https://github.com/huggingface/diffusers/blob/275869dcae4ebcfee6a80253fdabc56033335020/src/diffusers/models/transformers/transformer_hunyuanimage.py#L445) op

### Root cause 

**The op:** RoPE builds its position tables with `torch.arange(0, 64)`. That value depends on **nothing** — no model input.
**What the compiler does with it:** Because it depends on nothing, the compiler pre-computes it into its own tiny graph that **takes no arguments** — `func.func @main() -> tensor<64xf32>`.

**The buggy decision:** The plugin has a rule (from PR #5236): *"a no-input graph that produces a plain replicated value runs on 1 device."* That rule exists to stop single-device models from being wrongly spread across all chips.

But the plugin makes that decision by looking **only at the tiny arange graph in isolation**. From there it can't tell the difference between:

- a genuinely **single-device** model (should be 1 device — #5236's case), and
- our **4-device sharded** model that just happens to contain a no-input arange (should be replicated on all 4).

So it picks **1 device** for the arange graph.

**The damage:** Picking 1 device makes the runtime shrink the mesh `[1,4] → [1,1]`. But the model's real weights were already placed across **4** devices. Now a 4-device buffer meets a 1-device mesh → **"Device count mismatch: 4 vs 1."**

**TLDR :** `torch.arange` has no input → becomes a no-argument graph → the plugin assumes "no input means single device" → collapses the 4-device mesh.

### Logs

- [jun24_arange.log](https://github.com/user-attachments/files/29295359/jun24_arange.log)
- [jun24_HunyuanImage_2_1_transformer_before_fix.log](https://github.com/user-attachments/files/29295358/jun24_HunyuanImage_2_1_transformer_before_fix.log)
- [test_arange.py](https://github.com/user-attachments/files/29295357/test_arange.py)
- [test_transformer.py](https://github.com/user-attachments/files/29295360/test_transformer.py)

kamalrajkannan78:
### Ticket

- fixes https://github.com/tenstorrent/tt-xla/issues/5360

### Problem description

- Sharded models that contain a no-input graph hit Device count mismatch: N vs 1 (e.g. HunyuanImage 2.1 transformer on llmbox, [#5360](https://github.com/tenstorrent/tt-xla/issues/5360)). A const-folded no-input graph (e.g. torch.arange in RoPE) is stamped with a 1×1 mesh, which reshapes the device mesh [1,N] -> [1,1] and breaks the live sharded buffers. Full root cause: [#5360](https://github.com/tenstorrent/tt-xla/issues/5360).

### What's changed

- A no-input graph now adopts the full mesh when either its result is device-sharded or the run targets more than one device. The execution device count (num_partitions, parsed from ExecutableBuildOptionsProto) is threaded through compileMlirProgram -> buildModule -> runCompilerStableHLOPipeline and used as the discriminator.

### What's changed

- A no-input graph now adopts the full mesh when the run targets more than one device. The execution device count (from the device assignment) is threaded through compileMlirProgram -> buildModule -> runCompilerStableHLOPipeline and used as the discriminator.
- Removed the now-unused `anyShardedAcrossDevices` helper (its only caller was the simplified condition).
- Added a minimal `torch.arange` sanity test (`tests/torch/ops/test_arange.py`) that reproduces the mismatch on a multi-device mesh as a regression guard.

### Checklist
- [x] Verify the changes through local testing on sanity and model in WH

### Logs

- [jun24_arange_before_fix.log](https://github.com/user-attachments/files/29303232/jun24_arange_before_fix_final.log)
- [jun24_HunyuanImage_2_1_transformer_before_fix.log](https://github.com/user-attachments/files/29303233/jun24_HunyuanImage_2_1_transformer_before_fix.log)
- [jul1_arange.log](https://github.com/user-attachments/files/29542222/jul1_arange.log)
- [jul1_hunyuanImage_2_1_transformer_after_fix.log](https://github.com/user-attachments/files/29542224/jul1_hunyuanImage_2_1_transformer_after_fix.log)

### DIFF
--- a/pjrt_implementation/src/api/client_instance.cc
+++ b/pjrt_implementation/src/api/client_instance.cc
@@ -478,13 +478,14 @@ tt_pjrt_status ClientInstance::populateMemories() {
 tt_pjrt_status ClientInstance::compileMlirProgram(
     const PJRT_Program *mlir_program, LoadedExecutableInstance **out_executable,
     const std::unordered_map<std::string, std::string> &compile_options,
-    const std::optional<std::vector<int64_t>> &replica_device_ids) {
+    const std::optional<std::vector<int64_t>> &replica_device_ids,
+    size_t target_num_devices) {
 
   std::string_view mlir_code(mlir_program->code, mlir_program->code_size);
 
   std::tuple<tt_pjrt_status, std::shared_ptr<ExecutableImage>> compile_result =
       m_module_builder->buildModule(mlir_code, m_cached_system_descriptor_path,
-                                    compile_options, this);
+                                    compile_options, this, target_num_devices);
   tt_pjrt_status status = std::get<tt_pjrt_status>(compile_result);
   if (!tt_pjrt_status_is_ok(status)) {
     return status;
@@ -840,6 +841,13 @@ PJRT_Error *onClientCompile(PJRT_Client_Compile_Args *args) {
     return *ErrorInstance::makeError(compile_options_status).release();
   }
 
+  // Execution device count from the device assignment (>1 for a multi-device
+  // run) tells the module builder to give no-input graphs the full mesh instead
+  // of collapsing it. Assumes pure SPMD (num_replicas == 1), which holds for
+  // the torch-xla / JAX SPMD compiles here.
+  size_t target_num_devices =
+      replica_device_ids.has_value() ? replica_device_ids->size() : 1;
+
   std::string_view program_format(args->program->format,
                                   args->program->format_size);
   if (program_format != module_builder::c_mlir_format_name) {
@@ -855,7 +863,7 @@ PJRT_Error *onClientCompile(PJRT_Client_Compile_Args *args) {
   tt_pjrt_status compile_status = client_instance->compileMlirProgram(
       args->program,
       reinterpret_cast<LoadedExecutableInstance **>(&args->executable),
-      compile_options_map, replica_device_ids);
+      compile_options_map, replica_device_ids, target_num_devices);
 
   return *ErrorInstance::makeError(compile_status).release();
 }
--- a/pjrt_implementation/src/api/module_builder/module_builder.cc
+++ b/pjrt_implementation/src/api/module_builder/module_builder.cc
@@ -98,19 +98,6 @@ moduleHasAnyFuncArguments(const mlir::OwningOpRef<mlir::ModuleOp> &m) {
   return public_func_ops[0].getNumArguments() > 0;
 }
 
-// Returns true if any sharding genuinely partitions a tensor across devices
-// (`MeshShardType::Devices`). Replicate / presharded-replicate shardings do not
-// count, since they do not require a multi-device mesh.
-static bool anyShardedAcrossDevices(
-    const std::vector<mlir::tt::sharding_utils::MeshSharding> &shardings) {
-  return std::any_of(
-      shardings.begin(), shardings.end(),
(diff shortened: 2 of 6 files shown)


pjrt_implementation/src/api/client_instance.cc
```diff
@@ -844,6 +845,13 @@ PJRT_Error *onClientCompile(PJRT_Client_Compile_Args *args) {
     return *ErrorInstance::makeError(compile_options_status).release();
   }
 
+  // num_partitions (>1 for a genuine multi-device SPMD compile) tells the
+  // module builder to give no-input graphs the full mesh instead of collapsing
+  // it.
+  size_t target_num_devices =
+      static_cast<size_t>(CompileOptionsParser::extractNumPartitions(
+          args->compile_options, args->compile_options_size));
```
sdjukicTT: `extractNumPartitions` looks avoidable - `onClientCompile` already parses `replica_device_ids` from the same `DeviceAssignmentProto`, so this could just be:
```
size_t target_num_devices =
    replica_device_ids.has_value() ? replica_device_ids->size() : 1;
```
I verified on a 2-device n300 with the new `test_arange` test.

pjrt_implementation/src/api/module_builder/module_builder.cc
```diff
@@ -752,24 +752,21 @@ tt_pjrt_status ModuleBuilder::runCompilerStableHLOPipeline(
...
-  // single device.
+  // A no-input graph adopts the full mesh when its result is device-sharded, or
+  // when the run targets >1 device: a no-input replicated value (e.g. a
+  // const-folded torch.arange) must still replicate across the mesh, else it
+  // defaults to 1x1, collapses the mesh, and breaks sharded buffers.
+  // See https://github.com/tenstorrent/tt-xla/issues/5360
   if (current_mesh_shape.has_value() && current_mesh_shape->size() == 2 &&
       !moduleHasAnyFuncArguments(mlir_module) &&
-      anyShardedAcrossDevices(output_shardings)) {
+      (anyShardedAcrossDevices(output_shardings) || target_num_devices > 1)) {
```
sdjukicTT: Is `anyShardedAcrossDevices` check still needed now?

pjrt_implementation/src/api/client_instance.cc
kamalrajkannan78: Done — switched to replica_device_ids.size() and dropped extractNumPartitions. Left a comment noting it assumes pure SPMD (num_replicas == 1), which holds for our torch-xla/JAX SPMD compiles. Thanks for catching this!

pjrt_implementation/src/api/module_builder/module_builder.cc
kamalrajkannan78: Removed — redundant now, since a sharded result implies target_num_devices > 1. Kept just that check and deleted the now-unused helper.

sdjukicTT (approved):

Issue #5216, opened 2026-06-15
[Model Bringup] coqui/XTTS-v2


saiarthiraguram:
References - https://huggingface.co/coqui/XTTS-v2, https://github.com/coqui-ai/TTS

PR #5376, opened 2026-06-25
Bring up coqui/XTTS-v2: per-component tests + e2e text-to-speech pipeline + demo

ashokkumarkannan1:
### Ticket
- https://github.com/tenstorrent/tt-xla/issues/5216 (Model Bringup: coqui/XTTS-v2)

### Problem Description
- XTTS-v2 has no single traceable forward, so every learned `nn.Module` is brought up as its own component and the modules are chained into a real end-to-end text-to-speech run. There is no e2e TTS pipeline tested on tt-xla so far.

### What's changed
- **Component `test_config` entries** (`tests/runner/test_config/torch/test_config_inference_single_device.yaml`): all 6 XTTS-v2 variants — `speaker_encoder`, `conditioning`, `gpt_prefill`, `gpt_decode`, `gpt_latents`, `hifigan_decoder` — each PCC-compared against CPU on the real tensors it receives end-to-end.
- **Nightly e2e pipeline test** (`tests/torch/models/xtts_v2/test_xtts_v2_pipeline.py`): drives the reusable `XTTSPipeline` from tt-forge-models with every learned module on TT (tokenizer, mel front-end and token sampling stay on CPU) and gates on three properties:
  - a finite, non-empty 24 kHz output waveform,
  - per-step decode PCC ≥ 0.99 against a CPU replay of the exact token sequence TT produced,
  - the decode loop settling onto one reused graph instead of recompiling per token.

- **Runnable demo** (`examples/pytorch/xtts_v2_pipeline.py`): thin wrapper over the same `run_xtts_pipeline()` entry point, following the `examples/pytorch/hunyuan_image_2_1.py` pattern. Discovered automatically by `tests/examples/test_examples.py`, so it needs no workflow change.
- **Per-example requirements** (`tests/examples/test_examples.py`): the demo needs `coqui-tts` + `torchaudio`, which the base test environment does not carry and the examples runner did not install. `EXAMPLE_REQUIREMENTS` maps an example to a `requirements.txt` and installs it through `RequirementsManager` for the duration of that example, rolling back afterwards. Examples without an entry take a `contextlib.nullcontext()` path, so their behaviour is unchanged.

### Results

| Check | Result |
|-------|--------|
| Component `test_config` (6) | **6/6 PASS** — `speaker_encoder`, `conditioning`, `gpt_prefill`, `gpt_decode`, `gpt_latents`, `hifigan_decoder` (per-component PCC vs CPU) |
| E2E pipeline test | **PASS** — 64/64 decode steps, per-step PCC ≥ 0.99 vs CPU replay, cumulative compile count flat at 11 across the whole tail, valid 24 kHz waveform |
| Demo (`examples/pytorch/xtts_v2_pipeline.py`) | **PASS** via `tests/examples/test_examples.py` — full utterance, 6.97 s of 24 kHz audio, peak 0.92 / no clipping |

### Checklist

### Logs
- [xtts_v2_demo_example_test.log](https://github.com/user-attachments/files/30542812/xtts_v2_demo_example_test.log), [xtts_v2_e2e_pipeline_test.log](https://github.com/user-attachments/files/30542814/xtts_v2_e2e_pipeline_test.log),
[conditioning.log](https://github.com/user-attachments/files/30542829/conditioning.log),[gpt_decode.log](https://github.com/user-attachments/files/30542831/gpt_decode.log),[gpt_latents.log](https://github.com/user-attachments/files/30542832/gpt_latents.log),[gpt_prefill.log](https://github.com/user-attachments/files/30542833/gpt_prefill.log),[hifigan_decoder.log](https://github.com/user-attachments/files/30542836/hifigan_decoder.log)
[speaker_encoder.log](https://github.com/user-attachments/files/30542837/speaker_encoder.log)


tests/torch/models/xtts_v2/test_xtts_v2_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+
+import os
+from pathlib import Path
+
+import pytest
+from infra import RunMode
+from loguru import logger
+from utils import BringupStatus, Category, ModelGroup
+
+pytestmark = [
```
kamalrajkannan78: is this pytestmark used somewhere below? I see only one occurence

tests/torch/models/xtts_v2/test_xtts_v2_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+    pytest.mark.nightly,
+    pytest.mark.model_test,
+    pytest.mark.single_device,
+    pytest.mark.large,
+]
+
+_REPO_ROOT = Path(__file__).resolve().parents[4]
+_LOADER_PATH = str(
```
kamalrajkannan78: why we can't simply import like `third_party.tt_forge_models.xtts_v2`

tests/torch/models/xtts_v2/test_xtts_v2_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+            pipeline.setup()
+            wav = pipeline.run()
+        except Exception as exc:  # deps missing / CPML weights ungated -> skip
+            if "COQUI" in str(exc) or "download" in str(exc).lower():
+                pytest.skip(f"Could not build/download XTTS-v2: {exc}")
+            raise
+        save_wav(wav, output_path)
+
+        # 1) Output artifact validity (mirrors the SDXL-Lightning e2e test).
+        assert output_file.exists(), f"Output WAV {output_path} was not created"
```
kamalrajkannan78, line 222: remove (mirrors the SDXL-Lightning e2e test). I feel no need to put a reference here

kamalrajkannan78 (approved):

tests/torch/models/xtts_v2/test_xtts_v2_pipeline.py
ashokkumarkannan1: Yes — `pytestmark` is a module-level name pytest picks up automatically and applies to every test in the file, so one occurrence is all it needs (it is not referenced by name anywhere below by design). It is what routes this test to the nightly / model_test / single_device / large jobs.

Verified on the pushed code:

```
$ pytest tests/torch/models/xtts_v2/test_xtts_v2_pipeline.py --collect-only -q \
    -m "nightly and model_test and single_device and large"
1 test collected
```

Without it the marker filter would collect 0. Left as is.

tests/torch/models/xtts_v2/test_xtts_v2_pipeline.py
ashokkumarkannan1: Good catch — done in 4232b441.

The value was never used as an import; it goes to `RequirementsManager.for_loader()`, which takes a *filesystem path* and does `os.path.dirname()` on it to find the `requirements.txt` sitting next to the loader. So it needed a path, which is why it was hand-built rather than imported.

But you are right that hand-building it was wrong — the repo convention is to import the loader module and let `inspect` supply the path, which is what suryaocr / janus_pro / diffusiongemma do:

```python
import inspect
import third_party.tt_forge_models.xtts_v2.pytorch.loader as xtts_loader
...
loader_path = inspect.getsourcefile(xtts_loader)
```

That replaces the fragile `Path(__file__).resolve().parents[4] / "third_party" / ...` (which silently breaks if the file ever moves depth). Confirmed the module-scope import is safe here: the xtts_v2 loader defers its `TTS` imports into methods, so importing it does not pull coqui-tts at collection time — it resolves to the same path and the `requirements.txt` next to it is found.

tests/torch/models/xtts_v2/test_xtts_v2_pipeline.py
ashokkumarkannan1: Agreed, removed in 4232b441 — the comment is now just `# 1) Output artifact validity.`

tests/torch/models/xtts_v2/test_xtts_v2_pipeline.py
ashokkumarkannan1: Switched to stacked decorators in 99f69e6d — you were right, and it turned out to be a consistency issue too: this file was the **only** one in `tests/torch/models/` using a module-level `pytestmark` list, while 70 files use stacked decorators.

```python
@pytest.mark.nightly
@pytest.mark.model_test
@pytest.mark.large
@pytest.mark.single_device
@pytest.mark.record_test_properties(...)
def test_xtts_v2_pipeline():
```

Ordering matches diffusiongemma (`nightly` -> `model_test` -> `large` -> device -> `record_test_properties`). Same four markers, collection behaviour verified unchanged:

```
$ pytest ... --collect-only -q -m "nightly and model_test and single_device and large"
1 test collected
```

tests/torch/models/xtts_v2/test_xtts_v2_pipeline.py
ashokkumarkannan1: Removed

tests/torch/models/xtts_v2/test_xtts_v2_pipeline.py
ashokkumarkannan1: it is used only once, moved all the markers above the test

sgligorijevicTT (approved):

jameszianxuTT (approved):

Issue #5275, opened 2026-06-18
[Investigation] Reconsider manylinux build flags added during June 26 uplifts


acicovicTT:
Two build flags have been recently added to push through manylinux builds during uplifts/serious nightly failures that should be reconsidered. Maybe it's cleaner to remove them and replace them with something else that is a better long term solution - this issue is used for tracking this investigation:

1. Compiler flag in `third_party/CMakeLists.txt`:
```
-DCMAKE_CXX_FLAGS=-Wno-error=pass-failed # Uplift hack, manylinux build threw because stablehlo autogenerated source had a failed loop unroll.s
```
  - No original issue, but it was an obvious compilation error in the new version of the compiler.

2. Linker flag `--no-rosegment` in `.github/workflows/call-build-manylinux.yml` (added in https://github.com/tenstorrent/tt-xla/pull/5181):
```
LDFLAGS="-Wl,--no-rosegment"
```
  - Original issue: https://github.com/tenstorrent/tt-xla/issues/5172

acicovicTT:
## Investigation: `-Wno-error=pass-failed` (part 1) — **cannot be removed**

Tested removal in #5408. The manylinux CI build **fails**:

```
build/stablehlo/stablehlo/dialect/StablehloOps.cpp.inc:322:30: error: loop not unrolled:
the optimizer was unable to perform the requested transformation ...
[-Werror,-Wpass-failed=transform-warning]
```
(also lines 437, 447, 589, …)

### Root cause
tt-mlir compiles StableHLO's sources **in-tree** (`add_subdirectory` of the toolchain checkout) under its global `-Werror` (tt-mlir `CMakeLists.txt:108`). The tablegen-generated `StablehloOps.cpp.inc` contains loop-unroll metadata that the optimizer cannot honor in a Release build → Clang emits `-Wpass-failed=transform-warning` → `-Werror` promotes it to a hard error. `-Wno-error=pass-failed` demotes just that one diagnostic.

### Why it still fires (key finding)
The **manylinux image compiles with Clang 21.1.8**, not the clang-20 used by the regular CI ubuntu image. I verified locally that **clang 20.1.8 unrolls these loops fine** — a standalone Release build of the current StableHLO sources (same pin) with `-Werror=pass-failed` is completely clean. So this is a **clang-21-specific** unroller regression, not something we changed.

The StableHLO/LLVM pins have been unchanged since the Nov 2025 uplift, so nothing on our side will make this go away.

### Conclusion
[...]

PR #4957, opened 2026-05-27
Uplift third_party/tt-mlir to 464a5f341908d5a3f790e697be5ffa6fd3fb6f32 2026-06-02

vmilosevic:
This PR uplifts the third_party/tt-mlir to the 464a5f341908d5a3f790e697be5ffa6fd3fb6f32

PR #5408, opened 2026-06-29
Remove -Wno-error=pass-failed manylinux build workaround

acicovicTT:
## Summary

Removes the `-DCMAKE_CXX_FLAGS=-Wno-error=pass-failed` workaround from the tt-mlir build in `third_party/CMakeLists.txt`. Addresses part 1 of #5275.

## Background / root cause

tt-mlir compiles StableHLO's sources **in-tree** (`add_subdirectory` of the toolchain's StableHLO checkout) under its global `-Werror` (`tt-mlir CMakeLists.txt:108`). The flag was added as an "uplift hack" in #4957: a Release **manylinux** build hit a Clang `-Wpass-failed` loop-unroll remark on an autogenerated StableHLO `.cpp`, and `-Werror` promoted it to a hard error. `-Wno-error=pass-failed` demotes just that one diagnostic back to a warning.

This is an **optimizer-heuristic remark** (there is no `#pragma clang loop` anywhere in the StableHLO source), so whether it fires depends on the exact Clang point release — precisely the kind of thing a compiler update resolves.

## Why it should be safe to remove now

- The StableHLO (`0a4440a5`) and LLVM pins in tt-mlir have been **unchanged since the Nov 2025 uplift** — i.e. before this flag was even added — so the autogenerated source in question is identical.
- The manylinux image has used **clang-20** since March 2026 (tt-mlir #7530).

## Verification

Built the current StableHLO sources standalone against the toolchain MLIR in **Release (`-O3`)** with **clang 20.1.8** and **`-Werror=pass-failed`** (precisely promoting only the diagnostic in question). All 135 build steps / 57 objects compiled cleanly — including the heavy autogenerated files `StablehloOps.cpp`, `ChloOps.cpp`, `VhloOps.cpp`, `StablehloBytecode.cpp`, `VhloBytecode.cpp`, and `TypeInference.cpp` — with **no `pass-failed` warning**.

The `-Wpass-failed` remark can still be marginally sensitive to the manylinux image's exact Clang minor and target flags, so the **manylinux CI job on this PR is the authoritative gate**. If it passes, the workaround is confirmed dead.


acicovicTT:
Closing: removal is not viable. The manylinux CI build fails because its compiler (Clang 21.1.8) cannot unroll loops in the tablegen-generated `StablehloOps.cpp.inc`, which tt-mlir's `-Werror` then turns into a hard error. clang 20 handles it fine, so the flag is still required. Full root-cause analysis recorded on #5275.

Issue #496, opened 2025-04-12
Implement PJRT_Executable_GetCompiledMemoryStats


mrakitaTT:
Add implementation for `PJRT_Executable_GetCompiledMemoryStats` API to `ExecutableInstance`.

It should return memory stats that allows callers to estimate device memory usage when running the executable. Relevant xla code [here](https://github.com/openxla/xla/blob/f80e7871e58bef66b0d9a10f10c269935f9251f2/xla/pjrt/pjrt_executable.h#L282).

PR #5411, opened 2026-06-29
Silence warning on unimplemented PJRT API

svuckovicTT:
### Ticket
Workaround for https://github.com/tenstorrent/tt-xla/issues/496 until it's properly fixed/implemented.

### Problem description
`PJRT_Executable_GetCompiledMemoryStats` API is not implemented for our device. This results in annoying warning:
```
2026-06-29 08:56:19.161 (  28.320s) [        56A7E300]              stubs.inc:81    WARN| STUB: PJRT_Executable_GetCompiledMemoryStats
```

### What's changed
Implement the API but return "unimplemented". This provides the same behaviour without the annoying log.

### Checklist

### DIFF
--- a/pjrt_implementation/inc/api/executable_instance.h
+++ b/pjrt_implementation/inc/api/executable_instance.h
@@ -94,6 +94,10 @@ PJRT_Error *onExecutableNumOutputs(PJRT_Executable_NumOutputs_Args *args);
 PJRT_Error *onExecutableSizeOfGeneratedCodeInBytes(
     PJRT_Executable_SizeOfGeneratedCodeInBytes_Args *args);
 
+// Implements PJRT_Executable_GetCompiledMemoryStats API function.
+PJRT_Error *onExecutableGetCompiledMemoryStats(
+    PJRT_Executable_GetCompiledMemoryStats_Args *args);
+
 // Implements PJRT_Executable_OutputElementTypes API function.
 PJRT_Error *
 onExecutableOutputElementTypes(PJRT_Executable_OutputElementTypes_Args *args);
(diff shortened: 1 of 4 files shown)


acicovicTT:
I would prefer it if this function didn't have special treatment just to avoid a log. These logs could be useful in the future for someone that is trying to understand what a newer version of pytorch/jax is trying to do (maybe we could see a potential for optimization behind an implemented API), or logs from nightly/weekly could be analyzed to see the frequency of unimplemented calls per API when we are deciding what to implement next.

Btw, this line should only show up in debug build, since it's `DLOG_F`, which I also think is appropriate. In release build, there should be no logs at all related to stubs.

svuckovicTT:
> I would prefer it if this function didn't have special treatment just to avoid a log. These logs could be useful in the future for someone that is trying to understand what a newer version of pytorch/jax is trying to do (maybe we could see a potential for optimization behind an implemented API), or logs from nightly/weekly could be analyzed to see the frequency of unimplemented calls per API when we are deciding what to implement next.
> 
> Btw, this line should only show up in debug build, since it's `DLOG_F`, which I also think is appropriate. In release build, there should be no logs at all related to stubs.

Thanks for the reply!

I agree that these logs *could* be useful at some point, but don't consider it a strong argument. Looking at the track record, this specific API has been unimplemented for >14 months, with an issue to track it. I haven't looked at the other API stubs, so can't comment on them. Is there an actual intent/willpower/energy to solve this or is it just a "hey I'll stash this screw because it might come in handy someday" situation?

To argue for the PR, these logs are annoying today - engineers running debug builds deserve nice readable outputs as well :)

acicovicTT:
> Is there an actual intent/willpower/energy to solve this or is it just a "hey I'll stash this screw because it might come in handy someday" situation?

Not currently, and we can always add it again later if needed, but what I am arguing is that removing it for just one function doesn't make sense to me. I would rather remove it completely for all unimplemented functions in case people find them annoying.

svuckovicTT:
Changed to remove all 65 api stub logs instead of the one.

acicovicTT (approved):

Issue #3608, opened 2026-03-05
Automate transformers uplift


AleksKnezevic:
We need a pipeline for uplifting transformers to
- latest release
- latest main

so that we're guaranteed we can run newly released models. The nightly pipeline that tests latest main should just run some sanity tests and if needed add API changes if needed, re-test and submit a PR. The version that tests latest release should also test all relevant nightly / benchmark tests and create a PR with needed changes / version bump. 

Once [this](https://github.com/tenstorrent/tt-xla/pull/3371) lands we'll only be behind one minor version, so it should be a good test bed for automatic uplifts. 

As a starting point, I experimented with a simple claude code version in `aknezevic/hf_uplift` to do the nightly uplift. Also, @nsmithtt has a more complicated n8n based version which does llvm uplifts, maybe that can be leveraged  if the simple version does not work.

PR #5412, opened 2026-06-29
Automating transformers uplift + fixes

jazpurTT:
### Ticket
#3608 

### Problem description

Each new `transformers` release breaks the model test suite. Detecting which models regress, classifying real failures vs. pre-existing ones, and patching the loaders + plumbing in tt-xla and `tt_forge_models` is a repetitive multi-day chore that gates newer `transformers` versions from landing on `main`.

### What's changed

End-to-end CI pipeline that helps automating the uplift + fixes:

- **`transformers-uplift-fix` Claude skill** (`.claude/skills/transformers-uplift-fix/SKILL.md`) — receives a `SCOPE` (`api-check` / `model-test-uplifts` / `model-perf-uplift`) and a captured failure context, consults the upstream changelog between `CURRENT_VERSION` and `TARGET_VERSION` to identify root causes, and edits source under tt-xla + `tt_forge_models` only. Hard rules: no monkey-patching, no `if version < X` shims, no drive-by refactors, no git ops (the orchestrator owns every commit and push). Writes a fix summary to `.github/transformers-uplift/fix-summary.md` that the orchestrator uses as the commit body.
- **`schedule-transformers-uplift.yml`** — nightly cron polls PyPI for the next stable `transformers` release, creates a `transformers-uplift/<ver>` WIP branch on tt-xla and `tt_forge_models`, bumps the pin in `venv/requirements-dev.txt`, and dispatches the orchestrator. Resolves a baseline `schedule-nightly` run on `main` to feed downstream filtering.
- **`workflow-transformers-uplift.yml`** — orchestrates the fix loop:
  - **api-check** — `pytest --collect-only` sweep to surface import / signature breaks; Claude fixes via a self-bounded sub-loop (up to 5 retries).
  - **test-and-fix-base** — runs the curated baseline subset (33 models marked `baseline_uplift`) for a fast framework-level signal before paying for the full suite.
  - **decide** — if anything still fails and we're under `MAX_ITERATIONS`, self-redispatches iteration N+1; otherwise advances to the full passes.
  - **test-and-fix-full** — full `model-test-passing.json` suite.
  - **test-and-fix-perf** — perf-benchmark sweep across all runners (n150 / p150 / n300-llmbox / galaxy-wh-6u / qb2-blackhole) with regression check against the baseline nightly.
  - Each stage hands its captured failure context to the new `transformers-uplift-fix` Claude skill. Patches land on both tt-xla and the `tt_forge_models` submodule. Branch is sourced from one place (`github.ref_name`).
- **`call-test-uplift.yml`** / **`call-perf-uplift.yml`** — reusable wrappers around `call-test.yml` and `call-filtered-perf-tests.yml`. Compare current failures against a baseline nightly run so Claude only sees uplift-induced regressions, not pre-existing failures.
- **`manual-test-uplift.yml`** / **`manual-perf-uplift.yml`** — user-dispatchable wrappers for debugging a single suite/runner against an existing WIP branch.
- **Supporting scripts** — `detect-new-version.sh` (PyPI next-stable picker), `bump-transformers.sh` (requirements update), `extract-failures.py` / `extract-perf-failures.py` (junit + log parsing into Claude-friendly context), `run-api-check.sh`, `run-claude-fix.sh`.

#### Why these 33 models for `baseline_uplift`

Curated for **architectural diversity** rather than count — one or two representatives per family that exercise the corners of the `transformers` API most likely to break on a release:

- **Encoder LMs** — BERT-base-uncased (masked LM), ALBERT, RoBERTa-XLM, DistilBERT
- **Decoder / causal LMs** — Falcon-3.1B, Mistral-Ministral-3B, Phi-2, Phi-3-mini-128K, Qwen-2.5-0.5B, Qwen-3-0.6B, Gemma-1.1-2B
- **Seq2seq** — BART-large, MusicGen-small (audio seq2seq)
- **Vision encoders** — ViT-base, Swin-S, DINOv2-small, ResNet-50, EfficientNet-B0, MobileNetV2
- **Detection / segmentation** — YOLOv4, YOLOv7, YOLOS-small, YOLOP, DETR, OWL-ViT, Segformer-Mit-B0, MaskFormer-Swin-base
- **Multimodal** — CLIP-base-patch16, SigLIP-base-patch16
- **Embeddings / NER** — BGE-large-en, Sentencizer-XLM-RoBERTa
- **Generative / audio** — Stable Diffusion UNet, SpeechT5 HiFiGAN vocoder

These hit Cache, attention, attention-mask, tokenizer, image-processor, audio-processor, and generation surfaces — i.e. the high-churn areas.

### Checklist


jazpurTT:
I ran an e2e test uplifting transformers 5.5.1 -> 5.9.0
- api-check [passed and needed no fixes](https://github.com/tenstorrent/tt-xla/actions/runs/28298200980/job/83843017348).
- Baseline [tests failed](https://github.com/tenstorrent/tt-xla/actions/runs/28298200980/job/83843227635) and Claude [applied a fix](https://github.com/tenstorrent/tt-xla/commit/2ab56ddf107ad45ed467d36cd47bd6c5d3e759e2), on the second run we saw a [full successful run](https://github.com/tenstorrent/tt-xla/actions/runs/28300224863/job/83848417940#step:19:45)
- Model passing tests ran, this is the test and this is the [commit with the fixes + message](https://github.com/tenstorrent/tt-xla/commit/a0994f5f515afab036c0d9c96847ab9e920559f6):
    - Claude applied fixes for some models
    - Claude was able to identify models that had the same failure on baseline and skipped them
    - A couple of models were identified as non transformers related. This would need human review.
- Perf benchmark ran, this is the test ant this is the [commit](https://github.com/tenstorrent/tt-xla/commit/253b585e2993946a4c93d996b6517de63e53c4d2)
    - Small PCC drift, Claude made the change but also flagged it for human review.
    - The models that failed failed on the baseline test and Claude identified them as so.

acicovicTT:
Pls check `.github/scripts/uplift/uplift-config.json` and the corresponding test matrix that is invoked when transformers uplift is detected (your 33 models are probably a better curated list than what is already there), but we should not have two separate quality gates for transformers uplift, so either see if you can integrate with `uplift-config.json`, which is a pretty simple generic uplift test selection mechanism, OR if transformers uplift requires a more complicated machinery, remove it from `uplift-config.json`.

jazpurTT:
> Pls check `.github/scripts/uplift/uplift-config.json` and the corresponding test matrix that is invoked when transformers uplift is detected (your 33 models are probably a better curated list than what is already there), but we should not have two separate quality gates for transformers uplift, so either see if you can integrate with `uplift-config.json`, which is a pretty simple generic uplift test selection mechanism, OR if transformers uplift requires a more complicated machinery, remove it from `uplift-config.json`.

Thanks for the feedback. I integrated the baseline tests, so `transformers-uplift-qualification.json` now includes `baseline-uplift.json`. The PR gate and the auto-pipeline now share the same 33-model list via the `baseline_uplift` pytest marker. Adding/removing an entry in one place propagates to both. I do feel like there is no need to have both `baseline-uplift` and `model-test-extended` in `transformers-uplift-qualification.json` but if you feel otherwise we can have both.

.github/scripts/filter-test-matrix.py
```diff
@@ -37,7 +37,8 @@ def filter_matrix_adv(matrix, adv_filter):
     Filter is JSON array with sets of conditions.
     Each condition might have elements:
       - "runs-on": machine(s) on which the test should run. If ommited, condition will be applied to all unskipped machines. Parameter can be string or array of strings.
-      - "filter": string that should be present in the test name.
+      - "filter": string (or comma-separated list / list of strings) that should be present in the test name (include match).
+      - "exclude": string (or comma-separated list / list of strings) — any test whose name contains one of these substrings is dropped. Applied AFTER "filter".
```
jameszianxuTT, line 41: I think this is a good addition to the test matrix manipulation interface

.github/workflows/call-test-uplift.yml
```diff
@@ -0,0 +1,263 @@
...
+
+      - name: Install Claude Code CLI
+        shell: bash
+        run: |
+          if command -v claude >/dev/null 2>&1; then
+            claude --version
+            exit 0
+          fi
+          curl -fsSL https://nodejs.org/dist/v20.11.1/node-v20.11.1-linux-x64.tar.xz | tar -xJ --strip-components=1 -C /usr/local/
+          npm install -g @anthropic-ai/claude-code
```
jameszianxuTT: Will installing claude code via npm be brittle? Why not use the one line installer? `curl -fsSL https://claude.ai/install.sh | bash`

.github/workflows/call-test-uplift.yml
```diff
@@ -0,0 +1,263 @@
...
+
+      - name: Install wheel + uplifted transformers
+        shell: bash
+        env:
+          TARGET_VERSION: ${{ inputs.target_version }}
+        run: |
+          source venv/activate
+          uv pip install wheels/*.whl
+          uv pip install --force-reinstall --no-deps \
+            "transformers==${TARGET_VERSION}"
```
jameszianxuTT: I can understand why we want to avoid installing deps alongside the transformers upgrade to test transformers in isolation, but I think there's still a chance that we'll have deps that legitimately need an upgrade, could causing failures at runtime or package incompatibility. How will that be dealt with/has this failure mode been observed before?

sgligorijevicTT (approved):
Rubberstamping the testconfig change

`baseline_uplift` is not a very descriptive name IMO, perhaps something more descriptive could be used, but it is also fine to leave as is

.github/workflows/call-perf-uplift.yml
```diff
@@ -0,0 +1,245 @@
...
+            popd >/dev/null
+          fi
+
+          # tt-xla commit
+          git add "$LAST_FAILED_FILE"
+          if ! git diff --staged --quiet || [[ -n "$(git status --porcelain)" ]]; then
+            git add -A
+            git commit "${COMMIT_FLAG[@]}"
+            git push origin "$BRANCH"
+          fi
```
nsumrakTT, line 285: If the idea is to commit to existing PR opened in scheduled workflow, what happens if someone runs manual uplift on main? If not, please add main branch detection.
Should manual run also commit inside the branch? When this is useful (metal->mlir uplift?)

.github/workflows/schedule-transformers-uplift.yml
```diff
@@ -0,0 +1,162 @@
...
+          if [[ -z "$CURRENT" ]]; then
+            echo "::error::could not parse 'transformers==X.Y.Z' pin from $REQ_FILE" >&2
+            exit 1
+          fi
+
+          RELEASE_TAGS=$(gh release list --repo huggingface/transformers \
+            --limit 100 --json tagName,isPrerelease \
+            -q '.[] | select(.isPrerelease == false) | .tagName')
+
+          NEW_VERSION=$(python3 - "$CURRENT" <<PY
```
nsumrakTT, line 52: When using python on ubuntu-latest it is always a good idea to add setup-python action step with intended version, eg
```
      - uses: actions/setup-python@v6
        with:
          python-version: '3.12'
```
Also, I'm not a big fan of inline python scripts.

.github/workflows/call-transformers-uplift.yml
```diff
@@ -0,0 +1,281 @@
+name: Transformers Uplift + Claude Fix
```
nsumrakTT, line 1: Please rename to `call-transformers-uplift.yml` to keep naming schema valid.

nsumrakTT (approved):

.github/workflows/call-test-uplift.yml
jazpurTT: done, thanks for the suggestion

.github/workflows/call-test-uplift.yml
jazpurTT: this was for when the job was called from `main` but this has changed since so it is not necessary since it's being called from the uplift branch, removed the reinstall line.

.github/workflows/call-perf-uplift.yml
jazpurTT: added main branch detection to all three commit/push sites, and kept the commit-by-default behaviour for other branches so manual runs can iterate on a WIP uplift branch.

.github/workflows/schedule-transformers-uplift.yml
jazpurTT: the inline python felt "necessary" because it's a clean way to make sure we extract the right verion. Added the setup-python step.

jazpurTT:
> Rubberstamping the testconfig change
> 
> `baseline_uplift` is not a very descriptive name IMO, perhaps something more descriptive could be used, but it is also fine to leave as is

I understand, but I wanted to keep it as simple as possible, `transformers_base_uplift` or other iteration like that felt  wordy and long. Also, given the criteria of trying to select models based on "architectural diversity" I felt like it could be used as a baseline for other uplifts to.

AleksKnezevic (approved):

PR #5435, opened 2026-06-30
[GLM] Rework MoE routing to avoid scatter_/gather lowerings on multi-row mesh

### LINKED ISSUE #5409: [Benchmark] GLM 4.7 bad pcc for users on non-zero device
Glm 4.7 currently has good pcc (0.99) only on first 16/64 users, which are users on first of 4 devices batch is sharded on.

Main reason is the use of `local iota` `(arange(0,16))` in places where global is expected `(arange(0,64))`. When using local iota for indexing global sized tensor (scatter and embedding ops in router part of MoE block indirectly do this), only first 1/4 of the full batch is correctly computed, while others are 0. This is further propagated to users 0-15 having good, and 16-65 having same, but somewhat wrong output.

With this change alone, pcc is improved 0.88->0.95, so there is still some error of this kind to be found.

Note: similar issue [previously found in gpt-oss](https://github.com/tenstorrent/tt-mlir/discussions/7334#discussioncomment-16480883) was [bypassed with this monkey-patch](https://github.com/tenstorrent/tt-xla/pull/4316), but never fixed

Note: this is an example of why we need [pcc check for all users](https://github.com/tenstorrent/tt-xla/issues/4547). It should also be added to perf benchmarks if that is how we bringup models now, and consider model fully brought up only when it has good pcc on all users

mvasiljevicTT:
### Ticket
Fixes https://github.com/tenstorrent/tt-xla/issues/5409

### Problem description

GLM 4.7 had good PCC (0.99) only on the first 16/64 users — those on the first of the 4 devices the batch is sharded across — because the MoE router indexes a *global* tensor with a large index that is either mis-sharded or precision-truncated. Two bugs of this class, both in `route_tokens_to_experts` (`sparse_mlp.py`):

1. **`group_mask.scatter_(1, group_idx, 1)`** — `scatter_` adds a token row-iota that Shardy shards *without* the per-shard offset, so the group mask marks only mesh-row 0 and zeros routing for rows 1-3 (`0.88 → 0.95`).
2. **`router_logits.gather(1, topk_indices)`** — lowers to `ttir.embedding` over the all-gathered `[tokens·E]` score table with a flat index `token·E + expert` computed at fp16-class precision. For rows 1-3 the large `token·E` offset rounds off the expert bits (e.g. `2609 → 2610`), gathering the wrong expert's score (`0.95 → 0.99`).

### What's changed

Replaces both ops with the `one_hot` pattern already used for XLA compat (`_topk_to_sparse_scores`), so the `arange` is over experts/groups (a model dim, **not** the sharded batch axis) and the index stays small + exact:

- **Group mask: `scatter_` → `one_hot` + `any`.**
- **Top-k weights: `gather` → `one_hot` + `einsum`.**

Both rewrites are mathematically equivalent on CPU and just workarounds around buggy `scatter` and `gather` ops to unblock glm pcc until they are fixed.

### Impact

- GLM 4.7 multi-device PCC **0.88 → 0.993 on all 4 mesh-rows** (full-batch decode PCC **0.9937**); benchmark PASSES.
- CI A/B (GLM perf, shared runner off): [no fix → fails rows 1-3](https://github.com/tenstorrent/tt-xla/actions/runs/28383288938) · [with fix → passes](https://github.com/tenstorrent/tt-xla/actions/runs/28383183382)

### DIFF
--- a/python_package/tt_torch/sparse_mlp.py
+++ b/python_package/tt_torch/sparse_mlp.py
@@ -984,8 +984,18 @@ class DeepseekV3MoEToA2AAdapter(nn.Module):
                 .sum(dim=-1)
             )
             group_idx = torch.topk(group_scores, k=topk_group, dim=-1, sorted=False)[1]
-            group_mask = torch.zeros_like(group_scores)
-            group_mask.scatter_(1, group_idx, 1)
+            # one_hot + sum instead of scatter_: torch scatter_ lowers to a
+            # stablehlo.scatter whose dim-0 row index is a token iota; Shardy
+            # shards that iota WITHOUT the per-shard offset, so the group mask
+            # marks only mesh-row 0's tokens and zeros routing for rows 1-3.
+            group_mask = (
+                (
+                    group_idx.unsqueeze(-1)
+                    == torch.arange(n_group, device=group_idx.device)
+                )
+                .any(dim=1)
+                .to(group_scores.dtype)
+            )
             score_mask = (
                 group_mask.unsqueeze(-1)
                 .expand(-1, n_group, n_routed_experts // n_group)
@@ -997,7 +1007,16 @@ class DeepseekV3MoEToA2AAdapter(nn.Module):
             topk_indices = torch.topk(scores_for_choice, k=top_k, dim=-1, sorted=False)[
                 1
             ]
-            topk_weights = router_logits.gather(1, topk_indices)
+            # one_hot + einsum instead of gather: torch.gather lowers to
+            # ttir.embedding over the all-gathered [tokens*E] score table with a
+            # flat index token*E+expert computed at fp16-class precision; for
+            # mesh-rows 1-3 the large index rounds off the expert bits -> wrong
+            # weights. einsum keeps the small expert index implicit and exact.
+            _ohw = (
+                topk_indices.unsqueeze(-1)
+                == torch.arange(n_routed_experts, device=topk_indices.device)
+            ).to(router_logits.dtype)
+            topk_weights = torch.einsum("be,bke->bk", router_logits, _ohw)
             if norm_topk_prob:
                 denominator = topk_weights.sum(dim=-1, keepdim=True) + 1e-20
                 topk_weights /= denominator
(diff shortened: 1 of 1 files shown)


mvasiljevicTT:
Even with [iota fixed](https://github.com/tenstorrent/tt-mlir/commit/d999f834e251b4cb1fda74c26c9354f4bd20589b), using `one_hot+any` instead of `scatter` is faster in this case:

```
OP                                    A one_hot/iotaOFF   B scatter/iotaON    ΔTime    ΔCount
                                        time_μs (n)         time_μs (n)
------------------------------------------------------------------------------------------------
SparseMatmulDeviceOperation           1,202,841 ( 6)      1,206,791 ( 6)      +3,951    +0   ← noise (same work)
AllGatherDeviceOperation                116,075 (28)        116,681 (30)        +606    +2   ← iota reshard
AllBroadcastDeviceOperation                 275 ( 3)            993 ( 7)        +718    +4   ← iota reshard
ScatterDeviceOperation                        0 ( 0)             95 ( 2)         +95    +2   ← raw scatter_ kernel
MeshPartitionDeviceOperation                578 ( 4)            591 ( 8)         +13    +4   ← iota reshard
UntilizeWithUnpaddingDeviceOperation     19,404 (56)         19,703 (86)        +299   +30   ┐
TilizeWithValPaddingDeviceOperation      14,848 (71)         15,021 (85)        +173   +14   │ layout/tiling churn
ReshapeViewDeviceOperation                6,764 (71)          6,925 (81)        +161   +10   │ around the reshard
ConcatDeviceOperation                     3,794 (75)          3,826 (81)         +32    +6   │
PermuteDeviceOperation                    4,466 ( 4)          4,505 ( 6)         +40    +2   │
SliceDeviceOperation                        973 (230)           990 (234)        +18    +4   ┘
ReduceDeviceOperation                     1,891 ( 8)          1,880 ( 6)         -11    -2   ← A's one_hot .any()
FillPadDeviceOperation                    2,802 (144)         2,724 (142)        -77    -2
TypecastDeviceOperation                   2,391 (247)         2,395 (245)         +5    -2
------------------------------------------------------------------------------------------------
```

jameszianxuTT:
> GLM 4.7 had good PCC (0.99) only on the first 16/64 users — those on the first of the 4 devices the batch is sharded across 

Interesting, Kimi also sees good results on first 4/16 users, maybe it is the same kind of error. Thanks for finding this!

PR #5485, opened 2026-07-01
[vLLM] Hoist TT attention batch metadata out of the compiled graph

### LINKED ISSUE #5154: [vLLM] Hoist prefill `fill_batch_idx` to a setup-time metadata input
### Problem
In `_handle_paged_attention`'s prefill branch (`integrations/vllm_plugin/vllm_tt/attention.py`), the batched `paged_fill_cache` index is built inside the traced forward on every call:

```python
batch_idxs = torch.arange(key_for_update.shape[0], dtype=torch.int32, device=k_cache.device)
```

This is an in-graph, device-side index constructor created per call (~`2 × num_layers` times per prefill). It's the same antipattern that drove the sharding collapse in #5108 — index tensors built in the traced forward can lift as graph operands and aren't cleanly-sharded inputs. The batched change in #4955 already cut the lifted-operand count ~32× vs the old per-user loop, so this is not a current correctness/perf blocker, just cleanup.

### Proposed fix
Per @mmanzoorTT: pass `batch_size` (num_users, fixed via batch padding for generative models) through `TTMetadata`, precompute the index tensor **once on CPU and move it to device** at build/init time (e.g. `attn_metadata.fill_batch_idx`), and reference it in the prefill branch instead of `torch.arange(..., device=...)`. Turns ~`2 × num_layers` in-graph constructors into one stable, properly-sharded graph input.

### Context
- Deferred from #4955 review ([thread](https://github.com/tenstorrent/tt-xla/pull/4955#discussion_r... )); applies the feedback from #5108 ([discussion](https://github.com/tenstorrent/tt-xla/pull/5108#discussion_r3377077308)).

mmanzoorTT:
### Ticket
closes #5154 

### Problem description
Batch index are generated for each prefill call

### What's changed
- Hoist batch index creation to runner constructer.
- Add num_user to attn_metadata and use it directly instead of deducing it

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
+++ b/integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
@@ -183,7 +183,6 @@ class TTMetadata:
     chunk_start_idx: torch.Tensor
     batch_idx: Optional[torch.Tensor] = None
     num_users: Optional[int] = None
-    num_tokens: Optional[int] = None
 
     def __init__(
         self,
@@ -195,7 +194,6 @@ class TTMetadata:
         chunk_start_idx: torch.Tensor | None = None,
         batch_idx: torch.Tensor | None = None,
         num_users: Optional[int] = None,
-        num_tokens: Optional[int] = None,
     ):
         self.cache_position = cache_position
         self.attn_mask = attn_mask
@@ -205,10 +203,7 @@ class TTMetadata:
             fill_page_table if fill_page_table is not None else page_table
         )
         self.chunk_start_idx = chunk_start_idx
-        assert num_users is not None, "num_users must be provided."
-        assert num_tokens is not None, "num_tokens must be provided."
         self.num_users = num_users
-        self.num_tokens = num_tokens
         self.batch_idx = batch_idx
 
 
@@ -336,6 +331,7 @@ class TTAttentionBackendImpl(AttentionImpl):
         from collections import namedtuple
 
         num_users = attn_metadata.num_users
+        assert num_users is not None, "num_users must be provided in attn_metadata."
         orig_query_shape = query.shape
         orig_query_ndim = query.ndim
 
--- a/integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
+++ b/integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
@@ -181,6 +181,9 @@ class TTMetadata:
     # across users by same-stage batching). Set only on a cached-prefix chunk;
     # consumed by the chunked_scaled_dot_product_attention op.
     chunk_start_idx: torch.Tensor
+    batch_idx: Optional[torch.Tensor] = None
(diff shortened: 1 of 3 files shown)


AleksKnezevic (approved):

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2442,6 +2463,14 @@ def _get_dummy_inputs(self, config: dict):
             attn_mask=None,
             fill_page_table=fill_page_table,
             chunk_start_idx=chunk_start_idx,
+            fill_page_table=page_table,
```
kmabeeTT: duplicate line already listed above, which one to keep?

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
```diff
@@ -199,6 +205,11 @@ def __init__(
             fill_page_table if fill_page_table is not None else page_table
         )
         self.chunk_start_idx = chunk_start_idx
+        assert num_users is not None, "num_users must be provided."
+        assert num_tokens is not None, "num_tokens must be provided."
```
kmabeeTT: do we need num_tokens? Seems it is wired in here but never actually uses `attn_metadata.num_tokens`?

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
```diff
@@ -199,6 +205,11 @@ def __init__(
             fill_page_table if fill_page_table is not None else page_table
         )
         self.chunk_start_idx = chunk_start_idx
+        assert num_users is not None, "num_users must be provided."
```
kmabeeTT: These asserts are unconditional, but the build() stub at line 74 constructs TTMetadata without num_users/num_tokens — if that path runs it'll AssertionError. Is build() reached at runtime? If so, pass values there or move the asserts onto the consuming paths.

kmabeeTT (commented):
Some real issues and couple questions

integrations/vllm_plugin/vllm_tt/model_runner.py
mmanzoorTT: Missed it during rebasing. fixed now.

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
mmanzoorTT: Moved the assertion onto the consuming path.

integrations/vllm_plugin/vllm_tt/attention_impls/attention.py
mmanzoorTT: I originally added this to verify that the deduced `query_num_tokens` matches the expected value, but I think it's overkill. Removed.

kmabeeTT (approved):
Thx for changes, and fix!

PR #5492, opened 2026-07-02
Add option for measuring compile time of each MLIR pass

umalesTT:
### Ticket
None

### Problem description
We didn't have a way of measuring per-pass compile time and, given that compile time is prohibitively slow for some models, we need a way to debug it.

### What's changed
Through `TTXLA_MLIR_PASS_TIMING = 1` (or any non-zero value), if set for running program, on output we will get timing report which looks like [this](https://gist.github.com/umalesTT/e2dd52db95f04b9b545dd4b3e6d6c2c8) (example for compiling training step of Wan 2.2 5b model). 

One caveat is that, if `TTXLA_LOGGER_LEVEL` is set to `Verbose`, the timing flag will be ignored, as `Verbose` logger level disables multithreading therefore they can't be used together.

### Checklist

### DIFF
--- a/pjrt_implementation/src/api/module_builder/module_builder.cc
+++ b/pjrt_implementation/src/api/module_builder/module_builder.cc
@@ -494,6 +494,7 @@ tt_pjrt_status ModuleBuilder::convertFromVHLOToSHLO(
   mlir::stablehlo::createStablehloDeserializePipeline(vhlo_to_shlo_pm);
 
   enableVerboseIRPrinting(vhlo_to_shlo_pm);
+  enableMlirPassTiming(vhlo_to_shlo_pm);
 
   if (mlir::failed(vhlo_to_shlo_pm.run(mlir_module.get()))) {
     LOG_F(ERROR, "Failed to convert from VHLO to SHLO module");
@@ -778,6 +779,7 @@ tt_pjrt_status ModuleBuilder::runCompilerStableHLOPipeline(
                                                stablehlo_pipeline_options);
 
   enableVerboseIRPrinting(stablehlo_pipeline_pm);
+  enableMlirPassTiming(stablehlo_pipeline_pm);
 
   if (mlir::failed(stablehlo_pipeline_pm.run(mlir_module.get()))) {
     LOG_F(ERROR, "Failed to run stablehlo pipeline");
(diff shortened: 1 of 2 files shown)


svuckovicTT:
Awesome PR!

One of the lowest hanging fruits that might help improve perf non-trivially:
https://github.com/tenstorrent/tt-mlir/issues/8940

@sdjordjevicTT it'd be great if we could find someone on our team to look into this!

pjrt_implementation/src/api/module_builder/module_builder.cc
```diff
@@ -1192,6 +1215,27 @@ void ModuleBuilder::enableVerboseIRPrinting(mlir::PassManager &pm) {
   pm.enableIRPrinting();
 }
 
+void ModuleBuilder::enableMlirPassTiming(mlir::PassManager &pm) {
```
nvukobratTT, line 1218: @sdjordjevicTT @nsumrakTT @vvukomanTT, is there a way to track these in Superset for a few chosen models? :))

sdjordjevicTT (approved):
Great!

pjrt_implementation/src/api/module_builder/module_builder.cc
sdjordjevicTT: Honestly, I don't know how our superset pipeline works, but I guess we can extract these numbers for sure. We will need some help from experts in this area. :)

nsumrakTT: I am not sure what is resulting database schema - is it easily searchable filed or it is json inside the filed.

nsumrakTT (approved):

pjrt_implementation/src/api/module_builder/module_builder.cc
nvukobratTT: We could just use the existing model schema and add additional information as part of the new "tags" field.

vvukomanTT: @nvukobratTT Yep, adding this to the tags field would be the easiest way to track it.

sgligorijevicTT (approved):

PR #5514, opened 2026-07-03
Erase repeat_kv for GQA SDPA (enable_gqa on device)

agobeljicTT:
### Ticket
N/A

### Problem
GQA sends inflated K/V into SDPA, materializing the repeat on device for nothing:

```
repeat_kv:  K[B, Hkv, S, D]  ->  K[B, Hq, S, D]      (Hq = Hkv * n_rep)
SDPA(Q[B,Hq,S,D], K[B,Hq,S,D], V[B,Hq,S,D])
```

ttnn SDPA already broadcasts KV heads (needs only `Hq % Hkv == 0`), so the repeat is wasted.

### What's changed
New `erase_repeat_kv` FX pass: rewires SDPA to the non-inflated K/V + `enable_gqa=True`.

```
SDPA(Q[B,Hq,S,D], K[B,Hkv,S,D], V[B,Hkv,S,D], enable_gqa=True)
```

- Matches only interleave ordering (HF `unsqueeze→expand→reshape`, `repeat_interleave(dim=-3)`) → bit-identical (`repeat_kv == enable_gqa`, diff `0.0`). Tile/`repeat` and `n_rep==1` are left untouched.
- No `eliminate_dead_code()`: at dynamo stage in-place cache writes (`Tensor.copy_`) are `call_method` nodes FX treats as pure, so a blanket DCE drops them + RoPE (→ PCC ~0.3). Cleanup left to `torch.export`. There is no automated guard for this — any pass added here must be verified end-to-end (real model, PCC); a comment at the call site flags it.

Measured on Llama 3.1 8B prefill:
| Before | After |
| -- | -- |
| 195ms | 182ms |

### Checklist


python_package/tt_torch/backend/passes.py
```diff
@@ -132,6 +132,247 @@ def _replace_multi_output_op(gm, node, output_variants):
...
+    if dim is None or out_shape is None:
+        return None
+    if _normalize_dim(dim, len(out_shape)) != head_axis:
+        return None
+    # `repeats` must be a uniform scalar to match enable_gqa's interleave ordering.
+    # A per-head tensor of repeats can satisfy the output-shape gate while producing
+    # a different head ordering, which would make the enable_gqa substitution wrong.
+    repeats = node.args[1] if len(node.args) > 1 else node.kwargs.get("repeats")
+    if not isinstance(repeats, int):
+        return None
```
vkovinicTT, line 423: Do we even need this check?

python_package/tt_torch/backend/passes.py
```diff
@@ -132,6 +132,247 @@ def _replace_multi_output_op(gm, node, output_variants):
...
+        if len(matched) != 2:
+            continue
+
+        src_heads = {
+            _fx_node_shape(src)[-_SDPA_HEAD_AXIS_FROM_END] for src in matched.values()
+        }
+        if len(src_heads) != 1:
+            # Deflated key and value disagree on head count -> not a valid GQA
+            # grouping for enable_gqa; leave the node untouched.
+            continue
```
vkovinicTT, line 560: If we checked this above:
```python
  if len(matched) != 2:
      continue
```
we don't need this check, right?

python_package/tt_torch/backend/passes.py
```diff
@@ -132,6 +132,247 @@ def _replace_multi_output_op(gm, node, output_variants):
...
+        # NOTE: deliberately do NOT call gm.graph.eliminate_dead_code() here.
+        # In KV-cache models the pre-inflation source is a cache tensor written
+        # by an in-place `update_cache` (RoPE'd K/V). Once SDPA is rewired to read
+        # that source directly, the bypassed repeat_kv (expand/reshape) becomes
+        # dead -- but a dynamo-stage eliminate_dead_code() is not mutation-aware
+        # and would also drop the in-place cache write (and its RoPE), silently
+        # corrupting attention (observed PCC ~0.3, all update_cache ops removed).
+        # The orphaned repeat_kv nodes are pure and are cleaned up later by the
+        # mutation-aware torch.export / run_decompositions pipeline, which
+        # preserves the cache mutations.
```
vkovinicTT: Hmmm is there really no way to somehow flag this out so that we prevent people from using `gm.graph.eliminate_dead_code()` when you activate this?
Maybe  add debug log here with warning.

vkovinicTT (approved):

python_package/tt_torch/backend/passes.py
agobeljicTT: perhaps not

agobeljicTT:
we should investigate if _check_sdpa_constraints should only require QKV to be bf16, not the mask as well.

python_package/tt_torch/backend/passes.py
agobeljicTT: regardless, one must not call gm.graph.edc() before export.
copy_ function is treated as pure (which is wrong in this context) so it gets deleted

Issue #5349, opened 2026-06-23
[Codegen] Add tensor parallel tests to EmitPy nightly


svuckovicTT:
EmitPy tests run in nightly, as defined here:
[.github/workflows/test-matrix-presets/model-test-emitpy.json](https://github.com/tenstorrent/tt-xla/blob/da06b2bd65ff9cea81959cb7055d8d665c43f47d/.github/workflows/test-matrix-presets/model-test-emitpy.json)

We should add tensor_parallel tests as well.

I'm not sure what machine targets are available in CI, but we should probably aim for QB2 and maybe a couple GLX tests.

PR #5515, opened 2026-07-03
Fix emitpy tests false-timeout

### LINKED ISSUE #5513: Emitpy jobs false-timeout
## Summary

Tests run with the `--emitpy` flag take significantly longer than standard runs.
A standard run does the flatbuffer path and then a golden comparison. An
`--emitpy` run does that same work first, and then additionally runs the codegen
path: emit the code, run the generated code, and compare against the flatbuffer
result.

However, when computing their timeout minutes, emitpy jobs are treated
identically to standard runs. Durations recorded on regular runs are
used to approximate the total test duration, and the standard 3× safety margin is
applied. Since emitpy actually runs at a multiple of the recorded durations, this can result in jobs false-timeout.

The `model-emitpy-emitpy` [run](https://github.com/tenstorrent/tt-xla/actions/runs/28518075732/job/84538378327) exposes the issue.

amilovanovicTT:
### Ticket
Closes #5513, #5349

### Problem description
Previously, emitpy tests could be triggered only manually by running the `model-test-emitpy.json` workflow. 

### What's changed
Added emitpy tests to the nightly. This includes the following:
1. **Distinct emitpy test designators + per-mode config.**
   `--emitpy` runs now get their own test ID, which is an`emitpy-` prefixed regular ID, so they no longer collide with the normal runs in `.test_durations`, and failures are attributable to the right mode. Added `emitpy_overrides` in the YAML config (restricted to `status`/`reason`, mirroring `arch_overrides`) so a model can be `EXPECTED_PASSING` normally but flipped under `--emitpy` (this is only temporary until all emitpy nightly tests are fixed).

2. **Nightly emitpy jobs + classified the current failures.**
   Added `test_forge_models_emitpy` (passing) and `test_forge_models_emitpy_xfail`
   (tracked xfails) to `schedule-nightly.yml`, mirroring the normal
   passing/xfail split. Introduced an `emitpy_xfail` marker so the xfail job
   selects only emitpy-specific failures.

3. **Multi-chip emitpy coverage.**
   Added qb2-blackhole tensor-parallel emitpy to the weekly pipeline
   (`schedule-weekly.yml`, matching qb2 TP's weekly cadence) and galaxy-wh-6u
   tensor-parallel emitpy to the nightly (limited to `gpt_oss/pytorch-20B` as an example).

### Follow-up
EmitPy test durations aren't recorded yet. They currently use the 240-min default
timeout until the first scheduled run + a `manual-update-durations` refresh.

### Checklist

### DIFF
--- a/.github/workflows/manual-test.yml
+++ b/.github/workflows/manual-test.yml
@@ -33,6 +33,7 @@ on:
           - 'model-test-lb-blackhole-weekly.json'
           - 'model-test-emitpy.json'
           - 'model-test-emitpy-xfail.json'
+          - 'model-test-emitpy-weekly.json'
           - 'basic-multihost-test.json'
           - 'Custom'
       test_suite_custom:
--- a/.github/workflows/manual-test.yml
+++ b/.github/workflows/manual-test.yml
@@ -32,6 +32,7 @@ on:
           - 'model-test-lb-blackhole-nightly.json'
           - 'model-test-lb-blackhole-weekly.json'
           - 'model-test-emitpy.json'
+          - 'model-test-emitpy-xfail.json'
           - 'basic-multihost-test.json'
           - 'Custom'
       test_suite_custom:
--- a/.github/workflows/schedule-nightly.yml
+++ b/.github/workflows/schedule-nightly.yml
@@ -95,6 +95,38 @@ jobs:
       dumped_irs_run_id: ${{ github.run_id }}
       target_time_min: 180
 
+  # EmitPy verification: re-run passing torch inference models via the codegen_py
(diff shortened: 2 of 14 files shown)


nsumrakTT (commented):
I think the best way forward is to revisit test_duration calculations as there are several constrains that appeared recently. The main one is that same tests are run on multiple different machines and especially architectures that produces totally different results and we should probably take the worst case result. If emitpy has the same test designator then we should tackle naming.
I don't think this is a way to actually solve the problem.

vvukomanTT (approved):

sgligorijevicTT (approved):

Issue #5495, opened 2026-07-02
vLLM Device count missmatch error on nighlty benchmarks


vkovacevicTT:
vLLM benchmark models on p150-perf (qb2 runners) fail with:
`flatbuffer_loaded_execu:172    ERR| Device count mismatch: 1 vs 4`a

[Example](https://github.com/tenstorrent/tt-xla/actions/runs/28557033742/job/84738819336)

This is the same error we solved in https://github.com/tenstorrent/tt-xla/issues/5102

vkovacevicTT:
tt-xla commit: https://github.com/tenstorrent/tt-xla/commit/b621dabc81061b152919346229f02f398f77721f

Run on previous commit (PASSED): https://github.com/tenstorrent/tt-xla/actions/runs/28582576686/job/84746314521
Run on mentioned commit (FAILED): https://github.com/tenstorrent/tt-xla/actions/runs/28583428341/job/84749142733

PR #5531, opened 2026-07-06
Reland: fix device-count mismatch for no-input graphs (gate on num_partitions)

### LINKED ISSUE #5360: [HunyuanImage-2.1-Distilled-Diffusers] Device count mismatch: 4 vs 1 in transformer

### LINKED ISSUE #5551: [Krea Realtime 14B] transformer - Device count mismatch: 8 vs 1
### Problem description 

- Model fails with `Device count mismatch: 8 vs 1` error - [jul7_krea_realtime_transformer.log](https://github.com/user-attachments/files/29751343/jul7_krea_realtime_transformer.log)

kamalrajkannan78:
### Ticket

- fixes https://github.com/tenstorrent/tt-xla/issues/5360
- relands #5366 (reverted in #5496)
- fixes https://github.com/tenstorrent/tt-xla/issues/5551

### Problem description

- Sharded models that contain a no-input graph hit `Device count mismatch: N vs 1` (e.g. HunyuanImage 2.1 transformer on llmbox, [#5360](https://github.com/tenstorrent/tt-xla/issues/5360)). A const-folded no-input graph (e.g. `torch.arange` in RoPE) is stamped with a 1×1 mesh, which reshapes the device mesh `[1,N] -> [1,1]` and breaks the live sharded buffers. Full root cause: [#5360](https://github.com/tenstorrent/tt-xla/issues/5360).

### What's changed

- A no-input graph now adopts the full mesh when the run targets more than one device (`num_partitions > 1`). `num_partitions` is parsed from `ExecutableBuildOptionsProto` and threaded through `compileMlirProgram -> buildModule -> runCompilerStableHLOPipeline` as the discriminator.
- Removed the now-unused `anyShardedAcrossDevices` helper (its only caller was the simplified condition).
- Added `tests/torch/ops/test_arange.py` with two guards: a sharded case (no-input graph must adopt the full mesh) and a single-chip-on-multichip case (must stay 1×1).

### Why this differs from the reverted #5366

- #5366 used the device-assignment size (`replica_device_ids.size()`), which regressed single-chip vLLM models on a multi-chip host (`p150-perf`): `size()` counts host devices (N) while `num_partitions` is 1, so the no-input graph was wrongly stamped with the host mesh -> `Device count mismatch: 1 vs N` (#5495).
- This reland uses `num_partitions`, which is 1 for a single-chip model regardless of host, so single-chip runs stay 1×1 while genuinely sharded runs (`num_partitions = N`) still get the full mesh.

### Checklist

- [x] Verify the changes through local testing on sanity and model in WH
- [x] Re-verify the regressed case: `vllm_llama_3_1_8b` + `vllm_llama_3_2_1b` on `p150-perf` - https://github.com/tenstorrent/tt-xla/actions/runs/28788094874

### Logs

- [jul6_arange_before_fix.log](https://github.com/user-attachments/files/29703438/jul6_arange_before_fix.log)
- [jul6_arange_after_fix.log](https://github.com/user-attachments/files/29703437/jul6_arange_after_fix.log)
- [jul6_HunyuanImage_2_1_transformer_before_fix.log](https://github.com/user-attachments/files/29703442/jul6_HunyuanImage_2_1_transformer.log)
- [jul6_HunyuanImage_2_1_transformer_after_fix.log](https://github.com/user-attachments/files/29703439/jul6_HunyuanImage_2_1_transformer_after_fix.log)
- [jul7_krea_realtime_transformer_before_fix.log](https://github.com/user-attachments/files/29751391/jul7_krea_realtime_transformer.log)
- [jul7_krea_realtime_transformer_after_fix.log.zip](https://github.com/user-attachments/files/29751389/jul7_krea_realtime_transformer_after_fix.log.zip)

### DIFF
--- a/pjrt_implementation/src/api/module_builder/module_builder.cc
+++ b/pjrt_implementation/src/api/module_builder/module_builder.cc
@@ -100,19 +100,6 @@ moduleHasAnyFuncArguments(const mlir::OwningOpRef<mlir::ModuleOp> &m) {
   return public_func_ops[0].getNumArguments() > 0;
 }
 
-// Returns true if any sharding genuinely partitions a tensor across devices
-// (`MeshShardType::Devices`). Replicate / presharded-replicate shardings do not
-// count, since they do not require a multi-device mesh.
-static bool anyShardedAcrossDevices(
-    const std::vector<mlir::tt::sharding_utils::MeshSharding> &shardings) {
-  return std::any_of(
-      shardings.begin(), shardings.end(),
-      [](const mlir::tt::sharding_utils::MeshSharding &sharding) {
-        return sharding.getShardType() ==
-               mlir::tt::ttcore::MeshShardType::Devices;
-      });
-}
-
 // Maps per-axis fabric config to TTNN mesh topology for CCL operations.
 static std::vector<mlir::tt::ttcore::Topology>
 fabricConfigToMeshTopology(const tt::runtime::MeshFabricConfig &fabricConfig) {
@@ -301,7 +288,7 @@ ModuleBuilder::buildModule(
     const std::string_view &mlir_code,
     const std::string &system_descriptor_path,
     const std::unordered_map<std::string, std::string> &compile_options_map,
-    ClientInstance *client_instance) {
+    ClientInstance *client_instance, size_t target_num_devices) {
   DLOG_F(LOG_DEBUG, "ModuleBuilder::buildModule");
 
   auto compile_options = CompileOptions::parse(compile_options_map);
@@ -399,10 +386,10 @@ ModuleBuilder::buildModule(
       parent_mesh ? std::make_optional(tt::runtime::getMeshShape(*parent_mesh))
                   : std::nullopt;
 
-  status = runCompilerStableHLOPipeline(
-      mlir_module, result_presharded, output_shardings,
-      compile_options.export_path, compile_options.export_model_name,
-      current_mesh_shape);
+  status = runCompilerStableHLOPipeline(mlir_module, result_presharded,
+                                        compile_options.export_path,
+                                        compile_options.export_model_name,
+                                        current_mesh_shape, target_num_devices);
   if (!tt_pjrt_status_is_ok(status)) {
     return {status, nullptr};
   }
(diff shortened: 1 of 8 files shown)


sshonTT (approved):

kamalrajkannan78:
@acicovicTT could you pls review this PR?

pjrt_implementation/src/api/module_builder/module_builder.cc
```diff
@@ -787,27 +774,21 @@ std::vector<int64_t> ModuleBuilder::collectResultPresharded(
...
-  // when a result is genuinely device-sharded; otherwise let it default to a
-  // single device.
+  // A no-input graph on a multi-device run must adopt the full mesh: a no-input
+  // replicated value (e.g. a const-folded torch.arange) would otherwise default
+  // to 1x1, collapse the mesh, and break the live sharded buffers.
+  // See https://github.com/tenstorrent/tt-xla/issues/5360
   if (current_mesh_shape.has_value() && current_mesh_shape->size() == 2 &&
-      !moduleHasAnyFuncArguments(mlir_module) &&
-      anyShardedAcrossDevices(output_shardings)) {
+      !moduleHasAnyFuncArguments(mlir_module) && target_num_devices > 1) {
```
acicovicTT, line 791: anyShardedAcrossDevices was added to specifically address https://github.com/tenstorrent/tt-xla/issues/5102 . Did you run this test to confirm it fixes the issue as well? i.e. is `target_num_devices > 1` an equivalent change?

acicovicTT: Are we sure that `num_partitions` is set correctly always, both by pytorch and JAX frameworks?

kamalrajkannan78: > anyShardedAcrossDevices was added to specifically address https://github.com/tenstorrent/tt-xla/issues/5102 . Did you run this test to confirm it fixes the issue as well? i.e. is target_num_devices > 1 an equivalent change?

I ran the shared test case on my branch and it passes — https://github.com/tenstorrent/tt-xla/actions/runs/29017420753/job/86120276056

On equivalence: for #5102's models this is equivalent — they're single-chip (num_partitions=1), so `target_num_devices > 1` is false and the no-input graph stays 1×1, same as `anyShardedAcrossDevices` did. They only diverge for a multi-device run with a replicated no-input result (the case this PR fixes).

kamalrajkannan78: > Are we sure that num_partitions is set correctly always, both by pytorch and JAX frameworks?

Verified on JAX by logging num_partitions locally — it is set correctly and tracks the SPMD device count: 8 for tensor-parallel across 8 devices, 1 for single-chip. Same as torch-xla SPMD (8 sharded / 1 single-chip). So the discriminator is reliable under both frameworks.

- JAX tensor-parallel (`tests/jax/multi_chip/llmbox/8_devices/graphs/tensor_parallel/test_dot_psum.py`):
  `[numpart-debug] num_partitions=8 replica_device_ids.size=8`
- JAX single-chip (`tests/jax/single_chip/test_ranks.py`):
  `[numpart-debug] num_partitions=1 replica_device_ids.size=1`

### Logs

- [jul9_jax_single_chip.log](https://github.com/user-attachments/files/29849964/jul9_jax_sc.log)
- [jul9_jax_tensor_parallel.log](https://github.com/user-attachments/files/29849965/jul9_jax_tensor_parallel.log)

acicovicTT (approved):

PR #5539, opened 2026-07-06
Jackzhang/tt lang integration

jackzhangTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/4813#event-25777791477)

### Problem description
Re landing the tt-lang integration change with an updated tt-lang (v1.1.5) version that drops numpy dependencies and has a more up to date version of metal. This is the first POC integration of tt-lang into the tt-xla compiler pipeline.

### What's changed
In the current flow, the user's tt-lang @operation kernel is captured at decoration time and stored in a Python kernel registry. The kernel is then emitted as an opaque stablehlo.custom_call in the graph and processed through the StableHLO → TTIR → TTNN pipeline without the kernel body, the body lives only in the Python registry, only metadata (kernel_id, arg_roles, etc.) are kept in the IR.

We defer invoking the tt-lang compiler until the TTNN stage, because tt-lang needs the post-Shardy, shard-local shapes and the final TTNN layout encoding on each operand. This is handled by two tt-mlir passes appended to the TTIR→TTNN pipeline (not by the plugin walking the module):

--ttnn-resolve-tt-lang-kernels walks each ttnn.tt_lang_op, invokes the tt-lang compiler via pybind11 (tt_torch.tt_lang.resolve_operation) with the final operand types and grid, and attaches the returned artifact as the kernel_artifact attribute (JSON produced by the Python side).
--ttnn-lower-tt-lang-to-generic parses that kernel_artifact and rewrites the ttnn.tt_lang_op into an equivalent ttnn.generic op (building a #ttnn.program with the kernels' C++ source inline and per-operand TensorAccessor markers).

By the time TTNN→flatbuffer emission runs, the op is already a plain ttnn.generic; the flatbuffer emitter needs no tt-lang-specific handling and does not read kernel_artifact. TensorAccessor compile-time args are filled in by the tt-mlir runtime at execute time.

### Tests added

**Python unit tests** (`tests/torch/ops/test_tt_lang_kernel.py`, 27 tests):
- `@tt_torch.tt_lang_operation` decorator: required `arg_roles`, rejection of non-XLA tensors, rejection of zero "out" operands.
- Kernel registry: registration and lookup by `kernel_id`.
- `torch.ops.tt.tt_lang_op` dispatch: XLA-only enforcement, `_validate_tt_lang_op_out_indices` rejects zero outs / duplicate indices / out-of-range indices.
- Device-less compile path: `_StubTtnnTensor` / `_StubTtnnDevice` plumbing and the L1-budget probe bypass.
- Artifact serialization: `_ARTIFACT_FORMAT_VERSION` round-trip, JSON schema validation.

**Hardware e2e tests** (`tests/torch/ops/test_tt_lang_kernel_e2e.py`, 2 tests × 2 shape params = 4 invocations, n150):

- `test_tt_lang_eltwise_add_e2e` — Bare-kernel path. A tt-lang tilewise elementwise-add (`examples/eltwise_add.py`-style: explicit reader/compute/writer threads, 2-block CB pipeline, `grid="auto"`) is wrapped with `@tt_torch.tt_lang_operation`, called from a `torch.nn.Module` on the XLA device, and exercised end-to-end:
  `@tt_torch.tt_lang_operation` → `stablehlo.custom_call @tt.tt_lang_op` → TTIR → TTNN → embedded-Python bridge resolves the kernel → `kernel_artifact` populated → flatbuffer `GenericOp` → execute on silicon → bf16 PCC against torch golden.

- `test_tt_lang_add_stitched_with_torch_add_e2e` — **Graph-stitching path.** Same kernel, but wrapped between two regular `torch.add` ops inside a single XLA program:

### DIFF
--- a/.github/workflows/call-test.yml
+++ b/.github/workflows/call-test.yml
@@ -198,16 +198,7 @@ jobs:
 
         echo "=== Installing wheel ==="
         source venv/activate
-        if [[ "${{ matrix.build.extra-wheel }}" == "vllm" ]]; then
-          # tt-lang's wheel metadata caps numpy<2, but tt-lang works with numpy 2.x.
-          # vllm (via opencv-python-headless>=4.13) requires numpy>=2, so pjrt_plugin_tt
-          # and vllm_tt cannot co-resolve. Override the stale cap; every other
-          # dependency still resolves normally.
-          echo 'numpy>=2' > wheels/numpy-override.txt
-          uv pip install --override wheels/numpy-override.txt wheels/*.whl
-        else
-          uv pip install wheels/*.whl
-        fi
+        uv pip install wheels/*.whl
 
     - name: Install custom torch_xla from workflow run
       if: ${{ inputs.torchxla_build_run_id != '' }}
--- a/.github/workflows/call-test.yml
+++ b/.github/workflows/call-test.yml
@@ -198,7 +198,16 @@ jobs:
 
         echo "=== Installing wheel ==="
         source venv/activate
-        uv pip install wheels/*.whl
+        if [[ "${{ matrix.build.extra-wheel }}" == "vllm" ]]; then
+          # tt-lang's wheel metadata caps numpy<2, but tt-lang works with numpy 2.x.
+          # vllm (via opencv-python-headless>=4.13) requires numpy>=2, so pjrt_plugin_tt
+          # and vllm_tt cannot co-resolve. Override the stale cap; every other
+          # dependency still resolves normally.
+          echo 'numpy>=2' > wheels/numpy-override.txt
+          uv pip install --override wheels/numpy-override.txt wheels/*.whl
+        else
+          uv pip install wheels/*.whl
+        fi
 
     - name: Install custom torch_xla from workflow run
       if: ${{ inputs.torchxla_build_run_id != '' }}
--- a/.github/workflows/pr-main.yml
+++ b/.github/workflows/pr-main.yml
@@ -42,7 +42,7 @@ jobs:
     with:
       docker_image: ${{ needs.build-image.outputs.docker-image }}
       docker_image_manylinux: ${{ needs.build-image.outputs.docker-image-manylinux }}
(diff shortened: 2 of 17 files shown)


acicovicTT (approved):

AleksKnezevic (approved):

.github/workflows/call-test-tools.yml
```diff
@@ -80,7 +80,10 @@ jobs:
 
           echo "=== Installing wheel ==="
           pip install --upgrade pip
-          pip install wheels/*.whl
+          # tt-lang url
+          pip install \
+            --extra-index-url https://pypi.eng.aws.tenstorrent.com/2026-07/ \
+            wheels/*.whl
```
jameszianxuTT: Why not point this to a specific wheel? it looks like there are several in the folder.

seems like the logic here is like
- install the local wheels in wheels/*.whl
- if any deps need to be located, eg. tt-lang as dep of tt-xla whl, then search for the newest rev in `https://pypi.eng.aws.tenstorrent.com/2026-07/`

I think the pinning to this folder rather than a specific wheel url will lead
- to RNG install and automatic uplift if anyone publishes to the pypi server or yanks a wheel
- lack of clarity on what to change during uplift

In this call-test-tools we could just specifically pip install the expected tt-lang wheel from an exact URL (even with no-deps) then install everything in local wheels/*.whl. In the rest of the places where this pattern is used I think we should just install `tt-lang@<exact URL>` like we do for torch-xla instead of pulling the latest thing from that folder

hshahTT: +1. This should be synced with the version of tt-lang present in requirements.txt and requirements-dev.txt (1.1.5.dev20260704+light)

nsumrakTT: Why add extra index here? This workflow test tools inside tt-xla wheel (ttnn, tracy...)

jackzhangTT: changed to exact url

jackzhangTT: sorry this was not needed I removed it

jameszianxuTT (approved):

PR #5542, opened 2026-07-06
[vLLM] Speculative decode with NGram

### LINKED ISSUE #5706: [vLLM] Add support for NGram speculative decode

mmanzoorTT:
### Ticket
closes #5706 

### What's changed
Add support for NGram speculative decoding

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -872,7 +872,7 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         # runtime graph may appear during inference.
         if (
             self.speculative_config is not None
-            and new_compiled_graphs == 1
+            and new_compiled_graphs > 0
             and case_str == "execute_model"
         ):
             logger.warning(
@@ -2017,6 +2017,8 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
 
         # Convert sampled_token_ids to list of lists for ngram proposer
         sampled_token_ids_list = sampled_token_ids.cpu().tolist()
+        if not sampled_token_ids_list:
+            return
         if not isinstance(sampled_token_ids_list[0], list):
             # If it's a 2D tensor with single column, wrap each id in a list
             sampled_token_ids_list = [[int(tid)] for tid in sampled_token_ids_list]
@@ -2031,6 +2033,10 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         for req_idx in discard_req_indices:
             if 0 <= req_idx < len(sampled_token_ids_list):
                 sampled_token_ids_list[req_idx] = []
+
+        # No accepted tokens remain after filtering/discard; skip drafting.
+        if not any(sampled_token_ids_list):
+            return
 
         # Get token counts for each request (number of tokens already in context)
         num_tokens_no_spec = np.array(
@@ -2316,10 +2322,7 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                 # gather_logprobs graph fails trace at opt_level=1, so
                 # fall back to CPU — post-processing is cheap and only
                 # the sampled-token id + logits need to move to host.
-                logits_for_logprobs = self._extract_sampling_logits(
-                    logits, sampling_metadata
-                )
-                logits_cpu = logits_for_logprobs.cpu()
+                logits_cpu = logits.cpu()
                 tokens_cpu = selected_token_ids.cpu()
                 logprobs = self.sampler.gather_logprobs(
                     self.sampler.compute_logprobs(logits_cpu),
@@ -2327,10 +2330,7 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                     token_ids=tokens_cpu.squeeze(-1),
                 )
             else:
-                logits_for_logprobs = self._extract_sampling_logits(
-                    logits, sampling_metadata
-                )
-                logprobs = self.gather_logprobs(logits_for_logprobs, selected_token_ids)
+                logprobs = self.gather_logprobs(logits, selected_token_ids)
 
             # Remove padding on cpu and keep dynamic op outside of xla graph.
             selected_token_ids = selected_token_ids.cpu()[:num_reqs]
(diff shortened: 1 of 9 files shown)


integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2105,12 +2410,61 @@ def sample_tokens(
...
+        logits: torch.Tensor,
+        sampling_metadata: XLASupportedSamplingMetadata,
+    ) -> torch.Tensor:
+        """Return one sampling-logit row per request.
+
+        In spec decode, postprocess computes logits for packed draft+bonus
+        positions with shape [batch * (num_spec_tokens + 1), vocab]. Sampling
+        always runs on the bonus row (last packed column) per request.
+        """
+        return logits
```
alinakhanTT: _extract_sampling_logits returns logits on the very first line, so the whole body below it never runs. Leftover from debugging maybe?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -3182,6 +3650,7 @@ def capture_model(self) -> None:
         """
         Precompile all the subgraphs with possible input shapes.
         """
+        # return
```
kmabeeTT: commented out return should be dropped

tests/integrations/vllm_plugin/sampling/test_speculative_decode.py
```diff
@@ -0,0 +1,142 @@
...
+    return logits
+
+
+def _make_rejection_sampler() -> RejectionSampler:
+    return RejectionSampler(torch.nn.Identity())
+
+
+@pytest.mark.push
+@pytest.mark.cpu
+def test_speculative_greedy_fallback_accepts_all_matching_drafts():
```
kmabeeTT, line 28: Couple test ideas

- These all run with a single request, so the `start = cu[req_idx - 1]` / `flat_idx` offset math that packs multiple requests into one flat tensor never gets exercised which seems like an easy part to get subtly wrong while still passing current tests.   Consider adding batch>1 case with 2 different draft lengths?


- How about `draft_len == 0` case too whichw ould hit `output[req, 0] = bonus` branch in `_rejection_sample_fallback` (currently uncovered).

integrations/vllm_plugin/vllm_tt/sampler.py
```diff
@@ -197,6 +203,11 @@ def sample(
                 sampling_metadata.repetition_penalties,
             )
 
+        if sampling_metadata.all_greedy:
+            return self.greedy_sample(logits)
+
+        assert sampling_metadata.temperature is not None
+
```
kmabeeTT, line 209: Wasn't sure about why this is here, but I guess it's because of the new call site for the bonus token sampling and an all-greedy request has `temperature=None` which would have fired the assert unless it got moved lower here?

tests/integrations/vllm_plugin/sampling/test_speculative_decode.py
```diff
@@ -0,0 +1,142 @@
...
+import numpy as np
+import pytest
+import torch
+from vllm_tt.metadata import XLASupportedSamplingMetadata
+
+from integrations.vllm_plugin.vllm_tt.model_runner import TTModelRunner
+from integrations.vllm_plugin.vllm_tt.rejection_sampler import (
+    _PLACEHOLDER_TOKEN_ID,
+    RejectionSampler,
+)
```
kmabeeTT: These imports are causing errors on CI during collection time.... Should we use the `vllm_tt.` root like the other tests here instead of `integrations.vllm_plugin.vllm_tt.` ?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -826,13 +842,25 @@ def _verify_num_xla_graphs(self, case_str):
 
         curr_cached_graph = xr.get_num_cached_compilation_graph()
         new_compiled_graphs = curr_cached_graph - self.num_xla_graphs
+        self.num_xla_graphs += new_compiled_graphs
+
+        # Spec decode runs rejection sampling outside compiled TT graphs, so one
+        # runtime graph may appear during inference.
+        if (
+            self.speculative_config is not None
+            and new_compiled_graphs == 1
```
kmabeeTT: Hmm eager rejection path does run a few device ops (the `bonus_logits` gather + `self.sampler` call), and their shapes vary with draft length, so I'm not sure the new-graph count is always exactly 1... would >=1 be safer?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1825,6 +1970,129 @@ def _restore_model_hidden_states(
...
+        """
+        self._draft_token_req_ids = []
+        self._draft_token_ids = []
+
+        if not self.drafter or not self.num_spec_tokens:
+            return
+
+        # Convert sampled_token_ids to list of lists for ngram proposer
+        sampled_token_ids_list = sampled_token_ids.cpu().tolist()
+        if not isinstance(sampled_token_ids_list[0], list):
```
kmabeeTT, line 2022: Not sure if valid, but would this ever hit `IndexError` if sampled_token_ids_list is empty? Consider adding to early return unless it's never going to happen.

kmabeeTT (commented):
Nice - overall looks great! A few comments and suggestions added

integrations/vllm_plugin/vllm_tt/model_runner.py
mmanzoorTT: Yes, you are right. Claude has added `_extract_sampling_logits` but we don't need this in my opinion. I am running tests with it and will remove it if we don't need it.

integrations/vllm_plugin/vllm_tt/sampler.py
mmanzoorTT: Yes, this is right.

integrations/vllm_plugin/vllm_tt/model_runner.py
mmanzoorTT: remvoed

integrations/vllm_plugin/vllm_tt/model_runner.py
mmanzoorTT: Removed. thanks for pointing it out.

tests/integrations/vllm_plugin/sampling/test_speculative_decode.py
mmanzoorTT: Fixed. thanks.

kmabeeTT (approved):
Thx for changes!

alinakhanTT (approved):

tests/integrations/vllm_plugin/sampling/test_speculative_decode.py
mmanzoorTT: Thanks for the suggestion. tests added.

PR #5557, opened 2026-07-08
[CI] Upgrade process irs workflow

nsumrakTT:
### Ticket
https://github.com/tenstorrent/tt-forge/issues/953

### Problem description
- Problem of huge data on the dump irs side
- Problem of how to slit processing on multiple machines 

### What's changed
- Removed ttnn's during collect-prune process
- Increased max size to 200MB
- During processing:
  * Remove duplicate models favoring n150 jobs
  * Calculate time based on IR's size: 1MB for 2 minutes of processing (average based on a few trials)
  * Split model processing on required number of jobs based on job time target

### Checklist
Process IRs workflow manual run:
https://github.com/tenstorrent/tt-xla/actions/runs/29006683783

### DIFF
--- a/.github/workflows/call-process-dumped-irs.yml
+++ b/.github/workflows/call-process-dumped-irs.yml
@@ -33,6 +33,10 @@ on:
         required: false
         type: boolean
         default: false
+      model_filter:
+        description: 'Optional filter for specific models to process'
+        required: false
+        type: string
 
 env:
   GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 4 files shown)


vvukomanTT (approved):

.github/workflows/call-process-dumped-irs.yml
```diff
@@ -45,37 +63,56 @@ jobs:
...
+          echo "Matched model names:"
+          find collected_irs -mindepth 2 -maxdepth 2 -type d -printf '%f\n' | sort -u | grep '${{ inputs.model_filter }}' || true
+
+          echo "=== Deduplicating models (prefer n150 copies) ==="
+          python3 .github/scripts/ir_process_analyzer.py dedupe \
+            --root collected_irs \
+            --preferred-machine n150
+
+          echo "=== Removing .ttnn files from collected_irs ==="
+          find collected_irs -type f -name "*.ttnn" -print -delete
```
jameszianxuTT, line 101: Does this silently drop the TTNN IR from the collection? What if someone wants the TTNN IR?

Also, the IR can be quite different on singledevice (n150) and multidevice (eg. llmbox) machines. Will this discard the multidevice IR if the model is also run on singledevice?

nsumrakTT: The workflow purpose is to process irs, ttnn is not needed and increase the size of artifacts significantly and makes processing of large (nightly, weekly) workflows difficult. If someone needs ttnn, one can run dump locally. If in the future ttnn is needed in CI, we'll make changes accordingly.

Yes, I agree, but currently this is the solution so that basic workflow exists and generates database reports. Whole op-by-op is still wip (should be debugged, run on bh, ...) and it is up to devs and leads to decide and resolve such issues in subsequent PRs.

PR #5581, opened 2026-07-10
Install vllm package for single-test workflow

### LINKED ISSUE #5580: Single-test workflow failure for vLLM tests
vLLM is not installed for single-test workflow and vLLM test(s) fail due to missing vllm package. 

https://github.com/tenstorrent/tt-xla/actions/runs/29063370401/job/86269851917

mmanzoorTT:
### Ticket
closes #5580 

### Problem description
vLLM package is never installed for single test workflow.

### What's changed
Add an option to install vllm wheel

### DIFF
--- a/.github/workflows/call-test.yml
+++ b/.github/workflows/call-test.yml
@@ -186,7 +186,7 @@ jobs:
           --repo ${{ github.repository }} \
           --dir wheels \
           --name ${{ steps.strings.outputs.wheel_artifact_name }}
-        if [[ "${{ matrix.build.extra-wheel }}" == "vllm" ]]; then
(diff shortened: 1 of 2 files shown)


vvukomanTT:
I'd avoid using the dir to infer whether we need the vllm wheel installed on not, since we already have a mechanism for that -> `matrix.build.extra-wheel`.
My idea is:
1. Add an input in the `manual-test-single.yml` workflow (can be boolean, can be a multichoice) which will give an option for the vllm wheel to be installed as well
2. In `generate-test-config` in the same file just expand the constructed `TEST_CONFIG` with the `extra-wheel` field

`call-test.yml` can remain unchanged

nsumrakTT (approved):

vvukomanTT (approved):

PR #5584, opened 2026-07-10
[CI] Add vLLM nightly model-output summary via Claude skill

devisettymahidhar608:
### Ticket
Link to Github Issue

### Problem description
Nightly runs include many vLLM jobs (`run_vllm_*_tests`), but unlike torch/jax
there is no useful model-output summary in the run UI. As more vLLM models are
added, it becomes hard to open each job log and check what each model generated.
### What's changed
- Added Claude skill `.claude/skills/vllm-nightly-log` to collect generated
  text from all vLLM jobs in a run and write a single summary log.
- Wired the skill into nightly CI (`schedule-nightly.yml`) so after
  `nightly_tests` finishes, a **vLLM nightly model outputs** job runs the skill
  with `${{ github.run_id }}` and publishes the summary to the job summary +
  artifact.
- Validated the same flow via **Run Test** (`manual-test.yml`) using
  `vllm-model-tests-nightly.json`:
  - Skill summary produced:
    https://github.com/tenstorrent/tt-xla/actions/runs/29189301865
    https://github.com/tenstorrent/tt-xla/actions/runs/29193245689

### Checklist

### DIFF
--- a/.github/workflows/schedule-nightly.yml
+++ b/.github/workflows/schedule-nightly.yml
@@ -188,6 +188,105 @@ jobs:
           path: nightly-reports/
           retention-days: 90
 
+  # Runs the vllm-nightly-log Claude skill after nightly_tests finishes and
+  # publishes model generation outputs to this job's GitHub Actions summary
+  # (sidebar: "vLLM nightly model outputs"). Unlike torch/jax, pytest's
+  # test-summary does not surface vLLM generations — this fills that gap.
+  vllm-nightly-log-report:
+    name: "vLLM nightly model outputs"
+    needs: [ nightly_tests ]
+    if: always() && needs.nightly_tests.result != 'cancelled'
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      contents: read
+      actions: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 1
+          submodules: false
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: '3.12'
+
+      - name: Install Claude Code CLI
+        run: |
+          npm install -g @anthropic-ai/claude-code
+          claude --version
+
+      - name: Run vllm-nightly-log skill via Claude
+        env:
+          GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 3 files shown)


sonalibaskaran2499 (approved):

vvukomanTT:
Why not just expand the skill that is ran in the step `Run ci-benchmark-analyzer skill via Claude` in `schedule-nightly.yml`?

devisettymahidhar608:
> Why not just expand the skill that is ran in the step Run ci-benchmark-analyzer skill via Claude in schedule-nightly.yml?

Hi @vvukomanTT , The goal here isn't to produce a detailed CI or performance report. Instead, it's intended to provide a quick nightly overview of vLLM generations so we can easily verify whether models are producing coherent output for the given prompts, without needing to open every run_vllm_*_tests job as the number of models continues to grow.
ci-benchmark-analyzer serves a different purpose:

it focuses on perf-benchmark jobs (metrics, regressions, and failures)
it does not process nightly_tests or run_vllm_*_tests
it does not extract the generated prompt: / output: text from vLLM logs

Because of these differences, extending that skill would require broadening its scope beyond its current purpose rather than making a small enhancement. Keeping vllm-nightly-log as a separate, lightweight skill/job allows it to remain focused on providing a concise nightly summary of vLLM generations for quick validation of output quality.

Thank You

kmabeeTT:
No strong opinions here, don't know what the best way to do this would be or what we currently have, but I welcome a change that makes monitoring health of tests easier, in general.  Perhaps there is a way to make the existing summary include vllm tests though, with a smaller change (but would lose the extra details like prompt/output, etc).

.github/workflows/schedule-nightly.yml
```diff
@@ -188,6 +188,105 @@ jobs:
           path: nightly-reports/
           retention-days: 90
 
+  # Runs the vllm-nightly-log Claude skill after nightly_tests finishes and
+  # publishes model generation outputs to this job's GitHub Actions summary
+  # (sidebar: "vLLM nightly model outputs"). Unlike torch/jax, pytest's
+  # test-summary does not surface vLLM generations — this fills that gap.
+  vllm-nightly-log-report:
```
vvukomanTT, line 195: Let's move this job to a separate workflow (e.g. call-vllm-nightly-log-report.yml) and just add a call to it from here.
(I know that this wasn't done for the identical job above, my opinion on it is the same, that will have to get refactored later)

PR #5586, opened 2026-07-10
[WAN] umt5 sharding

vkovinicTT:
### Ticket
Closing #4864 

### What's changed
Refactored sharding a bit so it's more coherent with megatron-style sharding. Metal also shards this text encoder on data parallel axis, but since this is run only once (at most twice for negative prompt) it's fine to stay as is.


sgligorijevicTT (approved):
looks fine

Issue #5578, opened 2026-07-09
[vLLM] Investigate tests/models running with opt_level=0


mmanzoorTT:
vLLM uses opt_level=1 as default option. However, few tests/models are running with opt_level=0 due to compilation/execution failures. We need to investigate the models running with opt_level=0 and may ask for adding support in tt-mlir optimizer.

alinakhanTT:
Investigated all the vLLM tests pinned at `opt_level=0`. Summary:

**Flipped to `opt_level=1`** (compile and generate coherent output), in #5589: Qwen3-0.6B (2D mesh and data parallel), Qwen3-32B, Mistral-Large-Instruct-2411, and Pixtral-Large-Instruct-2411 (confirmed on the WH 6U galaxy).

**Genuinely fail at `opt_level=1`, kept at `opt_level=0`** with a tracking issue:
- Mistral-Small-24B, mm-encoder DRAM OOM: tenstorrent/tt-mlir#9006
- gemma-4-31B `[1,4]` and `[8,4]`, SDPA decode tree-reduction limit: tenstorrent/tt-mlir#9007 (`[8,4]` confirmed by bh_galaxy CI, same `Bad StatusOr / Error code: 13` at the fused prefill postprocess compile)
- gemma-4-26B-A4B, SDPA decode tree-reduction limit (same as 31B): tenstorrent/tt-mlir#9007, single-device 2D-mesh TP on BH QuietBox, 110 cores/head, confirmed in CI run [29358116829](https://github.com/tenstorrent/tt-xla/actions/runs/29358116829)
- gemma-4-E4B, SDPA decode circular buffers overflow L1 at opt_level=1: opt-level=1 SDPA decode circular buffers exceed L1 on gemma-4-E4B-it (2318272 B > 1572864 B) tenstorrent/tt-mlir#9039 (distinct from #9007, L1 CB overflow, not the tree-reduction limit; confirmed in CI run [29358116829]( https://github.com/tenstorrent/tt-xla/actions/runs/29358116829))
- DeepSeek-V2-Lite, MoE all_to_all_dispatch needs row-major layout at opt_level=1: tenstorrent/tt-mlir#8920 (closed but still reproduces)
- Llama-3.1-70B, garbage output at `opt_level=1`: #4325

PR #5589, opened 2026-07-10
[vLLM] Flip stale opt_level=0 pins to opt_level=1

alinakhanTT:
### Ticket
Part of #5578

### Problem
Several vLLM tests explicitly set optimization_level=0 even though the plugin now defaults to 1.

### What changed
Verified that the following models compile and produce correct output with opt_level=1, then removed the redundant optimization_level=0 override so they use the plugin default:
- Qwen3-0.6B (2D mesh)
- Qwen3-0.6B (all data-parallel configs)
- Qwen3-32B
- Mistral-Large-Instruct-2411
- Pixtral-Large-Instruct-2411

Also renamed the two galaxy_wh_6u_large tests to ..._mistral_large and ..._pixtral_large, since they previously shared a function name and only one test was collected.
Models that still fail with opt_level=1 remain pinned to optimization_level=0, with comments linking the corresponding tracking issues.


tests/integrations/vllm_plugin/generative/test_tensor_parallel_generation.py
mmanzoorTT: Let's remove opt_level completely. 
```suggestion
        pytest.param("Qwen/Qwen3-0.6B", True),
       pytest.param("Qwen/Qwen3-0.6B", False),
   ],
)
def test_tensor_parallel_generation_llmbox_small(
model_name: str,
use_2d_mesh: bool,
):
```

alinakhanTT: Done

mmanzoorTT (approved):

mmanzoorTT (approved):

Issue #4708, opened 2026-05-14
integrate tt-lang into tt-xla


jackzhangTT:
### Is your feature request related to a problem? Please describe

integrate tt-lang into tt-xla flow

### Describe the solution you'd like

tt-lang kernels are registered in a process global registry for all kernels and compiled via a seperate pybind11 process and stored in a cache

PR #5590, opened 2026-07-10
[tt-lang] Add custom Shardy sharding rules for tt-lang ops

jackzhangTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/4708)

### Problem description
tt-lang operations lower to `stablehlo.custom_call @tt.tt_lang_op`, which is opaque to Shardy's built-in sharding-rule inference: without a user-supplied rule, propagation is forced to replicate operands and results, so users have no way to guide how a tt-lang kernel is partitioned across devices.

### What's changed
On the tt-xla side this adds an optional sharding_rule parameter to the @tt_lang_operation decorator, which is emitted as the xla.sdy.sharding_rule frontend attribute on the op's stablehlo.custom_call, plus a make_sharding_rule builder that constructs the rule text via jaxlib's Shardy (sdy) bindings.

NOTE: This should not go in until this [PR](https://github.com/tenstorrent/tt-mlir/pull/8986) is landed and uplifted into tt-xla

### Checklist


python_package/tt_torch/custom_ops.py
```diff
@@ -2360,6 +2371,7 @@ def tt_lang_op_dispatch(
     version_tag: str,
     shard_spec: str,
     out_indices: Sequence[int],
+    sharding_rule: str = "",
```
sshonTT, line 2426: Nit: Can we add a warning for tensors that will be implicitly replicated when the sharding_rule is empty?

python_package/tt_torch/sharding_rule.py
```diff
@@ -0,0 +1,197 @@
...
+def make_sharding_rule(
+    operand_mappings: Sequence[Sequence[str]],
+    result_mappings: Sequence[Sequence[str]],
+    factor_sizes: Mapping[str, int],
+    *,
+    reduction_factors: Iterable[str] = (),
+    need_replication_factors: Iterable[str] = (),
+    permutation_factors: Iterable[str] = (),
+    blocked_propagation_factors: Iterable[str] = (),
+    is_custom: bool = True,
```
sshonTT: I assume that tt_lang always set this `is_custom` arg to True?

python_package/tt_torch/custom_ops.py
sshonTT: In addition to this, I wonder if we could always pass a sharding rule for every tt-lang op. Even if the rule is empty, we could explicitly set the factors for all arguments to be replicated (perhaps using around [here](https://github.com/tenstorrent/tt-xla/pull/5590/changes#diff-c52912c97fbf37decc7a285a62325a202e9d858bed0609edeeb7d6fe24e2c81bR106)). What do you think?

I understand tt-mlir stablehlo pipeline will take care of this case, but I'd like to keep tt-mlir side logic simple.

tests/torch/ops/test_sharding_rule.py
```diff
@@ -0,0 +1,516 @@
...
+
+def test_make_sharding_rule_matmul_shape():
+    """A canonical matmul rule matches the printer format Shardy uses."""
+    rule = make_sharding_rule(
+        operand_mappings=[("B", "M"), ("M", "N")],
+        result_mappings=[("B", "N")],
+        factor_sizes={"B": 8, "M": 16, "N": 32},
+        reduction_factors=["M"],
+    )
+    assert rule == (
```
sshonTT: Couldn't we check the sharding rule logic without an exact string check? I think a string check might be vulnerable to future jaxlib upgrades. If it's hard to find an alternative right now, feel free to leave it as is and just open an issue for it.

sshonTT: I dug with claude code and found this. Not validated yet so just for reference.


```
from jaxlib.mlir import ir
from jaxlib.mlir.dialects import sdy

def assert_rule(text, *, factor_sizes, operands, results,
                reduction=(), need_replication=(), permutation=(),
                blocked=(), is_custom=True):
    with ir.Context() as ctx, ir.Location.unknown():
        sdy.register_dialect(ctx)
        tm = lambda dims: sdy.TensorMappingAttr.get(
            [sdy.DimMappingAttr.get([i]) for i in dims])
        want = sdy.OpShardingRuleAttr.get(
            factor_sizes=list(factor_sizes),
            operand_mappings=[tm(d) for d in operands],
            result_mappings=[tm(d) for d in results],
            reduction_factors=list(reduction),
            need_replication_factors=list(need_replication),
            permutation_factors=list(permutation),
            blocked_propagation_factors=list(blocked),
            is_custom=is_custom,
        )
        assert ir.Attribute.parse(text) == want
```

And from the test
```
def test_make_sharding_rule_matmul_shape():
    rule = make_sharding_rule(
        operand_mappings=[("B", "M"), ("M", "N")],
        result_mappings=[("B", "N")],
        factor_sizes={"B": 8, "M": 16, "N": 32},
        reduction_factors=["M"],
    )
    assert_rule(rule, factor_sizes=[8, 16, 32],
                operands=[[0, 1], [1, 2]], results=[[0, 2]], reduction=[1])
```

tests/torch/ops/test_sharding_rule.py
```diff
@@ -0,0 +1,516 @@
...
+    a_xla = a_cpu.to(device)
+    b_xla = b_cpu.to(device)
+    out_xla = torch.zeros_like(a_cpu).to(device)
+
+    xs.mark_sharding(a_xla, mesh, ("model", None))
+    xs.mark_sharding(b_xla, mesh, ("model", None))
+    xs.mark_sharding(out_xla, mesh, ("model", None))
+
+    result_xla = add_op(a_xla, b_xla, out_xla)
+    xm.mark_step()
```
sshonTT: nit: we don't need this if we move tensor to cpu from next line.

tests/torch/ops/test_sharding_rule.py
```diff
@@ -0,0 +1,516 @@
...
+    b_cpu = torch.randn(rows, cols, dtype=torch.bfloat16)
+    golden = a_cpu + b_cpu
+
+    device = xm.xla_device()
+    a_xla = a_cpu.to(device)
+    b_xla = b_cpu.to(device)
+    out_xla = torch.zeros_like(a_cpu).to(device)
+
+    xs.mark_sharding(a_xla, mesh, ("model", None))
+    xs.mark_sharding(b_xla, mesh, ("model", None))
```
sshonTT, line 665: If the purpose of this test is to verify propagation through Shardy, we don't need this explicit mark_sharding. It should be propagated automatically.

tests/torch/ops/test_sharding_rule.py
```diff
@@ -0,0 +1,516 @@
...
+    xs.mark_sharding(out_xla, mesh, ("model", None))
+
+    result_xla = add_op(a_xla, b_xla, out_xla)
```
sshonTT: This is just out of curiosity, why do we need result_xla here? Isn't the output(out_xla)  inplaced for this op?

tests/torch/ops/test_sharding_rule.py
```diff
@@ -0,0 +1,516 @@
...
+    def add_op(a, b, out):
+        return _ttl_add(a, b, out)
+
+    return add_op
+
+
+@pytest.mark.push
+@pytest.mark.dual_chip
+@pytest.mark.llmbox
+def test_tt_lang_eltwise_add_custom_sharding_rule_e2e(clean_registry):
```
sshonTT: Can we add an e2e test where propagation is prohibited by the need_replication case? Something like this:

```
sharding_rule = make_sharding_rule(
        operand_mappings=[("i", "j"), ("i", "j")],
        result_mappings=[("i", "j")],
        factor_sizes={"i": rows, "j": cols},
        need_replication_factors=[("i, j")]
    )

xs.mark_sharding(a_xla, mesh, ("model", None))
```

tests/torch/ops/test_sharding_rule.py
jackzhangTT: Adopted this approach — thanks for the sketch. It's now the `assert_rule` helper in `test_sharding_rule.py`.

tests/torch/ops/test_sharding_rule.py
jackzhangTT: Removed — `.to(\"cpu\")` is enough to sync.

tests/torch/ops/test_sharding_rule.py
jackzhangTT: Updated — only `a` is `mark_sharding`'d now; the pointwise rule is expected to propagate onto the result (asserted via `_get_xla_sharding_spec`).

tests/torch/ops/test_sharding_rule.py
jackzhangTT: Its really a quirk of the way I modeled the tt-lang ops, as a custom SHLO op it's not DPS so it has a return value. However, the actual kernel when lowered in the MLIR side is actually DPS so if I remoevd the result_xla and only used out_xla it would also work. But having the result_xla just makes it a bit more clear and easier to follow from the SHLO side

python_package/tt_torch/sharding_rule.py
jackzhangTT: yea actually good catch going to drop this arg

tests/torch/ops/test_sharding_rule.py
```diff
@@ -0,0 +1,703 @@
...
+        )
+        assert ir.Attribute.parse(text) == want
+
+
+# ---------------------------------------------------------------------------
+# make_sharding_rule: happy-path structure (via Shardy attr equality)
+# ---------------------------------------------------------------------------
+
+
+def test_make_sharding_rule_matmul_shape():
```
sshonTT, line 91: This might be a silly question, but can we run pytest in CI while excluding marker like `@pytest.mark.push`?

tests/torch/ops/test_sharding_rule.py
sshonTT: I still see `xs.mark_sharding(b_xla, mesh, ("model", None))`. Did you forget to push the update?

third_party/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 option(USE_CUSTOM_TT_MLIR_VERSION "Flag to use TT_MLIR_VERSION set by the user" OFF)
 
 if (NOT DEFINED TT_MLIR_VERSION OR NOT USE_CUSTOM_TT_MLIR_VERSION)
-    set(TT_MLIR_VERSION "3abca428352b5fad791dfbe69f6a3147d2e44d0b")
+    set(TT_MLIR_VERSION "9b4014a3e54ec4eaab201a7d108db107a2ec9040")
```
AleksKnezevic: reminded to remove this before merging.

jackzhangTT: yes will do sorry it was the only way to test with the change since tt-mlir hasn't been uplifted yet

AleksKnezevic: no worries, just reminding you 😄

tests/torch/ops/test_sharding_rule.py
```diff
@@ -0,0 +1,769 @@
...
+    import torch_xla.distributed.spmd as xs
+    import torch_xla.runtime as xr
+    from infra.evaluators import ComparisonConfig, PccConfig, TorchComparisonEvaluator
+    from infra.utilities.torch_multichip_utils import enable_spmd
+
+    enable_spmd()
+    xr.set_device_type("TT")
+
+    num_devices = xr.global_runtime_device_count()
+    if num_devices < 2:
```
sshonTT: nit: now that you've added [these attributes](https://github.com/tenstorrent/tt-xla/pull/5590/changes#diff-1113602bebf7b6f90f48de0b3685e18d0200997189b474f2b24155eb77254d66R525-R526), it shouldn't hit this. I see other tests are doing this check.

tests/torch/ops/test_sharding_rule.py
```diff
@@ -0,0 +1,769 @@
...
+
+    comparison_config = ComparisonConfig(pcc=PccConfig(required_pcc=0.9999))
+    comparator = TorchComparisonEvaluator(comparison_config)
+    comparator.evaluate(result, golden)
+
+
+@pytest.mark.push
+@pytest.mark.dual_chip
+@pytest.mark.llmbox
+def test_tt_lang_eltwise_add_custom_sharding_rule_reshards_unmarked_operand_e2e(
```
sshonTT, line 526: If the only difference between this test and [that one](https://github.com/tenstorrent/tt-xla/pull/5590/changes#diff-1113602bebf7b6f90f48de0b3685e18d0200997189b474f2b24155eb77254d66R527) is the explicit mark sharding on b, we might want to remove the other test. The goal here is to verify that shardings are properly propagated through operands.

sshonTT (approved):

PR #5591, opened 2026-07-11
[vLLM] Fix inconsistent batch size keying for preallocated buffers 

### LINKED ISSUE #5416: [vLLM] Inconsistent batch_size usage for pre-allocated buffers creation
These three buffers (`_input_ids_dev`, `_position_ids_dev`, `_logits_indices_dev`) are keyed by `{min_num_reqs, max_num_reqs}`, but `_prepare_inputs` clamps the lookup key down to the SMEM row limit: `target_num_reqs = min(target_num_reqs, num_reqs_max_model_len)`. When `num_reqs_max_model_len < max_num_reqs`, the lookup misses → `KeyError`.

The page-table buffers already account for this (keyed by `{min_num_reqs, num_reqs_max_model_len}` / `num_reqs_most_model_len` after 99adbc280) — these three just didn't get the same treatment.

Note this isn't behind the feature flag: with `min_num_seqs` unset it resolves to `max_num_seqs`, so the clamp still runs and a plain decode hits the `KeyError` whenever `max_num_seqs` exceeds the SMEM seq limit (e.g. >128K context at `max_num_seqs=32`). On main this worked since input_ids was always sized `max_num_reqs` and decoupled from the page-table row count.

Cheap fix: key these the same way as the page tables. Could also add a constructor-time assert that every reachable `target_num_reqs` is present in each buffer dict, to turn this into a clear startup error instead of a deep KeyError.

Not a hard blocker given it needs an aggressive long-context config, but worth doing before merge since it's a few lines and isn't gated behind the feature.

Was easily reproduced in a device-free test (asserting the buffer keys cover every clamped target_num_reqs), which passes with the suggested fix.

_Originally posted by @kmabeeTT in https://github.com/tenstorrent/tt-xla/pull/5363#discussion_r3492758324_

alinakhanTT:
### Ticket
closes #5416

### Problem
The batch row count is capped by on chip memory to `num_reqs_max_model_len`, which drops below `max_num_seqs` at long context. The page tables and the compiled warmup graphs already key off that clamped count, but the input buffers (`_input_ids_dev`, `_position_ids_dev`, `_logits_indices_dev`) and `batch_idx` were keyed by `max_num_reqs`. When the two differ the runtime batch shape silently disagrees with the page tables, and a distinct `max_prefill_num_reqs` produces a `KeyError` on the buffer lookup. Normally the two counts are equal, so this only bites long context configs.

### Fix
Key every per batch buffer off the same SMEM clamped count the page tables use, reintroduce the row count clamp in `_prepare_inputs`, and make the fused warmup compile the clamped shapes (the backbone warmup already did). Row count bucketing is pulled into small pure helpers so the choice is testable without a device. No behavior change when `num_reqs_max_model_len == max_num_reqs` (the common case).

### Verification
* [x] device free unit test on the bucketing helpers (22 cases)
* [x] on device init test reproducing the clamp scenario (3 cases)
* [x] `test_opt_generation` push suite (single device common case)
* [x] `test_data_parallel_generation` llmbox padding (DP `batch_idx` sharding)
*  [x] No new failures on nightly caused by this PR ([run 29610901485 ](https://github.com/tenstorrent/tt-xla/actions/runs/29610901485))


integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2546,9 +2559,10 @@ def _precompile_model_fused(self) -> None:
                 f"decode_only is set, only compiling decode path with num_tokens=1"
             )
 
-        num_reqs_options = sorted(
-            {self.min_num_reqs, self.max_prefill_num_reqs, self.max_num_reqs}
-        )
+        # The fused dummy inputs always use the max_model_len page table
```
mmanzoorTT, line 2564: Add support for `most_model_len` as well.

mmanzoorTT (commented):
Update `_precompile_backbone` to use the updated bucket sizes.

integrations/vllm_plugin/vllm_tt/model_runner.py
alinakhanTT: Done!

mmanzoorTT (approved):

PR #5599, opened 2026-07-13
[vLLM] Don't force fp32_dest_acc_en=False graph-wide in benchmark configs

### LINKED ISSUE #5116: [vLLM] Per-user last-token gather index computed via bf16 matmul → wrong logits for batched users past flat-index 512

dgolubovicTT:
### Ticket
Closes #5116

### Problem description
TLDR: Inside `tests/benchmark/test_vllm_benchmarks.py ` we were setting `fp32_dest_acc_en` to False by default which caused divergence in batched prefil outputs. Setting this knob inside benchmark config is wrong. Here is the logic that this knob follows inside mlir compiler part when it is not explicitly set by frontend:
- For opt level 0: we don't care about perf and default value of `fp32_dest_acc_en` is True.
- For opt level >0: we care about perf and compiler leaves `fp32_dest_acc_en` unset, so each ttnn op can choose its value based on inputs etc.
Same is true for math_fidelity knob. For opt level 0 its default is HiFi4, but when opt level is >0, we leave it unset because of the perf.

Here is more detailed explanation why setting `fp32_dest_acc_en` to False globally caused per batch output mismatch:

Batched greedy decoding of N identical prompts diverged into contiguous groups: users past flat-index 512 emitted a different token than users before it. Root cause is a graph-wide fp32_dest_acc_en=False override that the benchmark _config passed by default. TTNNSetComputeKernelConfig stamps that value onto every matmul exposing a compute-kernel-config, including the tiny fp32 index matmul that lowers the per-user last-token gather (flat_index = [batch_idx, pos] @ [stride, 1]). With bf16 destination accumulation, flat indices >= 512 round to a multiple of 4, so the wrong hidden-state row is gathered -> wrong logits -> divergent argmax token. At opt>0 the pipeline resets the override to unset (so it was masked in nightly); at opt=0 the override survives and it breaks.

A compute-kernel-config knob should only ever be set deliberately, never carried as a frontend default. Leaving fp32_dest_acc_en unset lets ttnn pick the per-op default (fp32 accumulation for fp32-output ops like the index matmul, bf16 for bf16 compute matmuls) -- exact index arithmetic with no memory regression on the big matmuls.

### What's changed

Removing default False value from vllm benchmarks and old llm benchmarks.
- tests/benchmark/test_vllm_benchmarks.py: _config default False -> None
- tests/benchmark/test_llms.py: drop hardcoded False on the two 8B configs
Verified (Llama-3.2-3B, 32 identical prompts, greedy, opt0): before the change all rows split (boundary row 5, tokens [320, 4815]); after, all 32 rows agree (720). Deliberate True/False overrides are still honored.

Also cleaned up two vLLM prefill integration tests that carried the same fp32_dest_acc_en=False default, and added a comment so it doesn't get reintroduced:

- tests/integrations/vllm_plugin/generative/test_b1_prefill.py and test_chunked_prefill.py: drop the fp32_dest_acc_en=False from their configs. Both run at opt level 1, where the compiler already resets the knob to unset — so the value was a dead no-op (not a functional change), just the same anti-pattern. Re-ran both without it and nothing changed: test_chunked_prefill.py 3/3 passed, test_b1_prefill.py passed with identical TTFT behavior.


mmanzoorTT (approved):

integrations/vllm_plugin/vllm_tt/platform.py
```diff
@@ -128,6 +128,17 @@ class TTConfig:
...
+    # PREFER LEAVING THIS None (unset). Do NOT set it to False as a default /
+    # convenience in a model config or test. It is a GRAPH-WIDE override applied
+    # to every matmul, including tiny index-arithmetic matmuls (e.g. the per-user
+    # last-token gather that lowers to `flat_index = indices @ strides` ->
+    # embedding). Forcing bf16 destination accumulation there rounds flat
+    # indices >= 512 to the wrong value -> wrong gathered row -> wrong logits ->
+    # per-user divergence in batched greedy decoding (tt-xla #5116). Unset is
+    # better for accuracy: ttnn picks fp32 accumulation for fp32-output ops
+    # (exact index math) and bf16 for bf16 compute matmuls (no memory
+    # regression). Only set True/False deliberately for a validated reason.
```
jameszianxuTT, line 141: Great find! 

Not to block now, but I think this is a footgun that should not be covered up by a comment.  Maybe even just a rename to something like `override_all_ops_fp32_dest_acc_en` to make someone question their usage of this, since from userland we use this option without much thought.

jameszianxuTT (approved):

integrations/vllm_plugin/vllm_tt/platform.py
jameszianxuTT: Or removing user ability to override this from compiler_options directly, as it seems that setting graph level fp32 accumulation to either TRUE/FALSE has little utility and can introduce confounding errors. idk if we have some mechanism to enforce this on op level, but maybe leaving this kind of feature up to dedicated tooling with a different integration point instead of as a first class compile option may be a good idea, though my context here is limited

bmalesevicTT (approved):

vkovacevicTT (approved):

mmanzoorTT (approved):

integrations/vllm_plugin/vllm_tt/platform.py
dgolubovicTT: Thanks, I will think about this. I got a request from training team to expose all other compute kernel cfg knobs because they need them for additional accuracy (for example math approx mode). So I will tie everything in next PR related to this...

PR #5607, opened 2026-07-13
Disable AOT Autograd for Segformer in benchmarks

### LINKED ISSUE #5600: Segformer PCC error on benchmark nightly
Segformer fails with low PCC assertion on [benchmark nightly](https://github.com/tenstorrent/tt-xla/actions/runs/29215068413/job/86709307437) n150-perf and p150-perf:
```
tests/benchmark/test_vision.py:267: in test_segformer
    test_vision(
tests/benchmark/test_vision.py:72: in test_vision
    results = benchmark_vision_torch_xla(
tests/benchmark/benchmarks/vision_benchmark.py:239: in benchmark_vision_torch_xla
    pcc_value >= required_pcc
E   AssertionError: PCC comparison failed. PCC=0.683770, Required=0.97
```

vkovacevicTT:
### Ticket
Closes #5600 

### What's changed
Added aot_autograd flag in benchmarks.
Disabled aot_autograd for Segformer to retain good PCC.

### Checklist
- [x] [Single-chip Segformer](https://github.com/tenstorrent/tt-xla/actions/runs/29273917385)

### DIFF
--- a/tests/benchmark/benchmarks/vision_benchmark.py
+++ b/tests/benchmark/benchmarks/vision_benchmark.py
@@ -102,6 +102,7 @@ def benchmark_vision_torch_xla(
     extract_output_tensor_fn,
     display_name=None,
     required_pcc=0.97,
+    use_aot_autograd=None,
 ):
     """
     Benchmark a vision model using PyTorch and torch-xla.
@@ -159,8 +160,10 @@ def benchmark_vision_torch_xla(
 
     torch_xla.set_custom_compile_options(options)
 
-    # Compile model
-    framework_model.compile(backend="tt")
+    compile_kwargs = {}
+    if use_aot_autograd is not None:
+        compile_kwargs["options"] = {"tt_use_aot_autograd": use_aot_autograd}
+    framework_model.compile(backend="tt", **compile_kwargs)
 
     device = torch_xla.device()
 
--- a/tests/benchmark/test_vision.py
+++ b/tests/benchmark/test_vision.py
@@ -32,6 +32,7 @@ def test_vision(
     input_size=DEFAULT_INPUT_SIZE,
     data_format=DEFAULT_DATA_FORMAT,
     required_pcc=DEFAULT_REQUIRED_PCC,
+    use_aot_autograd=None,
 ):
     """Test vision model with the given configuration.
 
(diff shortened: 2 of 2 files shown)


bmalesevicTT (commented):
Don't have much context on the benefits/downsides of using autograd, but I think whatever caused the PCC degradation should be tracked on some level issue as it could potentially manifest on other models? 
@sgligorijevicTT any ideas what happened?

sgligorijevicTT:
> Don't have much context on the benefits/downsides of using autograd, but I think whatever caused the PCC degradation should be tracked on some level issue as it could potentially manifest on other models? @sgligorijevicTT any ideas what happened?

@bmalesevicTT Not really, I haven't had the time to dig into the details. What I know for certain is that AOTAutograd tends to change shapes, usually by squeezing, unsqueezing or squashing. In Segformer I distinctly remember having a case where a batch matmul ended up having it's shapes rewritten, causing consteval to blow up the memory and tank perf. And trough the haze I remember something about different ranks impacting accuracy, maybe from precisely this investigation... A brief glance at IRs attached on the issue made me suspect that is the case here too.
Either way, it's the right call to track that AOTAutograd can impact numerics, but if my hypothesis is true then this isn't really a frontend bug, more a case of superficial changes exposing lack of robustness underneath.
I had a PR in mlir to handle that exact case, I'm not sure why it wouldn't be triggering here.
As for benefits vs downsides, you only need AOTA iff you plan on calling `.backward()` over a compiled region, which isn't the case for inference benchmarks. The reason why it impacts inference graphs too despite being useful only for training is that it has to form a joint forward-backward graph and cut it appropriately, and in that step the divergence already occurs. But it would be wise to keep it on for inference too, lest we end up with Buda-esque config spaghetti.

vkovacevicTT:
To conclude: Should we proceed to disable AOTAutograd for Segformer only in benchmarks?
We didn't observe any significant numerical drop in other benchmark models.

@bmalesevicTT @sgligorijevicTT

sgligorijevicTT:
@vkovacevicTT My vote is for yes, and that we track the removal of this hack + rootcausing and fixing/filing. I personally don't have the bandwidth to pick this up soon tho.

bmalesevicTT (approved):
Then we can merge this, but, if closing the existing issue, keep track in a new issue to be transparently aware of this hack. 
I think it would be good to have a system for each per-model tweaked option in benchmarks, as an umbrella issue or a similar doc where we can easily track changed defaults per models and eventually align them

vkovacevicTT:
Issue for tracking AOTAutograd flag in Segformer: https://github.com/tenstorrent/tt-xla/issues/5677

Issue #5426, opened 2026-06-30
Facing Huge regressions in Nightly Pipeline


pdeviTT:
### SUMMARY

Facing huge Performance drop in nightly [pipeline](https://github.com/tenstorrent/tt-xla/actions/runs/28411925471)

<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
<html>
<head>

</head>
<body>

Job | Hardware | Yesterday (sps) | Today (sps) | Delta | Change%
-- | -- | -- | -- | -- | --
mistral_7b | n150-perf | 11.90 | 4.41 | -7.48 | -62.9%
mistral_7b | p150-perf | 21.72 | 8.47 | -13.25 | -61.0%
llama_3_1_8b_instruct | n150-perf | 6.23 | 3.31 | -2.92 | -46.9%
qwen_3_8b | n150-perf | 4.94 | 2.79 | -2.14 | -43.4%
qwen_3_4b | n150-perf | 5.40 | 3.10 | -2.31 | -42.7%
llama_3_1_8b_instruct | p150-perf | 9.98 | 5.78 | -4.20 | -42.1%
qwen_3_8b | p150-perf | 8.43 | 5.01 | -3.42 | -40.6%
qwen_3_4b | p150-perf | 9.27 | 5.63 | -3.65 | -39.3%
llama_3_2_3b_instruct | n150-perf | 6.83 | 4.29 | -2.53 | -37.1%
qwen_2_5_7b_instruct | n150-perf | 5.43 | 3.45 | -1.99 | -36.6%
llama_3_2_1b_instruct | n150-perf | 8.48 | 5.72 | -2.76 | -32.6%
qwen_2_5_7b_instruct | p150-perf | 8.46 | 5.73 | -2.73 | -32.3%
tiiuae_falcon3-7b-base | n150-perf | 6.30 | 4.28 | -2.02 | -32.1%
llama_3_2_3b_instruct | p150-perf | 11.00 | 7.49 | -3.51 | -31.9%
qwen_3_0_6b | n150-perf | 6.38 | 4.46 | -1.92 | -30.1%
qwen_2_5_3b_instruct | n150-perf | 6.05 | 4.25 | -1.80 | -29.8%
tiiuae_falcon3-3b-base | n150-perf | 6.73 | 4.84 | -1.88 | -28.0%
[...]

PR #5609, opened 2026-07-13
Catch SDPA composite for bool attention masks behind opt-in flag

### LINKED ISSUE #4780: [HunyuanImage-2.1-Distilled-Diffusers] OOM in transformer due to native SDPA decomposition
### Describe the bug

- On 4 chips, transformer fails with `Out of Memory: Not enough space to allocate 771162112 B DRAM buffer across 12 banks, where each bank needs to store 64266240 B, but bank size is 1071821792 B (allocated: 1014289568 B, free: 57532224 B, largest free block: 47172576 B)` - [jun1_HunyuanImage_2_1_transformer_4_chips.log](https://github.com/user-attachments/files/28458892/jun1_HunyuanImage_2_1_transformer_4_chips.log)

### Root cause analysis

- 4 chips, fp32 & bf16: OOM on a 771 MB buffer = native-SDPA attn_weight + mask add producing [1,7,5224,5224] f32 joint-attention scores. Verified: 0 SDPA composites survive in the IR (composite never engages).
- Why composite skipped: _check_sdpa_constraints ([composite_ops.py:430-444](vscode-webview://1l6oi0mrmo380rdlh5jeli9at5o1d5ilhptp9kritsbpbbn0302i/python_package/tt_torch/composite_ops.py#L430-L444)) needs all SDPA inputs bf16. fp32 → Q/K/V fail; bf16 → the attn_mask is bool (i1, built at [transformer_hunyuanimage.py:818-819](vscode-webview://1l6oi0mrmo380rdlh5jeli9at5o1d5ilhptp9kritsbpbbn0302i/venv/lib/python3.12/site-packages/diffusers/models/transformers/transformer_hunyuanimage.py#L818-L819)) → fails. Confirmed 1x1x1x5224xi1 mask in IR.
- Sharding is fine (not the bug): tensor-parallel weights + heads 28→7/chip, verified in IR. OOM is accumulated residency (~11.1 GB/12 GB; 45 graphs execute first), so it's a late joint-attention block, not the first SDPA.
- 8 chips (fp32 & bf16): different/earlier failure — 'sdy.collective_permute' op requires the same type for all operands and results. Distinct from the OOM.
- Root caused is clearly explained in below comments

### TODO

- [x] **Extract the problematic SDPA's details** — shape `[1, 28→7, 5224, 5224]`, dtype (f32 scores), the `attn_mask` dtype (`bool`/`i1`), and which block it lands in.
- [x] **Build a minimal sanity repro** — a single joint-attention call (seq 5224, 28 heads, bf16 Q/K/V + **bool** mask) that reproduces both (a) the composite being skipped and (b) the f32 score-matrix OOM, without the full 17.45B model.
- [x] **Fix on the sanity repro** — make the composite engage with a non-bf16 mask: relax `_check_sdpa_constraints` to gate only on Q/K/V (or cast the mask to bf16 additive), and confirm the fused TTNN SDPA is used.
- [x] **Verify the fix** — score matrix no longer materialized; OOM cleared (and check it doesn't just shift downstream).
- [x] **Raise the PR** (and link the companion tt-mlir change if the fused-mask path needs work there).

### Logs

- [jun1_HunyuanImage_2_1_transformer_8_chips_bfp16.log](https://github.com/user-attachments/files/28458987/jun1_HunyuanImage_2_1_transformer_8_chips_bfp16.log)
- [jun1_HunyuanImage_2_1_transformer_8_chips.log](https://github.com/user-attachments/files/28458986/jun1_HunyuanImage_2_1_transformer_8_chips.log)
-  [jun1_HunyuanImage_2_1_transformer_4_chips_bfp16.log.zip](https://github.com/user-attachments/files/28458916/jun1_HunyuanImage_2_1_transformer_4_chips_bfp16.log.zip)

kamalrajkannan78:
### Ticket

- fixes #4780

### Problem description

- `HunyuanImage-2.1`'s transformer feeds `F.scaled_dot_product_attention` bf16 Q/K/V together with a **bool** key-padding mask of shape `[1, 1, 1, S]`. The SDPA composite was never engaging on it, so the op fell back to the native decomposition, which materialises a full `[·, ·, S, S]` **f32 score matrix → OOM** (the 771 MB DRAM buffer in [#4780](https://github.com/tenstorrent/tt-xla/issues/4780)).
- Root cause: `_check_sdpa_constraints` required **every** SDPA input — including the mask — to be bf16. A bool mask disqualified the whole node, so the composite (and thus the fused `ttnn.scaled_dot_product_attention`) never formed.

### What's changed

- **`_check_sdpa_constraints`** — always gate **Q/K/V** on bf16 (positional or named). The attention mask is gated separately: `bf16` is accepted as-is; a `bool` mask is accepted only under the flag (see below); any other dtype (`int`, `f32`, …) still skips the composite and falls back to native, since the fused kernel only accepts a float mask.
- **`composite_scaled_dot_product_attention`** — convert a bool mask to an **additive bf16 bias** (`True → 0`, `False → -inf`). The fused TTNN SDPA *adds* the mask to the scores and `TT_FATAL`s on non-float masks; a plain bool→float cast (`False → 0`) would fail to suppress padding, so the additive form is required for correctness.
- Both changes are **opt-in** behind `TT_XLA_CATCH_BOOL_MASK_SDPA` (default off). Enabling bool-mask capture globally regressed perf (#5426), so for now it's turned on per-test for models with no perf concern.
- Added `test_sdpa_bool_mask_broadcast` in `tests/torch/ops/test_composite_ops.py` — tensor-parallel (head-sharded Q/K/V), parametrized over the shapes the model produces, covering the **query-broadcast `dim2 == 1`** masks (e.g. `(1,1,1,5224)`) as well as the full per-head mask `(1,28,1000,1000)`. Verified these fuse to `ttnn.scaled_dot_product_attention` and match golden PCC.
- `test_transformer_sharded` now sets the flag; with the fix the transformer fuses SDPA across all attention blocks and the 771 MB score-matrix OOM is gone.

### Checklist

- [x] Verified through local testing on WH

### Logs

- [jul13_sdpa_mc_before_fix.log](https://github.com/user-attachments/files/29969720/jul13_sdpa_mc_before_fix.log)
- [jul14_sdpa_mc_after_fix.log](https://github.com/user-attachments/files/30015065/jul14_sdpa_mc_after_fix.log)
- [jul13_HunyuanImage_2_1_transformer_before_fix.log.zip](https://github.com/user-attachments/files/29969715/jul13_HunyuanImage_2_1_transformer_1.log.zip)
- [jul14_HunyuanImage_2_1_transformer_after_fix.log.zip](https://github.com/user-attachments/files/30015064/jul14_HunyuanImage_2_1_transformer_after_fix.log.zip)

### DIFF
--- a/python_package/tt_torch/composite_ops.py
+++ b/python_package/tt_torch/composite_ops.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import itertools
+import os
 from typing import List, Optional, Union
 
 import torch
@@ -287,6 +288,17 @@ def composite_scaled_dot_product_attention(
     )
 
     if attn_mask is not None:
+        # Fused SDPA adds the mask to the scores, and ttnn.scaled_dot_product_attention
+        # accepts only a float mask (BF16/BFP8/BFP4), so convert a bool mask to additive
+        # form (True -> 0, False -> -inf). Gated by the same flag that allows capturing
+        # a bool mask in _check_sdpa_constraints.
+        # torch reference: https://github.com/pytorch/pytorch/blob/9f48642d4bf84566ca3191576c7ff30138953af8/torch/nn/functional.py#L6365
+        # ttnn check: tt-metal .../transformer/sdpa/device/sdpa_device_operation.cpp ("must be in BF16, BFP8, or BFP4 dataformat")
+        if (
+            os.environ.get("TT_XLA_CATCH_BOOL_MASK_SDPA", "0") == "1"
+            and attn_mask.dtype == torch.bool
+        ):
+            attn_mask = torch.where(attn_mask, 0.0, float("-inf")).to(query.dtype)
         query, key, value, attn_mask = builder.mark_inputs(query, key, value, attn_mask)
     else:
         query, key, value = builder.mark_inputs(query, key, value)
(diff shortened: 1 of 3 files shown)


python_package/tt_torch/composite_ops.py
```diff
@@ -509,12 +521,24 @@ def _check_sdpa_constraints(node: torch.fx.Node) -> bool:
...
-    tensor_args = list(node.args) + [
-        v for v in node.kwargs.values() if isinstance(v, torch.fx.Node)
-    ]
+    # By default all inputs (including attn_mask) are gated on bf16, so a bool mask
+    # skips the composite. Catching bool masks by default regressed perf (issue #5426),
+    # so it is opt-in via TT_XLA_CATCH_BOOL_MASK_SDPA for models with no perf concern
+    # at the moment.
+    # When set, gate only Q/K/V (positional or named); the bool mask is converted to
+    # additive form in the composite.
+    if os.environ.get("TT_XLA_CATCH_BOOL_MASK_SDPA", "0") == "1":
```
mstojkovicTT: What happens when mask is `int` for example? I think we should make sure that the mask is bool for this workaround.

mstojkovicTT (approved):

python_package/tt_torch/composite_ops.py
kamalrajkannan78: Great catch! Fixed — the mask is no longer blanket-exempted. The gate now only lets `bf16` (as-is) or `bool` (converted to additive bf16) masks through; any other dtype (`int`, `f32`, …) skips the composite → native fallback. Tested sanity + model on current change; description and logs updated.

sgligorijevicTT (approved):
I don't like introducing yet another env var, and especially not for something like this. But I don't see a better path forward short term. But this should be cleaned up eventually, and ideally sooner rather than later.

sgligorijevicTT:
cc: @nvukobratTT this touches upon discussions we had previously

PR #5615, opened 2026-07-13
[vLLM] Teacher-forced decode accuracy testing in vLLM benchmarks

### LINKED ISSUE #5179: Add Top1/Top5 accuracy testing to vLLM benchmark path
Add Top1/Top5 accuracy testing to vLLM benchmark path same way we do it in llm benchmarks.

dgolubovicTT:
### Ticket
Closes #5179

### Problem description
The vLLM benchmark path had no token-accuracy coverage. TOP1/TOP5 accuracy was only measured by the custom torch-xla path (`tests/benchmark/test_llms.py`), which runs a teacher-forced autoregressive decode loop against precomputed reference outputs. We want the same TOP1/TOP5 signal through vLLM so the vLLM backend's decode path is validated against the reference model.

The hard part is faithfully reproducing the custom path's behavior through vLLM. `llm_benchmark.py` does two things at once that vLLM's public API does not expose together:
- **Teacher-forcing** — at each decode step it feeds the reference ground-truth token instead of the model's own sampled token, which isolates per-position accuracy (one wrong token doesn't corrupt all downstream context).
- **Decode-kernel coverage** — those injected tokens flow through the real single-token decode loop (KV-cache reads, decode-time RoPE).

vLLM's `generate()` always feeds back the model's own sampled token, and neither TT sampler (device or CPU) invokes user logits processors, so there is no public hook to teacher-force the decode loop. An earlier prototype sidestepped this by measuring accuracy from a single prefill pass (`prompt_logprobs` over the full ground-truth sequence). That is teacher-forced, but every prediction is computed by the **prefill** kernel — it never exercises the decode kernel/KV-cache, so decode-path regressions (especially under `bfp_bf8`/`bfp_bf4` weights) could pass while the real decode path is broken.

### What's changed
Teacher-forced decode through vLLM, matching `llm_benchmark.py` semantics (1 prefill-position prediction + N-1 decode-kernel predictions, all teacher-forced):

- **`integrations/vllm_plugin/vllm_tt/`** — opt-in `teacher_forced_token()` helper (`vllm_utils.py`) wired into `TTModelRunner.sample_tokens`. When a request carries `SamplingParams.extra_args["teacher_forcing_tokens"]`, the sampled token is overridden with the reference ground-truth token (bounds-checked, indexed by decode step) so each decode step sees the reference context. No-op on the production path when the key is absent. The override lands **after** `gather_logprobs`, so the reported per-step logprobs keep the true device argmax — which is what gets scored. Teacher forcing applies only to the one-token-per-step path: under speculative decode the accepted draft tokens have already been through the verify pass, so their KV entries are written and overriding the token ids would desync the cache from `token_ids_cpu`. The runner asserts there rather than reporting a plausible but meaningless number.
- **`tests/benchmark/benchmarks/vllm_benchmark.py`** — the accuracy path (`_benchmark_vllm_accuracy()`) prompts with the prefill half of the reference sequence and generates the decode half (`max_tokens = decode length`, `logprobs=1`, `extra_args` carrying the ground-truth decode tokens). The device per-step argmax is read from `output.outputs[0].logprobs` and scored via the existing `TokenAccuracy`. Keeps the same measurement names (`top1/top5_accuracy_p5`, `..._mean`) and `evaluation_score = top1_p5` as the custom path, so dashboards stay compatible. Asserts the full decode window of logprobs is present, so a backend gap fails loudly instead of reporting a false 0% regression. Also prints a golden-vs-device decoded-text preview so TOP1 can be eyeballed against the text.
- **`tests/benchmark/llm_utils/token_accuracy.py`** — per-user TOP1/TOP5 aggregation extracted into `score_token_accuracy()` and shared with `llm_benchmark.py`, replacing a near-verbatim copy of that block. Both paths now report identically-computed numbers under the same measurement names. This touches the existing torch-xla accuracy path, so it was re-verified on device (see below).
- **`.github/workflows/perf-bench-matrix.json`** — single-device vLLM accuracy matrix entries.
- **`.github/workflows/test-matrix-presets/basic-test.json`** — a `push and cpu` job for `tests/integrations/vllm_plugin`, so the CPU unit tests below are actually collected. Without it nothing in that tree runs by mark on a CPU runner.
- **`tests/benchmark/test_vllm_benchmarks.py`** — plumbs the existing `--accuracy-testing` flag through `_run_vllm_benchmark` to both the single-device and TP benchmark tests.
- Unit tests, no device required: `tests/integrations/vllm_plugin/sampling/test_teacher_forcing.py` (override helper) and `tests/integrations/vllm_plugin/test_vllm_benchmark_accuracy_unit.py` (prediction extraction, fail-loud assertion).

Verified on-device (Qwen3-0.6B): `TOP1 95.31% / TOP5 100.00%` — 61 of the 64 decode positions match, which is exactly the reported 95.31%. The shared-scoring refactor was separately re-verified on the torch-xla path (`test_llms.py::test_qwen_2_5_0_5b --accuracy-testing`: 1 passed, `TOP1 95.31% / TOP5 100.00%` over 32 users). `.refpt` reference files are regenerated on demand and are not committed.

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -129,6 +129,7 @@ from .vllm_utils import (
     apply_hidden_layer_override,
     determine_mesh_shape,
     prev_power_of_2,
+    teacher_forced_token,
 )
 
 
@@ -2648,8 +2649,20 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
             for i in discard_sampled_tokens_req_indices:
                 valid_sampled_token_ids[i].clear()
 
-            # Append sampled tokens
+            # Append sampled tokens. For accuracy runs, teacher-force the decode
+            # by overriding the sampled token with the reference ground-truth
+            # token (via SamplingParams.extra_args). The override happens after
+            # gather_logprobs above, so `logprobs_lists` still carries the true
+            # device argmax (what accuracy scores), while the next decode input
+            # and recorded output become ground truth. No-op in production.
             for i, req_state, seq_len in request_seq_lens:
+                sampling_params = req_state.sampling_params
+                forced = teacher_forced_token(
+                    sampling_params.extra_args if sampling_params is not None else None,
+                    len(req_state.output_token_ids),
+                )
+                if forced is not None:
+                    valid_sampled_token_ids[i][0] = forced
                 token_id = valid_sampled_token_ids[i][0]
                 self.input_batch.token_ids_cpu[i, seq_len] = token_id
                 req_state.output_token_ids.append(token_id)
@@ -2667,6 +2680,21 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
             self.input_batch.num_tokens[:num_reqs] += gen_lens
             self.input_batch.num_tokens_no_spec[:num_reqs] += gen_lens
             for i, req_state, seq_len in request_seq_lens:
+                # Teacher forcing is only implemented for the one-token-per-step
+                # path. Accepted draft tokens have already been through the verify
+                # pass, so their KV entries are written; overriding the token ids
+                # here would desync the cache from token_ids_cpu. Warn rather than
+                # assert, so a stray extra_args can never break a decode run.
+                extra_args = (
+                    req_state.sampling_params.extra_args
+                    if req_state.sampling_params is not None
+                    else None
+                )
+                if extra_args is not None and "teacher_forcing_tokens" in extra_args:
+                    logger.warning_once(
+                        "teacher_forcing_tokens is not supported when sampling "
+                        "multiple tokens per step (e.g. speculative decoding)."
+                    )
                 if not valid_sampled_token_ids[i]:
                     continue
                 target_slice = slice(seq_len - gen_lens[i] + 1, seq_len + 1)
--- a/tests/benchmark/test_vllm_benchmarks.py
+++ b/tests/benchmark/test_vllm_benchmarks.py
@@ -290,7 +290,7 @@ TP_CONFIGS = [
 ]
 
 
-def _run_vllm_benchmark(config, output_file, request):
+def _run_vllm_benchmark(config, output_file, request, accuracy_testing=False):
     resolved_display_name = resolve_display_name(request=request, fallback=config.model)
     display_name = (
         resolved_display_name
@@ -308,7 +308,7 @@ def _run_vllm_benchmark(config, output_file, request):
         "export_model_name", sanitize_model_name(display_name)
     )
 
-    results = benchmark_vllm(config, display_name)
+    results = benchmark_vllm(config, display_name, accuracy_testing=accuracy_testing)
 
     if output_file:
         results["project"] = "tt-forge/tt-xla"
@@ -422,13 +422,13 @@ def _run_vllm_embedding_benchmark(config, output_file, request):
 
 
 @pytest.mark.parametrize("config", SINGLE_DEVICE_CONFIGS)
-def test_vllm_benchmark(config, output_file, request):
-    _run_vllm_benchmark(config, output_file, request)
+def test_vllm_benchmark(config, output_file, request, accuracy_testing):
+    _run_vllm_benchmark(config, output_file, request, accuracy_testing=accuracy_testing)
 
 
 @pytest.mark.parametrize("config", TP_CONFIGS)
-def test_vllm_tp_benchmark(config, output_file, request):
-    _run_vllm_benchmark(config, output_file, request)
+def test_vllm_tp_benchmark(config, output_file, request, accuracy_testing):
+    _run_vllm_benchmark(config, output_file, request, accuracy_testing=accuracy_testing)
 
 
 @pytest.mark.parametrize("config", EMBEDDING_CONFIGS)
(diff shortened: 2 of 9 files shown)


vvukomanTT (approved):

bmalesevicTT (approved):

tests/benchmark/test_vllm_benchmarks.py
```diff
@@ -404,13 +426,15 @@ def _run_vllm_embedding_benchmark(config, output_file, request):
 
 
 @pytest.mark.parametrize("config", SINGLE_DEVICE_CONFIGS)
-def test_vllm_benchmark(config, output_file, request):
-    _run_vllm_benchmark(config, output_file, request)
+def test_vllm_benchmark(config, output_file, request, accuracy_testing):
+    _run_vllm_benchmark(config, output_file, request, accuracy_testing=accuracy_testing)
 
 
-@pytest.mark.parametrize("config", TP_CONFIGS)
```
mmanzoorTT, line 429: is `@pytest.mark.parametrize("config", TP_CONFIGS)` removed intentionally?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2074,8 +2075,20 @@ def sample_tokens(
             for i in discard_sampled_tokens_req_indices:
                 valid_sampled_token_ids[i].clear()
 
-            # Append sampled tokens
+            # Append sampled tokens. For accuracy runs, teacher-force the decode
```
mmanzoorTT, line 2652: We have recently landed speculative decode. `else` block will be trigger in case of multiple accepted tokens (generated with speculative decode). What do you think about update 'else' block to use provided tokens?

tests/benchmark/test_vllm_benchmarks.py
dgolubovicTT: removed by mistake

integrations/vllm_plugin/vllm_tt/model_runner.py
dgolubovicTT: For now, I would just skip accuracy testing when speculative decode is ON. What do you think?
Then add it later if we have the need

mmanzoorTT: Then add a `warning_once` log when both accuracy_testing and speculative_decode are enabled.

.github/workflows/test-matrix-presets/basic-test.json
```diff
@@ -16,7 +16,7 @@
...
-  {"runs-on": "cpu",                "name": "run_vllm_cpu_tests",      "dir": "./tests/integrations/vllm_plugin",        "test-mark": "push and cpu",            "shared-runners": "true", "extra-wheel": "vllm"},
   {"runs-on": "n300",               "name": "run_examples",           "dir": "./tests/examples",                        "test-mark": "push and not bhqb", "require":"alchemist", "parallel-groups": 3, "shared-runners": "true" },
-  {"runs-on": "cpu",                "name": "run_mixed_precision_cpu","dir": "./mixed_precision",                       "test-mark": "push and cpu",            "shared-runners": "true", "forked": true }
+  {"runs-on": "cpu",                "name": "run_mixed_precision_cpu","dir": "./mixed_precision",                       "test-mark": "push and cpu",            "shared-runners": "true", "forked": true },
+  {"runs-on": "cpu",                "name": "run_vllm_cpu_tests",     "dir": "./tests/integrations/vllm_plugin",        "test-mark": "push and cpu",            "shared-runners": "true", "forked": true, "extra-wheel": "vllm" }
```
mmanzoorTT: Please rebase your branch. `cpu` job is added for vLLM recently. https://github.com/tenstorrent/tt-xla/blob/0bb36ceac4492db5e0cd71772c59af684a28afc7/.github/workflows/test-matrix-presets/basic-test.json#L19

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -2667,6 +2437,23 @@ def sample_tokens(
...
+                            if req_state.sampling_params is not None
+                            else None
+                        ),
+                        len(req_state.output_token_ids),
+                    )
+                    is None
+                ), (
+                    "teacher_forcing_tokens is not supported when sampling "
+                    "multiple tokens per step (e.g. speculative decoding)."
+                )
```
mmanzoorTT, line 2692: My suggestion is just print a warning instead of assert. 
```suggestion
extra_args = (
    req_state.sampling_params.extra_args
    if req_state.sampling_params is not None
    else None
)

if extra_args is not None and "teacher_forcing_tokens" in extra_args:
    logger.warning_once(
        "teacher_forcing_tokens is not supported when sampling "
        "multiple tokens per step (e.g. speculative decoding)."
    )
```

mmanzoorTT (approved):

integrations/vllm_plugin/vllm_tt/model_runner.py
dgolubovicTT: In that case we would still calculate some top1/top5 metrics but we would warn that result can be meaningless. I'd rather we assert when both teacher forcing and speculative decode are ON. What do you think?

mmanzoorTT: make sense. Agree.

mmanzoorTT (approved):

nsumrakTT (approved):

Issue #5579, opened 2026-07-09
Uplift regression chunked-prefill SDPA regressed by tt-mlir #8953: "Page table must be row major"


sgligorijevicTT:
**Test:** `tests/integrations/vllm_plugin/generative/test_prefill_recompile.py::test_no_prefill_recompile` (Qwen3-0.6B, chunked prefill, n300).

**Regressed by:** tt-mlir `758bdc56` — *"[Optimizer] Insert corrective to_layout when RowMajor propagation stops at an interior op (#8953)"*. Bisected by build-and-run (tt-xla PR #5555): reverting #8953 alone makes the test pass; every other commit in the uplift range is good.

**Symptom:** during warmup of the chunked-prefix graph,
```
TT_FATAL sdpa_device_operation.cpp:212: page_table.layout() == Layout::ROW_MAJOR
"Page table must be row major"
```
in `ttnn::transformer::chunked_scaled_dot_product_attention`.

**Suspected root cause:** #8953's new `insertTiledFixup` converts an operand back to Tiled when RowMajor propagation stops at an `OpModelExempt` op. `ChunkedScaledDotProductAttentionOp` is `OpModelExempt`, but its `page_table`/`chunk_start_idx` operands are pinned to RowMajor by its operand-workaround (and required row-major by tt-metal). The fixup tilizes the page_table it must not. Likely fix: skip operands a TTNN workaround pins to RowMajor. (Paged-flash-decode SDPA looks similarly exposed.)

ddilbazTT:
Are we sure it's n300? This test is marked for single-device.

sgligorijevicTT:
> Are we sure it's n300? This test is marked for single-device.

My bad, I let claude autofile this and was testing on n300 so it must have gotten confused. The bug should be parallelization agnostic anyway. The error reproduces on n300 too so in some sense it's not wrong, but yeah the test is single chip

kmabeeTT:
Thanks for flagging/catching (glad test was in push), and for working on fix @ddilbazTT! Since this broke all productized models (rely on chunked prefill), need to get fix uplifted asap (looks like https://github.com/tenstorrent/tt-xla/pull/5592 is kind of close). Going to mark this as blocking several models.

kmabeeTT:
[...]

PR #5618, opened 2026-07-13
[vLLM] Pin opt_level=0 in chunked prefill tests to keep page_table ROW_MAJOR

alinakhanTT:
The three chunked prefill tests fail in nightly with a `TT_FATAL page_table.layout() == Layout::ROW_MAJOR` abort in the chunked SDPA op. The tests never set `optimization_level`, so they run at the default opt1 where the layout optimizer retiles the page_table, but the op requires it to stay row major.

This sets `optimization_level=0` in both LLM configs (the `_generate` helper and the batch test), the same workaround already used by `test_llama3_3b_generation_trace_opt0`.

Verified on device (Wormhole) against main's tt-mlir pin: at opt1 all three tests hit the ROW_MAJOR abort, at opt0 all three pass with matching output.

### DIFF
--- a/tests/integrations/vllm_plugin/generative/test_chunked_prefill.py
+++ b/tests/integrations/vllm_plugin/generative/test_chunked_prefill.py
@@ -91,6 +91,9 @@ def _generate(max_num_batched_tokens: int) -> str:
             # length this is a single chunk (the oracle); with a small budget the
             # prompt is split into block-aligned chunks.
             "prefill_chunk_size": max_num_batched_tokens,
+            # opt1 retiles the page_table so the chunked-SDPA op aborts
+            # (page_table must stay ROW_MAJOR), so pin opt0.
+            "optimization_level": 0,
         },
     )
     sp = vllm.SamplingParams(temperature=0.0, max_tokens=_MAX_TOKENS, ignore_eos=True)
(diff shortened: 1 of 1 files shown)


kmabeeTT:
Hmm at first thought this doesn't sound great, we rely on opt-level=1 for perf and other things in production models, and also chunked prefill feature... did something regress here (was passing, and is now failing)? If so, we better not cover it up with a workaround like this, and should push for proper fix wherever that is...

AleksKnezevic:
@ddilbazTT was this the same layout issue you were seeing?

mmanzoorTT:
It is the same problem. [tt-xla issue](https://github.com/tenstorrent/tt-xla/issues/5579) -[tt-mlir issue](https://github.com/tenstorrent/tt-mlir/issues/9003)

ddilbazTT:
@AleksKnezevic Yes, it looks like the same error.
@alinakhanTT Could you try out either of these branches?

https://github.com/tenstorrent/tt-mlir/pull/9027 --> Adds opmodel for chunked sdpa op
https://github.com/tenstorrent/tt-mlir/compare/main...ddilbaz/chunked-sdpa-page-table-rowmajor-optimizer --> whitelists chunked sdpa op workaround so it is applied even when opt level >= 1 (workarounds are not applied when opt level >=1 unless whitelisted)

alinakhanTT:
@ddilbazTT tested against ddilbaz/chunked-sdpa-page-table-rowmajor-optimizer and it fixes it!
Closing this PR now since the opt0 workaround becomes unnecessary once that fix lands

PR #5622, opened 2026-07-14
[pjrt] Support rank-0 (scalar) complex host buffers

devisettymahidhar608:
### Ticket
Link to Github Issue

### Problem description
Running `deepseek-ai/DeepSeek-V2-Lite-Chat` on the TT backend crashed during
host-to-device transfer with:

RuntimeError: TT_THROW @ .../pjrt_implementation/src/api/buffer_instance.cc:398: tt::exception
info:
Complex tensor with num_dims == 0 is not supported.

`BufferInstance::calculateShape` explicitly threw whenever it received a complex
buffer with `num_dims == 0` â i.e. a rank-0 complex **scalar**. DeepSeek-V2's
rotary embedding produces exactly such a value, which reaches a `mul` and gets
transferred to device via `PjRtComputationClient::TransferShardsToDevice`,
triggering the throw and aborting the run.

### What's changed
Removed the `num_dims == 0` guard in `calculateShape` so scalar complex buffers
flow through the standard complex-tensor path.

The guard was unnecessary. tt-mlir represents complex tensors by appending a
trailing dimension of 2 (`tensor<...xcomplex<f32>>` â `tensor<...x2xf32>`), which
maps a rank-0 complex scalar cleanly to a rank-1 `tensor<2xf32>` â consistent
with how complex tensors of every other rank are already handled. The runtime
also already tolerates the shape/strides length mismatch that complex tensors
create (`stride.size() != shape.size()` is explicitly treated as contiguous in
`runtime.cpp`), so the scalar case (shape `[2]`, strides `[]`) is already
supported downstream.

### Checklist

### DIFF
--- a/pjrt_implementation/src/api/buffer_instance.cc
+++ b/pjrt_implementation/src/api/buffer_instance.cc
@@ -399,12 +399,7 @@ BufferInstance::calculateShape(const std::int64_t *dims, size_t num_dims,
   }
 
   // Complex tensors have no runtime equivalent dtype, so they are stored as
-  // float tensors with a trailing dimension of 2 for interleaved real/imag
-  // (tt-mlir maps tensor<...xcomplex<f32>> to tensor<...x2xf32>). This applies
-  // uniformly across ranks, including rank-0 complex scalars, which become a
-  // rank-1 tensor<2xf32>. The resulting shape/strides mismatch (strides has no
-  // entry for the trailing dim) is expected by the runtime, which treats such
-  // complex tensors as contiguous.
+  // float tensors with a trailing dimension of 2 for interleaved real/imag.
   if (data_type_utils::isComplexPJRTType(data_type)) {
     shape.push_back(2);
   }
--- a/pjrt_implementation/src/api/buffer_instance.cc
+++ b/pjrt_implementation/src/api/buffer_instance.cc
@@ -393,18 +393,18 @@ void BufferInstance::allocateUninitialized(const std::int64_t *byte_strides,
 std::vector<std::uint32_t>
 BufferInstance::calculateShape(const std::int64_t *dims, size_t num_dims,
                                PJRT_Buffer_Type data_type) {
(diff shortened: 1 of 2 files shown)


sonalibaskaran2499 (approved):

jameszianxuTT (approved):
I think this is fine

cc @ddilbazTT who should also look and was the author of this feature - I recall that we blocked this before because calculateShape returned a hardcoded {1} before, but it seems that this is no longer the case so it should be fine

pjrt_implementation/src/api/buffer_instance.cc
```diff
@@ -393,18 +393,18 @@ void BufferInstance::allocateUninitialized(const std::int64_t *byte_strides,
...
   std::vector<std::uint32_t> shape;
   for (size_t i = 0; i < num_dims; ++i) {
     shape.push_back(dims[i]);
   }
 
   // Complex tensors have no runtime equivalent dtype, so they are stored as
-  // float tensors with a trailing dimension of 2 for interleaved real/imag.
+  // float tensors with a trailing dimension of 2 for interleaved real/imag
+  // (tt-mlir maps tensor<...xcomplex<f32>> to tensor<...x2xf32>). This applies
+  // uniformly across ranks, including rank-0 complex scalars, which become a
```
jameszianxuTT: I think it is okay if we don't add all this justification here. We can sort it out in the PR discussion

kmabeeTT (commented):
Does DeepSeek-V2-Lite-Chat progress past this error now? Any updated link or log file? Worth mentioning in PR Desc.  Agreed, @ddilbazTT should look here too, I am not codeowner so my approval carries little weight :)

devisettymahidhar608:
hi @kmabeeTT , 
With these changes, the model is now passing. I've raised the PR: https://github.com/tenstorrent/tt-forge-models/pull/813

Thanks

ddilbazTT (approved):

ddilbazTT:
Thanks for the ping @jameszianxuTT @kmabeeTT I think it looks good.

PR #5631, opened 2026-07-14
Fix whisper E2E: forward keyword inputs + bool cumsum cast

ctr-pmuruganTT:
## Problem
`tests/torch/models/whisper/test_whisper.py::test_torch_whisper_inference[*]` did not pass end-to-end. Two independent issues were in the way:

1. **`forward()` signature** — the test failed immediately with:
   ```
   TypeError: WhisperWrapper.forward() got an unexpected keyword argument 'input_features'
   ```
2. **bool `cumsum`** — once past `forward()`, compilation failed on the `attention_mask` path, where an `aten.cumsum` runs on a **bool** tensor (tt-metal cumsum does not support the bool/uint8 input that this lowering produces).

## Root cause
**(1) forward signature.** `ModelLoader.load_inputs()` (whisper) returns a **dict** (`{"input_features": ..., "decoder_input_ids": ...}`, plus `attention_mask` for Large_v3_Turbo). `TorchModelTester._get_forward_method_kwargs` unpacks any `Mapping` into keyword args, so the model is invoked as `model(input_features=..., decoder_input_ids=...)`. But `WhisperWrapper.forward(self, *inputs)` only accepted **positional** args, so the keyword arguments were rejected.

**(2) bool cumsum.** Whisper's attention-mask construction lowers to `aten.cumsum` on a bool tensor. tt-metal's cumsum has no bool/uint8 kernel, so the graph failed to compile.

## Fix
**(1)** Change the signature to `forward(self, input_features, decoder_input_ids, attention_mask=None)` so it matches the keyword inputs. `attention_mask` is optional, covering both the 2-key (tiny / base / Large_v3) and 3-key (Large_v3_Turbo) input dicts. Behavior is otherwise unchanged.

**(2)** Add a `cast_bool_cumsum_to_int32` backend pass (`python_package/tt_torch/backend/passes.py`) that, before argument-type markers are inserted, rewrites any `aten.cumsum` whose input is bool to cast the input to `int32` (via `aten._to_copy`). It is wired into `torch_pass_pipeline` in `backend.py` at the point common to both the AOTAutograd and export paths, so the cumsum runs on a supported dtype.

## Commits
- `Fix WhisperWrapper.forward to accept keyword inputs`
- `Cast bool cumsum inputs to int32 in tt backend`

## Verification — E2E CI passing
Ran the full whisper inference suite in CI with **both** fixes applied:

**✅ https://github.com/tenstorrent/tt-xla/actions/runs/29389357454 — `success`** (n150)
```
whisper_tiny: 1 passed, 6 deselected, 16 warnings in 97.08s
```
The model now compiles and runs end-to-end; the original `TypeError` and the bool-cumsum compilation failure are both resolved.


sgligorijevicTT (approved):
LGTM. Perhaps the cumsum dtype problem could have been fixed on the mlir level to benefit Jax too, but this is fine for now. Moving the fix could be a low priority followup work for somebody else.

mstojkovicTT (approved):

Issue #5504, opened 2026-07-02
Reduce tracy overhead in default path (host memory leak)


jameszianxuTT:
Ref: https://github.com/tenstorrent/tt-inference-server/issues/4430

I think Tracy default enablement in CMakeLists with `-DTTMLIR_ENABLE_PERF_TRACE=ON` is the cause of a possibly significant host memory leak, see discussion in above issue. I do not know about perf implications but I think it would have a nonzero impact.

We may want to either:
- Ship a tracy-disabled wheel; and do all tests with tracy on but just produce an extra wheel with tracy removed. Under the justification that enabling tracy is much more likely to induce regressions that disabling it.
- Find another way to drain tracy's in-memory buffers to prevent unbounded host memory accumulation in a long running program. We should find a minimally complex way to implement this; but no matter what we wouldn't be able to eliminate the host runtime overhead of having tracy sampling unless it is disabled outright

jameszianxuTT:
cc @pilkicTT , would appreciate your thoughts on this. I will ask Mo about the draining path

edit: No easy drain solution other than spinning up a capture server which will just transfer the memory leak into another proc's memory space (or buffer it to disk, depending on how capture server saves live trace data at runtime); neither of which is a desirable outcome. So we should just disable.

pilkicTT:
@jameszianxuTT 

yeah, it makes sense. we can definitely have a `release` wheel which we can ship with all profiler stuff turned off.

i am curious to know if `tt-metal` does the same? if not, then we probably have more overhead due to `tt-mlir` tracy messages. i've tried to remove those in the past, but haven't ultimately done that due to `ttrt perf` dependency on those.

kmabeeTT:
Assign to @pilkicTT - need some solution since models are generating hundreds of thousands of tokens in tt-shield evals + benchmarks and this leak hurts.

jameszianxuTT:
Update for today:
- TRACY_ON_DEMAND is a tt-metal tracy compile time option that may do what we want (bounding tracy), requiring testing to see if it works, but tracy docs suggest it is meant to bound host memory by disabling tracing until a server connects. This might be a good default for forge where we don't care about the trace data, but Mo points out that it is a potential footgun as it may cause lost perf trace data at the start, and letting users hold this footgun by exposing it as a compile command is maybe a bad idea
  - Also mlir / metal would need to be patched and uplifted, blocked at least by #5606 which I am also working on
[...]

PR #5635, opened 2026-07-14
Add --disable-perf-trace CLI option to tt-xla wheel build 

jameszianxuTT:
### Ticket
Partial towards temporary resolution to #5504 

### Problem description
No path to disable tracy in build

### What's changed
- Add cli option to setup.py bdist_wheel with --disable-perf-trace sets DTTMLIR_ENABLE_PERF_TRACE inside wheel
- It also sets wheel version to have `.tracy` at the end  - this will affect all builds; not sure if anything consumes the whl version
- Pipe input to call-build-ttxla

### Checklist

### DIFF
--- a/.github/workflows/call-build-ttxla.yml
+++ b/.github/workflows/call-build-ttxla.yml
@@ -18,6 +18,11 @@ on:
         default: release
         type: string
         required: false
+      disable_perf_trace:
+        description: 'Build tt-mlir without Tracy perf tracing. Set true for a perf-trace-off wheel.'
+        type: boolean
+        required: false
+        default: false
       force_rebuild:
         description: 'Force rebuild even if artifact exists'
         type: boolean
@@ -81,7 +86,7 @@ jobs:
         run: |
             source venv/activate
             cd ${{ steps.strings.outputs.work-dir }}/python_package
-            python setup.py bdist_wheel --build-type ${{ inputs.build_type }}
+            python setup.py bdist_wheel --build-type ${{ inputs.build_type }} ${{ inputs.disable_perf_trace && '--disable-perf-trace' || '' }}
 
       - name: Validate build type of wheel build
         shell: bash
--- a/python_package/setup.py
+++ b/python_package/setup.py
@@ -47,6 +47,10 @@ class SetupConfig:
     # --- Dataclass fields ---
     build_type: str = "release"
 
+    # Whether tt-mlir/tt-metal is built WITHOUT Tracy performance tracing.
+    # Perf tracing is on by default; Set by --disable-perf-trace
+    disable_perf_trace: bool = False
+
     # --- Necessary wheel properties ---
 
     @property
@@ -65,7 +69,11 @@ class SetupConfig:
             .strip()
         )
 
-        return "0.1." + date + "+dev." + short_hash
+        version = "0.1." + date + "+dev." + short_hash
+        if not self.disable_perf_trace:
+            # Distinguish perf-trace-disabled wheels within the local version segment.
+            version += ".tracy"
+        return version
 
     @property
     def requirements(self) -> list:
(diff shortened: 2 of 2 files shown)


jameszianxuTT:
cc @vvukomanTT - this adds the option to disable perf trace in wheel build but does not change anything except version metadata inside wheel unless --disable-perf-trace is added to build a dupe wheel in CI

sgligorijevicTT (approved):

kmabeeTT (approved):
Thanks, looks good to me - had a chat with claude about it, and seems safe.

kmabeeTT:
For the follow up work to use this and make the tt-forge releases use the non-tracy wheels... here is what claude thinks need to happen. It could go into this PR or a subsequent one after this is merged.. FYI @vvukomanTT + DevEx folks

--

This is a good foundation. One thing to flag for the follow-up that will actually wire the non-tracy wheel into the tt-forge release: the `.tracy` marker set here at **build time** gets erased at **release time**. `modify_wheel_version.sh` rewrites every wheel's version to the flat `release_tag`, so a tracy wheel and a non-tracy wheel both become `pjrt_plugin_tt-<release_tag>-cp312-cp312-manylinux_2_34_x86_64.whl` — identical filename/version, which collides on the s3pypi bucket.

### Follow-up PR would need to:

1. **Add the second build job** — a `build-ttxla` variant in `schedule-nightly.yml` with `disable_perf_trace: true` and a distinct artifact name.

2. **Preserve the distinction through the rename (the key change)** — in `call-publish-release-wheel.yml` (and `call-build-and-test-release-docker.yml`), publish both pjrt wheels and give the **tracy** one a suffixed tag at the `modify_wheel_version.sh` step, e.g. `<release_tag>+tracy` (PEP 440 local segment). Non-tracy keeps the plain `<release_tag>`. This yields distinct filenames and makes `pip install pjrt-plugin-tt==<tag>` resolve to the non-tracy wheel automatically.

3. **tt-forge pin needs no change** — `forge-wheel-setup.py` pins pjrt by exact filename URL for the plain `version_tag`. As long as **non-tracy owns the plain tag**, the forge wheel ships non-tracy for free. (Keep it this way — don't invert; `+` is a reserved URL char and would need `%2B` if the tracy filename were ever URL-pinned.)

4. **Decide the "default" variant** for the docker image and the documented `pip install ==<tag>` in the GitHub release notes (`call-publish-release.yml`). Non-tracy as default; document `==<tag>+tracy` for perf users.

### Watch-outs
- Flipping the released/default wheel to non-tracy is a behavior change — confirm the nightly `perf-benchmark` job still consumes a tracy wheel.
- The `vllm_tt` wheel is also pinned in the forge wheel and rewritten to `release_tag`. If it links tracy tt-mlir, it may need a matching non-tracy variant too, otherwise the forge wheel mixes a non-tracy pjrt with a tracy vllm.
- Two full builds ~doubles nightly build time (the flag forces a tt-mlir/tt-metal rebuild).

nsumrakTT:
@kmabeeTT @jameszianxuTT 
I don't think we should release another wheel without tracy. We already have 3 different builds and imho I think the "real" customer release (manylinux) should be built without tracy profiling, while ubuntu release (the one used by devs with default local build) and especially explorer, which is kind debug wheel, should use tracy.

This is not new to us as manylinux wheel doesn't support alchemist and multihost, so CI already support multi wheel behavior which makes it easy to group tests that require tracy (such as perf benchmarks) into test groups that would use ubuntu release instead of manylinux wheel. 

Please let me know what you think.

jameszianxuTT:
We are not going down this path - for now, disable the tracy build unconditionally. We are waiting for metal -> mlir -> tt-xla uplift to get a proper fix to native tracy behaviour that would obviate this change.

PR #5637, opened 2026-07-15
Fix composite layer_norm dtype mismatch with affine params under AOTAutograd

ctr-pmuruganTT:
## Problem
`tests/torch/ops/test_composite_ops.py::test_patched_layer_norm_module[*-True]` (affine LayerNorm) crashes on n150 with a SIGABRT:
```
FATAL | destination data type must match the source data type
LOG_ASSERT tt-mlir/runtime/lib/ttnn/runtime.cpp:988: dstDataType == getTensorDataType(src)
transferFromDevice -> INTERNAL: Error code 13 -> Aborted
```
Failed at `39ab41b7` (07-13 nightly), passed at its parent `bf9774cf`. The `*-False` (non-affine) variants are unaffected.

## Root cause
`composite_layer_norm` calls `F.layer_norm(bf16_input, ..., fp32_weight, fp32_bias)`. With **AOTAutograd on by default (#3795)**, functionalization/decomposition promotes the traced output dtype (due to the fp32 affine params) so it no longer matches the device tensor's dtype. Copying the result to host then fails the runtime dtype assert. Only affine LayerNorm hits this because the `-False` variants have no fp32 weight/bias.

## Fix
Cast the affine `weight`/`bias` to the input dtype inside `composite_layer_norm` before the composite, keeping the output dtype consistent:
```python
if weight is not None and weight.dtype != input.dtype:
    weight = weight.to(input.dtype)
if bias is not None and bias.dtype != input.dtype:
    bias = bias.to(input.dtype)
```

## Verification
- **Local (n150):** the 3 `-True` shapes pass (`3 passed`).
- **CI (n150, fresh build):** full `test_patched_layer_norm_module` — **6 [passed](https://github.com/tenstorrent/tt-xla/actions/runs/29340872800)** (run 29340872800), covering all `-True` and `-False` variants.


mstojkovicTT (approved):
fyi @sgligorijevicTT

Issue #5520, opened 2026-07-03
TP multi-chip greedy decode not run-to-run reproducible at near-tie logits (flaky vLLM stop-token / min-tokens tests)


alinakhanTT:
## Summary

Under 2D/TP sharding, greedy (`temperature=0`) decode is **not run-to-run reproducible** when the top-2 logits at a position are near-tied. The cross-chip fp reduction on the lm_head/logit path is not bit-reproducible, so the `argmax` winner can flip between two otherwise-identical `generate()` calls. This is a pre-existing multi-chip greedy-determinism limitation that TP sharding exposes at a near-tie, orthogonal to the DP/TP feature in #4947.

The two tests below assume cross-call greedy bit-reproducibility (they take `baseline_tokens[2]` from a greedy baseline, re-run greedy with it as a `stop_token_id`, and assert generation halts by position 3), so they flake intermittently on `n300_llmbox`

## Affected tests (n300_llmbox, TP)
- `tests/integrations/vllm_plugin/sampling/test_sampling_params.py::test_stop_token_ids[n300_llmbox]`
- `tests/integrations/vllm_plugin/sampling/test_sampling_params.py::test_min_tokens[n300_llmbox]`

## Proposed fix 
**fp32 dest-acc on the lm_head matmul + its reduce** (analogous to #5116, where a bf16 dest-acc matmul feeding an `argmax`/gather flipped the index and forcing fp32 dest-acc fixed it). Makes the near-tie insensitive to reduction order and keeps the batch placement that compiles

PR #5647, opened 2026-07-15
[vLLM] Relax greedy TP tests to smoke checks on n300_llmbox

alinakhanTT:
### Ticket

Relates to #5520

### Problem description

Three greedy (temperature=0) vLLM sampling tests flake intermittently on n300_llmbox. Each takes a token from a greedy baseline generate() and asserts on a second greedy generate(), i.e. they assume greedy is bit-reproducible across calls. Under TP that assumption breaks: at a near-tie logit the cross-chip reduction order is not stable run-to-run, so argmax (itself deterministic) can flip between two otherwise-identical calls. The tests were being masked with xfail(strict=False), which hid the flake and left them proving nothing on n300_llmbox.

### What's changed

Removed the xfail markers and scoped the reproducibility-dependent assertions to the targets where greedy is stable (single_device, n300), keeping only robust checks on n300_llmbox:

- test_greedy_determinism : keeps the strict "3 runs identical" assert off-TP; on n300_llmbox smoke-checks that each run is non-empty. Degenerate-output guard kept for all targets.
- test_stop_token_ids : keeps the strict "halts at/before position 3" assert off-TP; on n300_llmbox smoke-checks that generation runs.
- test_min_tokens : the core output length >= min_tokens contract runs on all targets; the two baseline-reproduction asserts (stops early without min_tokens; suppressed stop token absent from the first min_tokens positions) are gated to off-TP.

Exact sampler semantics (min_tokens / stop-token / logit-bias masking) remain fully proven with fixed logits, no hardware, in the synthetic suite (test_sampling_params_synthetic.py), and the strong contracts still run end-to-end on single_device and n300.

#5520 stays open to track the underlying cross-chip reduction-order nondeterminism; if bit-accurate greedy under TP is ever required, the fix belongs lower down (deterministic all-reduce in tt-metal), not in these tests.

### DIFF
--- a/tests/integrations/vllm_plugin/sampling/test_sampling_params.py
+++ b/tests/integrations/vllm_plugin/sampling/test_sampling_params.py
@@ -210,25 +210,19 @@ def test_sampling_has_diversity_when_temp_positive(llm, prompt):
     assert_diverse(texts)
 
 
-@for_targets(
-    single_device="push",
-    n300="push",
-    n300_llmbox=(
-        "push",
-        pytest.mark.xfail(
-            strict=False,
-            reason=(
-                "Flaky under TP: greedy decode is not run-to-run reproducible at "
-                "a near-tie logit (same-graph cross-chip reduction nondeterminism; "
-                "argmax is deterministic, the logits themselves flip). Tracked in "
-                "tt-xla #5520; proper fix (fp32 dest-acc on the logit path) in a "
-                "follow-up PR."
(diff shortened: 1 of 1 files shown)


mmanzoorTT (commented):
@kmabeeTT

kmabeeTT (approved):
Thanks! Looks good.

mmanzoorTT (approved):

PR #5652, opened 2026-07-16
[op-by-op] Add extract-unique-ops for global op dedup across all model IRs

ctr-lelanchelianTT:
### Ticket

Part of the op-by-op analysis setup — implements Nikola's item: *"extract unique ops across all model IRs, and run only those."*

### Problem description

Today op-by-op dedup only happens **within a single processing job** (via `filter_and_deduplicate_ops` in `op_by_op_test.py`), and the pipeline splits models across many jobs. So the same op (same shapes/attrs) is compiled and executed many times across jobs — a lot of redundant device time. There was no way to compute the set of ops that are unique **across all models** and run only those.

### What's changed

Adds an `extract-unique-ops` subcommand to `.github/scripts/ir_process_analyzer.py`:

- Reads every matched IR under `--root`, splits each into constituent ops, and **deduplicates globally by `op_string`** (same semantics as `filter_and_deduplicate_ops`).
- Writes each unique op as a standalone single-op MLIR module under `<output-root>/unique_ops/`, in the **exact directory layout `plan` and `op_by_op_test` already consume** (`unique_ops/<op_dir>/<prefix_dirs>/<file_prefix>_<idx>.mlir`), so it drops into the existing pipeline with no changes to the test.
- Emits a JSON manifest (per-op: op_name, origin models, file) plus reduction stats.
- **Continue-on-failure:** IRs that can't be read/parsed/split are logged, recorded, and skipped (never abort the run). The `op_by_op_infra` import is lazy, so `plan`/`dedupe`/`materialize` still run on runners without the explorer wheel.

Wiring this into `call-process-dumped-irs.yml` (extract once globally, then plan/distribute the unique ops across jobs) is a follow-up, since it needs the explorer wheel available at plan time — will be validated via the workflow's `workflow_dispatch`.

### Verification (local, against real nightly IR dumps)

- Output **round-trips cleanly** through `op_by_op_test` (`--folder=<output>/unique_ops`): ops extracted and executed on n150, 0 extraction failures.
- Smoke set: 26 ops → **12 unique (53.8% reduction)**.
- At scale over ~1k IRs, unparseable multi-chip/Shardy modules (e.g. `sdy.manual_computation … unknown mesh`) are skipped and the rest continue.

### Checklist

### DIFF
--- a/.github/scripts/ir_process_analyzer.py
+++ b/.github/scripts/ir_process_analyzer.py
@@ -5,6 +5,7 @@
 import argparse
 import json
 import math
+import re
 import shutil
 import sys
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


ctr-lelanchelianTT:
**Reduction metric (from production data):** using the current `ops-status` DB as ground truth, unique ops (keyed by op + in/out shapes) = 15,122 vs 30,871 total op-executions across all models → **~51% global reduction** (each unique op is shared by 1.7 models on average, up to 12). This matches the ~53.8% seen on the local smoke set, so running only globally-unique ops roughly halves device executions.

Note/finding: on very large graphs (e.g. MoE models with thousands of near-identical `tt.moe_expert_token_remap` custom_calls), the split step is single-threaded and very slow — worth adding a per-IR timeout/guard when wiring this into CI so one huge IR can't stall the extraction job.

nsumrakTT (approved):
Great job, thank you!

Just a note: You can also use manual "Run Test" workflow with checked "Dump IRs" and have irs dumped and processed.
Also, I wonder when unique ops are extracted how these ops will be reported - I see that you collect origin_models but I'm not sure how it is translated into json that ends up in database? Could you please attach test run link?

PR #5663, opened 2026-07-16
Temporarily disable profiling for manylinux (released) whl to solve tracy host memory leak

jameszianxuTT:
To be REVERTED when https://github.com/tenstorrent/tracy/pull/40/ is merged and uplifted to tt-xla.

### Ticket
#5504 

### Problem description
Temporary fix for #5504 until tracy patch lands and can be uplifted.

### What's changed
manylinux build takes is_ci() path -> under this path, disable tracy which is on by default

### Checklist

### DIFF
--- a/.github/workflows/pr-main.yml
+++ b/.github/workflows/pr-main.yml
@@ -84,6 +84,8 @@ jobs:
       regression_check: false # toggle on #5486 completion
       artifact_release_run_id: ${{ needs.build-ttxla.outputs.artifacts_run_id }}
       artifact_suffix: ${{ needs.build-ttxla.outputs.artifact_suffix }}
+      # temporary override due to tt-xla #5663
+      wheel_build: release
 
   perf-benchmark-accuracy:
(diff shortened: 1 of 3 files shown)


nsumrakTT (approved):

vvukomanTT:
Thanks for the change James.
Please double check that all the workflows that trigger Performance Benchmark still work (they might need tracy in some cases, please validate)

fyi @pilkicTT

nsumrakTT:
Perf benchmarks require profiling, so it should use "release" wheel. This is also true for nightly:
https://github.com/tenstorrent/tt-xla/blob/2630ce96e2a48cbfc5acd5cb7721484ececa6f59/.github/workflows/schedule-nightly.yml#L110

Also here:
https://github.com/tenstorrent/tt-xla/blob/2630ce96e2a48cbfc5acd5cb7721484ececa6f59/.github/workflows/pr-main.yml#L87
and here:
https://github.com/tenstorrent/tt-xla/blob/2630ce96e2a48cbfc5acd5cb7721484ececa6f59/.github/workflows/pr-main.yml#L104
Should supply wheel_build: "release". These perf benchmarks are run only on uplift, so it is skipped in "normal" PR.

kmabeeTT (approved):

jameszianxuTT:
tbd - running some perf benchmarks with release wheel instead of multilinux to see if they break. they really shouldn't...


https://github.com/tenstorrent/tt-xla/actions/runs/29605672100

pilkicTT (approved):

kmabeeTT:
Looks good... tt-metal -> tt-mlir -> tt-xla uplifts are quite healthy lately, so hopefully this is short lived and solves real memory leak affecting production models.  Apologies to anyone using tracy functionality in nightly releases (doubtful anyone will notice), there is good chance that this can be undone before next monthly release.

PR #5672, opened 2026-07-16
[vLLM] Fix AscendScheduler decode deadlock on partial-prefill continuation

### LINKED ISSUE #5664: [vLLM] AscendScheduler.update_from_output never clears scheduled_req_ids for a partial-prefill continuation, permanently deadlocking decode
tt-xla miror of https://github.com/tenstorrent/tt-inference-server/issues/4521

## Problem

`vllm_tt.scheduler.AscendScheduler` (`integrations/vllm_plugin/vllm_tt/scheduler/ascend_scheduler.py`) can permanently deadlock: decode scheduling stops entirely, with zero device activity and no error, until something external (e.g. a client disconnect aborting in-flight requests) force-clears the stuck internal state.

- Surfaced downstream as [tenstorrent/tt-inference-server#4521](https://github.com/tenstorrent/tt-inference-server/issues/4521) ("Forge single-chip LLMs intermittently hang under concurrent load"), initially misdiagnosed as a tt-metal device-read stall — the only thread visible in gdb at hang time is idling in `FDMeshCommandQueue::read_completion_queue`, which is actually the normal, healthy idle state once the scheduler stops feeding it work.
- **Introduced by** commit `39200a196`, "Enable chunked prefill: decouple buckets + runtime chunked SDPA" (#4986 / #5283), merged 2026-07-02.
- **Not a rare corner case in practice:** any wide-enough simultaneous admission batch (at `max_num_seqs` scale) whose shared per-step `token_budget` doesn't divide evenly across every admitted request will hit it — routine under real concurrent load (`vllm bench serve`, `lm-eval-harness` with `num_concurrent` at or above `max_num_seqs`, etc.), and not tied to being the first request ever served.
- Two things broke the downstream case open: (1) removing any client-side timeout and watching the server directly showed it never resolves on its own, ruling out a device-side race that eventually clears up by itself; (2) `py-spy` on the process showed the main thread spinning in vLLM core's idle-loop path, which only fires when the scheduler returns a completely empty schedule — pointing at `AscendScheduler`, not the device.

## Root cause

`schedule()`'s decode path is gated on `len(self.scheduled_req_ids) == 0`. A request that gets a partial prefill chunk (token budget ran out partway through it) is correctly excluded from `self.running` until it finishes — but its ID is unconditionally added to `self.scheduled_req_ids` in the same prefill-loop pass. `update_from_output()` only ever clears `scheduled_req_ids` for requests found in `self.running`:

```python
for request in self.running:
    req_id = request.request_id
    if scheduler_output.num_scheduled_tokens.get(req_id, 0) == 0:
        continue
    if req_id in self.scheduled_req_ids:
        self.scheduled_req_ids.remove(req_id)
```

- A partial/continuation request is never in `self.running` until it completes, so this loop can never clear its entry.
- Decode is then blocked for *every* running request, not just the partial.
- The partial itself can never be re-admitted to finish either, once `self.running` is at capacity (the prefill loop's own concurrency-cap check blocks it) — so `scheduled_req_ids` never clears.
- Permanent deadlock: no exception, no device activity, no timeout.

## Repro / regression test

Added `tests/integrations/vllm_plugin/test_ascend_scheduler_deadlock.py` (`test_partial_prefill_chunk_does_not_deadlock_decode`) — standalone, deterministic, hardware-free (no TT device, no model weights, no serving stack, ~3s). Drives the real `AscendScheduler` directly: a few small requests fully admitted and running, then one more given a token budget too small for its whole prompt, forcing a partial chunk. Asserts `scheduled_req_ids` clears right after the partial's own step and that the next scheduling step still makes decode progress for the other running requests.

Confirmed it fails against the pre-fix code and passes with the fix:

```
>   assert not sched.scheduled_req_ids, (...)
E   AssertionError: scheduled_req_ids still has stuck entries after the partial's own step: {'partial-victim'} (this is the #4521 deadlock)
E   assert not {'partial-victim'}
1 failed in 4.07s
```

`partial-victim`'s ID is exactly the one left stuck — everything else in that test run completes and clears normally.

## Fix and validation

- Candidate fix: clear `scheduled_req_ids` from everything actually scheduled that step (`scheduler_output.num_scheduled_tokens`, which includes partials) instead of iterating `self.running`. Will post as a follow-up comment linking the branch where the candidate diff + test live.
- Validated downstream: 21 consecutive conc=64/96/128 runs against a real Qwen3-8B server (including the unmodified production config — full model, full context) with the fix applied, zero hangs — versus reliably hanging on the very first attempt without it.

alinakhanTT:
### Ticket
closes #5664

### Problem description
`AscendScheduler` can permanently deadlock decode under concurrent load, the server goes quiet with no device activity and no error, recovering only when something external (e.g. a client disconnect) clears the stuck state. A partial prefill chunk is kept out of `self.running` until it finishes, but `update_from_output()` only cleared `scheduled_req_ids` by iterating `self.running`, so the partial's id lingered forever. Since decode only runs when `scheduled_req_ids` is empty, one stuck partial starves decode for every running request. Regression from #5283 (chunked prefill)

### What's changed
Clear `scheduled_req_ids` from what was actually scheduled this step (`num_scheduled_tokens`, which includes partials) instead of iterating `self.running`. One line, and it drops the old membership-check bookkeeping. Adds a standalone regression test.

### Checklist
- [x]  New regression test (`test_ascend_scheduler_deadlock.py`) reproduces the trigger directly against `AscendScheduler` on CPU (no device/weights, ~4s): fails on old code, passes with the fix.
- [x] Full CPU scheduler suite (deadlock + abort + batching): `6 passed`.
- [x] Validated downstream (per @kmabeeTT, #5664): 21 consecutive conc=64/96/128 runs against a real Qwen3-8B server, zero hangs.
- [x] Rerun Falcon3-7B-Instruct and Qwen3-4B release flow in tt-inference-server multiple times, all passing no hangs like without this fix (details in issue)

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/scheduler/ascend_scheduler.py
+++ b/integrations/vllm_plugin/vllm_tt/scheduler/ascend_scheduler.py
@@ -802,16 +802,10 @@ class AscendScheduler(Scheduler):
                 stale_req_ids[:3],
             )
 
-        # NOTE(woosuk): As len(self.running) can be up to 1K or more, the below
-        # loop can be a performance bottleneck. We should do our best to avoid
-        # expensive operations inside the loop.
-        for request in self.running:
-            req_id = request.request_id
-            num_tokens_scheduled = num_scheduled_tokens.get(req_id, 0)
-            if num_tokens_scheduled == 0:
-                # The request was not scheduled in this step.
-                continue
-            if req_id in self.scheduled_req_ids:
-                self.scheduled_req_ids.remove(req_id)
+        # Clear by what was scheduled, not self.running: a partial-prefill
+        # continuation stays out of self.running, so iterating it would leak the
+        # partial's id and block decode forever (gated on empty). (tt-xla #5664)
+        for req_id in num_scheduled_tokens:
+            self.scheduled_req_ids.discard(req_id)
 
         return super().update_from_output(scheduler_output, model_runner_output)
--- a/tests/integrations/vllm_plugin/test_ascend_scheduler_deadlock.py
+++ b/tests/integrations/vllm_plugin/test_ascend_scheduler_deadlock.py
@@ -146,6 +146,5 @@ def test_partial_prefill_chunk_does_not_deadlock_decode():
     sched.max_num_scheduled_tokens = _BLOCK_SIZE - 1
     out3 = sched.schedule()
     assert out3.total_num_scheduled_tokens > 0, (
-        "decode did not proceed for the other running requests (the #5664 "
-        "deadlock)"
+        "decode did not proceed for the other running requests (the #5664 " "deadlock)"
     )
--- a/tests/integrations/vllm_plugin/test_ascend_scheduler_deadlock.py
+++ b/tests/integrations/vllm_plugin/test_ascend_scheduler_deadlock.py
@@ -1,0 +1,151 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Regression test for tt-xla #5664: a partial prefill chunk must not deadlock
+decode. A partial is kept out of ``self.running`` but its id was added to
+``scheduled_req_ids``; ``update_from_output`` only cleared ids found in
+``self.running``, so the partial's id lingered and the decode path (gated on
+``scheduled_req_ids`` being empty) starved every other running request. Drives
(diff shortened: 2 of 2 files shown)


kmabeeTT:
Thanks! Added a bullet in PR desc to mention new results from late yesterday where I ran this fix against more models in tt-inference-server that are now passing.

kmabeeTT (approved):
Thanks!

mmanzoorTT (approved):

PR #5675, opened 2026-07-17
[CI] Auto-install vllm wheel for Run Test Single vllm_plugin paths

devisettymahidhar608:
### Ticket
Link to Github Issue

### Problem description
[CI] Auto-install vllm wheel for Run Test Single vllm_plugin paths

### What's changed
Run Test Single left extra-wheel empty unless install_vllm_wheel was checked, so vllm_plugin conftest failed with ModuleNotFoundError: vllm.
Enable extra-wheel=vllm when Dir contains vllm_plugin (or the flag is on).

Ci run:( ModuleNotFoundError: No module named 'vllm')
https://github.com/tenstorrent/tt-xla/actions/runs/29515498702/job/87689201030

Ci run:(Passing)
https://github.com/tenstorrent/tt-xla/actions/runs/29593576470

### Checklist

### DIFF
--- a/.github/workflows/manual-test-single.yml
+++ b/.github/workflows/manual-test-single.yml
@@ -70,7 +70,7 @@ on:
         required: false
         type: string
       install_vllm_wheel:
-        description: 'Install vllm-tt wheel in the test job'
+        description: 'Install vllm-tt (+ upstream vllm) in the test job. Auto-enabled when Dir contains vllm_plugin.'
         required: false
         default: false
         type: boolean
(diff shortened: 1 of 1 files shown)


sonalibaskaran2499 (approved):

vvukomanTT:
I don't understand the point of the PR, why don't we just use the checkbox to install the vllm wheel when needed?

devisettymahidhar608:
Hi @vvukomanTT 

> I don't  understand the point of the PR, why don't we just use the checkbox to install the vllm wheel when needed?


The checkbox still works — use it when you need the vllm wheel outside a vllm_plugin path.

This PR is for the common case where people set Dir to something under vllm_plugin but leave install_vllm_wheel unchecked. That leaves extra-wheel empty, so the job fails with ModuleNotFoundError: vllm ([example](https://github.com/tenstorrent/tt-xla/actions/runs/29515498702/job/87689201030)). Matrix presets already always set "extra-wheel": "vllm" for those paths; auto-detect makes Run Test Single match that so you don’t have to remember the checkbox every time.

kmabeeTT:
I think we can drop this PR and just use the checkbox , it's why it exists, and if you didn't know about it before, now you do :) - less code to maintain :)

vvukomanTT:
@devisettymahidhar608 
I agree with @kmabeeTT here, the functionality already exists, let's just use it.

PR #5689, opened 2026-07-17
Big failing test cleanup

sgligorijevicTT:
### Problem description
We let test failures accumulate on nightly. This is making it harder to keep track of new regressions.

### What's changed
I let an agent xfail or skip many tests.

### Checklist


jameszianxuTT (approved):
Thanks for the housekeeping. small nit is that the phrase "part of the big failing-test cleanup" makes it seem like this is some kind of event and can probably go without saying

Issue #5404, opened 2026-06-29
Update vLLM Benchmarks to run high seq len + b32 + high gpu_mem_util


kmabeeTT:
## Problem

Currently vllm benchmarks run max_model_len=128 - but after https://github.com/tenstorrent/tt-xla/issues/4986 is in place we have ability to target much higher seq len (have seen 64K successfully) for models, which matches production configs desired to be served in tt-inference-server, so we should do that.

## Request

- For at least these models, they should exist in vllm benchmarks for single chip, run as batch-32 and should target highest max_model_len supported
- Optimize highest `gpu_mem_utilization` setting for maximum KV Cache size and concurrency (but not too tight to cause minor DRAM changes in fragmentation resulting in OOM

 | Model | native max |
  |---|---:|
  | Qwen3-4B | 40960 |
  | Qwen3-8B | 40960 |
  | Llama-3.2-3B-Instruct | 131072 |
  | Llama-3.1-8B-Instruct | 131072 |
  | Falcon3-7B-Instruct | 32768 |

kmabeeTT:
chunked prefill change https://github.com/tenstorrent/tt-xla/issues/4986 got merged today, so this should be in good shape to do now (high seq len, b32, bfp8 weights, bfp8 kv cache).

kmabeeTT:
some settings that could be enabled here... but maybe we should reconsider since it will impact compile time (2-4x for single chip and QB2 models).
 
- chunked prefill
- b1 prefill + b16 prefill
- high max_model_len
- batch 32 for single chip
- high gmu

PR #5692, opened 2026-07-17
[vLLM] Broaden chunked-prefill coverage across push and nightly vLLM tests

### LINKED ISSUE #5691: Improve Chunked Prefill coverage in tt-xla CI (single chip, multichip)
### Problem

- Chunked prefill feature from https://github.com/tenstorrent/tt-xla/issues/4986 is great and important for large models to fix on device
- Coverage is not very high for this chunked prefill feature, and it is used (successfully) in single chip models productized in tt-inference-server (when it works) and multichip models as soon as it works well (reports from @ssaliceTT about hangs existing)
- It recently broke in single chip models due to https://github.com/tenstorrent/tt-xla/issues/5579 and only a single test in push failed: `tests/integrations/vllm_plugin/generative/test_prefill_recompile.py::test_no_prefill_recompile`

### Request

- Add some more single chip, multichip models with chunked prefill enabled (keep runtime in mind) to onPush and onNightly as appropriate, sometime to make it clear that if these break, we are in trouble.  Previous test above was xfailed for a week or so which then made the same models break in tt-shield CI.

There is request already to add it to vllm benchmark tests via https://github.com/tenstorrent/tt-xla/issues/5404

FYI @ssaliceTT @mmanzoorTT @ddilbazTT

alinakhanTT:
### Ticket

Closes #5691

### Problem description

Chunked prefill (#4986) is relied on by productized single-chip models and is coming online for multichip, but CI coverage was thin. The only per-PR guard was test_no_prefill_recompile; the three real correctness tests ran nightly-only and single-chip, and no multichip test exercised chunked prefill at all (every TP/DP test used max_model_len=32 with prompts that fit in a single chunk).

This left the feature fragile. When the tt-mlir uplift regressed chunked prefill in #5579 (insertTiledFixup tilized the SDPA page_table → TT_FATAL: Page table must be row major, which fatals at engine init and breaks all chunked prefill), only that single push test caught it. Once it was xfailed, the breakage went dark and later surfaced as broken productized models in tt-shield CI. The feature needs enough push coverage that no single xfail can hide a regression.

### What's changed

Broaden chunked-prefill coverage so it's protected on every PR across the single-chip, tensor-parallel, and data-parallel paths:
- Promoted the existing opt-125m batch chunked-prefill correctness test to push (was nightly-only), tiny model keeps it fast, and it's a real correctness signal on every PR alongside the recompile guard.
- Added n300 tensor-parallel chunked-prefill test (push + nightly).
- Added n300 data-parallel chunked-prefill test (push + nightly).
- Factored a shared long prompt (CHUNKED_PREFILL_PROMPT) into conftest.

The multichip tests set prefill_chunk_size well below the prompt length so the prompt splits into block-aligned chunks that route chunks 2..N through the cached-prefix chunked-SDPA path, with KV cache and attention sharded across chips, a path no prior multichip test covered. They generate greedily and assert output coherence, which catches a corrupted cached-prefix (garbage) or a shard-side hang. Impact: a #5579-class regression now fails multiple push tests across all three parallelism modes instead of slipping through on one xfailed test. (Benchmark coverage, #5404, is a deliberate follow-up in a separate PR.)

### Checklist

- [x]  this PR is the coverage; all three tests validated green on n300 hardware (single-chip push, TP, and DP) against the #5579-fixed tt-mlir pin

### DIFF
--- a/tests/integrations/vllm_plugin/generative/conftest.py
+++ b/tests/integrations/vllm_plugin/generative/conftest.py
@@ -2,7 +2,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 """Shared conftest for vLLM generative tests."""
-
 import json
 import math
 import re
--- a/tests/integrations/vllm_plugin/generative/conftest.py
+++ b/tests/integrations/vllm_plugin/generative/conftest.py
@@ -113,21 +113,6 @@ def assert_batch_grounded(outputs, checks=GROUNDED_BATCH_CHECKS) -> None:
         ), f"expected {expected!r} for {prompt!r}, got {text!r}"
 
 
-# A prompt comfortably longer than a small prefill_chunk_size (e.g. 128) so the
-# prompt splits into several block-aligned chunks and chunks 2..N exercise the
-# cached-prefix chunked-SDPA path (tt-xla #4986/#5691). Shared by the multichip
-# chunked-prefill tests.
-CHUNKED_PREFILL_PARA = (
-    "The history of computing spans many centuries. Early mechanical "
-    "calculators gave way to electromechanical machines, and then to the "
-    "electronic digital computers that define the modern era. Each generation "
-    "brought dramatic improvements in speed, reliability, and cost. "
-)
-CHUNKED_PREFILL_PROMPT = (
-    "Continue in English. " + (CHUNKED_PREFILL_PARA * 5) + "\nContinuation:"
-)
-
-
 def check_host_memory(model_name: str) -> float:
     """Assert child process RSS is below the known threshold for a model.
 
--- a/tests/integrations/vllm_plugin/generative/conftest.py
+++ b/tests/integrations/vllm_plugin/generative/conftest.py
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 """Shared conftest for vLLM generative tests."""
+
 import json
 import math
 import re
@@ -110,6 +111,21 @@ def assert_batch_grounded(outputs, checks=GROUNDED_BATCH_CHECKS) -> None:
         assert (
             expected.lower() in text.lower()
         ), f"expected {expected!r} for {prompt!r}, got {text!r}"
+
+
+# A prompt comfortably longer than a small prefill_chunk_size (e.g. 128) so the
+# prompt splits into several block-aligned chunks and chunks 2..N exercise the
+# cached-prefix chunked-SDPA path (tt-xla #4986/#5691). Shared by the multichip
+# chunked-prefill tests.
+CHUNKED_PREFILL_PARA = (
+    "The history of computing spans many centuries. Early mechanical "
+    "calculators gave way to electromechanical machines, and then to the "
+    "electronic digital computers that define the modern era. Each generation "
+    "brought dramatic improvements in speed, reliability, and cost. "
+)
+CHUNKED_PREFILL_PROMPT = (
+    "Continue in English. " + (CHUNKED_PREFILL_PARA * 5) + "\nContinuation:"
+)
 
 
 def check_host_memory(model_name: str) -> float:
--- a/tests/integrations/vllm_plugin/generative/test_data_parallel_generation.py
+++ b/tests/integrations/vllm_plugin/generative/test_data_parallel_generation.py
@@ -7,9 +7,10 @@ enable_data_parallel=True / enable_tensor_parallel=False builds an SPMD mesh
 (dp_size, 1): weights replicated, the input batch sharded on the "batch" axis
 so each replica sees a disjoint subset of sentences.
 """
+
 import pytest
 import vllm
-from conftest import assert_output_coherent, check_host_memory
+from conftest import CHUNKED_PREFILL_PROMPT, assert_output_coherent, check_host_memory
 
 
 @pytest.mark.push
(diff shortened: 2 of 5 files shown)


tests/integrations/vllm_plugin/generative/test_data_parallel_generation.py
```diff
@@ -86,6 +87,46 @@ def test_data_parallel_generation_n300_wider_batch(model_name: str):
     check_host_memory(model_name)
 
 
+@pytest.mark.push
+@pytest.mark.nightly
```
mmanzoorTT: Marking a test with `push` and `nightly` will make it run in both cases. Is this intentional?

tests/integrations/vllm_plugin/generative/conftest.py
```diff
@@ -112,6 +113,21 @@ def assert_batch_grounded(outputs, checks=GROUNDED_BATCH_CHECKS) -> None:
...
+# chunked-prefill tests.
+CHUNKED_PREFILL_PARA = (
+    "The history of computing spans many centuries. Early mechanical "
+    "calculators gave way to electromechanical machines, and then to the "
+    "electronic digital computers that define the modern era. Each generation "
+    "brought dramatic improvements in speed, reliability, and cost. "
+)
+CHUNKED_PREFILL_PROMPT = (
+    "Continue in English. " + (CHUNKED_PREFILL_PARA * 5) + "\nContinuation:"
+)
```
mmanzoorTT: `conftest.py` is meant for pytest configuration, fixtures, hooks, and test utilities. Adding a constant here changes the role of it. My suggestion is to move this constant some other file, may be `chunked_prefill_data.py` or `data.py` (for general purpose use).

tests/integrations/vllm_plugin/generative/test_data_parallel_generation.py
```diff
@@ -86,6 +87,46 @@ def test_data_parallel_generation_n300_wider_batch(model_name: str):
...
+@pytest.mark.parametrize("model_name", ["Qwen/Qwen3-0.6B"])
+def test_data_parallel_chunked_prefill_n300(model_name: str):
+    """Chunked prefill under DP on n300 (tt-xla #4986/#5691).
+
+    Each DP replica prefills its own sentence; prefill_chunk_size << prompt
+    length splits each into several block-aligned chunks so both replicas run
+    the cached-prefix chunked-SDPA path in parallel. Greedy for determinism;
+    coherence per replica catches a corrupted cached-prefix or a DP hang.
+
+    On push + nightly to protect chunked prefill in the DP path per #5691.
```
mmanzoorTT: comment on line 102-103 not relevant.

tests/integrations/vllm_plugin/generative/test_tensor_parallel_generation.py
```diff
@@ -77,6 +78,47 @@ def test_tensor_parallel_generation_llmbox_small(
...
+
+    prefill_chunk_size << prompt length splits the prompt into several
+    block-aligned chunks, so chunks 2..N route through the cached-prefix
+    chunked-SDPA path while the KV cache and attention are sharded across the 2
+    chips. No existing multichip test exercised this (all use max_model_len=32
+    with prompts that fit one chunk). Greedy for determinism; coherence catches
+    a corrupted cached-prefix (garbage) or a TP-shard hang.
+
+    On push + nightly to protect chunked prefill in the TP path per #5691 (a
+    #5579-class break fatals at engine init here too).
```
mmanzoorTT: No need for line 95-97

mmanzoorTT (commented):
These tests should run with either `push` or `nightly` but not both. My suggestion is to run `test_chunked_prefill.py` and one of DP/TP with `push` and last one with `nightly`.

tests/integrations/vllm_plugin/generative/test_data_parallel_generation.py
alinakhanTT: Split so nothing runs in both, test_chunked_prefill.py::test_chunked_prefill_batch_all_users_match (fast opt-125m) and the DP chunked-prefill test now run on push, and the TP chunked-prefill test runs on nightly!

tests/integrations/vllm_plugin/generative/conftest.py
alinakhanTT: moved CHUNKED_PREFILL_PARA/CHUNKED_PREFILL_PROMPT into a new chunked_prefill_data.py and updated the DP/TP imports to pull from there, so conftest.py stays fixtures/hooks/utilities only

tests/integrations/vllm_plugin/generative/test_data_parallel_generation.py
alinakhanTT: Removed!

tests/integrations/vllm_plugin/generative/test_tensor_parallel_generation.py
alinakhanTT: Removed!

tests/integrations/vllm_plugin/generative/conftest.py
```diff
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 """Shared conftest for vLLM generative tests."""
+
```
mmanzoorTT: Please revert this.

mmanzoorTT (approved):

tests/integrations/vllm_plugin/generative/conftest.py
alinakhanTT: Done!

PR #5673, opened 2026-07-16
[vLLM][MRv2] Add Model Runner v2 for TT (TTModelRunnerV2, single-device generate)

alinakhanTT:
### Ticket
closes  #5629

### Problem description
vLLM's Model Runner v2 splits the old monolithic runner into four pieces (Runner / ModelState / RequestState / InputBatch), but the v2 runner is Triton/CUDA-only with no non-GPU template, so TT can't subclass it, it has to fork and adapt, swapping every Triton/CUDA/UVA piece for a host-side equivalent. The plugin currently only runs on the legacy v1 runner; this brings up the v2 architecture on TT.

### What's changed
- The split v2 state: TTModelState, TTRequestState, TTInputBatch, and TTSamplingStates (+ a bridge that feeds TT's sampler), all host-side numpy/torch instead of Triton/UVA.
- The runner (model_runner_v2.py, TTModelRunnerV2): a fork of the upstream v2 GPU runner for the single-device generate path, request lifecycle, host substitutes for the Triton input-prep and attention slot-mapping, the @torch.compile(backend="tt") forward+sample graph, KV cache alloc, and the full worker-facing surface.
- Worker wiring: TT_USE_V2_MODEL_RUNNER selects the v2 runner, default off, so v1 stays the path until v2 is proven out.
- Tests: 81 CPU unit/integration tests covering the state pieces and every host stage of the runner.

### Verification
- [x] 81/81 CPU tests pass.
- [x] On-device (Wormhole): KV allocation and the compiled forward+sample path validated.
- [x] End-to-end, v2 vs v1 (Qwen2.5-0.5B): with the toggle on, v2 produces byte-identical greedy tokens to v1, both a single prompt and a 4-way concurrent batch (identical on all four).
- [x] End-to-end, v2 vs v1 (Qwen3-8B, num_hidden_layers=1): identical KV sizing (314,560 tokens) and byte-identical greedy tokens; no OOM.

### Notes
- Default off; v1 unchanged.
- Deferred (each guarded): multi-device SPMD, multimodal, LoRA, and grammar/cpu-sampling/prompt-logprobs. Genuinely large models still need multi-device sharding to fit, same as v1.

Issue #5629, opened 2026-07-14
[vLLM] Investigate and add support for Model Runner states


mmanzoorTT:
## Summary

Extend the tt-xla vLLM plugin to support vLLM Model Runner state management so that model execution can correctly initialize, maintain, and restore runner-specific state throughout the inference lifecycle.

## Motivation

The current plugin integration focuses on model execution but does not fully support Model Runner states used by vLLM. As vLLM continues to evolve, Model Runner state handling becomes important for correctness, compatibility with newer APIs, and enabling advanced execution workflows.

Supporting these states will allow the plugin to:

* Align with the current vLLM execution model.
* Improve compatibility with recent and future vLLM releases.
* Enable proper initialization and restoration of runner state.
* Reduce the need for plugin-specific workarounds.

## Proposed Changes

* Add support for Model Runner state initialization.
* Track and manage runner state throughout the inference lifecycle.
* Implement serialization/deserialization or checkpointing hooks where required by the vLLM API.
* Ensure state transitions are correctly synchronized with tt-xla execution.
* Update the plugin interface to expose any required state management APIs.
* Add compatibility handling for supported vLLM versions, if necessary.

## Acceptance Criteria

* [ ] The tt-xla vLLM plugin supports Model Runner state initialization.
* [ ] Runner state is correctly maintained across inference requests.
* [ ] State restoration works as expected where supported.
* [ ] Existing functionality remains backward compatible.
* [ ] Unit tests cover Model Runner state lifecycle.
* [ ] Integration tests validate end-to-end execution with Model Runner states enabled.
* [ ] Documentation is updated to describe the implementation and any limitations.

## Additional Context

The implementation should closely follow the upstream vLLM Model Runner abstractions to minimize maintenance overhead and simplify future upgrades.

PR #5693, opened 2026-07-17
[vLLM][MRv2] Add multi-device (TP/DP/DP+TP) support to TTModelRunnerV2

alinakhanTT:
### Ticket
Part of #5629

### Problem description
The V2 runner from #5673 only ran on a single device. Production models are sharded across chips, so V2 can't replace V1 until it does tensor-parallel and data-parallel too. This adds TP, DP, and DP+TP to the V2 runner.

### What's changed
- Build the SPMD mesh in the runner __init__ (TP/DP mode selection, dp_size, max_num_seqs rounding), and have the worker pass the real parallel config through.
- Shard the weights in load_model (shard_model plus the vocab TP-rank patch).
- Add the forward SPMD annotations (input pinning, hidden-state and logits sharding constraints).
- Shard the KV cache and the per-step paged tensors on the DP batch axis.
- Pin each request's DP batch row to its slot so it keeps the same replica across prefill and decode (this was the tricky one).
- Warm up the graph in capture_model with valid dummies.

### Verification
- [x] 88 CPU tests pass.
- [x] On device (8x Wormhole): TP gives byte-identical greedy tokens to V1.
- [x] On device: DP and DP+TP give byte-identical greedy tokens to V1 on both batch slots (Qwen3-0.6B).

### Notes
- Stacked on #5673; I'll retarget the base to main once that lands.
- Still default off behind TT_USE_V2_MODEL_RUNNER, so V1 is untouched.
- The rest of the V2 features (multimodal, LoRA, grammar, prompt-logprobs, mrope, spec-decode) are still deferred and will land together in one more PR.

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner_v2.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner_v2.py
@@ -558,9 +558,22 @@ class TTModelRunnerV2:
         1-token decodes ahead of the prefills so the multi-pass loop can emit
         pure decode passes (dispatched to the decode graph) before prefills.
         """
+        sched = scheduler_output.num_scheduled_tokens
+        if self.dp_size > 1:
+            # DP: a request's batch position determines its DP replica, so it must
+            # be stable across prefill and decode (KV is written and read on the
+            # same replica). Order all active slots ascending by slot and pad the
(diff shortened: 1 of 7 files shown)


alinakhanTT:
Superseded by #5726, which consolidates the full v2 transition (single-device + multi-device TP/DP + all features) into one PR.

PR #5717, opened 2026-07-20
[benchmark] Use TP-only mesh for llama_3_1_70b_tp_qb2 to fix decode PCC (#5487)

### LINKED ISSUE #5487: llama_3_1_70b_tp PCC error on QB2
llama_3_1_70b_tp_qb2 [fails](https://github.com/tenstorrent/tt-xla/actions/runs/28512285954/job/84545923972) on  QB2 with PCC error and thrash output:

```
Prefill PCC verification passed with PCC=0.996464, rel_l2=7.732948e-02
FAILED

=================================== FAILURES ===================================
____________________________ test_llama_3_1_70b_tp _____________________________
tests/benchmark/test_llms.py:1874: in test_llama_3_1_70b_tp
    test_llm_tp(
tests/benchmark/test_llms.py:251: in test_llm_tp
    test_llm(
tests/benchmark/test_llms.py:139: in test_llm
    results = benchmark_llm_torch_xla(
tests/benchmark/benchmarks/llm_benchmark.py:849: in benchmark_llm_torch_xla
    decode_pcc_value >= required_pcc
E   AssertionError: First decode PCC failed. PCC=-0.000000, Required=0.94
```

Next steps:
- check for opt-level < 2

rpavlovicTT:
## Problem

`test_llama_3_1_70b_tp_qb2` uses the loader's default 70B mesh `(2, N//2)`. Despite the `"batch"` axis name, this is **not** data-parallel — inputs are replicated and the weights are sharded across **both** axes (2D tensor parallelism; the only qb2 model doing so, and the only one exercising `distributed_rms_norm`). That multi-axis weight sharding corrupts the first-decode KV-cache read (prefill fine ~0.995, first-decode logits garbage → decode PCC ≈ 0). See #5487.

Per the investigation in #5487, this is **not** the decode op, opt level, bf8 cache, bf4 MLP, or trace — all ruled out by experiment. `test_falcon3_10b_tp_qb2` uses the same decode op + bf8 cache but a **single-axis TP `(1,N)` mesh** and decodes fine.

## Fix

Use a single-axis TP-only `(1, N)` mesh for this test: decode PCC recovers to ~0.999. This also unblocks `optimization_level=2` (previously pinned to `1` due to an opt-2 hang).

This is a temporary workaround — collapsing to a single axis drops CI coverage of the multi-axis weight-sharding path and `distributed_rms_norm` on qb2. Restoring the 2D weight-sharded mesh is tracked in **#5738**.

## Verification

This change is byte-identical to branch `rpavlovic/fix-llama70b-qb2-decode-tp-mesh`, which was validated in CI as documented in #5487 (qb2-blackhole, 80 layers, opt-2 — decode PCC recovered to ~0.999): run https://github.com/tenstorrent/tt-xla/actions/runs/29535389034. Locally in this session only a syntax parse was performed; the accuracy numbers above come from that prior CI run.

Fixes #5487
Refs #5738

### DIFF
--- a/tests/benchmark/test_llms.py
+++ b/tests/benchmark/test_llms.py
@@ -1878,6 +1878,16 @@ def test_llama_3_1_70b_tp_qb2(
     )
 
     variant = ModelVariant.LLAMA_3_1_70B_INSTRUCT
+
+    # The loader's default 70B mesh on QB2 is (2, N//2) = a 2-way data-parallel
+    # "batch" axis x tensor-parallel "model" axis. That DP batch axis corrupts the
+    # first-decode KV-cache read (garbage/NaN logits, decode PCC ~0 -> issue #5487),
+    # even though the SDPA-decode op itself is correct (Falcon-10B, which uses a
+    # TP-only mesh + the same op, decodes fine). Use a TP-only mesh here: decode PCC
+    # recovers to ~0.999, and TP=N uses less memory/chip than DP=2 x TP=(N/2).
+    def _qb2_tp_only_mesh(model_loader, num_devices):
+        return (1, num_devices), ("batch", "model")
+
     test_llm_tp(
         ModelLoader,
         variant,
@@ -1888,11 +1898,12 @@ def test_llama_3_1_70b_tp_qb2(
         batch_size=batch_size,
         max_output_tokens=max_output_tokens,
         decode_only=decode_only,
+        mesh_config_fn=_qb2_tp_only_mesh,
         weight_dtype_overrides={
             "model.layers.*.mlp.gate_proj.weight": "bfp_bf4",
             "model.layers.*.mlp.up_proj.weight": "bfp_bf4",
         },
-        optimization_level=1,  # flaky: occasionally hangs in CI with optimization_level=2
+        optimization_level=2,  # TP-only mesh fixes the opt-2 decode failure (#5487)
     )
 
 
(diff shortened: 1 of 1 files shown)


tests/benchmark/test_llms.py
```diff
@@ -1878,6 +1878,16 @@ def test_llama_3_1_70b_tp_qb2(
     )
 
     variant = ModelVariant.LLAMA_3_1_70B_INSTRUCT
+
+    # The loader's default 70B mesh on QB2 is (2, N//2) = a 2-way data-parallel
```
mvasiljevicTT: One note here: it was never data parallel, as in batch sharded on separate devices, it was tensor parallel with weights sharded on both axes and inputs replicated.

mvasiljevicTT (approved):
As far as I know, this was the only model we have for qb2 that shards weights on multiple axes, so for example this was the only model that has distributed_rms_norm op. It is ok to disable it temporarily, but please make an issue tracking this.

tests/benchmark/test_llms.py
rpavlovicTT: Good catch — you're right, it was never data-parallel. Fixed the comment in 0a6ec0c: it now describes the default `(2, N//2)` mesh as replicated inputs with weights sharded across both axes (2D tensor parallelism, the only qb2 model doing so / using `distributed_rms_norm`), rather than a "DP batch axis".

Opened #5738 to track root-causing decode under the 2D weight-sharded mesh so we can restore it (and regain `distributed_rms_norm` coverage). Referenced it in both the code comment and the PR.

Issue #5606, opened 2026-07-13
[uplift] Multihost test failing on uplift


sgligorijevicTT:
Run link: https://github.com/tenstorrent/tt-xla/actions/runs/29223626701/job/86812236290?pr=5592
A llama multihost job is failing in uplift. Rerunning the job doesn't seem to help.

jameszianxuTT:
Small update - cannot reproduce locally. Trying to run it in CI with extra debug info but stalling on getting a runner, will update when I can see it failing in CI

jameszianxuTT:
Also passes in an isolated CI test - http://github.com/tenstorrent/tt-xla/actions/runs/29267462925/job/86876641192... which makes no sense.

jameszianxuTT:
Also failing in tt-metal uplift; expected if it subsumed the mlir uplift - https://github.com/tenstorrent/tt-xla/actions/runs/29251183579/job/86874593841

jameszianxuTT:
This makes even less sense - from the base `uplift` branch with no changes it also passes https://github.com/tenstorrent/tt-xla/actions/runs/29269851294/job/86890748276. I will rerun it in the onPR context and maybe experiment with the wheel, maybe that is corrupted somehow.

jameszianxuTT:
I have spent most of the day today trying to figure out WTF is going on in this issue, I am not that close to finding an answer:

1. This works locally from `uplift` branch
2. [This works on CI](https://github.com/tenstorrent/tt-xla/actions/runs/29269851294/job/86890748276.) when run from a debug uplift branch, and also run from the uplift branch alone using the `Run Test Single` workflow 
3. This also fails in every rerun of mlir uplift
4. It also fails in tt-metal uplift (which is expected since it would include the contents of this ttmlir uplift) https://github.com/tenstorrent/tt-xla/actions/runs/29251183579/job/86874593841
5. Downloading the failing original wheel used in (3) and running a local test causes failure with kInternal, with a non-helpful stacktrace though it suggests this happens at initialization and not at runtime. This detail is corroborated by the 1m error time shown in GH logs, which is very short and unlikely to have made it to runtime.
<img width="1139" height="292" alt="Image" src="https://github.com/user-attachments/assets/058dfe5c-336a-4601-be84-6096e146b142" />
[...]

PR #5725, opened 2026-07-20
Run llmbox multi-host test via topology infra, drop subprocess wrapper

jameszianxuTT:
### Ticket
None

### Problem description
After recent multihost infra improvements are merged, the local subproc multihost test can be shifted onto that architecture instead of creating a custom env-var setter test wrapper.

### What's changed
- This reduces an extra level of subprocess indirection in tests which makes for easier debug (less lost stdout) and cleaner integration with the rest of CI. Especially since this test is failing randomly + inscrutably more often now, eg. #5606 
- Remove old tests/torch/multi_host/llmbox/test_multihost.py

### Checklist
- [x] Example run - https://github.com/tenstorrent/tt-xla/actions/runs/29758923461


jameszianxuTT:
cc @sgligorijevicTT , this may help with diagnostics from the multihost llmbox test if we see ND failures again as we won't have one pytest capturing the results from another pytest + failing to regurgitate

vvukomanTT (approved):

sgligorijevicTT (approved):
Very cool, thanks

AleksKnezevic (approved):

Issue #5257, opened 2026-06-17
[Bringup]-  Bring up image+text path on-device at TP=8


saiarthiraguram:
**Description:**

Arch: n300-llmbox (8× Wormhole_b0) · Mesh: (1, 8) Megatron 1D TP · Dtype: bf16 · real weights.
Harness: `mm_harness.py --modality image` · CPU sanity: `mm_cpu_sanity.py` → `mm_cpu_sanity.json`, `mm_golden_image.pt`.
Inputs built by the loader: `ModelLoader.load_inputs(include_image=True)` → `load_image_inputs`.

**Status:** CPU sanity ✅ and on-device TP=8 ✅ both done. Image+text path runs on the n300-llmbox at TP=8 with **last-token PCC 0.9937** vs the CPU golden (§3); through the runner the **full-tensor PCC is 0.9675** (§5).

## 1. Gemma4 image interface (what feeds the vision path)

Gemma4Unified uses an encoder-free vision embedder (no SigLIP/ViT tower): raw merged pixel patches → LayerNorm → Dense → factorized 2D positional embedding → RMSNorm → Linear, projected into the 3840-d text embedding space and `masked_scatter`-ed into the token stream at the image-token positions.

`AutoProcessor.from_pretrained("google/gemma-4-12B")` (`Gemma4UnifiedProcessor`) turns `"<|image|>" + text` + a PIL image into:

| input | shape | dtype | role |
|-------|-------|-------|------|
| input_ids | (1, P) | int64 | text tokens with image-token (258880) span, bracketed by BOI 255999 / EOI 258882 |
| attention_mask | (1, P) | int64 | standard mask |
| mm_token_type_ids | (1, P) | int64 | 0=text, 1=image — marks multimodal positions |
| pixel_values | (1, 280, 6912) | float32 | merged model patches (48×48×3), padded to 280 |
| image_position_ids | (1, 280, 2) | int64 | per-patch (x,y); (-1,-1) = padding (stripped in get_image_features) |

One image → up to 280 soft tokens; the processor inserts exactly the valid patch count as image tokens (`#image_tokens == #valid_patches`).

## 2. CPU sanity (reference) — ✅ PASS

`mm_cpu_sanity.py` ran one image+text forward on CPU (bf16) with a synthetic 224×224 gradient image and prompt `"<|image|>Describe the image."`:

| field | value |
|-------|-------|
| processor | Gemma4UnifiedProcessor |
[...]

PR #5726, opened 2026-07-20
[vLLM][MRv2] Transition to Model Runner v2: single + multi-device (TP/DP) + features

alinakhanTT:
## Transition to vLLM Model Runner v2 (MRv2) for the TT plugin

The old monolithic `TTModelRunner` (a fork of upstream's TPU runner) is replaced by MRv2's four piece split: runner (`TTModelRunnerV2`) plus `TTModelState` plus `TTRequestState` plus `TTInputBatch`, forked and adapted for TT (upstream's v2 is Triton/CUDA gated, so we cannot subclass `GPUModelRunner`). Parallelism uses `mark_sharding`/GSPMD (no explicit collectives).

### Base

Rebased onto `main` at vLLM **0.25.1** (post #5744). This PR previously targeted 0.22.1; the branch was force-pushed onto the post-uplift main, so earlier inline review comments may show as outdated. Five commits:

1. Forward-port MRv2 runner onto vLLM 0.25.1
2. Fix logger initialization in MRv2
3. Split forward/sampling graphs; pin mrope + image tests to v1
4. Adapt v2 sampling bridge to the uplifted `from_input_batch`
5. Reject `speculative_config` on the v2 runner

### Default runner
`TTConfig.use_v2_model_runner` now defaults to True, so MRv2 is the default path. Set it False in `additional_config` to fall back to v1. Pooling models always use the v1 runner.

### What is included
* Phase 1 `TTModelState`, Phase 2 `TTRequestState` plus `TTInputBatch`, Phase 3 `TTModelRunnerV2` (host state machine: lifecycle, decode first batch select, host input and attn prep, two phase `execute_model`/`sample_tokens`, KV alloc, compiled forward plus sample, warmup).
* Multi device: SPMD device mesh, weight sharding, forward/KV/DP input sharding for TP, DP, and DP+TP.
* Features: prompt logprobs, cpu_sampling (Gumbel max host sampling), text only mrope, LoRA (mixin driven), multimodal `get_mm_embeddings`, structured output and grammar (bitmask masking), plus operator logging.
* Separately compiled forward->logits and sampling graphs, so post processing no longer re-specializes the model forward.

### 0.25.1 adaptations

* `TTModelState`: sets `inputs_embeds_size` itself (the 0.25.1 `ModelState` base sets it in its concrete `__init__`, which TT does not call because it would build a CUDA `EncoderRunner`), and `postprocess_state` matches the new `(idx_mapping, num_sampled, num_computed_tokens)` signature.
* `SamplingBatchView`: the uplift extended `XLASupportedSamplingMetadata.from_input_batch` to read `logits_processing_needs_token_ids`, `spec_token_ids`, `logitsprocs`, and `logitsprocs_need_output_token_ids`. The v2 batch view now exposes all four with the values the v2 generation path implies: pooling-only token-id needs are always False, `num_speculative_steps` is pinned to 0 so there are no drafts, and neither TT runner builds custom logits processors.
* `allowed_token_ids`: the uplift split the mask into a bool `allowed_token_ids_mask` (upstream contract) and a float `allowed_token_ids_additive_mask`. `sampler.py` already prefers the additive variant, so v2 keeps the additive path that avoids on-device `masked_fill` with bool tensors; only the CPU test's assertions needed widening to cover both.
* Speculative decode now raises instead of silently degrading (see below).

### Validation

* 109 MRv2 CPU tests pass on 0.25.1 (`tests/integrations/vllm_plugin/mrv2/`).
* Single device prefill + decode smoke runs on TT hardware on 0.25.1.
* The generative/sampling parity sweeps below were run on the 0.22.1 base; re-validation on 0.25.1 is via this PR's CI run.
  * Generative and sampling parity sweeps green on a single device n150 CI runner under MRv2 (v2 matches v1).
  * grammar and structured output: TP validated (a phone number regex produces the expected formatted output).
  * prompt logprobs and cpu_sampling: TP validated (v2 matches v1).
  * mrope: passes on a single device n150 CI runner.
  * Gemma 4 text only: passes on a single device Blackhole under MRv2.

### Known gaps (present on v1 too, not blockers)
* Gemma 4 multimodal image path OOMs on a single device: the vision forward needs more per bank DRAM than remains after the unsharded weights fill the device. Reproduces on v1 identically, so v2 does not regress it. Tracked separately; the image path already runs at TP=8 (#5257).
* LoRA on device e2e deferred: LoRA lowering is broken plugin wide on 0.20.2 and 0.22.1 and fails on v1 too, so v2 as default does not regress it.
* Speculative decode: the v2 runner pins `num_speculative_steps=0`, so a `speculative_config` would have had its drafts silently dropped now that v2 is the default. It raises `NotImplementedError` pointing at `use_v2_model_runner=False`. The NGram spec decode work in #5542 is v1 only and has no e2e test, so nothing regresses; porting it to v2 is tracked in #5795 Phase 3.
* Encoder decoder attention and multi type hybrid KV cache raise `NotImplementedError` guards, matching v1.


vvukomanTT (approved):

tests/integrations/vllm_plugin/conftest.py
```diff
@@ -17,11 +17,42 @@
...
+
+    def patched_init(self, *args, **kwargs):
+        additional_config = dict(kwargs.get("additional_config") or {})
+        additional_config["use_v2_model_runner"] = True
+        kwargs["additional_config"] = additional_config
+        return orig_init(self, *args, **kwargs)
+
+    monkeypatch.setattr(vllm.LLM, "__init__", patched_init)
+
+
```
mmanzoorTT: Why do we need them? Isn't plugin using `v2` as default for generative models and `v1` for pooling models?

integrations/vllm_plugin/vllm_tt/model_state.py
```diff
@@ -0,0 +1,232 @@
...
+        self.max_num_reqs = self.scheduler_config.max_num_seqs
+        self.max_num_tokens = self.scheduler_config.max_num_batched_tokens
+        self.dtype = self.model_config.dtype
+
+        # Rope/mrope state: upstream uses get_rope_state(), which is Triton-backed
+        # and does not run on TT. The common text-generation path uses plain 1D
+        # positions (rope_state is None -> prepare_inputs returns {}), matching
+        # DefaultModelState's common case. mrope models require a TT rope
+        # substitute; deferred to a later phase.
+        # TODO(mrv2): TT rope/mrope state for models with uses_mrope.
```
mmanzoorTT, line 57: Use `use_model_runner_v2=False` for `test_mrope.py`.

integrations/vllm_plugin/vllm_tt/model_state.py
```diff
@@ -0,0 +1,232 @@
...
+        self.rope_state = None
+
+        # TODO(mrv2): construct the TT multimodal encoder runner here when
+        # get_mm_embeddings is implemented (Phase 3). Upstream builds an
```
mmanzoorTT: Few tests in `test_gemma4_generation.py` and `test_tensor_parallel_generation.py` are using images as input. They will not run with V2. Add the flag to use V1.

integrations/vllm_plugin/vllm_tt/model_runner_v2.py
```diff
@@ -0,0 +1,2128 @@
...
+                want_prompt_hs,
+                inputs_embeds,
+                apply_grammar,
+                require_struct_decoding,
+                grammar_bitmask,
+                bitmasks,
+            )
+
+    @torch.compile(backend="tt", fullgraph=True, dynamic=False)
+    def _forward_and_sample_compiled(
```
mmanzoorTT: Compiling everything together will generate one big graph for each possible scenario. This will be problematic once we start adding different post-processing combinations like apply_grammar=True/False, greedy/non-greedy sampling, chunked prefill, speculative decoding etc.

tests/integrations/vllm_plugin/conftest.py
alinakhanTT: this flag predates making v2 the default, so it's stale. Removed the pytest_addoption and the autouse fixture. Tests that can't run on v2 opt out explicitly with additional_config={"use_v2_model_runner": False}. Done in 9392e14e4

integrations/vllm_plugin/vllm_tt/model_state.py
alinakhanTT: Done

integrations/vllm_plugin/vllm_tt/model_state.py
alinakhanTT: Done

integrations/vllm_plugin/vllm_tt/model_runner_v2.py
alinakhanTT: Thanks for pointing that out!  Split into two separately-compiled graphs in 9392e14e4

tests/integrations/vllm_plugin/conftest.py
```diff
@@ -10,6 +10,7 @@
 holding the TT device outlives the test — hanging the next one. Shut
 them down explicitly on failure.
 """
+
```
mmanzoorTT: Please revert it.

integrations/vllm_plugin/vllm_tt/model_runner_v2.py
```diff
@@ -0,0 +1,2445 @@
...
+    for _, module in model.named_modules():
+        if isinstance(module, BaseLayerWithLoRA):
+            module._original_set_lora = module.set_lora
+            module._original_reset_lora = module.reset_lora
+            module.set_lora = _tpu_set_lora.__get__(module, module.__class__)
+            module.reset_lora = _tpu_reset_lora.__get__(module, module.__class__)
+
+
+class TTModelRunnerV2(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
+    """Tenstorrent MRv2 model runner (see module docstring)."""
```
mmanzoorTT:
```suggestion
    """Tenstorrent MRv2 model runner."""
```

integrations/vllm_plugin/vllm_tt/model_runner_v2.py
```diff
@@ -0,0 +1,2445 @@
...
+        # 1-byte accounting stand-in so vLLM budgets blocks for the BFP8 footprint.
+        self.kv_cache_spec_dtype = (
+            torch.uint8
+            if getattr(tt, "experimental_kv_cache_dtype", None) == "bfp_bf8"
+            else self.kv_cache_dtype
+        )
+
+        self.block_size = self.cache_config.block_size
+        self.max_model_len = self.model_config.max_model_len
+        self.most_model_len = envs.VLLM_TPU_MOST_MODEL_LEN
```
mmanzoorTT: `most_model_len` is removed in v1 with vllm 0.26.0 uplift. Please remove all its uses/paths in v2.

mmanzoorTT (commented):
- There is no need to mention `mirrors v1` or specific vLLM version info for the changes in the comments.
- Comments are too much verbose and does not add any value to the code.

mmanzoorTT (commented):
Please run it with vllm v0.26.0. `MultiGroupBlockTable` is not matching with upstream.

tests/integrations/vllm_plugin/conftest.py
alinakhanTT: Done

integrations/vllm_plugin/vllm_tt/model_runner_v2.py
alinakhanTT: Done

integrations/vllm_plugin/vllm_tt/model_runner_v2.py
alinakhanTT: Done

integrations/vllm_plugin/vllm_tt/model_runner_v2.py
```diff
@@ -0,0 +1,2520 @@
...
+
+    def _remove_request(self, req_id: str) -> None:
+        """Free a request's slot across every per-slot table. Idempotent."""
+        slot = self.req_states.req_id_to_index.get(req_id)
+        if slot is None:
+            return
+        self.req_states.remove_request(req_id)
+        self.sampling_states.remove_request(slot)
+        self.block_table.clear_row(slot)
+        self.num_prompt_logprobs.pop(req_id, None)
```
kmabeeTT, line 586: `_remove_request` pops `num_prompt_logprobs` but not `in_progress_prompt_logprobs`, so a request with `prompt_logprobs` enabled that gets aborted and resubmitted under the same `req_id` (already a supported path, see `test_readd_same_id_clears_stale_slot`) picks up the stale `LogprobsTensors` sized for the old prompt in `_get_prompt_logprobs_dict` and writes the new request's logprobs into it, wrong data or an out-of-bounds write if the new prompt is longer. v1's `input_batch.py` pops both dicts in `remove_request`, this should match. Can we get a one line fix plus a test for abort+resubmit with `prompt_logprobs` on included in this PR before merge given it's silent corruption not a crash.

integrations/vllm_plugin/vllm_tt/model_runner_v2.py
```diff
@@ -0,0 +1,2520 @@
...
+
+                if (is_embed := pos_info.is_embed) is not None:
+                    is_embed = is_embed[start_idx:end_idx]
+                    mm_embeds_item = encoder_output[curr_embeds_start:curr_embeds_end]
+                else:
+                    mm_embeds_item = encoder_output[start_idx:end_idx]
+
+                # Column of this request's row where the placeholder begins
+                # (input_ids stores scheduled tokens from column 0).
+                col_base = start_pos - num_computed_tokens
```
kmabeeTT, line 1026: `col_base` here assumes column 0 of every row is `num_computed_tokens`, but `_prepare_input_tokens` shifts a row's column 0 to `computed - lead` when `row_lead` is nonzero under spec decode block realignment, and this function never reads `row_lead`. Looks like it's masked in the common case because by the time a row is drafting its mm placeholder is usually already behind `num_computed_tokens` and gets skipped, but nothing prevents a request with a not-yet-fully-computed mm segment from also being mid-draft, and there's no test crossing mm and spec decode either direction. Either add the `row_lead` offset here or assert the two features can't combine yet, matching how DP+spec decode is handled elsewhere in this file.

integrations/vllm_plugin/vllm_tt/input_batch_v2.py
```diff
@@ -0,0 +1,160 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+import numpy as np
+import torch
+from vllm.utils import random_uuid
+
+
+class TTInputBuffers:
```
kmabeeTT: `self.input_buffers` is constructed once in `model_runner_v2.py` but never passed anywhere. `TTInputBatch.make_dummy` is the only thing that consumes a `TTInputBuffers`, and it's never called from `model_runner_v2.py` at all, only exercised directly by `test_input_batch.py` with its own separate instance. Please either wire `self.input_buffers` into `make_dummy` if that's the intended use, or delete `self.input_buffers` (and `make_dummy`/`TTInputBuffers` too if nothing else is meant to call them) — as-is it's dead device allocation with no caller in sight.

integrations/vllm_plugin/vllm_tt/model_runner_v2.py
```diff
@@ -0,0 +1,2520 @@
...
+        Decode always runs at max_num_reqs; prefill runs at the min / max prefill
+        buckets. Each request-count bucket is compiled at every token-length
+        padding, matching the shapes _select_batch can produce at runtime.
+        """
+        targets = sorted(
+            {self.max_num_reqs, self.min_num_reqs, self.max_prefill_num_reqs}
+        )
+        return [(t, q) for t in targets for q in self.num_tokens_paddings]
+
+    def capture_model(self) -> None:
```
kmabeeTT, line 2828: `capture_model`'s docstring says warmup exists so "the first real request pays no compile latency", but `_warmup_buckets`/`_precompile_bucket` never vary on `spec_decode_metadata`, they always trace a flat `logits_indices` rather than the packed `[reqs, spec_width]` shape spec decode uses, and I don't see a separate warmup path for it anywhere else. Does that mean enabling spec decode ends up compiling that packed shape live on the first draft step at each bucket?  In the past that was not allowed/expected...

kmabeeTT (changes requested):
Thanks Alina - impressively large PR ... I was not originally familiar with, so I took a lot of help from claude reviewing this.  Seems good overall, left few comments.

- Requesting changes: `_remove_request` doesn't clear `in_progress_prompt_logprobs`, causing stale logprobs data on abort+resubmit under the same `req_id` (see inline comment), one-line fix plus a test before merge. 

- Also have a question on whether warmup covers the spec-decode graph shape or if that ends up compiling live during serving, which we don't allow, please confirm (inline). 

- One lower-severity gap flagged inline (mm+spec-decode column offset) is fine to track rather than block on, plus a dead-code cleanup ask on `TTInputBuffers`. 

Lastly in addition to getting onPR green, please run vllm nightly against branch and subset of vllm performance tests with `test_filter production,gemma4` (same set as run [31107292080](https://github.com/tenstorrent/tt-xla/actions/runs/31107292080)) for a throughput baseline against v1 before this becomes the default runner.

Hit me up when changes land and I'll re-review.

integrations/vllm_plugin/vllm_tt/model_runner_v2.py
alinakhanTT: Fixed in 021d29263!

integrations/vllm_plugin/vllm_tt/model_runner_v2.py
alinakhanTT: Done in ea8bbcd2c, asserts in _gather_mm_embeddings when a row has an in-flight placeholder and a nonzero lead. It only fires on the actual collision, a fully computed placeholder on a drafting row still skips normally, and text-only requests on an mm model never reach the inner loop. Test covers both directions!

integrations/vllm_plugin/vllm_tt/input_batch_v2.py
alinakhanTT: Deleted!

integrations/vllm_plugin/vllm_tt/model_runner_v2.py
alinakhanTT: Thanks for pointing that out! The drafting pass traces a different graph than warmup covered: logits_indices is packed [reqs, spec_width] rather than flat, the attention carries a shared prefix offset (chunk_start_idx, None off the spec path), and sampling is skipped because spec rejection-samples on the host. All three fall out of the same pass, so 9a0b23e42 handles them together.
_warmup_buckets now returns (reqs, query_len, spec) and compiles both variants at every multi-token bucket when num_spec_tokens > 0. The flat shape is still warmed too, a spec run takes it whenever no request carries drafts. query_len == 1 has no spec variant, since a drafting row is always multi-token. Warmup roughly doubles with spec decode on, unchanged otherwise

kmabeeTT (approved):
Thanks for comments and changes! To the future!

nsumrakTT (approved):

alinakhanTT:
Nightly re-run on 971924dde: 31638170403. 7 failures in 4 jobs left to debug before merge 

llmbox
- Llama-3.1-70B : gets far enough to die on device DRAM, failing a 2.1 GB allocation
- DeepSeek-V2-Lite : Can't get a single buffer from host storage distributed over mesh shape MeshShape([2,4]) in DistributeTensorOp during warmup. Config has flat_model_io: True, which may beinvolved.

galaxy
- Mistral-Large : host memory is flat at 566 GB across both runs, and the log is full of TT_FATAL: Read unexpected run_mailbox value from core 25-16 (×16). Might be a device/dispatch wedge rather than a memory problem

bhqb
- gemma4_31b : unknown mesh- @mesh on a reshape in _forward_to_logits_compiled. Same family as the prompt-logprobs bug I fixed: a sharding constraint lands in a graph with no operand that declares the mesh, and pure TP on a (1,4) mesh makes every batch-axis weight mark a no-op
- gemma4_26b_a4b : timed out at 896s, and it's not the budget, burns the whole time in the first warmup bucket compile, where the test used to finish end-to-end in 447s.

n150
- llama3_3b_generation_trace / _trace_opt0 : device DRAM OOM, 96 MiB allocation, in _sample_compiled during warmup with the card 99.5% full. Repros locally

PR #5731, opened 2026-07-21
[vLLM] Tensor Parallel: size KV cache via gpu_mem_utilization across all chips instead of ~one chip's worth

### LINKED ISSUE #5728: [vLLM] gemma-4-31B-it TP: KV cache capped at ~1 chip's memory (gpu_memory_utilization under-fills each chip)
### Summary
Under tensor parallelism, `gpu_memory_utilization` (GMU) is applied per chip, but vLLM sizes the KV cache using the full, **un-sharded** `num_kv_heads`. Since the cache is actually sharded `tp_size`-ways across chips, each chip ends up filled to only `GMU / tp_size` of its DRAM — a quarter of the requested GMU on a 1×4 mesh. Net effect: **the total KV cache is capped at ~one chip's worth of memory** no matter how many TP chips there are, so usable context / concurrency are ~`tp_size`× lower than the hardware allows.

The cache **is** correctly sharded on device (`mark_sharding`, confirmed in the compiled IR) — there is no duplication, and vLLM's reported token count matches what's physically on the device. It's *under-utilizing* the mesh, not miscounting.

*(This also originally surfaced a separate bug: gemma4-31b-it-tp was accidentally running on a 2×2 mesh instead of the intended 1×4 — regressed by #5224, fixed in #5758.)*

### Repro (baseline: QB2 4-chip TP, bf16, GMU=0.2, `max_model_len=128`)
```bash
pytest -svv "tests/benchmark/test_vllm_benchmarks.py::test_vllm_tp_benchmark[gemma4-31b-it-tp]"
```
```
GPU KV cache size: 7,584 tokens
```
7,584 tokens ≈ the full-16-head per-token cost fit into one chip's `0.2 × 32 GB` budget. Because each chip only stores its 4-head shard, the same memory could hold ~4× → **~30,368 tokens** if sized for the sharded layout.

### Root cause
`KVCacheSpec.num_kv_heads` (used by vLLM core's `max_memory_usage_bytes()` / `page_size_bytes`) is the full, un-sharded head count, and `determine_available_memory()` budgets per chip. So vLLM assumes the whole un-sharded per-token cost must fit on each chip and divides nothing by `tp_size` — even though `mark_sharding()` splits the cache `tp_size`-ways afterward.

### Fix
#5731 — scale the available-memory budget by `tp_size` in `determine_available_memory()` so GMU fills each chip as intended, and the KV cache scales across all TP chips (`tp_size`× larger). Verified on gemma4-31b-it-tp (1×4): `GPU KV cache size` 7,584 → 30,368 tokens (4× = `tp_size`).

### Caveat / related
- Fixing the accounting makes the KV cache larger, and today the host **materializes the entire un-sharded cache** before sharding it to devices. So at large context / GMU / `tp_size` the host build exhausts host RAM during compile — a separate problem tracked in **#5764**.
- #5727 (hybrid / sliding-window KV-cache manager) separately caps very-long-context KV.

kmabeeTT:
### Ticket
Closes #5728

### Problem

Under tensor parallelism the KV cache is correctly sharded across chips (`tp_size`-ways via `mark_sharding`, confirmed in the compiled IR), but vLLM's block-budget math (`max_memory_usage_bytes` / `page_size_bytes`) sizes blocks using the full, **un-sharded** `num_kv_heads`. In effect it assumes the entire un-sharded cache must fit on a **single chip**, so `gpu_memory_utilization` only fills `GMU / tp_size` of each chip (e.g. ~22.5% at GMU=0.9 on a 1×4 mesh).

Net effect: **the KV cache was capped at roughly one chip's worth of memory** — the aggregate DRAM of the TP group went unused, so usable context / concurrency were ~`tp_size`× lower than the hardware allows.

### What's changed

- `determine_available_memory()` scales the KV budget by the shard factor, so `gpu_memory_utilization` fills each chip as intended (GMU per chip, not GMU/`tp_size`) and the cache spans the mesh's DRAM instead of one chip's. Only the budget figure changes — the tensor's shape and sharding were already correct.
- That scaling is gated on whether the cache is **actually sharded**, via a new `kv_cache_shard_factor(runner)` in `vllm_distributed_utils.py`, rather than on `enable_tensor_parallel`. The two differ only under DP+TP, where `initialize_kv_cache()` leaves the cache replicated (the TP-only spec puts `block_size` on the DP axis and fails `ttir.paged_update_cache`); scaling there would multiply each chip's allocation instead of reconciling it, and #5779's gemma4-31b DP+TP test would have asked for 1.2× DRAM per chip. The helper returns `1` for no-TP and DP+TP, leaving both unchanged.
- `initialize_kv_cache()`'s divisibility assert now uses the same helper. It previously read `tp_size` from `original_parallel_config.tensor_parallel_size` — always `1` under this plugin's single-process SPMD model, making it a no-op. The stale `TODO: Handle kv cache duplication under SPMD mode` comment is dropped; `mark_sharding` already shards correctly.
- The `KV cache sizing` log line reports the factor and per-chip budget (visible under `tests/integrations`, whose conftest sets `TTXLA_LOGGER_LEVEL=INFO`; `tests/benchmark/` stays at the `WARN` default, where vLLM's own `GPU KV cache size` is the signal).
- `mistral-small-3.1-24b-tp`'s benchmark GMU drops 0.65 → 0.16. 0.65 was tuned against the old accounting and would now put 8.4 GiB of KV on a 12 GiB wormhole galaxy chip; 0.16 = 0.65/4 keeps the allocation its 8192-context validation ran at. The other eight TP benchmarks keep their GMUs — they have enough per-chip headroom — and single-device, embedding and DP-only benchmarks are unaffected entirely.

### Result

- gemma4-31b-it-tp (QB2, 1×4 mesh): `GPU KV cache size` **7,584 → 30,368 tokens** (~4× = `tp_size`) — i.e. the cache now spans all 4 chips instead of ~one. Run passes.
- Shard factor resolves as expected across every parallel mode: `1` for DISABLED / DP-only / DP+TP, `4` for TP-1D `(1,4)` and TP-2D `(8,4)`, `8` for TP-2D `(4,8)`.

- This changes how the KV cache is **sized** (the memory budget), not how it's **allocated** — and it deliberately makes the cache **larger** (`tp_size`× more device memory, which is the point). The allocation path is unchanged: the full un-sharded cache is still built on **host** before being sharded to devices, and this PR enlarges that build. So at high GMU / long context / large `tp_size`, the host build can now push **host** RAM to OOM during compile — a separate problem tracked in **#5764**.

### Checklist

- [x] vLLM Nightly on CI: https://github.com/tenstorrent/tt-xla/actions/runs/30219509277
- [x] passing vLLM TP Perf benchmarks, QB2 on CI: https://github.com/tenstorrent/tt-xla/actions/runs/30219140002 (all 8 qb2-blackhole TP benchmarks, incl. gemma4-31b-it and llama-3.1-70b)
- [x] passing vLLM TP Perf benchmark, Galaxy on CI: https://github.com/tenstorrent/tt-xla/actions/runs/30219140895 (`vllm_mistral_small_3_1_24b_tp` at the new GMU of 0.16)
- [x] DP+TP regression check on bh_galaxy: https://github.com/tenstorrent/tt-xla/actions/runs/30218962902 (#5779's gemma4-31b DP+TP test, confirming the guard leaves it unchanged) - ran locally.

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/worker.py
+++ b/integrations/vllm_plugin/vllm_tt/worker.py
@@ -309,6 +309,19 @@ class TTWorker:
         usable_memory_size = int(
             total_memory_size * self.cache_config.gpu_memory_utilization
         )
+        # vLLM's KV-cache budget math (max_memory_usage_bytes) uses the full,
+        # un-sharded num_kv_heads with no TP awareness, even though the cache
+        # tensor is already correctly sharded tp_size-ways via mark_sharding
+        # (confirmed in the compiled IR). Counterbalance by inflating the
+        # available budget here instead of touching the cache tensor's shape
+        # or sharding.
+        if self.model_runner.enable_tensor_parallel:
+            mesh = self.model_runner.mesh
+            if hasattr(mesh, "shape"):
+                tp_size = mesh.shape()["model"]
+            else:
+                tp_size = dict(zip(mesh.axis_names, mesh.mesh_shape))["model"]
+            usable_memory_size *= tp_size
         tpu_kv_cache_bytes = max(usable_memory_size - profiled, 0)
         head_size = self.model_config.get_head_size()
         if head_size > 0:
--- a/integrations/vllm_plugin/vllm_tt/worker.py
+++ b/integrations/vllm_plugin/vllm_tt/worker.py
@@ -44,6 +44,7 @@ from .logger import tt_init_logger
 from .model_runner import TTModelRunner
 from .platform import TTConfig
 from .pooling_runner import TTPoolingModelRunner
+from .vllm_distributed_utils import kv_cache_shard_factor
 
 logger = tt_init_logger(__name__)
 
@@ -314,14 +315,10 @@ class TTWorker:
         # tensor is already correctly sharded tp_size-ways via mark_sharding
         # (confirmed in the compiled IR). Counterbalance by inflating the
         # available budget here instead of touching the cache tensor's shape
-        # or sharding.
-        if self.model_runner.enable_tensor_parallel:
-            mesh = self.model_runner.mesh
-            if hasattr(mesh, "shape"):
-                tp_size = mesh.shape()["model"]
-            else:
-                tp_size = dict(zip(mesh.axis_names, mesh.mesh_shape))["model"]
-            usable_memory_size *= tp_size
+        # or sharding. Returns 1 where the cache is not actually sharded
+        # (no TP, or DP+TP), leaving the budget untouched.
+        kv_shard_factor = kv_cache_shard_factor(self.model_runner)
+        usable_memory_size *= kv_shard_factor
         tpu_kv_cache_bytes = max(usable_memory_size - profiled, 0)
         head_size = self.model_config.get_head_size()
         if head_size > 0:
@@ -335,10 +332,12 @@ class TTWorker:
             tpu_kv_cache_bytes = tpu_kv_cache_bytes * head_size // padded_head_size
         logger.info(
             "KV cache sizing: device DRAM = %.2f GiB, gpu_memory_utilization = %.3f, "
-            "KV cache budget = %.2f GiB",
+            "kv_shard_factor = %d, KV cache budget = %.2f GiB (%.2f GiB per chip)",
             total_memory_size / 1024**3,
             self.cache_config.gpu_memory_utilization,
+            kv_shard_factor,
             tpu_kv_cache_bytes / 1024**3,
+            tpu_kv_cache_bytes / kv_shard_factor / 1024**3,
         )
         return int(tpu_kv_cache_bytes)
 
--- a/integrations/vllm_plugin/vllm_tt/worker.py
+++ b/integrations/vllm_plugin/vllm_tt/worker.py
@@ -248,8 +248,8 @@ class TTWorker:
 
                 # Use an empty tensor instead of `None`` to force Dynamo to pass
                 # it by reference, rather by specializing on the value ``None``.
-                tpu_kv_cache = torch.tensor([0], dtype=dtype).to(self.device)
-                kv_caches[layer_name] = tpu_kv_cache
+                kv_cache = torch.tensor([0], dtype=dtype).to(self.device)
+                kv_caches[layer_name] = kv_cache
             else:
                 raise NotImplementedError(
                     f"Unsupported KV cache spec '{type(layer_spec)}'"
@@ -306,7 +306,7 @@ class TTWorker:
         # use the heuristic of 2% of weights.
         profiled = current_mem * 1.02
 
-        # Calculate the TPU KV cache size based on profiling.
+        # Calculate the KV cache size based on profiling.
         usable_memory_size = int(
             total_memory_size * self.cache_config.gpu_memory_utilization
         )
@@ -319,7 +319,7 @@ class TTWorker:
         # (no TP, or DP+TP), leaving the budget untouched.
         kv_shard_factor = kv_cache_shard_factor(self.model_runner)
         usable_memory_size *= kv_shard_factor
-        tpu_kv_cache_bytes = max(usable_memory_size - profiled, 0)
+        kv_cache_bytes = max(usable_memory_size - profiled, 0)
         head_size = self.model_config.get_head_size()
         if head_size > 0:
             padded_head_size = (
@@ -329,17 +329,17 @@ class TTWorker:
                 logger.warning_once("head size is padded to %d", padded_head_size)
             # We adjust the usable memory size for the KV cache to prevent OOM
             # errors, even after padding the head_size.
-            tpu_kv_cache_bytes = tpu_kv_cache_bytes * head_size // padded_head_size
+            kv_cache_bytes = kv_cache_bytes * head_size // padded_head_size
         logger.info(
             "KV cache sizing: device DRAM = %.2f GiB, gpu_memory_utilization = %.3f, "
             "kv_shard_factor = %d, KV cache budget = %.2f GiB (%.2f GiB per chip)",
             total_memory_size / 1024**3,
             self.cache_config.gpu_memory_utilization,
             kv_shard_factor,
-            tpu_kv_cache_bytes / 1024**3,
-            tpu_kv_cache_bytes / kv_shard_factor / 1024**3,
+            kv_cache_bytes / 1024**3,
+            kv_cache_bytes / kv_shard_factor / 1024**3,
         )
-        return int(tpu_kv_cache_bytes)
+        return int(kv_cache_bytes)
 
     def sample_tokens(self, grammar_output: "GrammarOutput") -> ModelRunnerOutput:
         return self.model_runner.sample_tokens(grammar_output)
--- a/tests/benchmark/test_vllm_benchmarks.py
+++ b/tests/benchmark/test_vllm_benchmarks.py
@@ -166,12 +166,12 @@ def _mistral_small_31_tp_config(model: str, batch_size: int):
     # text-only (limit_mm_per_prompt zeroed so the vision tower never compiles),
     # mirroring _gemma4_tp_config. Runs on galaxy-wh-6u in a 8x4 mesh.
     #
-    # Validated max_model_len of 8192 at GMU of 0.65, but the current default
+    # Validated max_model_len of 8192 at GMU of 0.16, but the current default
     # of max_model_len is 128, so it needs to be overriden through the env. var.
     cfg = _config(
         model,
         batch_size,
-        gpu_memory_utilization=0.65,
+        gpu_memory_utilization=0.16,
         optimization_level=1,
         experimental_weight_dtype="bfp_bf8",
         experimental_kv_cache_dtype="bfp_bf8",
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -3411,12 +3411,18 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                 elif isinstance(kv_cache_spec, AttentionSpec):
                     if self.enable_tensor_parallel:
                         num_kv_heads = kv_cache_spec.num_kv_heads
-                        assert self.original_parallel_config is not None
-                        tp_size = self.original_parallel_config.tensor_parallel_size
-                        # TODO: Handle kv cache duplication under SPMD mode.
+                        if hasattr(self.mesh, "shape"):
+                            tp_size = self.mesh.shape()["model"]
+                        else:
+                            tp_size = dict(
+                                zip(self.mesh.axis_names, self.mesh.mesh_shape)
+                            )["model"]
+                        # mark_sharding() below does the real sharding; this
+                        # just fails loudly instead of it silently falling
+                        # back to replication (see safe_mark_sharding).
                         assert num_kv_heads % tp_size == 0, (
                             f"num_kv_heads {num_kv_heads} must be divisible by "
-                            f"tp_size {tp_size} under SPMD mode"
+                            f"tp_size {tp_size} for correct KV-head sharding"
                         )
                     kv_cache_shape = TTAttentionBackend.get_kv_cache_shape(
                         num_blocks,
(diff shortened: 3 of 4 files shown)


kmabeeTT:
As a bonus of switching back to mesh_shape=(1,4) for gemma4-31b-it on QB2 we get:

- decode jumps from 5.5 tok/sec to 7.0 tok/sec for me locally
- garbage output seen with mesh_shape=(2,2) is gone, now it looks sensible

FYI @ddilbazTT

tests/benchmark/test_vllm_benchmarks.py
```diff
@@ -134,15 +134,19 @@ def _gemma4_tp_config(model: str, batch_size: int):
...
+    #     without this explicit False, determine_mesh_shape() picks a 2D
+    #     (2, 2) mesh for 4 devices instead of the intended 1D (1, 4).
     cfg = _config(
         model,
         batch_size,
         gpu_memory_utilization=0.2,
         # opt-level 2 fails with an L1 out-of-memory TT_FATAL; see #5440.
         optimization_level=1,
         enable_tensor_parallel=True,
+        use_2d_mesh=False,
```
mmanzoorTT, line 149: You can also use `mesh_shape=[1, 4]` to use specific mesh shape instead of relying on pre-defined 2D mesh shapes or 1D mesh shape.

mmanzoorTT (approved):

tests/benchmark/test_vllm_benchmarks.py
kmabeeTT: This may get re-used for llmbox in future w/ 8 chips 1d mesh (it was desired/requested, started to be worked on, but didn't complete) so will leave use_2d_mesh=False for now.

integrations/vllm_plugin/vllm_tt/worker.py
```diff
@@ -309,6 +309,19 @@ def determine_available_memory(self) -> int:
...
+        # (confirmed in the compiled IR). Counterbalance by inflating the
+        # available budget here instead of touching the cache tensor's shape
+        # or sharding.
+        if self.model_runner.enable_tensor_parallel:
+            mesh = self.model_runner.mesh
+            if hasattr(mesh, "shape"):
+                tp_size = mesh.shape()["model"]
+            else:
+                tp_size = dict(zip(mesh.axis_names, mesh.mesh_shape))["model"]
+            usable_memory_size *= tp_size
```
jameszianxuTT: Is this allowing vLLM to allocate larger host buffers?

kmabeeTT: Possibly yes...  I have been chasing/observing host RAM doubling from ~120GB to ~230GB in this model, and it is looking like it is coming from this line, so I am going to hold off from merging this until better understood - darn.

kmabeeTT: Gotta step out now will circle back to this and read it closer, but claude says it makes sense, so an alternative solution is needed.

--

**Yes — `usable_memory_size *= tp_size` does cause larger host buffers, and it's the direct cause of the host-RAM OOM.**

  **Mechanism (traced end-to-end):**
  1. `determine_available_memory()` returns the KV budget inflated by `tp_size` (worker.py:324).
  2. vLLM divides that budget by per-block bytes → **`num_blocks` grows ×`tp_size`** (×4 on the 1×4 mesh).
  3. In `initialize_kv_cache()` each layer's K and V are allocated as `torch.zeros((num_blocks, num_kv_heads, block_size, head_size)).to(self.device)` (model_runner.py:3416-3429) — the shape uses the **full, un-sharded `num_kv_heads`** and the **inflated
  `num_blocks`**, and `torch.zeros(...)` **materializes on host CPU first**, then transfers.
  4. `mark_sharding(..., (None, "model", None, None))` shards dim-1 (`num_kv_heads`) across the model axis, but only **after** allocation (model_runner.py:3450-3461). So sharding shrinks the *device* footprint (~39 GiB/chip, correct) but never the host
  `torch.zeros`.
  5. Net: the host builds the entire un-sharded, ×`tp_size`-inflated KV cache → host RSS scales ×`tp_size`.

  **Empirical proof (one-variable isolation, gemma4-31b-it-tp, mml=4096, GMU=0.9, bfp8, 1×4 mesh — only difference is the `*= tp_size` line):**

  | `*= tp_size` | GPU KV cache | Host peak RSS | Result |
  |---|---|---|---|
  | ON | 273,440 tok | 228.7 GiB | OOM (SIGKILL) |
  | OFF | 68,352 tok | 122.6 GiB | pass |

  Exactly ×4 on the KV pool, +106 GiB host RSS (~541 KB/token, matching the model's ~570 KB/token KV footprint). Reverting the 1×4→2×2 mesh gives the identical 68,352 tok / 122.9 GiB, so the mesh is irrelevant — the inflation is the sole driver.

  **Compounding detail:** the host staging tensor uses `self.kv_cache_dtype` (bf16, 2 B/elem), while the `uint8` spec dtype is only vLLM's 1-byte block-accounting stand-in — so the host copy is ~2× the on-device bfp8 footprint even before the `tp_size`
  inflation.

  **Suggested fix direction:** the inflation corrects vLLM's *accounting* but leaks into *allocation* because both flow from `num_blocks` and the host tensor is sharded only post-hoc. Cleaner to make the KV spec TP-aware (`num_kv_heads // tp_size`) so the
  per-chip budget is right **and** the host allocation is `1/tp_size`, instead of inflating the budget — or allocate the KV cache already-sharded so the host never materializes the full copy.

mmanzoorTT: We need support for Memory stats to make KV cache allocation more reliable. https://github.com/tenstorrent/tt-xla/issues/1414

jameszianxuTT: If this LLM answer is true, it would help to understand the lifetime of the host kv cache and whether it can be freed or if something in vllm is holding a handle to it

kmabeeTT: Thx guys, looked more into this and tried some alternative changes that are more involved, so likely going to close this PR, merge the mesh_shape(1,4) fix for test seperately, and update the original issue with new scope of problems and hand it off.

@mmanzoorTT — agreed we want #1414, but I think that's a separate axis: it makes *device* KV sizing reliable, whereas this OOM is *host* RAM during compile. 

@jameszianxuTT — it's not a freeable buffer / dangling handle. The per-layer host staging from `torch.zeros(...).to(device)` is already released (KV init completes fine — if it weren't, we'd OOM at ~313GB during init, not later). The OOM actually hits at the first prefill `tt` compile: the XLA/SPMD lowering rebuilds the full *un-sharded* logical KV tensor on host to partition it. vLLM does keep a persistent handle (`self.kv_caches` / `bind_kv_cache`), but that points at the sharded *device* tensor (~39GB/chip), not a host copy — so there's nothing to free. Confirmed by: allocating on-device instead of `.to(device)` didn't help, but allocating only the per-chip shard dropped host RSS from ~230GB → ~97GB.

So the only real fix is to never hand the compiler a full-logical tensor — allocate the KV cache already-sharded (per-chip). I have a prototype doing this (host stays flat), currently blocked on a tt-mlir mesh-resolution issue for the shard→full path; will write that up separately. This matters more as we scale TP (host cost currently grows ~linearly with tp_size, so galaxy/TP-8+ makes it much worse).

(Numbers: 230/97 GB are measured host peak RSS at mml=4096, GMU=0.9, bfp8, 1×4 — inflation vs. per-shard. 313 GB is the computed full un-sharded KV at bf16 for the 273K-token pool; it exceeds the 249 GB host, so it OOM-kills at ~230 en route.)

kmabeeTT: After looking more, this increase in memory consumption makes sense, going to mark this conversation resolved, re-open this PR. I've updated the title, PR Description and rebased, and will drop a comment in body - the fix is correct in the end.

kmabeeTT:
Update - Re-opening this PR after rebasing, update title and PR Description.

- Yesterday thought this wasn't the right fix, but now understanding is improved and I see it is, apologies for back-and-forth.
- My mistake was I thought GMU=0.2 x 32GB = 6.4GB KV Cache per device sharded already but VLLM was reporting entire KV Cache as 6.4GB, but now I see that was not true - both device KV Cache and VLLM reporting were in sync and correct, but since GMU maxes out at 1.0 we were limited to 1-chips-worth (32GB) of KV Cache.  
- This fix solves the issue, allows GMU to be used as overall TP-sized sharded KV Cache to get up to 4 chips x 32GB = 128GB in case of QB2.  Reporting shows the increase, and any increase in host memory is expected and make sense (unfortunately) currently.   https://github.com/tenstorrent/tt-xla/issues/5764 will improve (reduce) host memory usage.

kmabeeTT:
Update - Rebased on latest main to get gemma4-31b-it galaxy DP+TP test which exposed issue (fixed), tested against existing TP vllm benchmark tests (links in PR desc, passing) to make sure increased KV Cache size is safe, and vLLM nightly again (passing), and updated one test `mistral-small-3.1-24b-tp` to have equivalent/smaller GMU to go with this change.  More details:

- **A gotcha from #5779**, which merged after this PR opened and added a gemma4-31b DP+TP test on bh_galaxy at GMU 0.3. DP+TP is the one mode where this PR's premise doesn't hold: `initialize_kv_cache()` leaves the cache un-annotated there (the TP-only spec puts `block_size` on the DP axis and fails `ttir.paged_update_cache`), so SPMD replicates it per chip instead of sharding it. My guard keyed off `enable_tensor_parallel`, true in that mode, so it would have scaled the budget by the model axis and asked each chip for 1.2× its DRAM. Kicked off test w/ fix on branch, link in PR Desc.

  - **The fix** (`4aa4336a6`): scale on whether the cache is *actually* sharded, not on whether TP is enabled. New `kv_cache_shard_factor(runner)` returns the mesh's `"model"` axis size where the cache is really sharded and `1` otherwise, leaving DP-only and DP+TP unchanged. The same helper now backs the divisibility assert, which was checking `num_kv_heads % mesh["model"]` under DP+TP where nothing is sharded, and the DP+TP branch carries a note to update it if that path ever starts sharding. CI confirms: factor 1 on the DP-only shards, and gemma4-31b-it-tp on QB2 still at 30,368 tokens (4× the pre-fix 7,584).

- Test`mistral-small-3.1-24b-tp` GMU 0.65 → 0.16. 0.65 was set against the old accounting and would now put 8.4 GiB of KV on a 12 GiB chip; 0.16 = 0.65/4 keeps the allocation it was validated at. The other eight TP benchmarks are unchanged, safe and CI validated.

@mmanzoorTT —  can you please take another look since these changes are new since you last viewed?

integrations/vllm_plugin/vllm_tt/worker.py
```diff
@@ -322,10 +332,12 @@ def determine_available_memory(self) -> int:
             tpu_kv_cache_bytes = tpu_kv_cache_bytes * head_size // padded_head_size
         logger.info(
             "KV cache sizing: device DRAM = %.2f GiB, gpu_memory_utilization = %.3f, "
-            "KV cache budget = %.2f GiB",
+            "kv_shard_factor = %d, KV cache budget = %.2f GiB (%.2f GiB per chip)",
             total_memory_size / 1024**3,
             self.cache_config.gpu_memory_utilization,
+            kv_shard_factor,
             tpu_kv_cache_bytes / 1024**3,
+            tpu_kv_cache_bytes / kv_shard_factor / 1024**3,
```
mmanzoorTT: You may consider replacing `tpu_kv_cache_bytes` with `kv_cache_bytes`.

mmanzoorTT (approved):

integrations/vllm_plugin/vllm_tt/worker.py
kmabeeTT: done

kmabeeTT:
Updates

- Pushed tiny PR feedback to remove tpu_ prefix from variable names (two of them)
- Opened issue https://github.com/tenstorrent/tt-xla/issues/5796 for adding sharding to DP+TP case and updated code to refer to it 
- bh-galaxy CI machine is offline (DevEx ticket opened in Slack) so @ssaliceTT ran gemma4-31b-it on bh-galaxy locally w/ this change, as expected no impact.

```
(EngineCore pid=15843) INFO:vllm_tt.worker:KV cache sizing: device DRAM = 31.88 GiB, gpu_memory_utilization = 0.300, kv_shard_factor = 1, KV cache budget = 9.56 GiB (9.56 GiB per chip)
(EngineCore pid=15843) WARNING 07-27 18:25:30 [kv_cache_utils.py:1339] Hybrid KV cache manager is disabled for this hybrid model, This means we do not enable any optimizations for saving KV cache memory (e.g., dropping the KV cache outside the sliding window). The compute of layers like sliding window is still saved.
(EngineCore pid=15843) INFO 07-27 18:25:30 [kv_cache_utils.py:1733] GPU KV cache size: 11,392 tokens
```

Going to merge tonight.

Issue #5727, opened 2026-07-20
[vLLM] gemma-4-31B-it: sliding-window layers pay full max_model_len KV cache cost


kmabeeTT:
### Problem

gemma-4-31B-it interleaves sliding-window and full-attention layers (60 layers: 50 sliding_attention @ window=1024, 10 full_attention). The TT vLLM plugin has no per-layer KV cache spec -- it sizes every layer as if it were full-attention, so the 50 sliding-window layers scale with `max_model_len` instead of being capped at their 1024-token window.

This is a plugin-wide gap, not specific to gemma-4: any model with sliding-window/hybrid attention layers served through the TT vLLM plugin hits the same over-allocation.

### Quick Repro

On QB2 (Blackhole Quietbox 2) needed to repro as-is (takes 8 mins):

```bash
pytest -svv "tests/benchmark/test_vllm_benchmarks.py::test_vllm_tp_benchmark[gemma4-31b-it-tp]" |& tee baseline.log
```

Shows:

```
(EngineCore pid=...) WARNING ... [kv_cache_utils.py:1339] Hybrid KV cache manager is disabled for this hybrid model, This means we do not enable any optimizations for saving KV cache memory (e.g., dropping the KV cache outside the sliding window). The compute of layers like sliding window is still saved.

(EngineCore pid=...) INFO ... [kv_cache_utils.py:1733] GPU KV cache size: 7,584 tokens
(EngineCore pid=...) INFO ... [kv_cache_utils.py:1734] Maximum concurrency for 128 tokens per request: 59.25x
```

Runs at the test's baseline config: bf16 weights+KV, `GPU_MEMORY_UTILIZATION=0.2`, `max_model_len=128`. No bfp8/chunked-prefill/high-context setup required.

#### Math: the baseline number already confirms the uniform (buggy) cost model

At bf16 KV (2 bytes/element), costing every layer uniformly as full-attention (today's bug):
- sliding: `2 * 16 * 256 * 2B = 16 KiB/token/layer` x 50 layers = 800 KiB/token
- full: `2 * 4 * 512 * 2B = 8 KiB/token/layer` x 10 layers = 80 KiB/token
- total = **880 KiB/token**

[...]

PR #5744, opened 2026-07-21
[vLLM][uplift] v0.25.1

mmanzoorTT:
### Ticket
closes #5743 

### Problem description
vLLM uplift

### What's changed
- vLLM uplift: target v0.25.1
- Transformers uplift (required for vLLM): target v5.14.1
- TorchAudio cpu version pinned in requirements to avoid vLLM installing non-cpu version
- Update Schedular to match upstream changes
- Update FusedMoE override implementation to match upstream changes.

### Checklist
- [X] New/Existing tests provide coverage for changes

PR #5786, opened 2026-07-24
[vLLM] Hybrid KV cache : per-layer sliding-window ring buffer

sshonTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/5727)

### Problem description
Sliding window requires only size of window length for cache, but current vLLM pays this for full attention which makes constraint on context length

### What's changed
Each sliding-window layer gets a per-user, window-sized ring buffer instead of paying the full max_model_len KV cost; window_blocks rounds up to a multiple of 8 for 32-byte page-table stick alignment. Also pin the gemma-4-31B TP benchmark to qb2's 1x4 mesh (mesh_shape=[1, 4]) -- the inherited use_2d_mesh=True 2x2 default deadlocks its RMS-norm all_gather at 128K.

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -74,6 +74,7 @@ from vllm.v1.kv_cache_interface import (
     KVCacheGroupSpec,
     KVCacheSpec,
     MLAAttentionSpec,
+    SlidingWindowMLASpec,
     SlidingWindowSpec,
     UniformTypeKVCacheSpecs,
 )
@@ -111,6 +112,13 @@ from .overrides import replace_modules
 from .platform import TTConfig
 from .rejection_sampler import RejectionSampler
 from .sampler import Sampler
+from .swa_cache_utils import (
+    assign_ring_slots,
+    build_sliding_ring_page_table,
+    sliding_page_table_width,
+    sliding_ring_phys_blocks,
+    sliding_window_blocks,
+)
 from .vllm_distributed_utils import (
     ParallelismMode,
     kv_cache_shard_factor,
@@ -395,7 +403,17 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                 <= scheduler_config.max_num_batched_tokens
             ), f"The max_num_batched_tokens {scheduler_config.max_num_batched_tokens} must be larger than or equal to max_model_len ({self.max_model_len}) * max_num_seqs ({scheduler_config.max_num_seqs})"
         self.most_model_len = envs.VLLM_TPU_MOST_MODEL_LEN
-        self.max_num_blocks_per_req = cdiv(self.max_model_len, self.block_size)
+        # Match vLLM's per-request block-table width, which rounds up to a
+        # multiple of (128 // block_size) for attention-backend alignment (vLLM
+        # #39324). Sizing the padded page-table buffer the same way keeps it >=
+        # that width (initialize_kv_cache asserts this).
+        _base_blocks_per_req = cdiv(self.max_model_len, self.block_size)
+        self.max_num_blocks_per_req = (
+            cdiv(_base_blocks_per_req, 128 // self.block_size)
+            * (128 // self.block_size)
+            if self.block_size <= 128
+            else _base_blocks_per_req
+        )
         self.num_blocks_per_most_len_req = (
             cdiv(self.most_model_len, self.block_size)
             if self.most_model_len is not None
@@ -564,6 +582,29 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
             TTAttentionBackend.get_max_num_seqs(self.max_model_len, self.block_size),
             self.max_num_reqs,
         )
+        # Hybrid (multi-group) KV cache state. Defaults describe a single
+        # full-attention group (identical to the pre-hybrid path); (re)set in
+        # initialize_kv_cache once the real kv_cache_groups are known.
+        # _layer_to_group maps each attention layer to its group index so
+        # _prepare_inputs can hand each layer its own group's page table.
+        self._num_kv_cache_groups = 1
+        self._layer_to_group: dict[str, int] = {}
+        # Per-user sliding-ring slots, keyed by req_id so a request keeps its
+        # physical sub-ring across InputBatch row-condensing. Freed on request
+        # exit and reused. See swa_cache_utils.assign_ring_slots.
+        self._req_ring_slot: dict[str, int] = {}
+        self._free_ring_slots: list[int] = list(range(self.max_num_reqs))
+        # Per-group block sizes. vLLM's page-size unification can assign a
+        # different block_size per group when head dims differ across attention
+        # types, so page tables are padded to the widest group's width and each
+        # group uses its own block_size for the prefix-roll offset.
+        self._group_block_sizes: list[int] = [self.block_size]
+        # Per-group sliding-window classification. Sliding groups get a small
+        # physical ring buffer (window_blocks) instead of the full pool, and
+        # _prepare_inputs feeds them a positional ring page_table rather than
+        # absolute block ids. Set in initialize_kv_cache.
+        self._group_is_sliding: list[bool] = [False]
+        self._group_window_blocks: list[int] = [0]
         self.query_start_loc_cpu = torch.zeros(
             self.max_num_tokens + 1,
             dtype=torch.int32,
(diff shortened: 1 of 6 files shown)


tests/benchmark/test_vllm_benchmarks.py
```diff
@@ -134,11 +134,15 @@ def _gemma4_tp_config(model: str, batch_size: int):
...
-    #     (2, 2) mesh for 4 devices instead of the intended 1D (1, 4).
+    #   - flat_model_io for Gemma-4's PLE forward
+    #   - mesh_shape=[1, 4]: this benchmark targets qb2-blackhole (4 devices),
+    #     which wants a 1D 1x4 mesh. On galaxy, gemma-4 TP wants a 2D mesh
+    #     (e.g. [8, 4] -- see _mistral_small_31_tp_config). use_2d_mesh=False
+    #     alone is not enough: mesh_shape=None then picks (1, num_devices),
+    #     but pinning the shape documents the qb2 target and matches the
+    #     galaxy config's convention. The TTConfig use_2d_mesh default is True,
+    #     which would fall to a 2x2 mesh whose degenerate 2-device-axis CCL
+    #     deadlocks the distributed RMS-norm all_gather at 128K (batch>=4 hangs).
```
mmanzoorTT: Why `use_2d_mesh=False` is not enough? this option will generate `mesh_shape=[1,4]`. 

How does this configuration `matches the galaxy config's convention`?

mmanzoorTT (commented):
Have you tried it out with v0.25.1 uplift? https://github.com/tenstorrent/tt-xla/pull/5744

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1372,57 +1492,128 @@ def _prepare_inputs(self, scheduler_output: "SchedulerOutput", start_index: int)
...
+            if actual_num_reqs > 0:
+                nsched = np.asarray(num_scheduled_tokens_per_req[:actual_num_reqs])
+                num_computed = np.asarray(
+                    self.input_batch.num_computed_tokens_cpu[:actual_num_reqs]
+                )
+                filling = nsched > 1
+                bad = filling & ((start_block > 0) | (num_computed > 0))
+                if np.any(bad):
+                    raise NotImplementedError(
+                        "sliding-window prefill for context > window or with "
```
mmanzoorTT, line 1566: This will not support `chunked-prefill` as well. right?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -3354,92 +3681,159 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
...
+        layer_to_spec: dict[str, KVCacheSpec] = {}
+        for kv_cache_group in groups:
+            gspec = kv_cache_group.kv_cache_spec
+            for layer_name in kv_cache_group.layer_names:
+                layer_to_spec[layer_name] = _get_layer_kv_cache_spec(gspec, layer_name)
+
+        def _kv_shape(spec, num_blocks):
+            backend = (
+                TTMLAAttentionBackend
+                if isinstance(spec, MLAAttentionSpec)
```
mmanzoorTT: This will misclassify `SlidingWindowMLASpec` as `TTAttentionBackend`. What do you think of adding explicit checking and assignment for each `Spec` type and assert for new/un-seen types to avoid any misclassification and silent failure.

tests/benchmark/test_vllm_benchmarks.py
sshonTT: Yeah this is unnecessary change. Removed this change from this pr

integrations/vllm_plugin/vllm_tt/model_runner.py
sshonTT: You're right. Added explicit comment

integrations/vllm_plugin/vllm_tt/model_runner.py
sshonTT: Good point. Added explicit logic.

sshonTT:
> Have you tried it out with v0.25.1 uplift? #5744

@mmanzoorTT, I've tried and run the test on my local machine, and it worked well.

integrations/vllm_plugin/vllm_tt/model_runner.py
mmanzoorTT: Thanks

mmanzoorTT (approved):

hshahTT:
There's a lot of SWA-specific logic in both model_runner.py and worker.py. Some of it is also duplicated (e.g. the window_block calculation in [worker.py](https://github.com/tenstorrent/tt-xla/pull/5786/changes#diff-88bce720a7696c4b1ce44718eca629ef98c650f6e0c8c00b4e3c88602ac54d5dR348-R351) and the same calculation in [model_runner.py](https://github.com/tenstorrent/tt-xla/pull/5786/changes#diff-9d59e239d048c9bf1316167b5fbeeb166e6b0bc088bdabca349a8801fe6e8ba3R4152-R4156)).

I would recommend creating a separate file called something like `swa_cache_utils.py` and storing this logic there. I have a [commit](https://github.com/tenstorrent/tt-xla/commit/43aed8bcffca852db23fef1d63696fe83cd9d7ca) showing what I mean.

Adding a test file for these cache utils would also be nice too.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1084,6 +1141,70 @@ def get_supported_tasks(self) -> tuple[SupportedTask, ...]:
...
+            self.num_reqs_max_model_len,
+        }
+
+        # Per-group page-table width: sliding groups only address their small
+        # window ring, so their page_table is that narrow (the paged ops require
+        # page_table_width <= cache_num_blocks). Full groups use the base width.
+        def _pt_width(g, base_width):
+            if g < len(self._group_is_sliding) and self._group_is_sliding[g]:
+                return min(self._group_window_blocks[g], base_width)
+            return base_width
```
hshahTT, line 1140: `base_width` here seems to always be `self.max_num_blocks_per_req`. So we can change this function to:
```suggestion
        def _pt_width(g):
            if g < len(self._group_is_sliding) and self._group_is_sliding[g]:
                assert self._group_window_blocks[g] <= self.max_num_blocks_per_req, "SWA window blocks is more than allowed"
                return self._group_window_blocks[g]
            return self.max_num_blocks_per_req
```

sshonTT:
@hshahTT I moved everything into vllm_tt/swa_cache_utils.py (window blocks, physical blocks, reserve bytes, page table width, slot assignment, ring rotation, etc.).
Now both worker.py and model_runner.py share this, so they won't get out of sync.
I also added unit tests for this (`tests/integrations/vllm_plugin/test_swa_cache_utils.py`).

One quick thing to note: I made one tweak from your initial sketch. We don't need to divide the reserve by the KV shard factor here. Please look at PR #5731

hshahTT (approved):

integrations/vllm_plugin/vllm_tt/model_runner.py
sshonTT: Good catch on the redundancy, but I ended up keeping the min(). A couple of reasons for this:

First, base_width isn't always max_num_blocks_per_req - the second call site actually passes num_blocks_per_most_len_req (for the VLLM_TPU_MOST_MODEL_LEN path).

Second, the assert would actually trigger on a real config. For example, max_num_blocks_per_req is calculated as round_up(cdiv(max_model_len, block_size), 128 // block_size). So with max_model_len=128 and block_size=32, it becomes 4, while Gemma-4's 1024-token window gives window_blocks=40. (This is what happens in test_tensor_parallel_generation_gemma4_31b in nightly). When max_model_len < sliding_window, the page table has to be capped by the buffer width, which is what min() does here — otherwise we'll hit that "block-table width > buffer width" assert again.

So I kept it as a pure helper in swa_cache_utils.sliding_page_table_width(), and added test_page_table_width_is_capped_by_the_buffer to cover both cases.

kmabeeTT:
Thanks for change! A small followup, sorry not caught in review, but I went to double check we could close parent issue and realized this PR broke the baseline test from the issue run in vllm benchmark suite because a clamp on max_model_len was missing in `sliding_window_blocks()` leading to this error with default KV cache size now not being big enough:

```
  WARNING vllm_tt.worker: Sliding ring reservation (31.27 GiB) >= KV budget (25.50 GiB); no room for full-attention pool.
  ValueError: No available memory for the cache blocks.
```

I opened quick fix PR https://github.com/tenstorrent/tt-xla/pull/5834 to solve it (test is back passing now).

Issue #5034, opened 2026-06-01
Forge LLM decode throughput regresses 16–43% from tt-forge 1.1.0 → 1.2.0 (P150)


kmabeeTT:
TLDR; Perf regression needs bisect or debug, benchmarked in tt-inference-server locally when qualifying tt-forge 1.2.0 wheel. Here is bunch of claude generated data.  Should be unrelated to https://github.com/tenstorrent/tt-xla/issues/4902 because the offending commit there wasn't included in tt-forge 1.2.0 release.

### Summary

Upgrading the forge inference stack from **tt-forge 1.1.0 → 1.2.0** drops steady-state decode throughput **16–43%** (tok/s) across all 4 forge LLMs on P150, with TPOT (per-output-token time) up **20–85%**. Prefill latency (TTFT) is roughly unchanged for the wheel bump. The serving config (model, 16K context, batch=4, KV pool, `optimization_level=0`, `cpu_sampling=true`) is held identical between the two runs, so this isolates to the **1.2.0 wheel itself** (tt-xla / tt-mlir / tt-metal + its mandated torch bump), not to our serving config or the `optimization_level=0` workaround.

This is a decode-path regression — eval accuracy is unaffected (verified, see below).

### Environment

- Hardware: Tenstorrent **P150** (Blackhole, 32 GB), single chip per model.
- Stack: `tt-inference-server` forge image (forge-vllm-plugin), vLLM 0.19.1.
- Benchmark: `vllm bench serve --dataset-name random --random-input-len 128 --random-output-len 128 --max-concurrency 1 --num-prompts 8` (single-stream smoke). A 4-stream concurrent test corroborated the same direction/magnitude.

### What was used in each run

The two changes (wheel bump, and the `optimization_level` workaround) were measured separately via a 3-way comparison. **A′ vs B is the clean wheel isolation** (same `optimization_level=0`, only the wheel differs).

| Run | tt-forge wheel | tt-xla | tt-mlir | tt-metal | torch | optimization_level | cpu_sampling |
|-----|----------------|--------|---------|----------|-------|--------------------|--------------|
| **A**  (1.1.0 baseline) | 1.1.0 (released) | `470f0fad8` | `f3ddbfb6b` | `90c914ef2` | 2.7.1+cpu | 1 | true |
[...]

Issue #4891, opened 2026-05-24
vLLM b32 prefill batches ~10× less efficiently than decode on Blackhole p150


kmabeeTT:
## TL;DR

Prefill TTFT at b32 on a single Blackhole p150 is ~17–19× the b1 TTFT, even though b32 prefill does 32× the work. That sounds like ~1.8× per-FLOP batching benefit. Decode on the same hardware and same models gets ~25× per-FLOP batching benefit (b32 system tok/s ≈ 25× b1). **Prefill batches about an order of magnitude less efficiently than decode does, on the same HW.** Filing to track investigation into why. Independent of #4880; reproducible on `main` today.

## Data

Single Blackhole p150, single device, `TT_BENCHMARK_MAX_MODEL_LEN=128`, warmed up (the benchmark runs one warmup iteration before measurement; numbers below are post-warmup):

| model | batch | TTFT (warm)¹ | b1 → b32 ratio | per-FLOP batching gain² |
|---|---:|---:|---:|---:|
| Llama-3.2-3B  (28 layers) | 1  | 52.5 ms | — | — |
| Llama-3.2-3B  (28 layers) | 32 | 910 ms | 17.3× | **1.85×** |
| Llama-3.1-8B-Instruct  (32 layers) | 1  | 96.8 ms | — | — |
| Llama-3.1-8B-Instruct  (32 layers) | 32 | 1810 ms | 18.7× | **1.71×** |

¹ Source: the existing pytest's per-request `Request 0: TTFT=…ms` log line (see Reproduction below). At b32 the benchmark's "mean" TTFT also includes ~31 trailing users that pay an extra ~900 ms of vLLM scheduling/delivery on top; that's a separate artifact, not device compute. Request 0 is the compute number.

² `32 ÷ ratio`. 1× would mean batching does nothing for per-FLOP cost; 32× would mean perfect batching (b32 in the same wall-clock as b1).

### Compare against decode

On the same models / hardware, with a vLLM server interactively (using the `examples/vllm/Llama-3.2-3B` reference test in this branch, see below):

| | 3B b1 | 3B b32 (32 parallel users) |
|---|---:|---:|
| decode tok/s/user | 28.6 | 22.7 |
| system tok/s | 28.6 | 319 |

[...]

Issue #4301, opened 2026-04-20
Increase Batch32 Perf for llama3.1-8B through VLLM (p150 demo)


kmabeeTT:
Can Batch32 perf be improved?

**Problem:**

From some experiments last week, was seeing:

- llama3.1-8B t/s/u Batch32 was ~80% of Batch1 for non-VLLM (n150 benchmark test in CI), not bad.
- but for VLLM I'm seeing bigger perf drop for Batch32, where it's t/s/u was ~60% of Batch 1.

```
// generated from tt-xla kmabee/vllm_perf_apr16 based on main 4/11 commit 866d0abdd

Batch 32 tok/sec/user is ~60% of batch 1 (true for single layer and full model).

llama_greedy_batch1_debug_1layer.log: | Sample per second: 135.97540900263093
llama_greedy_batch32_debug_1layer.log:| Sample per second: 85.24502669556571

Full Model:
llama_greedy_batch1_debug.log:| Sample per second: 19.016332758780692
llama_greedy_batch32_debug.log:| Sample per second: 10.890453293849982
```

Needs debug... ideally perf would be closed to non-VLLM.   For tt-cloud-console demos have been sticking with lower batch count (1 or 2) until batch-32 perf can be increased, but we want to move to batch-32 asap.

**Details:**

From quick look at single-layer graphs + tracy (logs, IR, screenshot attached) I see rope fusion not working for batch-32 (though it seems like noise), greedy argmax sampler is ~0.9ms (up from near ~0ms), but still something unknown around ~0.9 ms slower in batch32 from quick sum of ops runtime in screenshot (7.5 ms per iter (b32) vs 5.7 ms per iter (batch1)

Tracy csv for each of the single layer runs (should be 10 tokens generated):

- [llama3.1_8b_b1_ops_perf.csv](https://gist.github.com/kmabeeTT/2801ea8a5e618af06655032c5738ca99#file-llama3-1_8b_b1_ops_perf-csv)
- [llama3.1_8b_b32_ops_perf.csv](https://gist.github.com/kmabeeTT/2801ea8a5e618af06655032c5738ca99#file-llama3-1_8b_b32_ops_perf-csv)

screenshot showing me compare batch1 vs batch32 tracy ops 
<img width="3620" height="1678" alt="Image" src="https://github.com/user-attachments/assets/6ccbe58a-4978-4528-8cf6-4df325c94bfa" />

**SHLO & TTIR IR:**

[...]

PR #5787, opened 2026-07-24
[vLLM][MRv2] Telemetry + visualizer (#5715)

### LINKED ISSUE #5715: vLLM telemetry
We often see poor performance when serving models through vLLM due to kv trashing and other scheduler related issues. These have been historically very difficult to debug, and some telemetry would be really helpful. Please add a way to publish:

- which slots (ie 32 slots for batch32) are used, what state (prefilling, decoding)
- how a new request stalls in-flight decodes
- actual decode rate
- scheduler level metrics (preemption, eviction, queue depth, when requests get re-admitted under chunked prefill)
- efficiency metrics (KV cache / batch utilization ie how much we're actually using), stall time somehow to figure out if scheduler making good decisions, number of prefill interruptions, percent of time doing decode, etc 

These should be either global or per-request depending on what makes sense. Publish doesn't need to be overly complicated, it can mean write out to json, for instance. 

As a follow-up, we'll need some way to consume / visualize these.

avorozcovsTT:
Part of #5715, for Model Runner v2. The v1 half landed on `main` as `b1f76f33d` (#5808), and brought the shared collector, the visualizer, the examples and the scheduler-side hook with it. This PR adds the runner-side collector to `TTModelRunnerV2`.

> **Base:** this PR targets `akhan/vllm-v2-features`, since the hooks live in `TTModelRunnerV2`. Retarget to `main` once MRv2 merges.

## What this PR changes

Five files. Two are the work:

| File | Change |
|---|---|
| `vllm_tt/model_runner_v2.py` | +89 — the collector, admit/complete hooks, the pass classifier, and the pass-loop tally |
| `tests/integrations/vllm_plugin/mrv2/test_runner_telemetry.py` | +337 — 14 test functions, 17 collected, marked `push` + `cpu` |

Three are already-merged shared files that carry a v2 delta:

- `vllm_tt/telemetry.py` — the `on_step` docstring attributed the multi-token accept path to the v1 runner; both runners run ngram spec decode now.
- `examples/vllm/telemetry/README.md` — names both runners, and its slot-output section covers v2's persistent slot table alongside v1's condensing batch.
- `examples/vllm/telemetry/run_telemetry_demo.py` — its comment named v1 as the default runner.

## How the v2 hooks work

- **They live in `_run_pass_loop`**, which this base introduced when it split `sample_tokens`. The loop returns `(prefill_passes, decode_passes, emitted_tokens)` and `sample_tokens` emits from that tally.
- **Pass classification survives speculative decode.** A pass counts as prefill when any scheduled row still has computed tokens below its prefill boundary, read before `postprocess`. Two simpler tests fail: row state *after* the pass calls a completed single-shot prefill a decode pass, and tokens-per-row calls a spec-decode verification pass a prefill pass, since a decoding row is scheduled `1 + num_speculative_tokens`.
- **Accepted-tokens/step is live here.** This base ports ngram speculative decode to v2 (`946d30a65`): `postprocess` commits every accepted draft, so a row can contribute several tokens in one step and the hook sums per-row lengths.
- **`readmission` fires rarely on v2.** The slot table holds a request across steps, so it triggers only when an id arrives while a stale slot still holds it (abort + resubmit). v1 re-admits routinely. A v2 request resumed from preemption arrives as a fresh request and shows as nonzero `num_cached_tokens` instead.
- **Off by default and zero-cost when off**: one cached bool on the hot path, plus a class-level disabled collector so instances built without `__init__` inherit it.

## What it builds on (already merged)

`main` carries the rest, so none of it is in this diff:

| Piece | Where |
|---|---|
| `SchedulerTelemetry` + the `AscendScheduler` hook | `vllm_tt/scheduler/ascend_scheduler.py` |
| `RunnerTelemetry` / `SchedulerTelemetry` collectors, sink and flush policy | `vllm_tt/telemetry.py` |
| Config resolution and env gating | `vllm_tt/platform.py` |
| Visualizer (`report` / `live`) | `scripts/telemetry/telemetry_viz.py` |
| README, three demos, HTTP load generator | `examples/vllm/telemetry/` |

The two collectors run in separate processes and write separate JSON-lines sinks (`scheduler.jsonl`, `runner.jsonl`, `runner_snapshot.json`), correlated offline on `request_id`. The scheduler answers *intent* — prefill vs decode, preemption, queue depth, KV/batch utilization, watermark rejects, b1-cap hits. This PR supplies the *reality* half on v2: slot occupancy, executed pass split, actual decode rate, per-request lifecycle.

## Testing

`-m "push and cpu"` over `tests/integrations/vllm_plugin` collects the new file; the mrv2 suite covers gating, env precedence, buffered flush, record schemas, disabled ⇒ zero files, and the cached-token and re-admission signals.

Measured on n150 (wh-17) under vLLM 0.26.0, Qwen3-0.6B, one run per scenario.

| Scenario | steps | peak slots | prefill / decode passes | tokens | tok/s median | notes |
|---|---|---|---|---|---|---|
| `run_telemetry_demo.py` | 33 | 4 / 4 | 2 / 31 | 128 | 40.0 | 2 prefill-stalled steps |
| `run_telemetry_oversubscribed.py` | 97 | 4 / 4 | 3 / 94 | 384 | 39.7 | max queue depth 4 |
| `run_telemetry_memory_pressure.py` | 200 | 16 / 16 | 12 / 188 | 768 | 27.1 | KV pinned at 95%, 1621 watermark rejects, 0 preemptions |
| `api_load_test.py` (live server) | 131 | 4 / 4 | 7 / 124 | 512 | 39.3 | 7 prefill-stalled steps |

The pass split agrees with the v1 runner to within one pass on every scenario (2/31, 3/94, 12/188, 7–8/124), from independently written hooks over different state models.

Running the same four scenarios with `use_v2_model_runner: False` also shows what the telemetry is for: under 0.26.0 the memory-pressure case costs v2 ~2.7x the decode rate it reached under 0.25.1 (73.3 → 27.1 tok/s), because v2 now fills all 16 slots and chunks prefill to a 95% KV peak instead of single-shot prefilling 8 rows at 75%. Deterministic across two runs; v1 is unchanged in that scenario. Under `vllm serve` the gap runs the other way — v1's median step is 443.2 ms against v2's 101.7 ms on identical work.

## What the output looks like

The merged visualizer rendering v2 telemetry from the three demos. These captures predate the pass-classification fix, so their prefill/decode split differs from the table above; the occupancy, decode-rate and utilization panels are unaffected.

Simple:

image

Oversubcribed:

image

Memory pressure:

image

## Notes

- Draft until MRv2 stabilises, per the review note on this PR.
- Carries a second commit that fixes the base's `MultiGroupBlockTable` call for vLLM 0.26.0. `TTModelRunnerV2.__init__` raises `TypeError` without it, so no engine starts on the default runner and 14 mrv2 init tests fail. Drop that commit once the base fixes the call.
- Pooling-runner telemetry is scoped separately in #5812 and deferred.
- Primary consumers to validate against: #4891 / #4301 (batch-32 prefill efficiency), #5034 (decode regression).
- Telemetry output defaults under `tt_telemetry/`, git-ignored.

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner_v2.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner_v2.py
@@ -151,6 +151,20 @@ class TTModelRunnerV2(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         self.lora_config = vllm_config.lora_config
         self.device = device
 
+        # Speculative decode (ngram). Mirrors v1: num_spec_tokens gates the whole
+        # path, drafter is only built for the ngram method.
+        self.speculative_config = getattr(vllm_config, "speculative_config", None)
+        self.num_spec_tokens = 0
+        self.drafter = None
+        self._draft_token_ids: list[list[int]] | None = None
+        self._draft_token_req_ids: list[str] | None = None
+        if self.speculative_config is not None:
+            self.num_spec_tokens = self.speculative_config.num_speculative_tokens
+            if self.speculative_config.method == "ngram":
+                from vllm.v1.spec_decode.ngram_proposer import NgramProposer
+
+                self.drafter = NgramProposer(vllm_config)
+
         # additional_config arrives as a plain dict; build the typed TTConfig
         # (as the v1 runner does) so the field reads below see real values.
         self.tt_config = TTConfig(**vllm_config.additional_config)
@@ -284,7 +298,7 @@ class TTModelRunnerV2(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
             max_num_reqs=self.max_num_reqs,
             max_model_len=self.max_model_len,
             max_num_batched_tokens=self.max_num_tokens,
-            num_speculative_steps=0,
+            num_speculative_steps=self.num_spec_tokens,
             vocab_size=self.vocab_size,
             device=self.device,
         )
@@ -432,6 +446,7 @@ class TTModelRunnerV2(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
 
         from .model_state import TTModelState
         from .overrides import repair_stale_moe_closures, replace_modules
+        from .rejection_sampler import RejectionSampler
         from .sampler import Sampler
 
         logger.info("Loading model %s ...", self.model_config.model)
@@ -500,6 +515,7 @@ class TTModelRunnerV2(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         self.model.compile(backend="tt", dynamic=False)
         logger.info("Model loaded and registered for tt compilation.")
         self.sampler = Sampler()
+        self.rejection_sampler = RejectionSampler(self.sampler)
 
         encoder_cache = self.encoder_cache if self.supports_mm_inputs else None
         self.model_state = TTModelState(
@@ -745,14 +761,20 @@ class TTModelRunnerV2(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         num_scheduled_tokens: np.ndarray,
         target_num_reqs: int,
         padded_query_len: int,
+        spec_decode_metadata=None,
     ) -> tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
         """Host substitute for the v2 Triton input-prep kernels.
 
         Builds 2D input_ids/positions [target_num_reqs, padded_query_len] plus flat
-        query_start_loc/seq_lens/logits_indices from TTRequestState, indexed by
-        idx_mapping_np[b] -> stable slot. Prefill and decode share one gather
+        query_start_loc/seq_lens from TTRequestState, indexed by idx_mapping_np[b]
+        -> stable slot. Prefill and decode share one gather
         (all_token_ids[computed:computed+n]); the sampled token was already
-        appended by the writeback and there is no spec decode. Padding stays zero.
+        appended by the writeback, and staged drafts sit right after it. Padding
+        stays zero.
+
+        logits_indices is flat [target_num_reqs] (one logit at each request's last
+        scheduled token), or packed [target_num_reqs, num_spec_tokens+1] under spec
+        decode so draft and bonus logits come out of a single pass.
         """
         num_reqs = len(idx_mapping_np)
         rs = self.req_states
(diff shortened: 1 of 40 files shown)


examples/vllm/telemetry/api_load_test.py
```diff
@@ -0,0 +1,310 @@
...
+    window, and build_prompt only approximates the token count (BOS + tokenizer
+    variance can push the real prompt a little past its target). Returns
+    ``(plan, clamped)`` where ``clamped`` is True if any input length was capped.
+    """
+    slack = max(64, int(0.03 * max_model_len))
+    budget = max_model_len - slack
+    clamped = False
+    plan = []
+    for _ in range(n):
+        o = rng.randint(osl_min, osl) if (osl_min and osl_min < osl) else osl
```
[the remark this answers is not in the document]
avorozcovsTT: False positive in this context.

`rng` is `random.Random(args.seed)`, and its draws only pick per-request input/output token *lengths* for a client-side load generator — they never become identifiers, tokens, nonces, or inputs to an access-control decision.

Reproducibility is the feature here, not an oversight: `--seed` is a documented CLI flag and `--dry-run` prints the exact plan it produces, so `secrets` / `random.SystemRandom()` would break the intended behaviour. `numpy.random.default_rng(seed)` would satisfy the rule while staying deterministic, but this module is deliberately stdlib-only ("Stdlib only (urllib + threads); no extra dependencies" in the header) and a dev-only script under `examples/` isn't worth a new dependency.

0397de5d4 adds a comment at the RNG site recording the intent. Resolving.

integrations/vllm_plugin/vllm_tt/model_runner_v2.py
```diff
@@ -37,6 +37,7 @@
 from vllm.v1.worker.lora_model_runner_mixin import LoRAModelRunnerMixin
 
 from .logger import tt_init_logger
+from .telemetry import RunnerTelemetry, resolve_config
```
mmanzoorTT, line 41: I suggest to add telemetry for ModelRunner V1 for now. V2 is not fully functional. We can add telemetry for V2 once it is ready.

mmanzoorTT (commented):
Any plan on adding telemetry for Pooling runner?

integrations/vllm_plugin/vllm_tt/model_runner_v2.py
avorozcovsTT: Thank you, added v1 telemetry in https://github.com/tenstorrent/tt-xla/pull/5808

avorozcovsTT: Added an issue for telemetry for the pooling runner here: https://github.com/tenstorrent/tt-xla/issues/5812

Issue #5778, opened 2026-07-23
[vLLM] Gemma4-31B-it DP+TP output degradation without `ignore_eos=True`


ddilbazTT:
## Summary

Running `google/gemma-4-31B-it` through the vLLM plugin in data-parallel + tensor-parallel mode on a Blackhole galaxy, a batch of **distinct** prompts produces corrupted (token-soup) output in the later batch positions, while the same run with a **homogeneous** batch (one prompt repeated) is clean. The corrupted text starts coherent and degrades into repeated punctuation / non-language tokens.

## Environment

- Hardware: Blackhole galaxy, 32 chips
- Model: `google/gemma-4-31B-it` (text-only path)
- Parallelism: `mesh_shape=[8, 4]` (DP=8, TP=4), `enable_data_parallel=True`,
  `enable_tensor_parallel=True`, `shard_weights_on_batch_axis=True`

## How it was discovered

While validating DP+TP generation, a large batch of distinct one-sentence prompts produced increasingly corrupted output toward the end of the batch. 

## Repro

```python
import vllm

MODEL = "google/gemma-4-31B-it"
PROMPTS = [
    "Describe Tenstorrent in one sentence.",
    "Explain what a neural network is in one sentence.",
    "What is the capital of France?",
    "Write one sentence about the ocean.",
    "Summarize the theory of relativity in one sentence.",
    "Give me a one-sentence description of photosynthesis.",
    "What is machine learning, in one sentence?",
    "Describe the sun in one sentence.",
]

llm = vllm.LLM(
    model=MODEL,
    limit_mm_per_prompt={"image": 0, "video": 0, "audio": 0},
    max_num_batched_tokens=8192,
    max_num_seqs=8,
    max_model_len=128,
    gpu_memory_utilization=0.3,
    additional_config={
        "enable_const_eval": True,
        "min_context_len": 32,
        "enable_data_parallel": True,
        "enable_tensor_parallel": True,
        "shard_weights_on_batch_axis": True,
        "experimental_weight_dtype": "",
        "mesh_shape": [8, 4],
        "cpu_sampling": False,
        "flat_model_io": True,
    },
)

sp = vllm.SamplingParams(temperature=0.0, top_p=1.0, max_tokens=32)
messages = [[{"role": "user", "content": p}] for p in PROMPTS]
for prompt, out in zip(PROMPTS, llm.chat(messages, sp)):
[...]

PR #5799, opened 2026-07-28
[vLLM] Fix DP condense() KV-cache corruption (#5778)

ssaliceTT:
### Ticket
#5778

### Problem description
Under DP+TP a batch slot is physically bound to a DP replica's (replicated) KV cache. Each replica holds KV only for its own slots. When a request finishes mid-batch, `condense()` relocates a surviving request into the freed slot to keep the batch dense, sometimes across the replica boundary. It moves the block-table mapping but not the KV bytes (which stay on the old replica), so the survivor then reads the wrong replica's cache and its output degrades into token-soup in later batch positions. It's deterministic: a homogeneous batch or `ignore_eos=True` avoids it (no staggered finishes → no relocation). Reproduced on Qwen3-8B.

### What's changed
Under DP, finished requests are no longer removed + condensed. Each finished slot is kept inert (the scheduler stops scheduling it → 0 tokens → masked) and reclaimed in place by the next new prefill, which writes that slot's own replica's KV. A slot's replica therefore never changes and no surviving request is ever relocated; `condense` is skipped entirely under DP. Non-DP and TP-only paths are unchanged (gated on `parallel_mode`). Adds `test_dp_condense_kv_affinity`.

Validated on Qwen3-8B DP+TP 2×4 mesh: the staggered-finish repro corrupts "survivor" positions pre-fix and with this change the accuracy issues are not to be seen. Slot reuse across generations also works. Also, the unscheduled path is unaffected (verified on-device locally with some tests).

### Checklist


integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -528,6 +528,8 @@ def __init__(
             if self.speculative_config.method == "ngram":
                 self.drafter = NgramProposer(vllm_config)
 
+        self._dp_inert_slots: list[int] = []
```
mmanzoorTT: do you mean `self._dp_insert_slots`?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -902,8 +904,15 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> bool:
...
         """
+        dp_active = self.parallel_mode in (
+            ParallelismMode.DATA_PARALLEL_ONLY,
+            ParallelismMode.DATA_TENSOR_PARALLEL,
+        )
+
         # Remove finished requests from the cached states.
         for req_id in scheduler_output.finished_req_ids:
+            if dp_active:
+                continue
```
mmanzoorTT, line 988: It seems that you are not removing/condensing requests. What will happen with a long-running server where requests keep coming?

mmanzoorTT (commented):
You

integrations/vllm_plugin/vllm_tt/model_runner.py
ssaliceTT: I did mean inert, not a typo for insert, but I see how it's confusing so I can change the naming to `self._dp_free_slots`

integrations/vllm_plugin/vllm_tt/model_runner.py
ssaliceTT: The freeing just happens differently under DP. We deliberately skip remove_request/condense here because condense is what corrupts the KV cache .

Slots are still recycled, just in place instead of by condensing:
- finish: we don't remove the request, we record its freed slot in the free-list (_dp_free_slots);
- new request: the add loop pops a free slot, evicts the old occupant from that exact index (requests.pop + input_batch.remove_request), and drops the newcomer there.

So the batch is bounded by `max_num_reqs` forever, every slot the scheduler counts as freed maps to exactly one free-list entry, so new admissions always find a slot to reclaim. 

I tried running tests with prompt list greater than `max_num_seqs` and it worked without any issue. I believe that handles the case of unscheduled requests that overflow and come in.

There is one tradeoff here techincally. Since we lose compaction, a finished slot rides along as decode padding until a new prefill reclaims it so we are doing extra compute. But also decode forward pass pads the batch to 32 anyways.

mmanzoorTT: Will the `finished request` be part of inputs during decoding (and will be updating kv cache etc) or just retaining a slot (and no inference will be made)?

ssaliceTT: Yeah it will be updating kv cache during decode too. I added some extra gating to ensure released cache blocks don't get written to as that would've have caused corruption as well. Thanks for asking that as it helped me catch it.

ssaliceTT: Note, we can make DP-aware condensing but it would be much more complex. And since we already pad to batch 32 during decode it made sense to make this the DP decode pathway, at least for the time being.

mmanzoorTT (approved):

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1098,9 +1112,18 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> bool:
         )
         for req_id in req_ids_to_add:
             req_state = self.requests[req_id]
-            # Fill the empty index or append to the end
-            req_index = removed_req_indices.pop() if removed_req_indices else None
-            self.input_batch.add_request(req_state, req_index)
+            if dp_active and self._dp_free_slots:
```
kmabeeTT: One case I'm not sure about: the vLLM NOTE at the top of `_update_states` (~L854) says `finished_req_ids` and `scheduled_new_reqs` can carry the same `req_id` — aborted and resubmitted under that ID, treated as two distinct requests. Since the finished branch no longer pops `self.requests`, the new `CachedRequestState` overwrites the old one, and then `old_id` here can be that same `req_id` — in which case `self.requests.pop(old_id, None)` drops the state we just installed.

The other ordering looks messier to me. If `pop(0)` returns a different slot first, the stale row for `req_id` survives at its old index, and when that index is popped later `remove_request(old_id)` resolves the row via `req_id_to_index` — which by then points at the live row, not the stale one. If I'm reading that right it would blank the live row and re-seat the request at another slot, which is the relocation we're trying to avoid.

Is this reachable in our engine config? If so, would a guard on `old_id == req_state.req_id` plus a staleness check (`req_id_to_index.get(old_id) != req_index`) be enough to cover it?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1817,6 +1840,13 @@ def _prepare_inputs(self, scheduler_output: "SchedulerOutput", start_index: int)
             if fill_page_table is not page_table:
                 fill_page_table[actual_num_reqs:, :] = 0
 
+            # A parked (finished/unscheduled) row's page_table may still point at
+            # blocks the scheduler freed and reassigned to a live request; sending
+            # its decode paged_update_cache write to the null block (as the fill
+            # path above does) keeps it from clobbering that request's KV.
+            if len(zero_sched_rows) > 0:
+                page_table[zero_sched_rows, :] = 0
```
kmabeeTT, line 1868: Two small things here:

The comment at ~L1396 ends with "the read-path page_table keeps the real blocks", which this change makes untrue — might be worth folding the new reasoning into that comment so the two don't disagree. (If it helps: my read is that zeroing the read path is safe because a zero-scheduled row's sampled token gets discarded in `execute_model` anyway, so what it reads out of block 0 never escapes — that part wasn't obvious to me.)

Also, now that `page_table` gets the same rows zeroed, the `page_table.clone()` above looks redundant in the non-roll case — the two end up identical, but `fill_page_table is not page_table` so L1428 takes the else branch and does an extra full page-table H2D. Since a parked row is the normal state under this PR, I think that's every decode step. Would this work instead?

```python
zero_sched_rows = np.nonzero(num_scheduled_tokens_per_req == 0)[0]
if len(zero_sched_rows) > 0:
    page_table[zero_sched_rows, :] = 0
    if fill_page_table is not page_table:
        fill_page_table[zero_sched_rows, :] = 0
```

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1114,7 +1137,7 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> bool:
                 )
 
         # Condense the batched states if there are empty indices.
-        if removed_req_indices:
+        if removed_req_indices and not dp_active:
```
kmabeeTT, line 1166: Could we get a short TODO + tracking issue here? The reasoning from your reply above is worth having in the file: condense is skipped under DP because relocation breaks slot->replica affinity, the cost is a finished slot riding along as decode padding until a new prefill reclaims it (largely free today since decode pads to the max request shape anyway), and DP-aware condensing is the eventual fix.

tests/integrations/vllm_plugin/generative/test_dp_condense_kv_affinity.py
```diff
@@ -0,0 +1,75 @@
...
+                temperature=0.0,
+                top_p=1.0,
+                max_tokens=(short_len if i < n // 2 else long_len),
+                ignore_eos=True,
+            )
+            for i in range(n)
+        ],
+    )
+
+    for i in survivors:
```
kmabeeTT: Since the prefix compare already handles unequal lengths (`m = min(...)`), could this loop just be `range(n)`? As written the early-finishing half is never checked against the reference, so a regression that corrupts those rows instead of the survivors would pass.

kmabeeTT (commented):
Left some comments. Good idea to run vLLM Nightly to double check safe and compare with latest.  Also two coverage questions, neither blocking:

1. Is `test_dp_condense_kv_affinity` expected to catch the parked-row `page_table` zeroing on its own? The case that seems to exercise it is a parked slot whose freed blocks get reassigned to a live request, and I'm not sure a 40-token run gets there — so I can't tell whether that part would regress silently.

2. The push llmbox job runs `tensor_parallel and llmbox and not data_parallel`, so everything here is nightly-only and ungated pre-merge. Not this PR's problem, but now that there's real DP correctness logic in `_update_states`, a DP+TP smoke test in push might be worth a follow-up.

ssaliceTT:
> Left some comments. Good idea to run vLLM Nightly to double check safe and compare with latest. Also two coverage questions, neither blocking:
> 
> 1. Is `test_dp_condense_kv_affinity` expected to catch the parked-row `page_table` zeroing on its own? The case that seems to exercise it is a parked slot whose freed blocks get reassigned to a live request, and I'm not sure a 40-token run gets there — so I can't tell whether that part would regress silently.
> 2. The push llmbox job runs `tensor_parallel and llmbox and not data_parallel`, so everything here is nightly-only and ungated pre-merge. Not this PR's problem, but now that there's real DP correctness logic in `_update_states`, a DP+TP smoke test in push might be worth a follow-up.

I have this PR https://github.com/tenstorrent/tt-xla/pull/5800 where I'll be adding some push and nightly tests for DP+TP chunked prefill cases. To be honest with that, the dp condense kv affinity test would be redundant so I'll take it off of this PR. 

Also ran these two CI runs https://github.com/tenstorrent/tt-xla/actions/runs/30972620281 and https://github.com/tenstorrent/tt-xla/actions/runs/30972636265 on the branch. Seems tests are fine and nothing currently running will be broken.

integrations/vllm_plugin/vllm_tt/model_runner.py
ssaliceTT: Added TODO with the reasoning and opened #5896 to track it

integrations/vllm_plugin/vllm_tt/model_runner.py
ssaliceTT: Updated the stale comment and dropped the redundant page table clone. So now it zeroes `page_table[zero_sched_rows]` directly and only touch `fill_page_table` when it's a separate object, so the normal path shares one buffer and skips the per-step page-table copy.

integrations/vllm_plugin/vllm_tt/model_runner.py
ssaliceTT: Yeah good catch. in an abort+resubmit case the id would land in both finished and scheduled reqs so when the slots collide either the new state gets dropped, or the id ends up in two slots (one live, one stale).

I handle it by reserving the slots for resubmitted ids instead of parking or freeing it. I compute `resubmitted = finished_req_ids & {n.req_id for n in scheduled_new_reqs}`, and for those ids, the finished loops don't park them and don't add the slot to `_dp_free_slots`, so nothing else can grab it. Then in the same update step when slots are reclaimed by new requests, I `add_request` the fresh state back into the same slot. Since the id is still in the batch, I also `remove_request` the old incarnation and before adding. Normal finishes (not resubmitted) still park as before.

Because it's a remove+add on the request's own slot, the id is never in two slots, nothing is dropped, and no hole is left below num_reqs, for any ordering. That makes the three-part guard you mentioned unnecessary.

PR #5800, opened 2026-07-28
[vLLM] Add DP+TP chunked-prefill tests

ssaliceTT:
### Problem description

#5692 broadened chunked-prefill coverage to single-chip, TP-only and DP-only, but skipped DP+TP. All three existing DP+TP tests use `max_model_len=32` with prompts that fit in one chunk, so the cached-prefix chunked-SDPA path was never exercised on a 2D mesh.

### What's changed

Two nightly llmbox tests (Qwen3-0.6B and Qwen3-8B), parametrized on `batch_size` 32 and 64, with `prefill_chunk_size=128` and `max_model_len=1024`. Each batch mixes prompts longer than the chunk (so they split and route chunks 2..N through the cached-prefix path) with short single-chunk prompts, so one step contains sequences at different prefill stages.

`assert_slots_agree` in conftest: the batch repeats a few base prompts, so under greedy sampling slots holding the same prompt must produce identical token ids and a mismatch names the corrupted slot. Coherence is asserted too, since equality alone passes if every slot is identically garbled.

### Test

Validated on n300-llmbox (2x4 mesh). Draft because these currently **fail**, due to two bugs found:

1. Device hang: `paged_fill_cache` got a tilized page table under DP+TP chunked prefill. Fixed by #5797; these tests need it to get past engine init.
2. Output corruption: with #5797 applied the tests run to completion, but slots holding the same prompt diverge. Short single-chunk prompts are corrupted only when batched alongside multi-chunk ones, so it looks related to rows sitting at different prefill stages in one step. Still under investigation. Not sure if #5799 alone fixes it.

Will un-draft once those land and the tests pass.

### DIFF
--- a/tests/integrations/vllm_plugin/generative/conftest.py
+++ b/tests/integrations/vllm_plugin/generative/conftest.py
@@ -112,6 +112,31 @@ def assert_batch_grounded(outputs, checks=GROUNDED_BATCH_CHECKS) -> None:
         ), f"expected {expected!r} for {prompt!r}, got {text!r}"
 
 
+def assert_slots_agree(outputs, prompts) -> None:
+    """Slots sharing a prompt must generate identical, coherent output.
+
+    The batch repeats a few base prompts, so greedy sampling makes slots holding
+    the same prompt directly comparable and a mismatch names the corrupted slot.
+    Coherence is checked too, since equality alone would pass if every slot were
+    identically garbled.
+    """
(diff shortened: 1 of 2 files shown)


ssaliceTT:
Will add some push tests as well to have onPR coverage. It will focus on running e2e without necessarily checking for accuracy as there may still be issue with accuracy as of today.

PR #5801, opened 2026-07-28
[vLLM] DP+TP chunked prefill: sharding and buffer-sizing fixes

ssaliceTT:
## Problem

Chunked prefill at batch 256 on an 8x4 DP+TP mesh (Qwen3-32B) compiles, then dies at the first generation step:

```
ValueError: provided out is the wrong size for the accumulation
```

`query_start_loc_cpu` and `seq_lens_cpu` are indexed per request but were sized by `max_num_tokens`. Without chunked prefill `max_num_tokens >= max_num_seqs` always held. Under chunked prefill `max_num_tokens` is the per-sequence chunk budget (128), smaller than `max_num_seqs` (256), so `np.cumsum` rejects the short `out=` slice.

Separately, `partition_column_parallel_linear` accepts `shard_weights_on_batch_axis` but ignores it, unlike its Merged and QKV siblings.

## What's changed

Size `query_start_loc_cpu` and `seq_lens_cpu` by `max_num_reqs`, since every use site is request-indexed and `num_reqs` is always clamped to it.

Honor `shard_weights_on_batch_axis` in `partition_column_parallel_linear`. No effect on Qwen3-32B, which instantiates no bare `ColumnParallelLinear`.

Verified on a 32 chip Blackhole galaxy, mesh (8,4), Qwen3-32B with chunked prefill. Batch 256 completes 256/256 prompts at 1024 and 16384 context on a 2 layer model. Batch 16 at 4096 context on the full 64 layer model generates coherent output, with identical single-chunk prompts in different batch rows producing byte-identical results.

_**Note**_: An earlier revision of this PR also pinned the users dim on `RowParallelLinear` output to work around a missing users factor in the tt-mlir chunked SDPA sharding rule. That workaround has been dropped in favour of fixing the rule itself, tenstorrent/tt-mlir#9135 and tenstorrent/tt-mlir#9142.

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -565,7 +565,7 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
             self.max_num_reqs,
         )
         self.query_start_loc_cpu = torch.zeros(
-            self.max_num_tokens + 1,
+            self.max_num_reqs + 1,
             dtype=torch.int32,
             device="cpu",
             pin_memory=self.pin_memory,
@@ -573,7 +573,7 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         self.query_start_loc_np = self.query_start_loc_cpu.numpy()
 
         self.seq_lens_cpu = torch.zeros(
-            self.max_num_tokens,
+            self.max_num_reqs,
             dtype=torch.int32,
             device="cpu",
             pin_memory=self.pin_memory,
--- a/integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
+++ b/integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
@@ -318,8 +318,8 @@ def partition_column_parallel_linear(
     layer: torch.nn.Module, mesh: xs.Mesh, shard_weights_on_batch_axis: bool = True
 ) -> torch.nn.Module:
     assert isinstance(layer, ColumnParallelLinear)
-    # Weight is [output, input]: output on the "model" (TP) axis, input replicated.
-    safe_mark_sharding(layer.weight, mesh, ("model", None))
+    batch_axis = "batch" if shard_weights_on_batch_axis else None
+    safe_mark_sharding(layer.weight, mesh, ("model", batch_axis))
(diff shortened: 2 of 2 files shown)


integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
```diff
@@ -318,18 +318,77 @@ def partition_column_parallel_linear(
...
+        # local sharded shape for result N". Leave them to Shardy — with a
+        # single user there is no batch split to preserve anyway.
+        if not batch_axis_size or tensor.shape[0] % batch_axis_size != 0:
+            return output
+        spec = ("batch",) + (None,) * (tensor.ndim - 1)
+        sdy_sharding = _partition_spec_to_sdy_sharding(mesh, spec)
+        pinned = torch.ops.tt.sharding_constraint(tensor, sdy_sharding)
+        return (pinned, *output[1:]) if is_tuple else pinned
+
+    return hook
```
mmanzoorTT: Can we move this to `python_package/tt_torch/sharding.py`?

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -564,16 +564,17 @@ def __init__(
             TTAttentionBackend.get_max_num_seqs(self.max_model_len, self.block_size),
             self.max_num_reqs,
         )
+        req_token_capacity = max(self.max_num_tokens, self.max_num_reqs)
         self.query_start_loc_cpu = torch.zeros(
-            self.max_num_tokens + 1,
+            req_token_capacity + 1,
```
mmanzoorTT: It should be always `self.max_num_reqs` as we are slicing it based on received requests. 
```suggestion
            self.max_num_reqs + 1,
```

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -564,16 +564,17 @@ def __init__(
...
             dtype=torch.int32,
             device="cpu",
             pin_memory=self.pin_memory,
         )
         self.query_start_loc_np = self.query_start_loc_cpu.numpy()
 
         self.seq_lens_cpu = torch.zeros(
-            self.max_num_tokens,
+            req_token_capacity,
```
mmanzoorTT: Same here. 

```suggestion
            self.max_num_reqs,
```

mmanzoorTT (approved):

mmanzoorTT (approved):

PR #5804, opened 2026-07-28
[vLLM] Fix global vs local row indexing in multi-pass prefill

ssaliceTT:
### Problem description

`_prepare_inputs` handles rows `[start_index, num_reqs)` of `input_batch`, but reads per-request state with **local** 0-based indices:

```python
+ self.input_batch.num_computed_tokens_cpu[i]                      # needs start_index + i
computed_tokens = self.input_batch.num_computed_tokens_cpu[i]
input_ids_cpu[i, :n] = self.input_batch.token_ids_cpu_tensor[i, ...]
self.seq_lens_np[:num_reqs] = self.input_batch.num_computed_tokens_cpu[:num_reqs] + ...
page_table[:actual_num_reqs] = ...block_table[0].get_cpu_tensor()[:actual_num_reqs]
```

The caller uses the right convention (`local_idx = i - start_index`, `model_runner.py:2328`), so on any pass with `start_index > 0` the requests are fed the **first** N rows' prompt tokens, computed-token offsets and block tables instead of their own.

Multi-pass fires whenever the prefill row cap trims the batch — i.e. when `max_prefill_num_seqs` is set, or `num_reqs_max_model_len` (SMEM) is below `max_num_seqs`. With the default (no cap) there is a single pass and this is dormant.

### What's changed

Offset the 8 `input_batch` reads by `start_index`. Destinations stay local — only the source indexing changes.

### Test

n300-llmbox, Qwen3-0.6B, TP-only, no chunking, batch 64, uniform 96-token prompt (so all 64 slots must produce identical greedy output), `max_prefill_num_seqs=16` to force 4 passes:

| | correct rows |
|---|---|
| before | 17/64 — only pass 1 (`start_index=0`) plus the 1-row warm step |
| after | **64/64** |

Corruption boundaries lined up exactly with the logged pass boundaries (`start_index=0/16/32/48`), and the failing rows re-emitted the *start* of their prompt, consistent with reading another row's state.

Found while investigating DP+TP chunked-prefill accuracy; unrelated to that root cause, filed separately since it is a standalone latent bug.

### Note

Setting `max_prefill_num_seqs` under data parallelism separately trips `assert max_num_scheduled_tokens_all_reqs > 0`: the cap slices the first N rows positionally, but DP intentionally retains already-prefilled requests as inert zero-scheduled rows (`model_runner.py:934-943`), so a pass can select N rows that are all inert. Not addressed here.

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -1362,7 +1362,7 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         for i, n in enumerate(num_scheduled_tokens_per_req):
             positions = (
                 torch.arange(n, dtype=torch.int32)
-                + self.input_batch.num_computed_tokens_cpu[i]
+                + self.input_batch.num_computed_tokens_cpu[start_index + i]
             )
             if self.uses_mrope:
                 arange[:, i, :n] = positions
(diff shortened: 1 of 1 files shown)


ssaliceTT:
Closing for now. Fixes a bug (_prepare_inputs reads per-request state with local indices while covering global rows [start_index, num_reqs), so every pass after the first feeds requests another row's prompt tokens and block tables; verified 17/64 -> 64/64 correct rows in a 4-pass config). But it is dormant unless max_prefill_num_seqs is set, and setting that under DP hits a separate assert (the cap slices rows positionally and can select only inert zero-scheduled rows). Branch ssalice/prefill-multipass-index-fix is kept; reopen if the row cap becomes usable.

PR #5825, opened 2026-07-29
Add galaxy-bh forge-model CI support and DeepSeek-V2-Chat TP

devisettymahidhar608:
## Summary

- Add galaxy-bh as a first-class forge-model architecture (ALLOWED_ARCHES, runner handling, weekly matrix) and bring up DeepSeek-V2-Chat TP with required_pcc: 0.98.

- Unify arch selection by renaming inverted bh_galaxy marks to galaxy_bh, matching forge’s galaxy-bh / galaxy_bh naming.

- Keep runner targeting as runs-on: galaxy-bh; standalone CI filters with galaxy_bh, forge-model CI with galaxy-bh.


sonalibaskaran2499 (approved):

vvukomanTT (approved):

nsumrakTT (approved):

kmabeeTT (approved):

jameszianxuTT (changes requested):
We already have blackhole galaxy tests and those are run by filtering for the mark `@pytest.mark.bh_galaxy`

I think it would be better to fold these "galaxy-bh" and "bh_galaxy" under the same mark/selection method so we don't have arch specified in different ways depending on the test. Might want to zoom out and take a look at what is happening here, I think the test_models.py script takes the supported arches and inserts it into the pytest marks. 

The runs on you've set is correct though, and to target a runner it should be `runs-on: galaxy-bh` which makes me wonder why bh_galaxy exists as a mark at all in our repo... It may be better to convert all the bh_galaxy marked tests to galaxy-bh

devisettymahidhar608:
@jameszianxuTT Addressed in 621cfb064.

Converted the inverted `bh_galaxy` mark to `galaxy_bh` everywhere (pytest marks, `pytest.ini`, nightly matrix filters, inspect-changes exclusion) so it matches the underscore form that forge-model collection already derives from `supported_archs: [galaxy-bh]` in `tests/runner/conftest.py` (hyphenated + underscore variants).

- Runner remains `runs-on: galaxy-bh`
- Forge-model matrix keeps selecting with `galaxy-bh` (arch string)
- Standalone tests/CI now use `galaxy_bh`, same pattern as `lb_blackhole` / `galaxy_wh_6u`

mmanzoorTT (approved):

jameszianxuTT (approved):
Thanks for making changes!

nvukobratTT (approved):

PR #5832, opened 2026-07-29
Fix bge-m3 flattened-IO regression: bf16 linears and host-side reshape

### LINKED ISSUE #5756: [vLLM] Perf Analysis on using flattened inputs
vLLM is using batched inputs as default option, [PR](https://github.com/tenstorrent/tt-xla/pull/5700) makes flattened inputs as default option. The purpose of this analysis to find out its impact on Performance Benchmark tests. 
CI job: https://github.com/tenstorrent/tt-xla/actions/runs/29731726925

alinakhanTT:
### Ticket
Closes #5756

### Problem description

bge-m3 regressed 30-54% with `flat_model_io` while other models were perf neutral. Two independent causes, both triggered by the flattened path handing the model rank-2 activations instead of rank-3.

**1. The encoder silently ran in f32.** A rank-2 activation makes `nn.Linear` lower to `aten::addmm` instead of `aten::mm`, and PyTorch's `addmm` decomposition routes every operand through float32 for opmath. With flattened IO that hit all 9216 linears: 288 bf16->f32 typecasts, weights transposed *and* upcast in const-eval, and no `transpose_b` folding or gelu fusion because converts sat between the ops.

`bypass_dtype_promotion_and_redundant_cast` already existed to undo exactly this ("Pytorch insists on casting nodes to float32 during decompositions"), but it only recognised the `prims::convert_element_type` spelling and read the target dtype from `args[1]`. These casts arrive as `aten::_to_copy` with dtype in kwargs, so every one slipped through.

**2. Flattening compiled a reshape program per input shape.** The flatten reshapes ran on the device, outside the compiled model. Eager ops there become their own XLA program, and the program is shape-specialised, so each distinct input shape needed its own compilation. `_dummy_run` produced one per warmup bucket (96, alongside the 96 model programs); `_prepare_inputs` produced more for real requests, so a shape first seen mid-run compiled then -- putting a 4.3s compile inside the benchmark's timed loop.

Worth noting for anyone reading the original numbers: the benchmark averages all iterations, so those compiles dominated the reported figure. Steady-state per-iteration latency was always ~0.33s in both modes. Identical code measured 41, 53, 60, 74 and 97 samples/sec across runs depending on how many compiles landed inside the window.

### What's changed

**`passes.py`** - read the cast's target dtype from either spelling. For `_to_copy`, strip only casts attributed to the fused add+matmul decompositions (`addmm`/`addmv`/`addbmm`/`baddbmm`), since `_to_copy` is also how an explicit user `.to()` is spelled, and leave alone any `_to_copy` carrying device/layout/memory_format kwargs (those relocate the tensor, not just retype it). Batched inputs lower to `aten::mm` and emit no casts at all, so their IR is byte-identical.

**`pooling_runner.py`** - reshape on the host, before the tensors reach the device, in both `_prepare_inputs` and `_dummy_run`, and restore the hidden states after the CPU transfer rather than before it. A host reshape needs no program at all. Data parallel marks sharding on the rank-2 tensors, so it keeps the existing device-side path.

bge-m3 batch32, trace disabled, on n150-class silicon:

| | samples/sec | graphs (reshape helpers) |
|---|---|---|
| neither fix | 41.6 (41-74 across runs) | 199 (103) |
| pooling fix only | 67.3 | 97 (1) |
| **both** | **98.7** | **97 (1)** |
| batched reference | 96.7-97.6 | 97 (1) |

Flattened now matches the batched graph count exactly and edges past it on throughput, which is the expected outcome -- flattening genuinely removes work from the model graph (9512 reshapes vs 18810). Flattened embeddings are bit-identical to batched. Batched is unaffected: same 97 graphs, same throughput, same embeddings.

### Checklist
- [x] New `test_embedding_flat_model_io` runs the pooling path with `flat_model_io`  enabled against the same baseline as the batched test. Nothing enabled the flag  before, which is how the f32 regression shipped unnoticed.
- [x] New filecheck test `test_linear_bias_rank2_stays_bf16` asserts a biased rank-2  bf16 linear keeps bf16 operands. The existing `test_model_filecheck` has a  rank-2 bf16 linear but with `bias=False`, which lowers to `mm` and never  exercised the `addmm` path.
- [x] vLLM pooling suites all green: 18 passed/1 skipped single-device, 7 DP,  9 TP + DP+TP.
- [x] 'tests/torch/ops` + `tests/torch/graphs` A/B'd with and without the dtype fix:  identical failure sets (153 failed / 274 passed either way, all pre-existing on  the test box), so the pass change introduces no new failures.

### DIFF
--- a/python_package/tt_torch/backend/passes.py
+++ b/python_package/tt_torch/backend/passes.py
@@ -670,6 +670,43 @@ def clamp_neg_getitem_bounds(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
     return gm
 
 
+# Decompositions of the fused add+matmul ops route every operand through float32
+# irrespective of the model dtype. A rank-2 activation makes nn.Linear lower to
+# addmm instead of mm, so with flattened model IO this hits every linear in the
+# graph and drags whole encoders into f32. Refs: tt-xla #5756.
+_FLOAT32_PROMOTING_MATMULS = (
+    "aten::addmm",
+    "aten::addmv",
+    "aten::addbmm",
+    "aten::baddbmm",
+)
+
+
+def _dtype_cast_target(node):
+    """
+    (spelling, target dtype) of a pure dtype-cast node, or None if not one.
+
+    Decompositions spell the cast either as prims::convert_element_type or as
+    aten::_to_copy with a dtype kwarg. For _to_copy, any kwarg that also relocates
+    the tensor (device/layout/memory_format) means it is not a pure cast and must
+    be left alone.
+    """
+    if node.op != "call_function" or not hasattr(node.target, "name"):
+        return None
+    if not node.args or not isinstance(node.args[0], torch.fx.Node):
+        return None
+
+    name = node.target.name()
+    if "prims::convert_element_type" in name:
+        dtype = node.args[1] if len(node.args) > 1 else node.kwargs.get("dtype")
+        return ("convert_element_type", dtype)
+    if "aten::_to_copy" in name:
+        if any(key not in ("dtype", "non_blocking") for key in node.kwargs):
+            return None
+        return ("_to_copy", node.kwargs.get("dtype"))
+    return None
+
+
 def bypass_dtype_promotion_and_redundant_cast(gm, example_inputs):
     """
     Removes casting of nodes to float32 unless they were explicitly set by the user.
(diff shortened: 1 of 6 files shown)


alinakhanTT:
Nightly green: https://github.com/tenstorrent/tt-xla/actions/runs/31188613932

python_package/tt_torch/backend/passes.py
```diff
@@ -670,6 +670,43 @@ def clamp_neg_getitem_bounds(gm: torch.fx.GraphModule) -> torch.fx.GraphModule:
     return gm
 
 
+# Decompositions of the fused add+matmul ops route every operand through float32
```
kmabeeTT, line 673: Is it rank that matters here, or bias? Tracing a bf16 Linear locally, `bias=True` gives the f32 `addmm` sandwich at both rank 2 and rank 3 (rank-3 just flattens via `view` first); `bias=False` gives a clean bf16 `mm` at both. If that's right, batched IR isn't byte-identical and this touches every biased `nn.Linear` rather than just flat IO — probably a win, but wider than the description scopes.

Does that square with bge-m3 batched coming out flat? XLM-R linears have bias, so I'd have expected those to move too, which makes me suspect I'm missing something about the pooling path. And if it is repo-wide, is `ops` + `graphs` the right A/B, or worth a nightly `tests/runner` run once rebased?

tests/integrations/vllm_plugin/pooling/test_single_device.py
```diff
@@ -29,6 +29,34 @@ def test_embedding_push(model_name: str, baseline_path, max_model_len: int):
...
+    """Flattened model IO must match the same baseline as the batched path.
+
+    Flattening hands the model rank-2 activations, which lower through a
+    different set of aten ops than rank-3 does. Refs: tt-xla #5756.
+    """
+    run_pooling_test(
+        model_name,
+        baseline_path,
+        max_model_len,
+        max_num_reqs=1,
```
kmabeeTT: Any reason for `max_num_reqs=1` here? At batch 1 the restore path is always `(1, N)` → `(N,)` → `(1, N, H)`, so the multi-row reshape never actually runs — and as far as I can tell nothing else in the suite sets `flat_model_io`, so this is its only coverage. `run_pooling_test` already defaults to 2, so just dropping the override would exercise it.

kmabeeTT (commented):
Looks good, couple questions

tests/integrations/vllm_plugin/pooling/test_single_device.py
alinakhanTT: You're right, batch 1 made the multi-row restore dead code, the override was copied from test_embedding_push. Changed to max_num_reqs=4 rather than dropping to the default 2, because batch 2 on this model is the combination test_batched_inference skips for random segfaults (#3094) and I'd rather not put that in the push gate, batch 4 runs unskipped there.
Cost is precompile going from 1 to 4 backbone programs (83s of a 102s test)

python_package/tt_torch/backend/passes.py
alinakhanTT: It's the device, not the rank. aten::linear sends rank 2 + bias to addmm unconditionally, but the rank-3 flatten-to-addmm path is gated on !input.is_xla(). So on XLA rank 3 falls through to matmul + add and never hits the promoting decomposition, bias and rank 2 and XLA. Your CPU trace is right for CPU, that gate is what squares bge-m3 batched coming out flat despite XLM-R having biases. (PJRT_DEVICE=CPU repros the XLA dispatch if you want to see it)
Your scope point stands anyway for a different reason: rank-2 into a biased linear happens outside flat IO too, pooler/classifier heads on [B, H], 2D MLP inputs. So "byte-identical" is a bge-m3 measurement, not a general rule, I reworded that and put the is_xla gate in the comment
Rebased and ran both suites: model-test-passing has 10 failed, all pre existing, and model-test-xfail is green with nothing flipping to XPASS

PR #5833, opened 2026-07-30
Add Falcon3-7B-Instruct production vLLM benchmark config

kmabeeTT:
## Ticket

Addresses #5404 (Falcon3-7B-Instruct only — not the other models listed in that issue)

## Problem description

The existing `falcon3-7b-base` vLLM benchmark runs a minimal smoke-test config (`max_model_len=128`), not representative of how Falcon3-7B-Instruct is actually served in production. #5404 asks for benchmark coverage at each model's native max sequence length, batch-32, with the highest safe `gpu_memory_utilization` — Falcon3-7B-Instruct's native max is 32768.

## What's changed

- `tests/benchmark/test_vllm_benchmarks.py`: added `_falcon3_7b_instruct_production_config()` and a new `falcon3-7b-instruct-production` single-device benchmark entry, mirroring the real production launch config (`launch_falcon3_7b_instruct_uvicorn.sh` / `cnn.yaml` forge spec): `max_model_len=32768`, batch-32, `gpu_memory_utilization=0.35`, `optimization_level=1`, bfp8 weights/KV cache, chunked prefill (`prefill_chunk_size=1024`), b1-prefill (`min_num_seqs=1`, `prefill_batch_threshold=16`), and a b16 prefill-graph cap (`max_prefill_num_seqs=16`, from #5541) — measured to cut ~22% off compile time with no throughput/TTFT regression.
- `optimization_level=1` is intentional, not a placeholder: at this config's `max_model_len=32768` + chunked-prefill shape ladder, `optimization_level=2` took 2:00:09 to complete (full run measured) vs. 13:07 at `optimization_level=1` — 9.2x slower to compile, for +23% samples/sec (19.86 vs 16.14) but +32% worse TTFT (1146.2ms vs 866.2ms). Avoid bumping this to 2 without accounting for that compile-time cost.

## Checklist
- [x] Verified `falcon3-7b-instruct-production` passes; samples/sec and TTFT consistent across repeated runs (~15.6-16.1 samples/sec, TTFT ~850-870ms).
- [x] Run on CI via Perf Benchmark too: https://github.com/tenstorrent/tt-xla/actions/runs/30505233947

### DIFF
--- a/tests/benchmark/test_vllm_benchmarks.py
+++ b/tests/benchmark/test_vllm_benchmarks.py
@@ -198,6 +198,40 @@ def _mistral_small_31_tp_config(model: str, batch_size: int):
     return cfg
 
 
+def _falcon3_7b_instruct_production_config():
+    # Mirrors the real single-chip (p150) production launch config from
+    # ~/scripts/model_servers/launch_falcon3_7b_instruct_uvicorn.sh (which
+    # mirrors workflows/model_specs/dev/cnn.yaml), not the minimal smoke-test
+    # defaults. Kept as the single test we maintain at production settings.
+    cfg = _config(
+        "tiiuae/Falcon3-7B-Instruct",
+        32,
+        gpu_memory_utilization=0.35,
+        optimization_level=1,
+        experimental_weight_dtype="bfp_bf8",
+        experimental_kv_cache_dtype="bfp_bf8",
+        enable_const_eval=True,
+        min_context_len=128,
+        prefill_chunk_size=1024,
+        # b1-prefill: serve prefills serially (small graph) when <=16 are
+        # pending instead of paying for a wasted-row b32 batch. Needs
+        # min_num_seqs < max_num_seqs (batch_size=32).
+        min_num_seqs=1,
+        prefill_batch_threshold=16,
+        # Cap prefill graphs at b16 instead of b32 (#5541): cuts prefill-trace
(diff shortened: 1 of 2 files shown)


vvukomanTT (approved):

tests/benchmark/test_vllm_benchmarks.py
```diff
@@ -198,6 +198,40 @@ def _mistral_small_31_tp_config(model: str, batch_size: int):
...
+        gpu_memory_utilization=0.35,
+        optimization_level=1,
+        experimental_weight_dtype="bfp_bf8",
+        experimental_kv_cache_dtype="bfp_bf8",
+        enable_const_eval=True,
+        min_context_len=128,
+        prefill_chunk_size=1024,
+        # b1-prefill: serve prefills serially (small graph) when <=16 are
+        # pending instead of paying for a wasted-row b32 batch. Needs
+        # min_num_seqs < max_num_seqs (batch_size=32).
```
mmanzoorTT, line 218: Where `max_num_seqs` is specified? is batch_size=32 a default option?

kmabeeTT: Not default, but it's the 2nd positional argument above to _config, set to 32.

mmanzoorTT: Thanks, I missed the positional argument.

mmanzoorTT (approved):

Issue #5867, opened 2026-08-03
[op-by-op] Remaining verification + data-fidelity gaps in the op-by-op analysis pipeline


ctr-lelanchelianTT:
Tracking issue for the verification and data-fidelity work left over from the op-by-op analysis setup, so it stops living in Slack threads.

### Context

The code-side items from the original hand-off are merged:

| Item | PR | Status |
|---|---|---|
| Continue processing when an IR fails to split | #5627 | merged |
| `extract-unique-ops` for global op dedup | #5652 | merged |
| Populate `frontend` on OpTest records | #5660 | merged |
| Wire `extract-unique-ops` into the processing pipeline | #5868 | open |

What remains is verification of what actually lands in the database and on the [ops-status dashboard](http://yyz2-forge-dash.local.tenstorrent.com:8000/ops-status), plus one data-fidelity consequence introduced by the unique-ops flow.

**Relationship to #5659:** that issue remains the source of truth for the dashboard-side gaps it already tracks (model count, empty `frontend`, coverage). Items 2–4 below restate them only so this issue gives one view of everything outstanding — close them there, not here. Items 1 and 5–7 are new.

---

### 1. Per-model attribution is lost in the unique-ops flow

**Blocking for the unique-ops rollout.** `op_by_op_test` derives an op's `origin_model` from the IR's parent directory name (`tests/op_by_op/op_by_op_test.py:152,186`). Once ops are extracted into per-op directories, that name becomes the op dir (`op_00042_stablehlo_add`), so the ops-status "models" column would no longer reflect the models an op came from.

The `unique-ops-manifest-*` artifact preserves the mapping per op: `op_name`, `op_dir`, `file`, `origin_models`, `origin_model_count`.

- [ ] Decide the fix: ingestion joins the manifest on `op_dir`, **or** `op_by_op_test` reads a sidecar and records `origin_model` correctly at record time
- [ ] Implement and confirm the dashboard's per-model view still resolves

### 2. Model count on the dashboard is inflated (~640 shown vs ~270 real) — tracked in #5659

Diagnosed earlier: the `models` field mixes two separators — `", "` between models within one op, and `" | "` across ops — so counting on `|` overcounts.

[...]

PR #5868, opened 2026-08-03
[op-by-op] Run only globally unique ops in the IR processing pipeline

ctr-lelanchelianTT:
### Ticket

Follow-up to #5652, which added `extract-unique-ops` but left the pipeline wiring open. Completes Nikola's item: *"extract unique ops across all model IRs, **and run only those**."*

### Problem description

`extract-unique-ops` landed in #5652 as a standalone subcommand, but `call-process-dumped-irs.yml` never called it — it still ran `dedupe` → `plan` → `materialize` over **whole model IR dumps**. So the pipeline kept compiling and executing the same op (identical shapes/attrs) once per model that contains it, across every processing job. Dedup only ever happened *within* a job, via `filter_and_deduplicate_ops`.

The wiring was deferred because of an ordering problem: extraction needs `op_by_op_infra` from the explorer wheel, but `plan` ran in `check_dumped_irs`, which runs on `ubuntu-latest` with no wheel — and the wheel build itself `needs: check_dumped_irs`.

### What's changed

**Job graph** — planning moves after the wheel build so extraction can precede it:

```
check_dumped_irs        (ubuntu)  detect ir-dumps-* artifacts only
  └── build_ttxla_explorer        build explorer wheel
        └── extract_unique_ops    NEW (n150 + wheel): dedupe → extract → plan
              └── process_dumped_irs  matrix over unique ops
```

- **`check_dumped_irs`** is now only an artifact-existence check. It no longer downloads IRs, dedupes, or plans, and no longer emits `ir_count` / `job_matrix`.
- **`extract_unique_ops`** (new) downloads the dumps once, runs model-level `dedupe`, extracts every op, deduplicates globally by `op_string`, then runs `plan` over the resulting unique-op tree. Emits `skip_ir_processing` / `ir_count` / `job_matrix`, and uploads two artifacts: the unique-op modules and the per-op manifest.
- **`process_dumped_irs`** takes its matrix from `extract_unique_ops` and downloads the **unique-op artifact** rather than re-downloading the full dump set and re-deduping per job. This resolves the existing `TODO: should each one download all irs dumps or can we do better?`.

**Planner** — `plan` gains `--seconds-per-model`:

Job sizing estimated runtime from IR file size (`SECONDS_PER_MB`). That is a fine proxy for whole-model dumps but wrong for single-op modules, which are ~1KB each while their cost is dominated by compile+execute. Left unchanged, the planner estimated ~1s per op and packed all 366 ops of the smoke set into one job. `--seconds-per-model` plans by op count instead; size-based estimation is unchanged when it is omitted (default `0`).

**`extract-unique-ops`** gains `--model-filter`:

The workflow's `model_filter` input matched on model-dir names in `plan`. After extraction the planner sees *op* dirs, so the filter has to be applied during extraction or it silently matches nothing. Also reports `matched_models` in the stats.

**New workflow input** — `seconds_per_unique_op` (default `30`). See the calibration note below.

### Verification (local, against real nightly IR dumps)

Two models from a real nightly dump (`ir-dumps-…-n150-…`), 54 matched IRs:

| Stage | Result |
|---|---|
| `extract-unique-ops` | 2110 ops → **366 unique (82.7% reduction)**, 0 extraction failures |
| `plan` over unique-op tree | 366/366 op dirs discovered, valid matrix, paths `unique_ops/op_XXXXX_<op>` |
| `plan --seconds-per-model 30`, 240min target | 10980s → 1 job |
| `plan --seconds-per-model 30`, 2000s target | 6 jobs, evenly split at 61 ops each |
| `materialize` job 0 | 366 op dirs → 61, 61 `.mlir` files |
| `op_by_op_test` over extracted layout | 4-op slice: all modules discovered, compiled and executed on n150, **PASSED** (test asserts 0 failed ops) |

Also: workflow YAML parses and the `needs`/`if` graph resolves; all three callers (`schedule-nightly`, `schedule-weekly`, `manual-test`) pass only pre-existing inputs, so the new input's default applies and no caller changes are needed; `black` clean.

The 82.7% figure is from two models only — it is a floor, not a projection. Cross-model dedup should improve substantially over the full ~148-model dump set, but the real number will come from the first CI run.

### Needs a dashboard-side follow-up 

`op_by_op_test` derives each op's `origin_model` from the IR's **parent directory name** (`op_by_op_test.py:152,186`). In the unique-ops layout that directory is the op dir, so recorded `origin_model` values become `op_00042_stablehlo_add` instead of the model the op came from. **Per-model attribution in the ops-status DB would be lost** unless ingestion joins against the manifest.

The `unique-ops-manifest-*` artifact preserves the mapping (`op_name`, `op_dir`, `file`, `origin_models`, `origin_model_count`) for exactly this reason. Options, in preference order:

1. Ingestion joins the manifest on `op_dir` to restore `origin_models` (no test changes).
2. Teach `op_by_op_test` to read a sidecar so `origin_model` is populated correctly at record time.

I did not pick one here because it depends on how the ingestion side wants to consume it. Happy to do either — flagging it before this changes what lands in the DB. Tracked as item 1 in #5867 along with the remaining verification work.

### Notes / follow-ups

- **`seconds_per_unique_op` default of 30 is a placeholder.** It only affects how ops are bucketed into jobs, not correctness, but it should be recalibrated from the first run's actual per-op durations.
- `extract_unique_ops` runs on an `n150` because the `op_by_op_infra` import is verified there today. Extraction itself only parses and splits MLIR — no device execution — so it could likely move to a CPU runner and free the device slot. Worth a follow-up once the flow is proven green.
- Best validated via `workflow_dispatch` against a known `dumped_irs_run_id` before it runs on a nightly.

### Checklist


.github/workflows/call-process-dumped-irs.yml
```diff
@@ -84,21 +92,90 @@ jobs:
...
+  # Extract every op from every model IR, deduplicate globally by op_string, and
+  # plan the processing jobs over the unique ops rather than over whole models.
+  # This runs on a device runner because op extraction needs op_by_op_infra from
+  # the explorer wheel, which is why planning happens here and not in
+  # check_dumped_irs (that job has no wheel).
+  extract_unique_ops:
+    needs: [ check_dumped_irs, build_ttxla_explorer ]
+    if: needs.check_dumped_irs.outputs.skip_ir_processing == 'false'
+
+    runs-on: ${{ !inputs.shared-runners && 'n150' || 'tt-ubuntu-2204-n150-stable' }}
```
nsumrakTT: Do we need hw machine just to deduplicate ops?

nsumrakTT (commented):
LGTM!
Could you please attach test run for modified workflow?

.github/workflows/call-process-dumped-irs.yml
ctr-lelanchelianTT: no, it does not. Fixed in 3961427.

Extraction is pure MLIR parsing (`MLIRModuleSplitter.split`); it never calls `_ensure_system_desc()`, and I confirmed zero device fds are opened. Moved the job to `tt-ubuntu-2204-large-stable` with a plain container. It still needs the explorer wheel for the ttmlir bindings, so the image and `needs: build_ttxla_explorer` stay — only `process_dumped_irs` holds a device now.

ctr-lelanchelianTT:
> LGTM! Could you please attach test run for modified workflow?

https://github.com/tenstorrent/tt-xla/actions/runs/30993073418

nsumrakTT (approved):

PR #5869, opened 2026-08-03
Add production configs for 4 vLLM benchmark models

### LINKED ISSUE #5404: Update vLLM Benchmarks to run high seq len + b32 + high gpu_mem_util

vkovacevicTT:
### Ticket
Closes #5404 

### What's changed
Added production config vLLM benchmark variants for Llama3.2-3B, Llama3.1-8B, Qwen3-4B and Qwen3-8B.

### Checklist
- [x] [qb2 benchmark run](https://github.com/tenstorrent/tt-xla/actions/runs/30644228196)

### DIFF
--- a/.github/workflows/perf-bench-matrix.json
+++ b/.github/workflows/perf-bench-matrix.json
@@ -636,6 +636,13 @@
         "skip-device-perf": true
       },
       {
+        "name": "vllm_llama_3_2_3b_instruct_production",
+        "pytest": "tests/benchmark/test_vllm_benchmarks.py::test_vllm_benchmark[llama-3.2-3b-instruct-production]",
+        "vllm": true,
+        "skip-device-perf": true,
+        "runs-on": "p150-perf"
+      },
+      {
         "name": "vllm_llama_3_2_1b_instruct",
         "pytest": "tests/benchmark/test_vllm_benchmarks.py::test_vllm_benchmark[llama-3.2-1b]",
         "vllm": true,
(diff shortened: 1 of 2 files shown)


vvukomanTT (approved):

nsumrakTT (approved):

mmanzoorTT (approved):

kmabeeTT (approved):
Thanks for change! I'm okay with this, but ideally `max_prefill_num_seqs=16` should be added to these 4 models like the Falcon3-7B-Intsruct I added late last week, it will be used in production and it will also somewhat reduce compile time by preparing b16 prefill graphs instead of b32 prefill graphs.

PR #5870, opened 2026-08-03
Release device weights before EmitPy run to avoid double residency

amilovanovicTT:
### Ticket
N/A

### Problem description
EmitPy run of `qwen_3/embedding/Embedding_8B` ran out of DRAM on n150.
The issue is that the runtime keeps preprocessed weights in a process-global const-eval cache and evicts an entry only when one of the input tensors that produced it is destroyed. Since the model stays on device, the flatbuffer reference run's weights remain resident when the generated EmitPy program streams its own copy, roughly doubling peak DRAM usage. 
A device with more DRAM (p150) masks the problem.

### What's changed
Move the model off device before the EmitPy run. That destroys the input tensors
which produced the flatbuffer run's const-eval entries, so the runtime evicts
them and frees their DRAM.

The EmitPy run then puts the model back: `DeviceRunner.run_on_device` re-places
the workload on device on every invocation.

### Checklist
- [Before the fix: qwen_3/embedding/Embedding_8B failed](https://github.com/tenstorrent/tt-xla/actions/runs/30834714297/job/91757210908)
- [After the fix: qwen_3/embedding/Embedding_8B passed](https://github.com/tenstorrent/tt-xla/actions/runs/30834626200)

### DIFF
--- a/tests/infra/testers/single_chip/model/torch_model_tester.py
+++ b/tests/infra/testers/single_chip/model/torch_model_tester.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import collections
+import gc
 import os
 import shutil
 import tempfile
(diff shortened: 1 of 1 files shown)


tests/infra/testers/single_chip/model/torch_model_tester.py
```diff
@@ -360,6 +361,27 @@ def _test_training(self) -> Tuple[ComparisonResult, ...]:
...
+        model off device destroys those tensors, so the cache drains through its
+        intended mechanism.
+
+        Safe to call before a subsequent device run: ``DeviceRunner.run_on_device``
+        re-places the workload on device on every invocation.
+        """
+        model = getattr(self._workload, "model", None)
+        if model is None or not hasattr(model, "to"):
+            return
+        model.to("cpu")
```
jameszianxuTT, line 380: I thought this was an out of place operation, so I'm not sure if calling model.to(cpu) does anything here. Is it possible to just drop the reference to your _workload model once you don't need it anymore and then gc.collect (or, is the original model already at zero refcount  and just awaiting GC)? What else is holding onto the self._workload.model at this point? (if nothing is holding it, then just a gc.collect() would work.)

PR #5872, opened 2026-08-04
[vLLM] Shard KV cache heads under DP+TP (issue #5796)

### LINKED ISSUE #5796: [vLLM] DP+TP: KV cache is replicated on every chip instead of sharded
Followup from code inspection during https://github.com/tenstorrent/tt-xla/pull/5731

## Problem

Under DP+TP the KV cache is left un-annotated, so SPMD replicates it across **both** mesh axes — every chip holds the entire block pool with the full, un-sharded `num_kv_heads`. Pure TP shards heads on the model axis; DP+TP gets nothing.

`initialize_kv_cache` (`integrations/vllm_plugin/vllm_tt/model_runner.py:3455`):

```python
if self.parallel_mode == ParallelismMode.DATA_TENSOR_PARALLEL:
    # DP+TP: leave the KV cache un-annotated (replicated under SPMD);
    # each device writes its own K/V slice via paged_update_cache. The
    # TP-only spec puts block_size on the DP axis and fails
    # ttir.paged_update_cache. Tracked as a follow-up.
    pass
elif self.enable_tensor_parallel:
    safe_mark_sharding(cache, self.mesh, (None, "model", None, None))
```

Correctness holds via divergent writes, but aggregate KV capacity across the mesh is only ~one chip's worth — on `test_data_tensor_parallel_generation_gemma4_31b` (mesh `[8, 4]`) the whole pool is replicated on all 32 chips. Introduced in #4947; #5731 codified it rather than causing it (`kv_cache_shard_factor()` returns `1` for DP+TP).

## Suggested fix

- **TP side — heads on `"model"`:** reuse the pure-TP spec `(None, "model", None, None)`, DP axis replicated. Recovers `tp_size`× (4×) on its own. This is the config reported to fail `ttir.paged_update_cache`, so re-capture that error first — the recorded reason doesn't match the current layout. Then drop the DP+TP case in `kv_cache_shard_factor()` (`vllm_distributed_utils.py:83`) so the #5731 budget inflation applies.

- **DP side — blocks on `"batch"`:** spec becomes `("batch", "model", None, None)`, for a further `dp_size`×. Deferrable to a follow-up — each replica would need to own a disjoint slice of the block pool, which the scheduler doesn't do today.

alinakhanTT:
## Ticket
Partially closes #5796

## Summary
Recover DP+TP KV capacity. vLLM sizes the block pool as if the cache were un-sharded, so `kv_cache_shard_factor()` returning `1` for DP+TP meant we allocated `tp_size`× fewer blocks than the mesh can hold. This returns `mesh["model"]`, and annotates the cache so that budget is safe to apply.

Note for reviewers: the heads were *already* head-sharded in the graph before this PR, `paged_fill_cache`'s head factor is `kPassThrough` linking cache dim1 ↔ fill_value dim1, so Shardy propagates the TP sharding from the QKV
projections onto an unannotated cache. Sharding correctness was never the issue, the budget math is.

## Why the annotation is still needed
Propagation shards the *operand*, not the *parameter*. Measured on device (Qwen3-0.6B, dp=2 tp=4), by entry-param `local_shape`, what a chip holds:

|        | prefill                            | decode |
|--------|------------------------------------|--------|
| before | `28x8` + 56 `ttir.mesh_partition`  | `28x2` |
| after  | `28x2`, 0 `mesh_partition`         | `28x2` |

The prefill graph takes the cache full-width and slices it (56 = 28 layers × K,V). Inflating the budget while the parameter is still full width would give 4× the blocks *and* still materialize a full-width buffer at prefill entry,
so the annotation and the multiplier are only correct together.

## What's changed
- `model_runner.py`: mark heads `(None, "model", None, None)` for TP-only and DP+TP alike, deduped by tensor identity for cross-layer sharing; MLA excluded
- `vllm_distributed_utils.py`: `kv_cache_shard_factor()` returns `mesh["model"]` for DP+TP, and takes the layer specs so MLA returns 1
- `worker.py`: pass the specs when inflating the KV budget
- `distributed/tpu_communicator.py`, `platform.py`: vLLM 0.26.0 compat
- tests: device TTIR regression + unit coverage incl. MLA

## MLA
The first revision hoisted the `num_kv_heads % tp_size` assert above the MLA branch, so DeepSeek-V2-Lite (`num_kv_heads=1`) failed engine init under TP, caught in nightly, not the push set. MLA holds a single *replicated* latent cache, so it must neither be asserted on nor counted as sharded. Fixed in 321e65254.

## Testing
- Unit + CPU green
- Device (dp=2, tp=4): cache `28x2x32x128` in every graph, 0 `mesh_partition`
- nightly TP after the MLA fix: <fill in>

Blocks sharding follow-up: #5910.

## Test timeout change (unrelated to KV sharding)
This branch's nightly vLLM job was failing on two timeouts, not on behavior. The autouse _test_timeout fixture arms signal.alarm(2 × recorded_duration), but the alarm also covers the setup of every fixture ordered after it, while .test_durations records only the call phase. So test_responses_api_basic[message_input] got a 2 s budget on a path whose shared-server fixture alone costs ~120 s, and test_opt_generation_large_batch[batch8] got 149 s against a measured 149.47 s.
Fix: floor the budget at 300 s and refresh the stale batch8 entry (74.46 -> 149.47). No test's budget is lowered , anything long-running keeps its 2×. Both entries are unchanged from main, so this branch didn't move them; the flake was pre-existing and re-running can't clear it (running batch8 in isolation makes it worse, since it then pays the job's cold first-compile).


ssaliceTT:
On DP+TP today I'm already seeing the KV cache head-sharded in the graph, before this PR. 
Ex. Qwen3-8B on a 2×4 mesh gives paged_fill_cache cache operand 682x2x32x128 (8 KV heads / tp 4 = 2) without any explicit mark_sharding on the cache.

I think Shardy is propagating it from `getPagedAttentionShardingRule`, whose head factor links cache dim 1 to fill_value dim 1, so the cache inherits the activations' head sharding. Could you double-check whether it's genuinely unsharded today?

If it already is, the `mark_sharding` here is would still be worth doing as it makes the inferred property explicit. However, the actual capacity recovery comes entirely from fixing `kv_cache_shard_factor()` for DP+TP since vLLM was sizing the pool as if the cache were unsharded, so we were allocating tp_size× fewer blocks than the the pool allocated. Might be nice adjusting the PR desc so reviewers focus on the budget math rather than the sharding correctness.

ssaliceTT:
> I think Shardy is propagating it from `getPagedAttentionShardingRule`, whose head factor links cache dim 1 to fill_value dim 1, so the cache inherits the activations' head sharding. Could you double-check whether it's genuinely unsharded today?

So it seems decode graph was getting the kv cache presharded but not prefill, and the explicit sharding fixes it. Thanks for updating the PR!

alinakhanTT:
Nightly green: https://github.com/tenstorrent/tt-xla/actions/runs/31425250011

.test_durations
```diff
@@ -3613,7 +3613,7 @@
     "tests/integrations/vllm_plugin/generative/test_max_prefill_num_seqs.py::test_max_prefill_num_seqs_uses_capped_prefill_shape": 87.23,
     "tests/integrations/vllm_plugin/generative/test_opt_generation.py::test_opt_generation_large_batch[batch16]": 70.26,
     "tests/integrations/vllm_plugin/generative/test_opt_generation.py::test_opt_generation_large_batch[batch32]": 64.13,
-    "tests/integrations/vllm_plugin/generative/test_opt_generation.py::test_opt_generation_large_batch[batch8]": 74.46,
+    "tests/integrations/vllm_plugin/generative/test_opt_generation.py::test_opt_generation_large_batch[batch8]": 149.47,
```
kmabeeTT, line 3617: Can you add entry for the newly added push test here too? Was going to ask for how long it ran, but checked previous onPR and fouind it is quick, so safe to keep, but having it here will avoid 3600s default timout and pytest splitting to work better.

```
"tests/integrations/vllm_plugin/generative/test_kv_cache_sharding.py::test_kv_cache_head_sharded_under_dp_tp": 117.04,
```

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
```diff
@@ -111,15 +112,20 @@ def kv_cache_shard_factor(runner) -> int:
...
+    whose single latent cache is replicated rather than head-sharded; pass it
+    wherever the specs are known.
     """
     if not runner.enable_tensor_parallel:
         return 1
-    if runner.parallel_mode == ParallelismMode.DATA_TENSOR_PARALLEL:
-        # DP+TP leaves the cache replicated, so per-chip usage already equals
-        # the full budget — no reconciliation to do. Sharding it properly
-        # (heads on "model", blocks on "batch") is tracked in #5796.
+    if kv_cache_spec and any(
```
kmabeeTT, line 126: Dropping the DP+TP early return activates a dormant assert: `initialize_kv_cache` asserts `num_kv_heads % tp_size == 0` only when `tp_size > 1`, which under DP+TP was always 1. A model with indivisible KV heads now goes from silently-replicated-but-running to a hard engine-init failure. Right behavior — just confirm nothing in the matrix hits it (gemma-4-31B on `[8, 4]`), and consider saying "DP+TP now head-shards too" in the assert message.

Separately: make `kv_cache_spec` required instead of defaulting to `None`. Your unit test documents the footgun (`kv_cache_shard_factor(runner) == 4` for MLA), and a wrong answer here over-allocates the pool and OOMs — the exact failure this PR prevents. The worker has the specs; the `model_runner` call site can pass `None` explicitly.

kmabeeTT (approved):
Looks good, thx for rebase + conflict drop and updating test_durations. Left a couple small comments.

.test_durations
alinakhanTT: Done!  also dropped the test's own subprocess timeout from 1800s to 240s so it trips before that alarm, otherwise a hang gets SIGALRM'd and the child's captured stdout/stderr is tossed

integrations/vllm_plugin/vllm_tt/vllm_distributed_utils.py
alinakhanTT: Done! nothing hits the assert either

nsumrakTT (approved):

PR #5875, opened 2026-08-04
Upgrade Jax to 0.11

sgligorijevicTT:
### Ticket
No issue, but there is an external request to support later versions of Jax

### Problem description
Old versions of Jax cause compatibility headaches for downstream users. Our plugin crashed when driven by newer versions of Jax.

### What's changed
Bumped requirements
Implemented a few more features of the PJRT API to allow newer versions of Jax to use the plugin, namely the API for iterating over payload errors(which we don't model), the vtable on MemoryInstance and logical addressable devices.

Beware: This PR is not complete. Our fork of TorchXLA should be modified in lockstep to depend on the same version of jax that we upgraded to.

### Checklist


sgligorijevicTT:
Companion PR for our torchxla fork https://github.com/tenstorrent/pytorch-xla/pull/41 , I won't merge that until I get some eyes on it, and build is failing here until torchxla is upgraded

jameszianxuTT:
btw, in case you didn't know, you can run tt-xla CI tests against a custom torch-xla commit, ref https://github.com/tenstorrent/tt-xla/pull/4177. You just need to build torch-xla via a custom workflow inside tt-xla and then set the run-id in the call-test manual dispatch, but more instructions are in the PR

sgligorijevicTT:
@jameszianxuTT Thanks for the tip James, I didn't know about this. It definitely came in useful, I just caught a build breakage with that that didn't show up on another CI flow.

Issue #5659, opened 2026-07-16
[op-by-op] ops-status dashboard data gaps (model count, empty frontend, coverage)


ctr-lelanchelianTT:
Verification of the ops-status dashboard (http://yyz2-forge-dash.local.tenstorrent.com:8000/ops-status) against the processing output. Snapshot: 24,431 unique-op rows, 61 op types, 53,503 executions (99.0% pass / 509 fail), 30 job-ids.

### Gap 1 (headline): analysed-model count inflated ~2.4x
Dashboard shows **640 models but only 270 are real**. The `models` field mixes two delimiters: `", "` (models sharing one op) and `" | "` (across ops), so any count that splits only on `" | "` overcounts. 942 rows carry comma-mashed model tokens.
- Root cause: `tools/op-by-op-infra/op_by_op_infra/utils.py:418` -> `origin_model = ", ".join(self.origin_model)` becomes `OpTest.model_name`; the DB then joins rows with `" | "`.
- Fix: split on both `","` and `" | "` in the ingestion/dashboard aggregation (tt-forge-dashboard side), or destructure origin models at the source.

### Gap 2: `frontend` empty on every row — FIXED
`op_by_op_test.py` never passed `frontend=` to `execute_extracted_ops`, so `OpTest.frontend` was `None` on all 30,347 pre-fix rows. Fixed by passing `frontend="tt-xla"` (matches `tests/runner/test_models.py:749`); verified on device (12/12 entries now `tt-xla`).

### Gap 3: model names contain spaces (cosmetic)
36 model names carry embedded spaces from HF ids (e.g. `jax_Qwen 3_0_6B_...`, `pytorch_Deformable DETR_...`) — breaks exact-match joins/filters.

### Coverage gap (related, FIXED)
In the run that populated the DB, 20 of 31 processing jobs produced 0 ops — all aborted at the first unparseable IR. Fixed by #5627 (merged); coverage should increase next weekly run.

### Failure hotspots to triage (genuine op failures, not pipeline bugs)
By op type: `tenstorrent.group_norm` 45.5% (51/112), `tenstorrent.layer_norm` 2.7%, `stablehlo.convolution` 2.2%, `tenstorrent.rms_norm` 1.1%.
By model: `Mamba_370M` 66%, `MobileNetV1_..._0.75_192` 69%, the DLA/timm family, `XGLM_1.7b`/`XGLM_564M`.

### Follow-ups
[...]

PR #5890, opened 2026-08-05
[runner] Sanitize IR-dump directory names so model names are joinable

ctr-lelanchelianTT:
### Ticket

Part of the op-by-op data-quality work — the naming gap logged on TEN-919 as the blocker for a clean expected-vs-ingested model set-diff. Related: #5659, #5867.

### Problem description

Model names come from `model_info.name` and are derived from HuggingFace ids, so many contain spaces. In a real nightly dump, **42 of 148** model directories look like this:

```
pytorch_Qwen 2.5 Coder_32B_Instruct_nlp_causal_lm_huggingface
pytorch_MaskFormer Swin-B_Swin_Base_Coco_cv_image_seg_huggingface
pytorch_Panoptic Segmentation_ResNet50_Backbone_1x_COCO_cv_panoptic_seg_detectron2
```

That directory name is not cosmetic. Op-by-op recovers the model name **from the directory name** — `match_and_extract_model_name` takes the path component before `irs` — so the spaces flow straight into the ops-status database. **67 of the 600** models currently in the DB carry embedded spaces, which breaks exact-match joins and filters against configs that use the underscored form.

### What's changed

Adds `sanitize_ir_dump_dir_name()` to `tests/runner/test_utils.py`: collapses each run of whitespace to a single `_`, and defensively replaces path separators and other path-hostile characters (`/ \ : * ? " < > |`). Dots and dashes are preserved, so legitimate names like `Swin-B` and `2.5` are untouched.

Applied at **both** sites that build the dump path — they have to agree, or the read side stops finding what the dump wrote:

- `test_models.py:120` — writes `collected_irs/<name>/`
- `test_models.py:730` — reads it back for the op-by-op subprocess test

The separator replacement is defensive rather than observed: no current name contains one, but a `/` would silently create a nested directory and break the model-name recovery above.

### Verification

Against all **148** model directory names from a real nightly dump (run 30724942468):

| Check | Result |
|---|---|
| Names changed | **42** — exactly the ones with spaces |
| Names returned unchanged | **106** — the sanitizer is a no-op for already-clean names, so this does not churn existing DB values |
| Collisions after sanitizing | **none** |
| Whitespace remaining | none |
| Round-trip through `match_and_extract_model_name` | recovered name == directory name for the sanitized forms |

Edge cases: `'  padded  '` → `'padded'`, `'multi   space'` → `'multi_space'`, `'a/b'` → `'a_b'`, `'already_clean-2.5'` unchanged.

`py_compile` clean, `black` clean.

### Note for the ingestion side

Model names recorded from the **next** dump onward will use the underscored form, so for the 67 affected models the DB will show the sanitized name alongside historical spaced rows. That is the intended outcome — the spaced values cannot be joined today — but worth knowing if anything keys on the old strings. cc @vladimirmilosevicTT @nsumrakTT

### Checklist


nsumrakTT (approved):

ctr-lelanchelianTT:
CI note: the only failing job on this PR was `test push and training (qb2-blackhole, run_torch_training)`, and the failure is unrelated to this change:

```
tests/torch/training/test_moe.py::test_gpt_oss_moe_multichip_backward
RuntimeError: TT_FATAL @ .../tt_metal/fabric/topology_mapper.cpp:546: mapping_result.success
Graph specified in MGD could not fit in the discovered physical topology.
```

That is a mesh/fabric topology-discovery failure on the multichip Blackhole runner. This PR touches only `tests/runner/test_utils.py` (adds `sanitize_ir_dump_dir_name`) and its two call sites in `tests/runner/test_models.py`; nothing in it is reachable from `tests/torch/training/test_moe.py`. I have re-run the failed jobs.

Ready for review whenever one of you has a moment — it is a small, self-contained change (one helper + two call sites that must agree, since op-by-op recovers the model name from the dump directory name).

ctr-lelanchelianTT:
Update: re-run is green. \`test_gpt_oss_moe_multichip_backward\` **passed** in 202s (job 92367373494), confirming the MGD topology failure was a runner flake, and \`check-all-green\` is SUCCESS.

The one remaining red mark, \`TT-XLA model-test-push\` ("34 tests run, 33 passed, 1 failed"), is a reporting artifact rather than a real failure: the test-reporter aggregates every \`test-reports-*\` artifact in the run, and attempt 1's failed artifact (\`...-1.7-qb2-blackhole--92269749376\`) is still present next to attempt 2's passing one (\`...-2.7-...-92367373494\`). It is not a required check.

So this is ready to merge on approval.

sgligorijevicTT (approved):

PR #5893, opened 2026-08-05
[vLLM] Fix DP row corruption: batch-axis sharding, dp_size-aligned buckets, active-row chunk offset

### LINKED ISSUE #5892: [vLLM] DP generation corrupts every replica but replica 0, and the coherence check passes it
## Bug

Under data parallelism, replicas other than replica 0 can return garbage, and the test that covers it passes anyway.

`test_data_parallel_chunked_prefill_n300` sends the same prompt twice with greedy decoding, so both replicas should say the same thing. On `main`:

```
replica 0:  The history of computing spans many centuries. Early mechanical calculators gave way ...
replica 1: BlockBlockBlockBlockBlockBlockBlockBlockBlockBlock ...
```

It has been green since #5692, because `assert_output_coherent` counts `BlockBlockBlock` as a single word.

Four things are wrong:

1. DP-only never shards the batch. The axis is picked via `use_2d_mesh`, which is false for a `(dp, 1)` mesh, so it shards on size-1 `"model"` and nothing happens.
2. Prefix caching is unsafe under DP. The block pool is global but each replica has its own KV cache, so a cross-replica hit reads zeros.
3. Request-count buckets are not multiples of `dp_size`, so the batch dim ends up indivisible by the mesh axis and gets replicated.
4. Chunked prefill reads its prefix offset off row 0, so an inactive row's offset can land on a row starting its first chunk.

## Steps to repro

```
pytest -svv tests/integrations/vllm_plugin/generative/test_data_parallel_generation.py \
  -k test_data_parallel_chunked_prefill_n300
```

Needs an n300 for a real dp=2. Mesh size follows the device count, so dp=2 cannot be forced on a 4 or 8 chip box. Read the printed `replica` lines, not the pass/fail.

## Expected behavior

Same prompt, same output on every replica. Garbage should fail the test.

## Environment information

- Ubuntu 22.04, n300 CI runner
- Python 3.12, vLLM 0.26.0
- tt-xla `10b111f38`, tt-mlir pin `724d2ff7`

alinakhanTT:
### Ticket
closes #5892

### Problem description

Under data parallelism, replicas other than replica 0 can return garbage, and the test covering it passes anyway.

`test_data_parallel_chunked_prefill_n300` sends the same prompt twice with greedy decoding, so both replicas should say the same thing. On `main` replica 0 is correct and replica 1 returns `BlockBlockBlock...`. It stays green because `assert_output_coherent` counts that as a single word.

Four things are wrong. See #5892 for detail.

1. DP-only never shards the batch: the axis is picked via `use_2d_mesh`, false for a `(dp, 1)` mesh, so it shards on size-1 `"model"` and nothing happens.
2. Prefix caching is unsafe under DP: the block pool is global but each replica has its own KV cache, so a cross-replica hit reads zeros.
3. Request-count buckets are not multiples of `dp_size`, so the batch dim ends up indivisible by the mesh axis and gets replicated.
4. Chunked prefill reads its prefix offset off row 0, so an inactive row's offset can land on a row starting its first chunk.

### What's changed

- **`model_runner.py`**: pick the batch axis from the mesh shape, since only pure 1D-TP lacks a batch axis. Align the prefill bounds up and the SMEM-capped counts down to a multiple of `dp_size`, and raise if a capped count rounds to zero. Take the chunked-prefill offset from rows with scheduled tokens, and raise on mixed stages instead of applying one row's offset to another.
- **`platform.py`**: disable prefix caching under DP. It cannot work as built, so this fails safe rather than serving wrong tokens.
- **`vllm_distributed_utils.py`**: warn when `safe_mark_sharding` gets a size-1 axis. That no-op is how defect 1 stayed invisible. Deduped, since pure-TP marks a size-1 `"batch"` axis legitimately.
- **`generative/conftest.py`**: teach `assert_output_coherent` to catch repeated tokens, punctuation soup and non-Latin garbage, all before the short-output early return. Checked against seven real garbage outputs and against normal generations with smart quotes and lists.

### Verification

On an n300 at dp=2:

- `test_data_parallel_chunked_prefill_n300` passes with both replicas byte-identical. Reverting only the offset commit brings `BlockBlockBlock...` back, so that fix is doing the work.
- The dp=2 matrix from bring-up is clean in all four shapes, including two rows per replica, which used to corrupt rows 1 through 3.

Defects 2 and 3 were validated on an 8-chip llmbox: dp=8 went from 7/8 garbage rows to 8/8 correct, and 4 requests at dp=8 went from a replication warning on every batch tensor to none.

Wider coverage on the rebased branch:

- dp=8 on an llmbox: `test_data_parallel_chunked_prefill_n300` passes, as do the llmbox padding and tight tests.
- DP+TP: `test_data_tensor_parallel_generation_push` passes in 118s against a 567s budget.
- Speculative decode, all four configs pass (single device, TP, DP, DP+TP). Worth checking because the chunked-prefill offset is shared with the spec-decode block-alignment path.

### Notes

Prefix caching under DP is disabled, not fixed. Making it work needs per-replica block accounting.

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -2244,8 +2244,10 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
             ParallelismMode.DATA_TENSOR_PARALLEL,
         ):
             return
-        # 2D mesh: batch dim -> "batch"; pure 1D-TP: "batch" is size 1, use "model".
-        batch_axis = "batch" if self.use_2d_mesh else "model"
+        # Only pure 1D-TP (mesh (1, N)) lacks a batch axis. Keying off
+        # use_2d_mesh instead picked size-1 "model" for DP-only (dp, 1), a
+        # silent no-op that left activations replicated.
+        batch_axis = "model" if self.mesh.mesh_shape[0] == 1 else "batch"
         if input_ids is not None:
             safe_mark_sharding(input_ids, self.mesh, (batch_axis, None))
         if inputs_embeds is not None:
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -485,6 +485,24 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
             if self.tt_config.max_prefill_num_seqs is not None
             else self.max_num_reqs
         )
+        # Resolved from the *unpadded* max_num_seqs in platform.py, so under DP
+        # they can land on a non-multiple of dp_size. Align them up.
+        if self.dp_size > 1:
+            for _name in ("min_num_reqs", "max_prefill_num_reqs"):
+                _old = getattr(self, _name)
+                _new = min(
+                    _align_num_reqs_for_dp(_old, self.dp_size, round_up=True),
+                    self.max_num_reqs,
+                )
+                if _new != _old:
+                    logger.warning(
+                        "Data parallel requires %s divisible by dp_size; "
+                        "adjusting from %d to %d.",
+                        _name,
+                        _old,
+                        _new,
+                    )
+                    setattr(self, _name, _new)
         if scheduler_config.max_num_batched_tokens < self.tt_config.min_context_len:
             logger.warning(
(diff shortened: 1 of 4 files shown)


AleksKnezevic (approved):

alinakhanTT:
All tests passing on nightly https://github.com/tenstorrent/tt-xla/actions/runs/31106845355

Issue #5897, opened 2026-08-06
[vLLM] Device hang at max_model_len=128K in`ttnn.paged_fill_cache` op with a wide page table


kmabeeTT:
> Followup from max_model_len=128K hang discussed in https://github.com/tenstorrent/tt-xla/issues/5727#issuecomment-5184131910

## Problem

vLLM benchmarks hang the device during warmup at `max_model_len=131072`. 96K and below pass. Reproduces on latest main (`10b111f38`), on a **single chip** with Llama-3.1-8B and on **TP=4** with gemma-4-31B-it. The device is left unrecoverable:

```
TT_THROW: TIMEOUT: device timeout, potential hang detected, the device is unrecoverable
  @ tt_metal/impl/dispatch/system_memory_manager.cpp
```

The hanging op is **`ttnn.paged_fill_cache`** — 10/10 attributed hangs (method below).

## Repro

**tt-xla `10b111f38`** (main, 2026-08-04 — "[vLLM] Make ngram speculative decode work on device (#5836)"), tt-mlir `724d2ff7b`, vllm 0.26.0. QB2: 2x Blackhole p300c (4 ASICs). Stock tree, no local changes beyond the debug cherry-pick noted below.

Cheapest: single chip, **1 decoder layer**, ~3 min to hang.

```bash
export TT_VISIBLE_DEVICES=0
export TT_MESH_GRAPH_DESC_PATH=<tt-metal>/tt_metal/fabric/mesh_graph_descriptors/p150_mesh_graph_descriptor.textproto

TT_BENCHMARK_GMU=0.35 TT_BENCHMARK_MAX_MODEL_LEN=131072 TT_BENCHMARK_BATCH_SIZE=32 \
TT_BENCHMARK_PREFILL_CHUNK_SIZE=1024 TT_BENCHMARK_WEIGHT_DTYPE=bfp_bf8 \
TT_BENCHMARK_KV_CACHE_DTYPE=bfp_bf8 TT_BENCHMARK_NUM_HIDDEN_LAYERS=1 \
TT_METAL_OPERATION_TIMEOUT_SECONDS=60 \
pytest -svv "tests/benchmark/test_vllm_benchmarks.py::test_vllm_benchmark[llama-3.1-8b]"
```

Same command at `65536` / `98304` passes.

[...]

PR #5901, opened 2026-08-06
[vLLM] Narrow the fill page table to the columns paged_fill_cache reads

sshonTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/5897)

### Problem description
Gemma4 31b benchmark test with 128k context hangs on qb2 device. The kernel fetches the row with a single `noc_async_read` of the entire stick, and a large single read from interleaved DRAM intermittently never completes, wedging the device in `async_read_barrier()`. The root cause belongs in tt-metal.

### What's changed
`paged_fill_cache` indexes at most `cdiv(step_tokens, block_size)` page-table entries -- the bound its own validator enforces -- but we passed it the read table, sized for the whole context: 4096 entries (16KB per row) at max_model_len=131072, where 128B does the job.

Allocate and populate the fill table at that width, rounded up to the 8-entry stick alignment this file already applies elsewhere. Prefix caching is unaffected -- `_prepare_inputs` already rolls each row so the chunk's blocks lead. The read table is untouched, and sliding-window groups were never exposed (their ring rows are already small).

The narrowed buffer can no longer alias the read table, so both branches now always copy into a dedicated one, through a helper that asserts the narrowing still spans the step. That buys back tens of MB of page-table reads the kernel no longer issues, since every core re-reads the row on every layer's fill.

Unit tests cover the width helper and the guard; the buffer-keying test checks that the allocated widths hold the bound.

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/model_runner.py
+++ b/integrations/vllm_plugin/vllm_tt/model_runner.py
@@ -1212,6 +1212,60 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
 
         return tuple(tasks)
 
+    def _fill_page_table_width(self, group_idx: int, read_width: int) -> int:
+        """Columns the fill page table needs, far fewer than the read table.
+
+        paged_fill_cache only indexes ``cdiv(step_tokens, block_size)`` entries --
+        the bound its validator enforces -- but reads the whole row with a single
+        noc_async_read, and a large single read from interleaved DRAM can hang the
+        device. The roll in _prepare_inputs puts the chunk's blocks up front, so the
+        leading columns are the ones it wants.
+
+        The bound is the per-step token budget, not the context length: without
+        chunked prefill the width comes back at the read table's.
+
+        Guarded here rather than at each caller so allocation and both warmup
+        graph builders fail on the host at startup, not on the device.
+        """
+        block_size = (
+            self._group_block_sizes[group_idx]
+            if group_idx < len(self._group_block_sizes)
+            else self.block_size
+        )
+        # Round to 8 entries: a ttnn page-table stick must be 32B-aligned, the
+        # same rule _chunked_sdpa_active checks against max_num_blocks_per_req.
+        width = cdiv(cdiv(self.max_num_tokens, block_size), 8) * 8
+        fill_width = min(width, read_width)
+        # Do not relax: paged_fill_cache's page-table lookup is unbounded, so a
+        # sliding ring shorter than the chunk hangs or silently corrupts KV.
+        # https://github.com/tenstorrent/tt-xla/issues/5911
+        assert fill_width * block_size >= self.max_num_tokens, (
+            f"group {group_idx} fill page table ({fill_width} cols x block_size "
+            f"{block_size}) cannot cover a {self.max_num_tokens}-token step"
+        )
+        return fill_width
+
+    def _copy_narrow_fill_page_table(
+        self,
+        dev_buf: torch.Tensor,
+        host_table: torch.Tensor,
+        block_size: int,
+        step_tokens: int,
+        group_idx: int,
+    ) -> None:
+        """Stage the leading columns of ``host_table`` into the narrow fill buffer.
+
+        Narrower than the read table (see ``_fill_page_table_width``), so this
+        always truncates; assert the truncation still spans the step.
+        """
+        fill_width = dev_buf.shape[1]
+        # Per-step backstop for the invariant _fill_page_table_width pins.
+        assert fill_width * block_size >= step_tokens, (
+            f"group {group_idx} fill page table ({fill_width} cols x block_size "
+            f"{block_size}) cannot cover this step's {step_tokens} tokens"
+        )
+        dev_buf.copy_(host_table[:, :fill_width])
+
     def _alloc_page_table_device_buffers(self, num_groups: int) -> None:
         """(Re)allocate per-group page_table / fill_page_table device buffers.
 
@@ -1240,6 +1294,11 @@ class TTModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                 )
             return base_width
 
+        def _fill_pt_width(group_idx, base_width):
+            return self._fill_page_table_width(
+                group_idx, _pt_width(group_idx, base_width)
+            )
+
         self._page_table_dev_max = {
             g: {
                 nr: _alloc_dev((nr, _pt_width(g, self.max_num_blocks_per_req)), dt)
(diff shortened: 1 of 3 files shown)


sshonTT:
vllm nightly pass : https://github.com/tenstorrent/tt-xla/actions/runs/31082633533

sshonTT:
perf regression test with filtering `production` and `gemma4`. [main](https://github.com/tenstorrent/tt-xla/actions/runs/31107508076) , [this pr](https://github.com/tenstorrent/tt-xla/actions/runs/31107292080)

| Test | samples/sec (main → PR) | TTFT ms (main → PR) |
| :--- | :--- | :--- |
| **llama-3.2-3b-instruct-production** | 29.11 → 29.31 (+0.7%) | 642.0 → 640.4 |
| **llama-3.1-8b-instruct-production** | 17.99 → 18.41 (+2.3%) | 870.4 → 867.0 |
| **qwen3-4b-production** | 24.07 → 24.16 (+0.4%) | 729.7 → 728.6 |
| **qwen3-8b-production** | 17.17 → 16.78 (−2.3%) | 939.8 → 935.4 |
| **falcon3-7b-instruct-production** | 17.60 → 17.67 (+0.4%) | 870.5 → 871.4 |
| **gemma4-31b-it-tp** | 7.085 → 7.064 (−0.3%) | 919.8 → 921.4 |

+-2% seems noise, so I'd like to say no regression.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1212,6 +1212,45 @@ def get_supported_tasks(self) -> tuple[SupportedTask, ...]:
...
+        step_tokens: int,
+        group_idx: int,
+    ) -> None:
+        """Stage the leading columns of ``host_table`` into the narrow fill buffer.
+
+        Narrower than the read table (see ``_fill_page_table_width``), so this
+        always truncates; assert the truncation still spans the step.
+        """
+        fill_width = dev_buf.shape[1]
+        assert fill_width * block_size >= step_tokens, (
```
kmabeeTT, line 1263: Worth flagging that this is the only capacity guard anywhere in the fill path — nothing checked it before your PR. It will look like an over-strict false positive on sliding groups, since a window ring can be shorter than the step. It isn't: when it fires, it's real.

A sliding group whose ring is shorter than the chunk makes `paged_fill_cache` index past the end of its page-table CB — the kernel's lookup is unchecked, and its bounded-window wrap (`cache_position_modulo`) is never set from tt-xla. Result is a device hang, or silent KV corruption if the garbage block id lands in range. gemma-4 at `prefill_chunk_size=2048` hits exactly this, hanging 2/2 on base as well as with your PR, so it is pre-existing and not yours to fix. Details and repro in #5911.

Two optional asks, neither blocking:

- a one-line code comment recording why this assert must not be relaxed;
- the same check in `_dummy_run` (~3185) and `_get_dummy_inputs` (~3656), which is where it actually bites — warmup builds the graph with no capacity check, so today the failure is a hang rather than a host error.

integrations/vllm_plugin/vllm_tt/model_runner.py
```diff
@@ -1212,6 +1212,45 @@ def get_supported_tasks(self) -> tuple[SupportedTask, ...]:
...
+        leading columns are the ones it wants.
+        """
+        block_size = (
+            self._group_block_sizes[group_idx]
+            if group_idx < len(self._group_block_sizes)
+            else self.block_size
+        )
+        # Round to 8 entries: a ttnn page-table stick must be 32B-aligned, the
+        # same rule _chunked_sdpa_active checks against max_num_blocks_per_req.
+        width = cdiv(cdiv(self.max_num_tokens, block_size), 8) * 8
```
kmabeeTT, line 1237: `max_num_tokens = num_tokens_paddings[-1]`, so this narrowing only bites when chunked prefill is on. Without `prefill_chunk_size` the width returns to `cdiv(max_model_len, block_size)` and the row is 16 KB again — unchanged from today.

That is fine for #5897, whose config sets `prefill_chunk_size=1024`, but the docstring currently reads as though the op always gets a small table. A line here saying the bound is the chunk budget, not the context length, would stop the next reader treating this as a general fix for wide page tables.

kmabeeTT (approved):
Thanks! Couple feedbacks left... and some testing details here.

Validated against the #5897 hang. Both arms on one build at this PR's base `1bd189463`, rebuilt from scratch with the PR diff as the only variable, tt-metal pinned at `f1f4ff75579` with the chunked-read kernel patch **not** applied — so the tt-metal defect is live in both arms and this PR alone suppresses it.

| config, 128K | base | +#5901 |
|---|---|---|
| Llama-3.1-8B, 1 layer, 1 chip, chunk 1024 | 3/4 hang | 8/8 pass |
| Llama-3.1-8B, 1 layer, 1 chip, chunk 2048 | 4/4 hang | 4/4 pass |
| gemma-4-31B-it, TP=4, 60 layers, chunk 1024 | 5/5 hang | 3/3 pass |
| gemma-4-31B-it, TP=4, 6 layers, chunk 1024 | 1/2 hang | 3/3 pass |

Op confirmed by attribution rather than inferred: with tt-mlir `6cc62ef5f2` cherry-picked and `TT_RUNTIME_SYNC_AFTER_OP=1` plus trace off, the hanging run's last logged op is `ttnn.paged_fill_cache` with no `Synced after operation` after it, page-table operand `tensor<32x4096xsi32>` = 16,384 B per row. Post-PR it is `tensor<32x32xi32>` = 128 B. Width scales with `max_num_tokens` rather than `max_model_len`, so unlike `block_size=64` it does not degrade as context grows.

tt-metal's chunked read fixes the same case independently (same binary, base, chunk 2048: stock 2/3 hang, patched 4/4 pass), so tenstorrent/tt-metal#52270 and this PR are alternatives rather than complements. #52270 is still worth landing for the read page table, which stays at 4096 columns / 16 KB, and for the missing bound check.

Caveats: `TT_RUNTIME_SYNC_AFTER_OP=1` with trace off drops the hang rate (4/4 to 1/3), so a clean run under sync is not on its own evidence of a fix; gemma at chunk 2048 still hangs, on base too, for the unrelated reason in #5911; and chunk 2048 does not fit at GMU 0.35 on 60 layers (30.4 GiB KV needed vs 28.99 available).

integrations/vllm_plugin/vllm_tt/model_runner.py
sshonTT: goot catch! Yes, I've updated the docstring to clarify that this bound applies to the per-step chunk budget rather than the full context length.

integrations/vllm_plugin/vllm_tt/model_runner.py
sshonTT: thanks for the feedback! I've updated the PR with the host-side guard and comments as suggested.

Issue #5876, opened 2026-08-04
Add  TTS Models Benchmarking Harness for coqui/XTTS-v2


ashokkumarkannan1:

ashokkumarkannan1:
https://github.com/tenstorrent/tt-xla/pull/5905 closes this

PR #5905, opened 2026-08-06
[XTTS-v2] TTS perf benchmark harness and nightly entry

ashokkumarkannan1:
### Ticket
- https://github.com/tenstorrent/tt-xla/issues/5216 (Model Bringup: coqui/XTTS-v2)
- https://github.com/tenstorrent/tt-xla/issues/5876

### Problem Description
- XTTS-v2 was brought up and PCC-gated in #5376, but its throughput was never measured: `perf-bench-matrix.json` had 162 entries and no audio model, and `tests/benchmark` carried no TTS harness.

### What's changed
- **Model-agnostic TTS harness** (`tests/benchmark/benchmarks/tts_benchmark.py`), sibling of the imagegen / AR-imagegen / video harnesses. It names no model: stage names come from the pipeline's own `_perf` dict, so a new TTS model needs a `*_pipeline.py` and a test entry, not a new harness. Headline throughput is the **real-time factor** (generated audio seconds per wall-clock second), mapped onto `total_samples / total_time` so the existing Superset regression check works unmodified.
- **Compilation inside a timed region is a hard failure**, not a warning — a recompile inflates the result by a plausible-looking margin and reads downstream as an ordinary perf regression. Three assertions cover it:
  - `assert_no_recompiles` — `CompileTime` counter delta bracketing the measured pass,
  - `assert_decode_graph_reused` — the per-step cumulative compile curve must be flat over the tail of the decode loop, which the aggregate check cannot see if warmup absorbed it,
  - `assert_stage_timings_plausible` — stage time must be ≤ wall time and ≥ 30% of it, catching a missing `torch_xla.sync()` / `wait_device_ops()`.
- **`compile_curve` is required**, not optional. Reading it needs torch-xla's counters, so the per-model wiring records it; making it optional would let a model silently opt out of the only guard that can see a decode loop recompiling per token.
- **Warmup is two full generations** (`WARMUP_PASSES = 2`, set empirically). Pass 1 compiles the graphs and specialises the traced code; pass 2 compiles a further ~5 once the reused KV-cache buffers reach their steady state (no dynamo retracing occurs there — verified); pass 3 onward is flat. One warmup pass would leave those inside the measured pass, which the recompile guard would correctly fail on.
- **Per-model wiring** (`tests/benchmark/benchmarks/xtts_v2_pipeline.py`, 74 lines) adds only `compile_curve`. Everything else — per-stage timing, device residency, KV-cache reuse, fixed generated length — lives in `XTTSPipeline` upstream, so the measured code path is the one the nightly e2e test validates.
- **Nightly matrix entry** — `{"name": "xtts_v2", "pytest": "tests/benchmark/test_tts.py::test_xtts_v2"}`, with no `runs-on` override so it inherits `test-defaults` and runs on both `n150-perf` and `p150-perf` (162 → 163 entries).
- **E2E PCC test updated for device residency** (`tests/torch/models/xtts_v2/test_xtts_v2_pipeline.py`). Its CPU replay previously relied on the pipeline evicting `decode_step` to the host as a side effect. With every module now resident, a wrapper built over `self.xtts` mid-run would share *device* weights and the "CPU" golden would silently run on TT — comparing the device against itself and passing for the wrong reason. The replay now happens after `run()` returns, when the device weights are finished with, so one `gpt_latents.to("cpu")` brings the whole `xtts.gpt` tree back and the original golden semantics hold with no second model in memory.

Config matches the nightly e2e path: float32 (submodules do not cast uniformly to bf16), `optimization_level` 0, trace disabled, `MAX_AUDIO_TOKENS` 128. Weights are CPML-gated behind `COQUI_TOS_AGREED`; deps install via `RequirementsManager`.

### Results

Verified on Blackhole (p150-class):

| Check | Result |
|-------|--------|
| Benchmark (`tests/benchmark/test_tts.py::test_xtts_v2`) | **PASS** — **0 compilations** in the measured pass, RTF 0.989x, 127 audio tokens (5.89 s @ 24 kHz), decode 27.1 ms/token, stage times 98.9% of wall time. Five runs: RTF 0.946 / 0.960 / 0.975 / 0.988 / 0.989, identical 127-token output, 0 compiles every time |
| E2E pipeline test | **PASS** — per-step decode PCC ≥ 0.99 vs CPU replay over 64 steps, cumulative compile count flat at 12 from step 3 |
| Demo (`examples/pytorch/xtts_v2_pipeline.py`), uncapped | **PASS** — stops on its own stop token at ~163 tokens, 6.97 s of 24 kHz audio |

Steady-state stage breakdown: `speaker_encoder` 0.043 s, `conditioning` 0.189 s, `gpt_prefill` 0.035 s, `gpt_latents` 0.046 s, `hifigan` 2.136 s, decode 27.1 ms/token. One `hifigan` forward costs more than all 127 decode steps combined — the first place to look for optimisation.

Token policy differs per test, deliberately: the benchmark runs three times so its length must be fixed (`stop_early=False`) or the length-shaped stages recompile mid-measurement; the PCC test and the demo run once, so they use the model's natural stop token.

### Checklist

### Logs
- [xtts_v2_e2e_pipeline_test.log](https://github.com/user-attachments/files/30784128/xtts_v2_e2e_pipeline_test.log)
- [xtts_v2_benchmark_output.wav](https://github.com/user-attachments/files/30784123/xtts_v2_benchmark_output.wav)
- [xtts_v2_benchmark_test.log](https://github.com/user-attachments/files/30784124/xtts_v2_benchmark_test.log)
- [xtts_v2_demo_example_test.log](https://github.com/user-attachments/files/30784125/xtts_v2_demo_example_test.log)
- [xtts_v2_demo_output.wav](https://github.com/user-attachments/files/30784126/xtts_v2_demo_output.wav)
- [xtts_v2_e2e_pipeline_output.wav](https://github.com/user-attachments/files/30784127/xtts_v2_e2e_pipeline_output.wav)

### CI benchmark run

- Both `xtts_v2` dispatches failed inside `get_file()` before any measurement: [32320607057](https://github.com/tenstorrent/tt-xla/actions/runs/32320607057) (`sh-runner=true`) with `ValueError: IRD_LF_CACHE environment variable is not set`, and [32323066818](https://github.com/tenstorrent/tt-xla/actions/runs/32323066818) (`sh-runner=false`, the nightly's config) with `FileNotFoundError` on `test_files/pytorch/whisper/1272-128104-0000.pt` — the perf pool's `/mnt/dockercache` carries no `test_files/` tree, and `vars.IRD_LF_CACHE` is a cluster-internal address that does not resolve from bare metal ([32326295221](https://github.com/tenstorrent/tt-xla/actions/runs/32326295221): DNS failure, `http_code=000`).
- Fixed in https://github.com/tenstorrent/tt-forge-models/pull/888, which reads the same LibriSpeech clip from `hf-internal-testing/librispeech_asr_dummy` instead — byte-identical input, so no baseline in this PR moves; re-verified on p150 after the change: PASSED, RTF 1.079x, 127 audio tokens, 0 compilations in the measured pass, deviation `0.000e+00`.
- **Merge order:** merge this only after https://github.com/tenstorrent/tt-forge-models/pull/888 lands and the `tt_forge_models` submodule is uplifted (`schedule-uplift-forge-models.yml`, daily 07:00 UTC or `workflow_dispatch`); merging sooner puts failing `n150-perf` / `p150-perf` legs into the nightly.

### DIFF
--- a/.github/workflows/perf-bench-matrix.json
+++ b/.github/workflows/perf-bench-matrix.json
@@ -91,6 +91,11 @@
         "group": "experimental"
       },
       {
+        "name": "xtts_v2",
+        "pytest": "tests/benchmark/test_tts.py::test_xtts_v2",
+        "group": "experimental"
+      },
+      {
         "name": "llama_3_1_8b_instruct",
         "pytest": "tests/benchmark/test_llms.py::test_llama_3_1_8b",
         "group": "regular"
--- a/tests/benchmark/README.md
+++ b/tests/benchmark/README.md
@@ -45,6 +45,7 @@ tests/benchmark/
 ├── test_llms.py             # LLM test definitions
 ├── test_vision.py           # Vision model test definitions
 ├── test_encoders.py         # Encoder model test definitions
+├── test_tts.py              # Text-to-speech test definitions
 ├── benchmarks/              # Core benchmark implementations
 ├── llm_utils/               # LLM-specific utilities
 ```
@@ -129,6 +130,22 @@ def test_new_model(
 ```
 
 Vision and encoder tests follow the same pattern — import the existing `ModelLoader`, pick a variant, and call `test_vision()` or `test_encoder()`.
+
+### Adding a text-to-speech model
+
+`benchmarks/tts_benchmark.py` is model-agnostic: a new TTS model needs a
+`build_pipeline_fn` and a pipeline that fills in a `_perf` dict (`components`,
+`steps`, `step_metric_name`, `total`, `audio_samples`), plus a `test_<model>` entry
+in `test_tts.py`. See `benchmarks/xtts_v2_pipeline.py` for a worked example.
+
+Two things differ from the other families:
+
+- **Headline metric.** `samples_per_sec` is the *real-time factor* — generated audio
+  seconds per wall-clock second — so `total_samples` is the audio duration, not a
+  count of items.
+- **Fixed work per run.** The generated token count is pinned, so an RTF change is a
+  pure wall-clock change and a graph compiling inside a timed region is a hard
+  failure (`assert_no_recompiles`) rather than a warning.
 
 ### Adding the test to CI
 
--- a/tests/benchmark/benchmarks/tts_benchmark.py
+++ b/tests/benchmark/benchmarks/tts_benchmark.py
@@ -1,0 +1,347 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Generic text-to-speech benchmark harness for torch-xla / TT.
+
+Sibling of ``imagegen_benchmark.py`` and ``ar_imagegen_benchmark.py``, for models
+whose output is a waveform. Per-model wiring supplies a ``build_pipeline_fn`` and
+the pipeline reports its own stage names, so a new TTS model needs a
+``*_pipeline.py`` plus a test entry, not a new harness.
+
+A TTS model is several compiled graphs chained by CPU orchestration rather than one
+traceable forward, so the harness times the chain and reports each stage as its own
+``<stage>_s`` measurement. Headline throughput is the real-time factor: generated
+audio seconds per wall-clock second.
+
+``WARMUP_PASSES`` *full* generations run before the timed one: the vocoder and
+latents stages are shaped by the generated token count, so anything shorter would
+compile inside the measured pass. Recompiles there are a hard error, not a warning
+-- they inflate the result by an amount that reads as an ordinary regression.
+
+The ``_perf`` contract a pipeline must populate on each generation:
+
+    _perf = {
+        "components": {<stage name>: seconds, ...},  # scalar per-stage times
+        "steps": [seconds, ...],                     # per audio-token decode step
+        "step_metric_name": "audio_token_step",
+        "total": seconds,                            # full generate() wall time
+        "audio_samples": int,                        # output waveform length
+        "text_tokens": int,                          # optional, for reporting
+        "compile_curve": [int, ...],                 # cumulative compiles per step
+    }
+"""
+
+import socket
+import time
+
+import torch_xla
+import torch_xla.core.xla_model as xm
+import torch_xla.debug.metrics as met
+import torch_xla.runtime as xr
+from utils import (
+    build_xla_export_name,
+    create_benchmark_result,
+    get_benchmark_metadata,
+    get_xla_device_arch,
+    print_benchmark_results,
+)
+
+xr.set_device_type("TT")
+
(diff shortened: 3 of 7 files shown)


kamalrajkannan78 (approved):
Reduce the verbose in comments and doc string to minimal

jameszianxuTT (approved):

kmabeeTT (approved):

nsumrakTT (approved):

.github/workflows/perf-bench-matrix.json
```diff
@@ -114,6 +114,10 @@
         "pytest": "tests/benchmark/test_imagegen.py::test_glm_image",
         "runs-on": "qb2-blackhole"
       },
+      {
+        "name": "xtts_v2",
+        "pytest": "tests/benchmark/test_tts.py::test_xtts_v2"
```
vkovacevicTT: Please rebase and add `"group": "experimetnal"` here :)

vkovacevicTT (changes requested):

.github/workflows/perf-bench-matrix.json
ashokkumarkannan1: thanks!, added @vkovacevicTT

vvukomanTT (approved):

ashokkumarkannan1:
## CI benchmark run
- Both xtts_v2 dispatches failed inside get_file() before any measurement: [32320607057](https://github.com/tenstorrent/tt-xla/actions/runs/32320607057) (sh-runner=true) with ValueError: IRD_LF_CACHE environment variable is not set, and [32323066818](https://github.com/tenstorrent/tt-xla/actions/runs/32323066818) (sh-runner=false, the nightly's config) with FileNotFoundError on test_files/pytorch/whisper/1272-128104-0000.pt — the perf pool's /mnt/dockercache carries no test_files/ tree, and vars.IRD_LF_CACHE is a cluster-internal address that does not resolve from bare metal ([32326295221](https://github.com/tenstorrent/tt-xla/actions/runs/32326295221): DNS failure, http_code=000).
- Fixed in https://github.com/tenstorrent/tt-forge-models/pull/888, which reads the same LibriSpeech clip from hf-internal-testing/librispeech_asr_dummy instead — byte-identical input, so no baseline in this PR moves; re-verified on p150 after the change: PASSED, RTF 1.079x, 127 audio tokens, 0 compilations in the measured pass, deviation 0.000e+00.
- Merge order: merge this only after https://github.com/tenstorrent/tt-forge-models/pull/888 lands and the tt_forge_models submodule is uplifted (schedule-uplift-forge-models.yml, daily 07:00 UTC or workflow_dispatch); merging sooner puts failing n150-perf / p150-perf legs into the nightly.

vkovacevicTT (approved):

Issue #5879, opened 2026-08-04
[vLLM] Sliding-window ring reserve scales with max_num_seqs, costing ~4x long-context concurrency


kmabeeTT:
**TL;DR** — `max_num_seqs` now consumes KV cache budget, and before #5786 it consumed none. At `max_model_len=65536` the KV pool is **3.9x larger at `max_num_seqs=8` than at 32** (373,255 vs 96,210 tokens) on identical hardware. Before the sliding-window work the pool was byte-identical at both batch sizes.

Follow-up to #5786. Not a regression — the ring works, and at small batch it delivers.

## Problem

The per-user ring reserve is taken off the top of the KV budget before the shared pool is sized, and it scales with `max_num_seqs` rather than with the number of requests the pool can actually keep resident:

```
reserve = n_sliding_layers * (window_blocks * max_num_reqs + 1) * page_size_bytes
```

At 64K the concurrency the pool can support is set by the 10 full-attention layers drawing on it — **1.47x** at `max_num_seqs=32`, not 32. So the reserve provisions 32 sub-rings for a pool that can keep barely one or two requests resident, and the rest can never be reached. It is independent of `gpu_memory_utilization`, so it cannot be tuned away that way. Consequences:

* Long-context runs at the default `max_num_seqs=32` leave **~4x** concurrency on the table for this model and configuration.
* The coupling is silent — nothing warns that the reserve is provisioning far more sub-rings than the pool can fill, and the tuning direction is counterintuitive.
* Workaround today: lower `max_num_seqs` for long-context configurations.

To be clear about what this is *not*: "fixed cache, more concurrent requests, less cache each" is ordinary arithmetic and has always been true. Here the **total pool itself** changes size with `max_num_seqs`, which it did not do before #5786.

## Evidence and repro

One configuration throughout — `google/gemma-4-31B-it`, 4x Blackhole (p300c), `max_model_len=65536`, `bfp_bf8` KV, `gpu_memory_utilization=0.30`, `prefill_chunk_size=1024`. The **only** variables are the commit and `max_num_seqs`. All four runs pass.

| | batch 32 | batch 8 | effect of batch |
|---|---|---|---|
| **before sliding window** (`591d66f52`) | 91,136 tok / 1.39x | **91,136 tok / 1.39x** | **none — identical** |
| **with sliding window** (`b2c90b4c3`) | 96,210 tok / 1.47x | **373,255 tok / 5.70x** | **3.9x** |

[...]

Issue #5860, opened 2026-07-31
[vLLM] Sliding-window KV regressions: pool got smaller at short context, and a 64K config no longer starts


kmabeeTT:
**TL;DR** — sliding-window KV support (#5786) made the KV cache **29% smaller** at the default `max_model_len=128` (60,768 -> 43,420 tokens), and made a 64K / batch-32 configuration that ran before it **fail to start**.

Follow-up to #5727 / #5786 / #5834. Two independent regressions against the commit before the sliding-window work. Neither is a correctness bug — outputs are unaffected — but both are a real loss of usable KV capacity, not a reporting artifact: the same 51.00 GiB of device DRAM is allocated either way, and 6.27 GiB of it now goes to ring buffers that hold nothing the shared pool would not have held. Throughput at Problem 1's shape is unchanged only because `max_model_len=128` x batch 32 needs 4,096 KV tokens and even the reduced pool holds 43,420; Problem 2 is the same defect where the headroom has run out.

## Setup

Branch `kmabee/issue_5727_followup` puts the two sliding-window commits back-to-back on an otherwise-untouched base, so one command isolates them:

```
b2c90b4c3  [vLLM] Cap sliding-window ring size by max_model_len (#5834)          <- tip = "with SWA"
0bb36ceac  [vLLM] Hybrid KV cache : per-layer sliding-window ring buffer (#5786)
591d66f52  Allow single-group DeepSeek-V2 MoE routing (#5823)                    <- "before SWA"
```

> **Why a branch and not main:** on main six PRs sit between these two commits — including the vLLM 0.26 uplift (#5827), serving telemetry (#5808) and DP+TP fixes
> (#5801) — so you cannot roll back only the sliding-window change there. This branch pins **vLLM 0.25.1**. `worker.py` and `swa_cache_utils.py` are byte-identical to
> main's, so the KV-sizing path under test is the one main has.

```bash
cd <tt-xla>
source venv/activate
git checkout kmabee/issue_5727_followup
```

Switch between the two refs with a plain checkout:

```bash
git checkout 591d66f52                     # -> before SWA
git checkout kmabee/issue_5727_followup    # -> back to with SWA
```

No rebuild needed either way: the only files that differ between these two refs are Python (the vLLM plugin plus two of its tests) — no C++, no submodules.

[...]

PR #5827, opened 2026-07-29
[vLLM][uplift] v0.26.0

mmanzoorTT:
### Ticket
closes #5826

### Problem description
vLLM v0.26.0 released

### What's changed
- Uplift vLLM
- Remove most_model_len path in runners and update relevant test
- Update MultiGroupBlockTable usage
- Update Schedular

### Checklist
- [X] New/Existing tests provide coverage for changes

PR #5908, opened 2026-08-06
[vLLM] Fix sliding-window KV sizing at short and long context

sshonTT:
### Ticket
#5860  
#5879  partially

### Problem description
#5786 introduced two sizing defects in per-layer sliding-window rings:
 - Short Context: At max_model_len <= sliding_window, ring overhead (straddle block & alignment) shrinks the KV pool by ~29% unnecessarily.
 - Long Context: Admission bound calculates memory using max_num_batched_tokens instead of tt_prefill_chunk_size, over-charging budget up to max_num_seqsx and incorrectly rejecting valid configs at startup.

### What's changed
 - Hybrid KV Cache Switch: Enabled the hybrid manager only when a ring is actually cheaper than full attention (otherwise fallback to a shared pool). Manual user settings still take priority.
 - Fixed Admission Bound Over-charging: Clamped the sliding-window admission bound to the per-request prefill chunk size in the spec method (used for both startup sizing and runtime admission).
 - Aligned Ring Reservation: Sized worker ring reservations based on the actual grouped spec, preventing ghost allocations. Added unit tests for sizing paths.
 

### Tests
New unit tests, all CPU-only (no device, no engine):

| file | covers |
|---|---|
| `test_hybrid_kv_profitability.py` | the ring-vs-full crossover and the cases the gate must leave alone (no sliding layers, explicit user setting, auto-fit `max_model_len`, larger block size); the admission clamp, its no-op when TT chunked prefill is off, that it never raises the bound, install idempotence, and the in-flight multiplier |
| `test_sliding_ring_reserve.py` | the four spec/flag combinations the reservation must get right, including the uniform all-sliding model that keeps its specs through unification; that the caller's spec dict is not mutated |
| `test_swa_cache_utils.py` (extended) | ring width capped by `max_model_len`, the profitability helper's boundary |

One test pins the config attribute names the gate reads against the real
`ModelConfig` / `CacheConfig` / `SchedulerConfig` / `VllmConfig`, so an upstream
rename breaks here rather than in a benchmark run.

Existing tests: `tests/integrations/vllm_plugin/` passes (46 tests).

### Benchmark results
`gemma-4-31B-it`, bhqb, `bfp_bf8`:

| config | before | after |
|---|---|---|
| `max_model_len=128`, batch 32, `gmu 0.40` | 43,420 tokens / 339x | **60,768 / 475x** (pre-#5786 size) |
| 64K, batch 32, `gmu 0.225`, chunk 1024 | fails to start | **passes**, 259,666 tokens / 3.96x |
| 64K, batch 32, `gmu 0.30` | 96,210 tokens / 1.47x | **449,958 / 6.87x** |

#5879's batch-size coupling drops from 3.9x to 1.5x; the reserve still scales
with `max_num_seqs`, which is what remains there.

### Checklist

### DIFF
--- a/integrations/vllm_plugin/vllm_tt/platform.py
+++ b/integrations/vllm_plugin/vllm_tt/platform.py
@@ -720,6 +720,7 @@ class TTPlatform(Platform):
                 # TT-internal per-sequence chunk cap, read by AscendScheduler
                 # (per-request chunk sizing) and TTModelRunner (bucket ladder).
                 scheduler_config.tt_prefill_chunk_size = per_seq_chunk
+                publish_tt_per_request_prefill_chunk(scheduler_config)
                 # Derived value: a user-supplied max_num_batched_tokens is
                 # intentionally overridden here (logged above).
                 scheduler_config.max_num_batched_tokens = budget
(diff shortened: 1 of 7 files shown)


integrations/vllm_plugin/vllm_tt/platform.py
```diff
@@ -722,6 +771,69 @@ def manual_seed_all(cls, seed: int) -> None:
...
+
+    original = SlidingWindowSpec.max_admission_blocks_per_request
+    if getattr(original, "_tt_patched", False):
+        return  # idempotent: never stack wrappers and clamp twice
+
+    def _tt_max_admission_blocks_per_request(
+        self, max_in_flight_tokens: int, max_model_len: int
+    ) -> int:
+        if _TT_PER_REQUEST_PREFILL_CHUNK:
+            max_in_flight_tokens = min(
```
kmabeeTT, line 836: The 0.26.0 uplift (#5827) renamed this arg from `max_num_batched_tokens`, and the rename hid a factor: `max_in_flight_tokens = max_concurrent_batches * max_num_batched_tokens`. So the per-request analogue is `max_concurrent_batches * tt_prefill_chunk_size`, not the bare chunk.

Exact today, since TT forces `async_scheduling = False` and PP is pinned to 1. But if either changes this under-charges silently — the deadlock (vllm#39734) / mid-prefill-OOM direction. An `assert max_concurrent_batches == 1` would make the assumption fail loudly rather than sit there load-bearing.

Scope, for the record: `FullAttentionSpec` doesn't define this method, so full attention is untouched; `SlidingWindowMLASpec` inherits it, so SWA-MLA is covered. `ChunkedLocalAttentionSpec` has the same bound on the same path and isn't patched — unreachable on TT today, but the same bug if such a model lands.

integrations/vllm_plugin/vllm_tt/platform.py
```diff
@@ -641,6 +689,7 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
                 # TT-internal per-sequence chunk cap, read by AscendScheduler
                 # (per-request chunk sizing) and TTModelRunner (bucket ladder).
                 scheduler_config.tt_prefill_chunk_size = per_seq_chunk
+                publish_tt_per_request_prefill_chunk(scheduler_config)
```
kmabeeTT: The function's docstring says it "always assigns, so a config without TT chunked prefill clears a value a previous config published" — true of the function, but this call sits inside `if chunk_size > 0:`, so three paths through `check_and_update_config` never reach it: the MLA branch above (which sets `tt_chunked_prefill_enabled = False` without publishing), this `elif` when `chunk_size == 0`, and the pooling / `model_config is None` case. On all three the module global keeps whatever a previous config left.

Harmless in the real flow, since `TTWorker.__init__` publishes unconditionally in EngineCore where both consumers live. It only bites when two configs share a process (`VLLM_ENABLE_V1_MULTIPROCESSING=0`, or several configs in one pytest process), and then it is last-writer-wins against a class-level patch on `SlidingWindowSpec` — a wrong-sized pool rather than an error.

One unconditional call after the whole branch chain would cover all of them; dedenting this one out of the `if` only fixes one of the three.

integrations/vllm_plugin/vllm_tt/worker.py
```diff
@@ -57,6 +58,48 @@
...
+    and reserving for them would strand the bytes. Running vLLM's own function
+    over a copy keeps the two in agreement, including the uniform all-sliding
+    case it leaves alone -- whose rings are real and must still be reserved.
+
+    Under pipeline parallelism this worker's spec is a subset of the merged spec
+    vLLM unifies, so a stage holding only sliding layers can still disagree.
+    """
+    effective_spec = dict(kv_cache_spec)
+    if hybrid_kv_cache_disabled:
+        unify_hybrid_kv_cache_specs(effective_spec)
```
kmabeeTT: `unify_hybrid_kv_cache_specs()` logs a WARNING every call ("Hybrid KV cache manager is disabled for this hybrid model... we do not enable any optimizations for saving KV cache memory"). Calling it here on a copy means users now see it **twice** — confirmed in the `max_model_len=128` run: 2 occurrences on this branch, 0 on the base commit.

It reads like something went wrong, when in fact the plugin deliberately chose the shared pool because it is cheaper. Worth suppressing for the probe call, or at least following it with the plugin's own line explaining the choice was intentional.

(For what it is worth, the shallow `dict(kv_cache_spec)` copy is safe: every spec is a `@dataclass(frozen=True)`, so unification has to replace entries rather than mutate them.)

tests/integrations/vllm_plugin/test_hybrid_kv_profitability.py
```diff
@@ -0,0 +1,198 @@
...
+
+import types
+
+import pytest
+from vllm_tt.platform import TTPlatform
+
+pytestmark = [pytest.mark.push, pytest.mark.cpu]
+
+
+def _gate(
```
kmabeeTT, line 23: The gate tests use a `types.SimpleNamespace` stand-in, so they'd stay green if `get_sliding_window()`, `original_max_model_len`, or `cache_config.block_size` got renamed or dropped upstream. We just hit that kind of break with the 0.26.0 uplift (#5827), when `max_num_batched_tokens` became `max_in_flight_tokens`. Maybe worth one test that runs the gate over a real `VllmConfig`, so the next uplift breaks here instead of in a benchmark run.

The admission-bound tests below don't have this problem — they call the real `SlidingWindowSpec` with the real keyword, so a rename would break them.

kmabeeTT (approved):
Thanks @sshonTT! Looks good, few non-blocking small change requests, can you also update PR Desc a bit to mention new tests and existing test results?

"#5879 partially" is right — here's how much. The reserve still scales with `max_num_seqs` in `sliding_ring_phys_blocks`. Same 64K config on this branch, only the batch changed:

| `max_num_seqs` | ring reserve | pool | concurrency |
|---|---|---|---|
| 32 | 15.637 GiB | 259,666 tok | 3.96x |
| 8 | 3.918 GiB | 492,855 tok | **7.52x** |

So 1.90x is still on the table for #5879. That's less than the ~3.9x I measured before this PR — the admission fix already recovers part of it.

integrations/vllm_plugin/vllm_tt/worker.py
sshonTT: Probe call now runs with vLLM's `kv_cache_utils` logger raised to ERROR, so the warning appears
once, from the real call.

integrations/vllm_plugin/vllm_tt/platform.py
sshonTT: Read it instead of asserting: publish now takes the `VllmConfig` and multiplies by
`max_concurrent_batches`, with a test pinning it. Left `ChunkedLocalAttentionSpec` unpatched —
unreachable on TT today.

integrations/vllm_plugin/vllm_tt/platform.py
sshonTT: Moved to one unconditional call after the branch chain, so all three paths clear the global.

tests/integrations/vllm_plugin/test_hybrid_kv_profitability.py
sshonTT: Added — pins the names against the real config classes. `original_max_model_len` is assigned in
`ModelConfig.__post_init__` rather than declared, so it's checked separately.

PR #5920, opened 2026-08-07
[vLLM] Add continuous batching test

### LINKED ISSUE #5813: [vLLM] Add test/demo for continuous batching
vLLM plugin scheduler supports continuous batching; add a demo or test to verify this behaviour.

- If new prompts arrive while older requests are already decoding, the scheduler prioritizes prefill for the new prompts. If the current requests are less than batch_size (max_num_seqs).
- On the next decode-only step, decode runs for both old and newly prefilled requests.

### Verification:
- [ ] Verify that kv cache is updated properly.
- [ ] Verify the generated output for each prompt. 

Try single device and multi device scenarios.

alinakhanTT:
### Ticket
closes #5813

### What's changed
Adds an on device continuous batching test. Two prompts start generating, and once they are decoding two more are submitted, so their prefill joins a live decode batch.

Every output is compared token for token against a reference pass where each prompt runs alone. That is what catches a request whose KV was clobbered by the joiner, or a joiner reading stale KV from its slot, since both keep producing fluent text that a coherence check would happily accept. The test also asserts the two waves really overlapped, so it cannot quietly turn into an ordinary batched run.

Covers single device, plus llmbox under TP and DP+TP. All three pass locally on device, with every output identical to its reference.

The scheduler half of the issue (prefill first on arrival, then decode for old and new together) is already covered on CPU by test_ascend_scheduler_batching.py, so this fills in the device side.

### Checklist

### DIFF
--- a/tests/integrations/vllm_plugin/generative/test_continuous_batching.py
+++ b/tests/integrations/vllm_plugin/generative/test_continuous_batching.py
@@ -1,0 +1,230 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Continuous batching on device (#5813).
+
+Wave B is submitted only once wave A is decoding, so its prefill joins a live
+decode batch. Every output must match a reference pass where each prompt runs
+alone: a request whose KV was clobbered by the joiner, or a joiner reading a
+slot's stale KV, diverges from that reference while still reading as fluent
+text, so coherence checks alone would miss it.
+
+The scheduler side (prefill first, then decode for old and new together) is
+covered on CPU by unit_tests/test_ascend_scheduler_batching.py.
+"""
+import asyncio
+
+import pytest
+from conftest import GROUNDED_BATCH_CHECKS, check_host_memory
+from vllm import AsyncEngineArgs, AsyncLLMEngine, SamplingParams
+from vllm.sampling_params import RequestOutputKind
+
+MODEL = "Qwen/Qwen3-0.6B"
+MAX_NUM_SEQS = 4
+MAX_MODEL_LEN = 64
+# Wave A must outlive B's arrival, so it generates for longer.
+MAX_TOKENS_A = 24
+MAX_TOKENS_B = 12
+DECODE_TOKENS_BEFORE_ARRIVAL = 4
+# The first request pays engine warmup and graph compilation.
+ARRIVAL_TIMEOUT_S = 900
+
+WAVE_A = GROUNDED_BATCH_CHECKS[:2]
+WAVE_B = GROUNDED_BATCH_CHECKS[2:]
+
+
+def _make_engine(additional_config: dict) -> AsyncLLMEngine:
+    args = AsyncEngineArgs(
+        model=MODEL,
+        max_model_len=MAX_MODEL_LEN,
+        max_num_batched_tokens=MAX_MODEL_LEN * MAX_NUM_SEQS,
+        max_num_seqs=MAX_NUM_SEQS,
+        gpu_memory_utilization=0.002,
+        # Off so the reference pass doesn't seed the cache for the staggered one
+        # and wave B really prefills on arrival. DP disables it anyway.
+        enable_prefix_caching=False,
+        additional_config={"min_context_len": 32, **additional_config},
+    )
+    return AsyncLLMEngine.from_engine_args(args)
+
+
+def _sampling_params(max_tokens: int) -> SamplingParams:
+    # Greedy and fixed length, so the reference is an exact oracle.
+    return SamplingParams(
+        temperature=0.0,
+        max_tokens=max_tokens,
+        ignore_eos=True,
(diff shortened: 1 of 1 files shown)


tests/integrations/vllm_plugin/generative/test_continuous_batching.py
```diff
@@ -0,0 +1,230 @@
...
+    ):
+        label = "A" if i < len(WAVE_A) else "B"
+        print(f"  wave {label} req {i}: {prompt!r} -> {text!r} (ref {ref_text!r})")
+        if expected.lower() not in ref_text.lower():
+            failures.append(
+                f"req {i} ({prompt!r}): reference itself is wrong, expected "
+                f"{expected!r}, got {ref_text!r} -- not a batching failure"
+            )
+            continue
+        if ids != ref_ids:
```
kmabeeTT, line 182: `ids != ref_ids` compares two separate `generate()` calls, which is exactly the property #5520 says is not guaranteed under TP: the cross-chip lm_head reduction is not bit-reproducible, so a near-tie argmax can flip between otherwise-identical calls. `cpu_sampling=True` does not cover it — #5520 is in the reduction producing the logits, not in the sampler op. The conftest you import from says the same thing at line 151, where `assert_batch_grounded` is deliberately "tolerant of the #5520 near-tie fp drift" on these same four prompts.

Suggest keeping the exact-token oracle for `single_device` (no cross-chip reduction, so #5520 does not apply) and dropping the TP / DP+TP variants back to `_assert_overlapped` plus a grounded check. That still catches the failure you are actually hunting — a joiner clobbering a neighbour's KV shows up as garbage or a repeat-loop, not as a one-token near-tie flip — without asserting bit-reproducibility the stack does not currently provide. Alternatively keep it and xfail the llmbox variants against #5520 so the flake is attributed when it lands.

tests/integrations/vllm_plugin/generative/test_continuous_batching.py
```diff
@@ -0,0 +1,230 @@
...
+MAX_NUM_SEQS = 4
+MAX_MODEL_LEN = 64
+# Wave A must outlive B's arrival, so it generates for longer.
+MAX_TOKENS_A = 24
+MAX_TOKENS_B = 12
+DECODE_TOKENS_BEFORE_ARRIVAL = 4
+# The first request pays engine warmup and graph compilation.
+ARRIVAL_TIMEOUT_S = 900
+
+WAVE_A = GROUNDED_BATCH_CHECKS[:2]
```
kmabeeTT, line 32: `WAVE_A`/`WAVE_B` slice a shared constant at a fixed index, and the test only works while `len(GROUNDED_BATCH_CHECKS) <= MAX_NUM_SEQS`. Today it is exactly 4 == 4, so both waves fit concurrently. If anyone appends a fifth check to `GROUNDED_BATCH_CHECKS` — it is shared with `assert_batch_grounded` and its other callers — `WAVE_B` becomes 3, the total exceeds `max_num_seqs`, and one B request queues instead of joining the live decode batch. The test keeps passing while no longer testing continuous batching.

One line at module scope pins it:

    assert len(WAVE_A) + len(WAVE_B) <= MAX_NUM_SEQS, "waves must fit in one batch"

tests/integrations/vllm_plugin/generative/test_continuous_batching.py
```diff
@@ -0,0 +1,230 @@
...
+    ]
+
+    gate_task = asyncio.create_task(gate.wait())
+    done, _pending = await asyncio.wait(
+        {gate_task, *a_tasks},
+        timeout=ARRIVAL_TIMEOUT_S,
+        return_when=asyncio.FIRST_COMPLETED,
+    )
+    if not gate_task.done():
+        gate_task.cancel()
```
kmabeeTT, line 120: The timeout path cancels `gate_task` but leaves `a_tasks` pending, so the `AssertionError` propagates while wave A is still generating into an engine that `_collect`'s `finally` is about to shut down. That yields "Task was destroyed but it is pending" noise on top of the real failure, right when the output is being read to diagnose a timeout. Cancelling `a_tasks` alongside `gate_task` keeps the failure message clean.

kmabeeTT (changes requested):
Looks good overall, few things to consider changing

tests/integrations/vllm_plugin/generative/test_continuous_batching.py
alinakhanTT: Added!

tests/integrations/vllm_plugin/generative/test_continuous_batching.py
alinakhanTT: TP / DP+TP variants dropped back to _assert_overlapped plus a grounded check and exact-token oracle for single_device kept

tests/integrations/vllm_plugin/generative/test_continuous_batching.py
alinakhanTT: Done

kmabeeTT (approved):
Thanks for changes!

PR #5923, opened 2026-08-10
[op-by-op] Parallelize unique-op extraction and bound the job

ctr-lelanchelianTT:
### Problem

The **2026-08-09 weekly run was the first op-by-op run after #5868 landed**, and it produced no data:

```
process_ir / extract_unique_ops     failure   09:09:42 → 14:55:04   (5h45m)
  ✓ Download dumped IR artifacts
  ⨯ Extract globally unique ops      ← killed here
process_ir / process_dumped_irs     skipped
```

No `timeout-minutes` was set, so GitHub's 360min default killed the job, and every processing shard was skipped. For comparison, the previous week's run on the old whole-model flow executed **27,125 ops across 24 shards**.

This is a regression from #5868, which I validated on 2 models / 54 IRs. The real deduped dump is **3,458 IR files across 148 models** — 117× that corpus.

**Root cause.** Extraction ran single-threaded, and its cost grows superlinearly with a module's op count. Measured per-file on real nightly IRs:

| model | ops | time |
|---|---|---|
| Phi-1 | 3049 | 53.8s |
| SigLIP Base | 2344 | 32.2s |
| DINOv2 Small | 1300 | 10.1s |
| *median file* | — | **0.36s** |

2.3× the ops costs 5.4× the time. A handful of large language models dominate hours of wall time, and the step printed **nothing** for 5h45m, so the timeout was diagnosable only from step durations.

### Change

- **Parallel per-IR-file extraction** across worker processes — `--jobs`, default `min(cpu_count, 8)`; the workflow passes `nproc`. Each worker deduplicates within its own file first, so only a small amount of module text crosses the process boundary.
- **Deterministic ordering.** Results are consumed in submission order, so op indices and op directory names match the serial ordering and don't depend on which worker finished first.
- **Progress logging** with elapsed and projected remaining time.
- **`--time-budget-seconds`** stops extraction and continues with the ops found so far, reporting how many IR files were skipped and recording each as an extraction failure. The workflow sets `9000s` inside an explicit `timeout-minutes: 240`, so a slow run ends by uploading a usable op set rather than being killed with nothing.

The budget is a **soft** bound — it's checked as each file's result is consumed, so in-flight workers finish and the overshoot can be as large as the slowest single IR file. The help text says this explicitly, and the workflow leaves headroom.

### Verification

On 26 real nightly IRs (8 models, 23,436 ops, 1068 unique):

| | result |
|---|---|
| serial (`--jobs 1`) | 305.6s |
| parallel (`--jobs 8`) | **67.9s — 4.5×** |
| output trees | **byte-for-byte identical** (all 1068 modules) |
| manifests | identical except `output_root` |
| totals | identical: 23,436 ops, 1068 unique, 0 failures |

The 4.5× rather than 8× is expected: the single 54s file sets the floor.

Budget path, 20s budget on the same corpus: stopped at 17/26 files, still wrote 691 modules, reported 9 skipped, set `extraction_complete: false`, exit 0.

`black` clean, `py_compile` clean, workflow YAML parses with all four jobs.

A full-corpus run over all 3,458 deduped IRs is in progress locally; I'll post the end-to-end wall time when it finishes. Note the CI runner has more cores than the 8 used above, so that figure is a conservative upper bound.

### Not fixed here

The superlinear cost itself lives in tt-mlir's `extract_ops_from_module` / `MLIRModuleSplitter.split`. Parallelism buys a constant factor; it does not change the curve. Worth a tt-mlir issue — as the model set grows, this will need revisiting.

Refs #5867

### DIFF
--- a/.github/scripts/ir_process_analyzer.py
+++ b/.github/scripts/ir_process_analyzer.py
@@ -3,17 +3,29 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import argparse
+import concurrent.futures
 import json
 import math
+import os
 import re
 import shutil
 import sys
+import time
 from pathlib import Path
 
 SECONDS_PER_MB = 120
 
 # Max model/op names listed inline per job in the human-readable plan summary.
 SUMMARY_MODEL_LIMIT = 25
+
+# Extraction is CPU-bound MLIR parsing and its cost grows superlinearly with the
+# op count of a module, so a handful of large models dominate the wall time. It
+# parallelizes cleanly per IR file. Capped rather than using every core because
+# each worker holds a fully parsed module in memory.
+DEFAULT_EXTRACT_JOBS = min(os.cpu_count() or 1, 8)
+
+# How often to print extraction progress, in files.
+EXTRACT_PROGRESS_EVERY = 25
 
 
 def estimate_seconds(size_bytes: int) -> int:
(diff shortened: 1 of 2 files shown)


nsumrakTT (commented):
As I see it, the problem experienced in https://github.com/tenstorrent/tt-xla/pull/5868 with timeout is due to the problem in the implementation. It was not tested as it should, eg. running with existing data weekly data (it is possible to launch it locally or with manual dispatch of workflow run call-process-dumped-irs.yml, modified in this PR).

Now this PR doesn't fix the problem but instead try to parallelize its execution. That doesn't make sense to me. Have you tested these changes? I don't see the link to the run of modified workflow. Extracting unique ops takes more time than duplicate processing? I don't think so, if that is the case then there is no sense in implementing global unique ops filter.

Please fix the problem before adding more complexity to it.

PR #5938, opened 2026-08-11
[CI] Fix perf benchmarks skipped by broken matrix_skip guard

### LINKED ISSUE #5918: CI: rerun-passed tests still reported as failed in TT-XLA basic-test / check-all-green
## Problem

`TT-XLA basic-test` (and the `check-all-green` gate that depends on it) can stay red on a PR even after a job rerun has fixed the failure. A test that failed in attempt 1 and passed on rerun (attempt 2) still gets reported as failed, because the test-summary step merges JUnit XML from every attempt without dropping superseded results.

Observed on tenstorrent/tt-xla#5850, run [31176545307](https://github.com/tenstorrent/tt-xla/actions/runs/31176545307):

- Attempt 1, job `run_torch, 5` ([92866738056](https://github.com/tenstorrent/tt-xla/actions/runs/31176545307/attempts/1/job/92866738056)): `test_qwen2_5_attention_prefill_push[7B_Instruct-1024-single_device]` **FAILED** — `PCC comparison failed. Calculated: pcc=0.8999734493840111. Required: pcc=0.98` (borderline, looks like run-to-run non-determinism, not a real regression).
- Rerun → attempt 2, same shard, job [92879876496](https://github.com/tenstorrent/tt-xla/actions/runs/31176545307/job/92879876496): same test **PASSED**.
- `TT-XLA basic-test` (check run [92881395273](https://github.com/tenstorrent/tt-xla/runs/92881395273)) still reports `"3940 tests run, 3454 passed, 485 skipped, 1 failed"`, pointing at the same test, and `check-all-green` fails because it gates on that check.

(Side note: `TT-XLA basic-test` has no expand arrow in the UI because it isn't an Actions job — `"steps":[]`, `"runner_id":null` — it's a check run posted directly via the Checks API by `mikepenz/action-junit-report`, not a job with logs.)

## Details

`.github/workflows/call-test.yml`

1. **Retry design (working as intended):** on a rerun, `check-rerun-attempt` (lines 255-294) downloads the previous attempt's report for the same shard, extracts failed test names via `find_all_failed_tests.py`, and `Set pytest command` (line 331) makes the new attempt run **only** those tests. So attempt 2's XML for shard 5 contains just the one retried test, now passing — not the full shard.
2. **Aggregation bug:** `test-summary` (lines 497-518) runs:
   ```yaml
   gh run download ${{ github.run_id }} \
     --pattern "test-reports-${{ needs.generate-matrix-test.outputs.test_matrix_hash }}-*" \
     --dir test_reports
   find test_reports -name "report_*.xml" -type f -exec mv {} test_reports/ \;
   ...
   uses: mikepenz/action-junit-report@v6
   with:
     report_paths: test_reports/report_*.xml
   ```
   The `--pattern` matches on `test_matrix_hash` only, not `github.run_attempt`, so it pulls the test-reports artifact from **every** attempt. Each shard's JUnit file is named `report_$JOB_ID.xml` (line 168), and job IDs differ per attempt, so the flatten step doesn't overwrite — both files survive:
   - `report_92866738056.xml` (attempt 1, full shard, FAILED testcase)
   - `report_92879876496.xml` (attempt 2, retried subset, same testcase PASSED)

   `action-junit-report` unions every `<testcase>` across all input files with no "same test, later attempt, take the newer verdict" logic. The failed record from attempt 1 is never superseded, so the merged check reports the test failed regardless of how many reruns fix it.

Nothing in the repo currently dedupes JUnit testcases across attempts — `find_all_failed_tests.py` only extracts failures (to decide what to rerun), and `summarize_junit_xmls.py` dedupes input files, not testcases across attempts.

## Suggested Fix

Before `test-summary` hands files to `action-junit-report` (`.github/workflows/call-test.yml:505-511`, before line 514), resolve each testcase to its result from the highest attempt that reports it, dropping the superseded copy:

- Artifact names already encode attempt + shard (`test-reports-<hash>-<attempt>.<job-index>-<runs-on>-<test_mark>-<check_run_id>`), so attempt number and shard identity are recoverable without touching the XML schema.
- A rerun's XML only contains the retried subset (not the full shard), so the merge must be per-testcase, not per-file: for each `(classname, name)`, keep the result from the highest attempt's XML that contains it, and strip that testcase out of any lower-attempt XML before the glob/flatten step.

FYI @sgligorijevicTT since I asked you about this earlier :)

pdeviTT:
## Problem

The nightly perf comparison has no data because **all perf-benchmark jobs are being skipped**. On run [31444942259](https://github.com/tenstorrent/tt-xla/actions/runs/31444942259), `perf-benchmark / filter-tests` succeeded but `run-perf-benchmarks` (and every per-model job) was skipped, even though the nightly passes a normal, non-empty `adv_filter`.

This is a regression introduced by #5925 (`cf1eb30`). The last nightly *before* that commit ([31343580920](https://github.com/tenstorrent/tt-xla/actions/runs/31343580920)) ran all 168 perf jobs; the first nightly *after* it skipped them all.

## Root cause

The guard added in #5925 reads a value that does not exist:

```yaml
if: ${{ !cancelled() && fromJson(needs.filter-tests.outputs.matrix.matrix_skip) != 'true' && needs.filter-tests.result == 'success' }}
```

1. **`matrix_skip` is never exported.** `filter-test-matrix.py` prints `{"matrix": [...], "matrix_skip": "true|false"}`, but the `filter-tests` step only extracts `.matrix` into `$GITHUB_OUTPUT`, and the job only declares a `matrix` output. `matrix_skip` is discarded.
2. **The access path is malformed.** `needs.filter-tests.outputs.matrix` is a JSON *string* (the array), so `.matrix_skip` on it does not resolve, and wrapping that in `fromJson(...)` never yields `'true'`/`'false'`.

Net effect: the middle term never reflects the real filter decision and collapses to "skip" for the nightly, so the entire perf matrix disappears.

## Fix

- Export `matrix_skip` as a real job output of `filter-tests`.
- Write it to `$GITHUB_OUTPUT` alongside `matrix`.
- Compare it directly as a string in the guard (it is already `"true"`/`"false"` from the Python script), so no `fromJson` is needed.

```yaml
if: ${{ !cancelled() && needs.filter-tests.result == 'success' && needs.filter-tests.outputs.matrix_skip != 'true' }}
```

This preserves the original intent of #5925 (cleanly skip when the filter matches nothing, instead of erroring on an empty matrix) while ensuring perf benchmarks actually run when the filter is non-empty.

## Test plan

- [x] Trigger the nightly (or a manual perf run with a non-empty `adv_filter`) and confirm `run-perf-benchmarks` runs and per-model perf jobs execute ([pipeline](https://github.com/tenstorrent/tt-xla/actions/runs/31477097226))
- [x] Confirm a filter that matches nothing (empty matrix / `matrix_skip == 'true'`) still cleanly skips `run-perf-benchmarks` without erroring ([pipeline](https://github.com/tenstorrent/tt-xla/actions/runs/31480195616)).

## Note

This does **not** address issue #5918 (rerun-passed tests still reported as failed in `TT-XLA basic-test` / `check-all-green`), which is an unrelated JUnit-aggregation problem in `call-test.yml`. #5925 was linked as "closes #5918" but did not fix it; that issue should be reopened and handled separately.

### DIFF
--- a/.github/workflows/call-filtered-perf-tests.yml
+++ b/.github/workflows/call-filtered-perf-tests.yml
@@ -63,6 +63,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       matrix: ${{ steps.set-perf-benchmarks.outputs.matrix }}
+      matrix_skip: ${{ steps.set-perf-benchmarks.outputs.matrix_skip }}
     steps:
     - name: Checkout repository
       uses: actions/checkout@v6
@@ -90,10 +91,12 @@ jobs:
         echo $result
         matrix=$(echo $result | jq -r -c '.matrix')
         echo "matrix=$matrix" >> $GITHUB_OUTPUT
+        matrix_skip=$(echo $result | jq -r -c '.matrix_skip')
+        echo "matrix_skip=$matrix_skip" >> $GITHUB_OUTPUT
 
   run-perf-benchmarks:
     needs: filter-tests
-    if: ${{ !cancelled() && fromJson(needs.filter-tests.outputs.matrix.matrix_skip) != 'true' && needs.filter-tests.result == 'success' }}
+    if: ${{ !cancelled() && needs.filter-tests.result == 'success' && needs.filter-tests.outputs.matrix_skip != 'true' }}
     secrets: inherit
     uses: ./.github/workflows/call-perf-test.yml
     with:


nsumrakTT (changes requested):
The fix for this problem is to move one character ')' to position before '.'

Issue #5575, opened 2026-07-09
SDXL Lightning fails nightly benchmarks


vkovacevicTT:
SDXL Lightning fails [nightly benchmarks](https://github.com/tenstorrent/tt-xla/actions/runs/28985068529/job/86113895508) on n150-perf and p150 perf with error:
```
tests/benchmark/test_imagegen.py:252: in test_sdxl_lightning
    test_imagegen(
tests/benchmark/test_imagegen.py:83: in test_imagegen
    results = benchmark_imagegen_torch_xla(
tests/benchmark/benchmarks/imagegen_benchmark.py:133: in benchmark_imagegen_torch_xla
    components = perf["components"]
                 ^^^^^^^^^^^^^^^^^^
E   KeyError: 'components
```

PR #5085, opened 2026-06-04
[Benchmark] Wire image-gen e2e pipelines into nightly + benchmark CI (SD1.5, SD3)

ctr-lelanchelianTT:
### Problem description

Following review feedback on #5044 (the merged Playground v2.5 PR): every model that passes bringup should be wired into **both nightly and benchmark CI** (#5047). #5044 mirrored this PR's original image-gen pattern and merged first, landing the canonical harness (`imagegen_benchmark.py`, `test_imagegen.py`) on `main`. This PR reconciles with that canonical harness and wires **SD 1.5** and **SD 3 Medium** into nightly + benchmark, alongside Playground v2.5.

### What's changed

**Benchmark** — mirrors the reviewed `imagegen_benchmark` pattern from #5044 (shared harness in `benchmarks/`, per-model config in `test_imagegen.py`):

- **`tests/benchmark/benchmarks/imagegen_benchmark.py`** — adopt the canonical two-pass harness and **generalize** its per-component reporting to a model-agnostic `_perf` schema so the same harness serves every image-gen pipeline:
  ```python
  _perf = {
      "components": {<name>: seconds, ...},   # scalar per-stage times (e.g. text_encoder, vae)
      "steps": [seconds, ...],                # per heavy-net-step times
      "step_metric_name": "unet_step" | "transformer_step",
      "total": seconds,                       # full generate() wall time
  }
  ```
  Two-pass scheme: **warmup** (1 step — triggers first-forward compile of every component) + **steady-state** (full step count, all cache hits). Emits per-component steady-state measurements + `cpu_overhead_s`, `e2e_latency`, `images_per_second`. The old `compile_time` / `per_step_latency_ms` are dropped (per @nvukobratTT on #5044).
- **`tests/benchmark/test_imagegen.py`** — config-driven `test_stable_diffusion_1_5`, `test_stable_diffusion_3`. Canonical API: `build_pipeline_fn(compile_options) -> (pipeline, generate_fn)`, `generate_fn(prompt, steps)`, images via `utils.save_image`. Imports consolidated at top (per review).
- **`tests/benchmark/benchmarks/playground_v2_5_pipeline.py`** — migrated to the generic `_perf` schema (emitted metric names unchanged).
- **`.github/workflows/perf-bench-matrix.json`** — `playground_v2_5` + the 2 image-gen matrix entries.

**Nightly** — mirrors #5044's `test_playground_v25_pipeline`:

[...]

PR #5954, opened 2026-08-14
[Benchmark] Wire the Voxtral-Mini-3B audio e2e pipeline into nightly + benchmark CI

ctr-lelanchelianTT:
### Problem description

Same policy as #5047 / #5085: every model that passes bringup gets wired into **both** nightly and benchmark CI. **Voxtral-Mini-3B-2507** (Mistral's audio-understanding model) passed bringup and its end-to-end audio+text -> text pipeline landed in tt-forge-models (tenstorrent/tt-forge-models#802), but nothing runs it in CI — no nightly e2e test, no benchmark entry, so a regression in the audio path would only surface by hand.

There is also no shared harness for audio models. Image-gen has two (`imagegen_benchmark.py` for diffusion, `ar_imagegen_benchmark.py` for autoregressive image tokens); an audio pipeline fits neither: the interesting throughput is decode tokens/second and there is no image to save.

### What's changed

**Benchmark** — mirrors the reviewed image-gen pattern (shared harness in `benchmarks/`, per-model config in the `test_*.py`):

- **`tests/benchmark/benchmarks/audio_benchmark.py`** (new) — shared harness for audio pipelines whose generation is an autoregressive decode loop: audio-understanding models that answer in text (Voxtral) and TTS models that emit audio tokens (XTTS-v2, Llasa) can both use it. It reads the same model-agnostic `_perf` schema the image-gen harness reads, with the seconds/counts split the XTTS-v2 pipeline already uses:
  ```python
  _perf = {
      "components": {<name>: seconds, ...},   # scalar per-stage device times
      "steps": [seconds, ...],                # one entry per decode step
      "step_metric_name": "decode_step",      # e.g. "audio_token_step"
      "total": seconds,                       # full generate() wall time
      <count_name>: int,                      # prefill_tokens, output_tokens, ...
  }
  ```
  Every extra scalar key is emitted verbatim as a dashboard measurement, so a pipeline can report its own counts without touching the harness. Emits `tokens_per_second` (headline), `ttft`, `e2e_latency`, `<step_metric_name>_mean_s`, `cpu_overhead_s` and one measurement per component; `total_time` / `total_samples` are the decode-loop totals, the same convention as `llm_benchmark.py`, so a generative audio model is comparable with the LLMs on the dashboard. No `compile_time` / `per_step_latency_ms` (per @nvukobratTT on #5044).

  Two-pass scheme as in the image-gen harnesses, with one difference worth calling out: the warmup pass uses the **same** token budget as the measured pass. These pipelines decode over a static right-padded window whose shape is `prefill + max_new_tokens`, so a 1-step warmup would compile a *different* shape and leave the compile inside the measured pass.

- **`tests/benchmark/test_audio.py`** (new) — config-driven `test_voxtral_mini_3b`. Canonical API: `build_pipeline_fn(compile_options) -> (pipeline, generate_fn)`, `generate_fn(max_new_tokens)`; imports at the top of the file (per review on #5044/#5085). The answer text is written next to the perf JSON.

- **`.github/workflows/perf-bench-matrix.json`** — `voxtral_mini_3b` entry, default runners (both n150 and p150 pass).

- **`venv/requirements-dev.txt`** — `mistral-common[audio]`. Voxtral tokenizes through the mistral_common backend (it ships a `tekken.json`, not a Jinja chat template), and this **cannot** be installed per-loader at test time: transformers resolves what is available while it is being imported and memoizes it, so a later install leaves the processor holding a tokenizer with no chat template and `apply_chat_template` raises `tokenizer.chat_template is not set`. A p150 nightly run caught exactly that; see the comment below. `venv/activate` installs this file, so both the test and perf jobs get it.

**Nightly** — mirrors #5085's `test_stable_diffusion_1_5_pipeline` and the XTTS-v2 audio test:

- **`tests/torch/models/voxtral/test_voxtral_pipeline.py`** (new) — full e2e audio -> text test: the Whisper-style audio tower and the audio/text merge stay on the host (the HF merge is a data-dependent `masked_scatter` that will not lower), the 30-layer Ministral-3B LM greedy-decodes on TT. Reuses the `tt_forge_models` pipeline shared with the benchmark, and installs the loader's own requirements through `RequirementsManager` exactly as the XTTS-v2 test does. Markers: `nightly`, `model_test`, `single_device`, `large`, `record_test_properties(MODEL_TEST, RED, INFERENCE, PASSED)`. No xfail.

  The test asserts the model actually *understood* the prompt: the answer has to name the sport in the audio (`baseball`, a word that appears nowhere in the prompt) and be a real sentence, rather than only checking that some text came back. It deliberately does not assert the rhyme's name or the specific game — n150 and p150 word the answer differently, so those would be device-dependent.

This PR carries **no `third_party/tt_forge_models` submodule bump** — the companion pipeline change lands separately (below) and is picked up via the normal submodule **uplift**.

**Blast radius:** nothing existing is touched. `audio_benchmark.py` and `test_audio.py` are new files with a single consumer, and no other model reads the Voxtral `_perf` (so nothing can repeat the `sdxl_lightning` `KeyError: 'components'` regression from #5575 — `imagegen_benchmark.py` and its models are untouched). The one shared file edited is `perf-bench-matrix.json`, where only a new entry is appended. Voxtral runs **single-chip** (3B at bf16 fits one Wormhole; verified on n150).

### Depends on (delivered via submodule uplift, not a bump in this PR)

- tt-forge-models companion: **tenstorrent/tt-forge-models#864** (`lelanchelian/voxtral-perf-schema`) — moves the Voxtral `_perf` counts out of `components` (the harness sums `components` as seconds) and adds the loader `requirements.txt` for `mistral-common`.
- **Land order:** merge the companion -> uplift `third_party/tt_forge_models` to the resulting main SHA -> this PR's benchmark + nightly imports resolve and run green.

### Verification

Every number below is from the code as it stands on this branch. The CI runs use `lelanchelian/voxtral-benchmark-ci-verify` (this branch + a temporary submodule bump to #864) for the benchmark, and the PR branch itself with `forge_models_override` for the nightly test — the pipeline is not in the current pin yet.

**Benchmark** — `tests/benchmark/test_audio.py::test_voxtral_mini_3b`, optimization_level 1, 32-token budget, 2-pass:

| | [n150-perf](https://github.com/tenstorrent/tt-xla/actions/runs/31816664740) | [p150-perf](https://github.com/tenstorrent/tt-xla/actions/runs/31816674271) |
|---|---|---|
| warmup pass (incl. compile) | 109.2 s | 133.2 s |
| steady-state pass (`e2e_latency`) | 44.5 s | 78.3 s |
| `ttft` | 1415 ms | 2436 ms |
| `decode_step_mean_s` | 1.4859 s | 2.4444 s |
| `tokens_per_second` | 0.67 | 0.41 |
| `cpu_overhead_s` | 0.002 s | 0.003 s |
| `prefill_tokens` | 765 | 765 |
| `output_tokens` | 29 (hit EOS) | 32 (hit budget) |
| pytest | **1 passed** (277 s) | **1 passed** (277 s) |

Answers, both correct:

- **n150** — "The sport referenced is baseball, specifically the 1969 World Series. The nursery rhyme referenced is "Mary Had a Little Lamb.""
- **p150** — "The audio references both baseball and a nursery rhyme. The baseball reference is about a home run hit by Edgar Martinez, ..." (cut off by the 32-token budget)

So the matrix entry keeps the **default runners** — no n150-only restriction like SD 1.5 / SD 3 needed in #5085.

**Nightly** — `tests/torch/models/voxtral/test_voxtral_pipeline.py`:

- p150 in CI: **1 passed** (256 s) — https://github.com/tenstorrent/tt-xla/actions/runs/31816684529
- n150 locally: **1 passed** (117 s), answer "The audio references both a nursery rhyme and a baseball game. The nursery rhyme is "Mary Had a Little Lamb," and the baseball game is a playoff game between the Baltimore Orioles and the Oakland Athletics."

The wording differs between the two because the benchmark sets `optimization_level=1` while the nightly test uses the default — different numerics, so greedy decode picks a different phrasing. Both answer correctly, which is why the assertion keys on the one content word that cannot come from the prompt (`baseball`) plus a minimum answer length, rather than an exact string.

On the throughput figure: 0.67 tokens/s is what this pipeline is, not a regression. There is no KV cache, so every decode step recomputes the full `prefill + max_new_tokens` window (797 tokens of a 3B LM) to keep the graph static. Wiring it into the benchmark means a future KV-cache version shows up as a step change instead of an anecdote. Also visible in the two runs: p150 is ~1.6x slower per decode step than n150 on the same graph.

### Checklist
- [x] Benchmark green in CI on n150 **and** p150 (with the companion pipeline change present), so the matrix entry keeps the default runners
- [x] Nightly e2e pipeline test green on n150 (local) and p150 (CI)
- [ ] Merge tenstorrent/tt-forge-models#864, then uplift the tt-xla submodule

### DIFF
--- a/.github/workflows/perf-bench-matrix.json
+++ b/.github/workflows/perf-bench-matrix.json
@@ -115,6 +115,10 @@
         "runs-on": "qb2-blackhole"
       },
       {
+        "name": "voxtral_mini_3b",
+        "pytest": "tests/benchmark/test_audio.py::test_voxtral_mini_3b"
+      },
+      {
         "name": "bge_m3_encode",
         "pyreq": "transformers==4.57.1 FlagEmbedding",
         "pytest": "tests/benchmark/test_encoders.py::test_bge_m3"
--- a/tests/benchmark/benchmarks/audio_benchmark.py
+++ b/tests/benchmark/benchmarks/audio_benchmark.py
@@ -1,0 +1,252 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Generic audio-pipeline benchmark harness for torch-xla / TT.
+
+Sibling of ``imagegen_benchmark.py`` / ``ar_imagegen_benchmark.py``, for models
+whose input or output is audio and whose generation is an autoregressive decode
+loop: audio-understanding models that answer in text (Voxtral) and TTS models
+that emit audio tokens (XTTS-v2, Llasa). The per-model configuration lives in
+``test_audio.py``; this module owns the reusable measurement logic.
+
+Two-pass scheme (same idea as the image-gen harnesses):
+
+  - Pass 1 (warmup): a full ``generate()``. It has to use the *same* token
+    budget as the measured pass: these pipelines decode over a static
+    right-padded window whose shape is ``prefill + max_new_tokens``, so a
+    shorter warmup would compile a different shape and leave the compile in the
+    measured pass.
+  - Pass 2 (steady-state): a second full ``generate()`` where every forward is
+    a cache hit. This is the pass that is reported.
+
+The pipeline reports its own per-stage timings in ``pipeline._perf``, in the
+schema shared with the image-gen harness — ``components`` and ``steps`` hold
+seconds, counts live in their own top-level keys::
+
+    _perf = {
+        "components": {<name>: seconds, ...},   # scalar per-stage device times
+        "steps": [seconds, ...],                # one entry per decode step
+        "step_metric_name": "decode_step",      # e.g. "audio_token_step"
+        "total": seconds,                       # full generate() wall time
+        <count_name>: int,                      # e.g. prefill_tokens, output_tokens
+    }
+
+Every extra scalar key is emitted verbatim as a dashboard measurement, so a
+pipeline can report its own counts (``prefill_tokens``, ``audio_samples``, ...)
+without touching this harness.
+
+Headline throughput is decode tokens/second, and ``total_time`` /
+``total_samples`` are the decode-loop totals — the same convention as
+``llm_benchmark.py``, so a generative audio model is comparable with the LLMs on
+the dashboard. ``steps[0]`` is reported as TTFT: these pipelines carry no KV
+cache, so the first step is the one that also computes the prompt.
+"""
+
+import socket
+import time
+
+import torch_xla
+import torch_xla.runtime as xr
+from utils import (
+    build_xla_export_name,
+    create_benchmark_result,
+    get_benchmark_metadata,
+    get_xla_device_arch,
+    print_benchmark_results,
+)
+
+xr.set_device_type("TT")
+
+MODULE_EXPORT_PATH = "modules"
+
+# Keys of the ``_perf`` schema that this harness interprets itself; any other
+# scalar key is passed through to the dashboard as-is.
+_PERF_RESERVED_KEYS = frozenset({"components", "steps", "step_metric_name", "total"})
+
+
+def benchmark_audio_torch_xla(
+    build_pipeline_fn,
+    model_info_name,
+    num_generation_steps,
+    optimization_level,
+    trace_enabled,
+    ttnn_perf_metrics_output_file,
+    model_type,
+    dataset_name,
+    display_name=None,
+    save_output_fn=None,
+):
+    """Benchmark an audio generation/understanding pipeline on the TT backend.
+
+    Args:
+        build_pipeline_fn: ``build_pipeline_fn(compile_options) -> (pipeline, generate_fn)``.
+            ``compile_options`` is forwarded so the pipeline can merge instead of
+            overwrite if it needs to switch an option inline.
+            ``generate_fn(num_generation_steps) -> output`` runs one full
+            generation and populates ``pipeline._perf``.
+        model_info_name: Model name for identification and reporting.
+        num_generation_steps: Token budget per generation (the decode-loop cap).
+        optimization_level: tt-mlir optimization level for compilation.
+        trace_enabled: Whether to enable tracing.
+        ttnn_perf_metrics_output_file: Base path for TTNN perf metrics files.
+        model_type: Dashboard model type (e.g. "audio-to-text").
+        dataset_name: Dashboard dataset name (e.g. "Audio Prompt").
+        display_name: Display name used for export naming / dashboard.
+        save_output_fn: If set, ``save_output_fn(output)`` persists the
+            steady-state output (a text answer, a waveform, ...).
+
+    Returns:
+        Standardized benchmark result dict (see ``create_benchmark_result``).
+    """
+    export_model_name = build_xla_export_name(
+        model_name=display_name or model_info_name,
+        num_layers=None,
+        batch_size=1,
+        input_sequence_length=None,
+    )
+
+    options = {
+        "optimization_level": optimization_level,
+        "export_path": MODULE_EXPORT_PATH,
+        "export_model_name": export_model_name,
+        "ttnn_perf_metrics_enabled": True,
+        "ttnn_perf_metrics_output_file": ttnn_perf_metrics_output_file,
+        "enable_trace": trace_enabled,
+    }
+    torch_xla.set_custom_compile_options(options)
+
+    # Build the pipeline (registers the "tt" backend and moves the model to the
+    # XLA device; kernels compile lazily on the first forward, i.e. during the
+    # warmup pass below).
+    pipeline, generate_fn = build_pipeline_fn(options)
+
+    # Pass 1 (warmup): same token budget as the measured pass, so the decode
+    # window has the shape the steady-state pass reuses.
+    print("Starting warmup pass (includes compile)...")
+    warmup_start = time.perf_counter()
+    generate_fn(num_generation_steps)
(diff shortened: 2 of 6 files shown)


ctr-lelanchelianTT:
### Benchmark CI: n150 ✅ and p150 ✅

Both runs are on `lelanchelian/voxtral-benchmark-ci-verify` — this branch plus a temporary submodule bump to tenstorrent/tt-forge-models#864, because the Voxtral pipeline is not in the current pin yet (same approach as #5085's verification run).

| | [n150-perf](https://github.com/tenstorrent/tt-xla/actions/runs/31810873189) | [p150-perf](https://github.com/tenstorrent/tt-xla/actions/runs/31810883101) |
|---|---|---|
| warmup pass (incl. compile) | 108.8 s | 134.5 s |
| steady-state pass (`e2e_latency`) | 44.2 s | 76.9 s |
| `ttft` | 1401 ms | 2320 ms |
| `decode_step_mean_s` | 1.4747 s | 2.4053 s |
| `tokens_per_second` | 0.68 | 0.42 |
| `cpu_overhead_s` | 0.001 s | 0.002 s |
| `prefill_tokens` | 765 | 765 |
| `output_tokens` | 29 (hit EOS) | 32 (hit budget) |
| pytest | **1 passed** (302 s) | **1 passed** (269 s) |

Answers, both correct:

- **n150** — "The sport referenced is baseball, specifically the 1969 World Series. The nursery rhyme referenced is "Mary Had a Little Lamb.""
- **p150** — "The audio references both baseball and a nursery rhyme. The baseball reference is about a home run hit by Edgar Martinez, which led to a celebration and a discussion" (cut off by the 32-token budget)

So the matrix entry keeps the **default runners** — no need to restrict it to n150 the way SD 1.5 / SD 3 were in #5085. Checklist item ticked.

Two follow-ups this run produced:

1. **The nightly assertion is now arch-independent.** p150 acknowledges the nursery rhyme without naming it inside the token budget, so the original "answer must contain both `baseball` and `mary`" check would have been phrasing- and device-dependent — a flake waiting to happen on the p150 nightly job. It now asserts the sport word only (`baseball`, which appears nowhere in the prompt, so it can only come from the audio) plus a minimum answer length. Re-verified on n150 after the change.

2. **p150 is ~1.6x slower per decode step than n150** on the same graph (2.4053 s vs 1.4747 s). Nothing in this PR addresses that — flagging it in case it is interesting, and the harness now tracks it either way.

ctr-lelanchelianTT:
### Update: p150 caught a real dependency bug; fixed and everything re-verified

The first p150 nightly run failed — worth writing down, because the cause is not obvious and it affects any model whose dependency gates transformers behaviour.

`RequirementsManager` **did** install `mistral-common` (visible in the job log), and the test still failed with `tokenizer.chat_template is not set`. The reason: transformers resolves what is installed *while it is being imported* — `is_<pkg>_available()` is `lru_cache`d and the tokenizer/auto mappings are built at import time — and conftest imports transformers long before any per-test install. So the Voxtral processor was already committed to a tokenizer with no chat template. Measured in isolation:

```
before install, find_spec: False
after install,  find_spec: True   _is_package_available: True   is_mistral_common_available: False   <- stale
after invalidate_caches + cache_clear:                          is_mistral_common_available: True
```

I first tried fixing that in-process (clearing those caches inside `RequirementsManager` after installs). The log shows it running — `cleared 131 transformers availability cache(s)` — and the test **still** failed, because the availability memo is not the only import-time-frozen state. So that change is reverted: no shared test-infra edits in this PR.

The fix that holds is environment-level: `mistral-common[audio]` in `venv/requirements-dev.txt`, which `venv/activate` installs before pytest starts, for both the test and the perf job. That let the nightly test get *simpler* — no `RequirementsManager` wrapper, no deferred imports, just a module-level import like #5085's SD 1.5 test — and made the matrix `pyreq` redundant, so it is gone too.

Because dropping `pyreq` changes how the perf job obtains the dependency, the earlier benchmark runs no longer covered the current code. All three runs below are on the code as it stands:

| run | result |
|---|---|
| [benchmark n150](https://github.com/tenstorrent/tt-xla/actions/runs/31816664740) | ✅ 1 passed — 1.4859 s/step, 0.67 tok/s, TTFT 1415 ms |
| [benchmark p150](https://github.com/tenstorrent/tt-xla/actions/runs/31816674271) | ✅ 1 passed — 2.4444 s/step, 0.41 tok/s, TTFT 2436 ms |
| [nightly test p150](https://github.com/tenstorrent/tt-xla/actions/runs/31816684529) | ✅ 1 passed (256 s) |

Also in this push: the nightly assertion no longer requires the answer to name the nursery rhyme (p150 acknowledges it without naming it inside the token budget, n150 names it — that would have been a device-dependent flake), and an unrelated JSON reformat of the `janus_pro_7b` entry was reverted, so the matrix diff is four added lines and nothing else.

vvukomanTT (approved):

nsumrakTT (approved):

.github/workflows/perf-bench-matrix.json
```diff
@@ -114,6 +114,10 @@
         "pytest": "tests/benchmark/test_imagegen.py::test_glm_image",
         "runs-on": "qb2-blackhole"
       },
+      {
+        "name": "voxtral_mini_3b",
+        "pytest": "tests/benchmark/test_audio.py::test_voxtral_mini_3b"
+      },
```
vkovacevicTT, line 120: Please rebase and add `"group" : "experimental"` here :)

vkovacevicTT (changes requested):

PR #5983, opened 2026-08-20
[HunyuanImage-2.1]  Replace distilled variant with base variant

kamalrajkannan78:
### Ticket

- https://github.com/tenstorrent/tt-xla/issues/5985

### Problem description

- The requirement is to support [HunyuanImage-2.1](https://huggingface.co/docs/diffusers/api/pipelines/hunyuanimage21). It ships three variants which share the same MMDiT architecture and differ in how guidance is applied:

| Variant | num_inference_steps | guidance |
| --- | --- | --- |
| HunyuanImage-2.1 | 50 | real CFG + APG, from the repo's `guider` / `ocr_guider` |
| HunyuanImage-2.1-Distilled | 8 | `distilled_guidance_scale=3.5`, embedded in the forward |
| HunyuanImage-2.1-Refiner | 4 | separate second-stage pipeline |

- The Distilled variant was added initially, as it reaches comparable output quality in far fewer steps. The current required variant is the base model, so the component loader and the e2e pipeline should both target it: guidance is no longer embedded in the transformer, so the denoising loop needs restructuring — two forwards per step with a host-side combine — not just a new checkpoint.

### What's changed

- Tracks [tt-forge-models#889](https://github.com/tenstorrent/tt-forge-models/pull/889), which carries the loader and pipeline changes. Once it lands and the auto-uplift picks it up, this PR will be rebased onto the new submodule pointer.
- `examples/pytorch/hunyuan_image_2_1.py`: dropped the `DISTILLED_GUIDANCE_SCALE` import and `generate()` kwarg, which no longer exist for the base model.
- `tests/benchmark/test_imagegen.py`: `num_inference_steps` 8 → 10 (10 for now, will be boosted to the model card's 50 later). Each step is now two transformer forwards, so device time per image roughly doubles at the same step count.
- `tests/torch/models/HunyuanImage_2_1/test_hunyuan_image_2_1_pipeline.py`: `num_inference_steps` 8 → 10, and docstring updated for the CFG call pattern — Qwen checked twice (conditional + unconditional), transformer twice per step.
- After this change, the e2e nightly PCC test fails with a PCC drop in transformer — tracked separately in [#5984](https://github.com/tenstorrent/tt-xla/issues/5984).

### Checklist

- [x] Cpu parity check - [cpu_parity.zip](https://github.com/user-attachments/files/31272893/cpu_parity.zip)
- [x] PCC gated nightly test - [PCC_check.log.zip](https://github.com/user-attachments/files/31272909/aug20_hunyuan_image_2_1_pipeline_base_pcc_check_1.log.zip)
- [x] Benchmark - https://github.com/tenstorrent/tt-xla/actions/runs/32376506316

### DIFF
--- a/tests/torch/models/HunyuanImage_2_1/test_hunyuan_image_2_1_pipeline.py
+++ b/tests/torch/models/HunyuanImage_2_1/test_hunyuan_image_2_1_pipeline.py
@@ -2,10 +2,12 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-"""HunyuanImage 2.1 (Distilled) — nightly e2e pipeline test, every TT component
-PCC-gated against a CPU twin in the same dtype. The Qwen and ByT5 encoders and
-the MMDiT transformer run bf16 on TT; scheduler and VAE stay on CPU. Encoders
-are checked once each, the transformer once per denoising step.
+"""HunyuanImage 2.1 — nightly e2e pipeline test, every TT component PCC-gated
+against a CPU twin in the same dtype. The Qwen and ByT5 encoders and the MMDiT
+transformer run bf16 on TT; scheduler, guider combine and VAE stay on CPU.
+
+Guidance is real CFG, so Qwen is checked twice (conditional + unconditional) and
+the transformer twice per denoising step.
 """
 
 import pytest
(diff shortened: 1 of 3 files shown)


nvukobratTT (approved):

tests/torch/models/HunyuanImage_2_1/test_hunyuan_image_2_1_pipeline.py
```diff
@@ -118,7 +120,14 @@ def wrapped_forward(*args):
     model_info=MODEL_INFO,
     parallelism=Parallelism.TENSOR_PARALLEL,
     run_mode=RunMode.INFERENCE,
-    bringup_status=BringupStatus.PASSED,
+    bringup_status=BringupStatus.INCORRECT_RESULT,
+)
```
nvukobratTT: Let's figure out failure (ideally fix it) before merging it to the CI :))

Issue #5479, opened 2026-07-01
[Cog Videox 5B] PCC Drop in Vae Decoder Component


meenakshiramanathan1:
`tests/torch/models/cog_videox/test_vae_decoder.py::test_vae_decoder_sharded` fails with 
`error: 'ttnn.group_norm' op flattened height must be tile-aligned, got 5400`

meenakshiramanathan1:
`tests/torch/models/cog_videox/test_vae_decoder.py::test_vae_decoder_sharded` now fails with pcc drop in latest main
` AssertionError: Evaluation result 0 failed: PCC comparison failed. Calculated: pcc=0.8352076483242976. Required: pcc=0.99.`

PR #6020, opened 2026-08-28
Skip cog_videox VAE decoder sharded test (hangs, 240-min job timeout)

ctr-pmuruganTT:
## Summary
- `test_vae_decoder_sharded` (`tests/torch/models/cog_videox/test_vae_decoder.py`) carried a stale `@pytest.mark.xfail` from an old compile-time bug (#5479, `ttnn.group_norm` tile-alignment error) that no longer reproduces.
- What actually happens now is a **hang**: the test never returns, so the `nightly_tests / test nightly and llmbox (n300-llmbox, run_torch_multi_chip)` job burns its full 240-minute action budget and times out with no result, every night since 2026-08-14 (confirmed recurring in nightlies 31756540956, 31980812865, 32199716272, 33029130061, 33136112144).
- `xfail` only evaluates *after* a test runs to completion, so it does nothing to prevent a hang — that's why this test kept hanging despite already being "marked."
- PR #5782 already fixed this exact class of hang for the `krea_realtime`, `flux2`, and `z_image` VAE decoders by switching them from xfail/unmarked to `@pytest.mark.skip(...)`. This applies the identical pattern to `cog_videox`, the one VAE decoder that PR missed (it wasn't unmarked — it already had an xfail, which looked "handled" but wasn't).

## Test plan
- [x] CI green on the `nightly_tests and llmbox` job (n300-llmbox) — verify the job completes well under 240 minutes and this test reports `SKIPPED`, not a timeout.

This is a test-marker-only change (no model/runtime code touched), so it cannot introduce a new failure — it only prevents an already-broken test from hanging the whole job.


ctr-pmuruganTT:
Verified with a targeted CI dispatch against this branch on real n300-llmbox hardware (not waiting for the next scheduled nightly): https://github.com/tenstorrent/tt-xla/actions/runs/33160456798

```
tests/torch/models/cog_videox/test_vae_decoder.py::test_vae_decoder_sharded SKIPPED
======================= 1 skipped, 14 warnings in 0.12s ========================
```

Job completed in ~4 minutes total (vs. the 240-minute timeout it used to hit every night). Checked off the test-plan item.

